PR #1, opened 2025-07-11
feat(Data/HasFresh): new constructor `HasFresh.ofNatEmbed`

kckennylau:


kckennylau:
I have two proofs here, which one do you prefer?

PR #5, opened 2025-07-13
Locally Nameless Lambda Calculus

chenson2018:
This PR would add the basic definitions for a locally nameless style lambda calculus. I have adapted this from my repo, but minimizing it by removing pairs and constants. I have tried to add more descriptive documentation.

I also made a couple of small changes in `HasFresh.lean`

- add a `simp` attribute to `HasFresh.fresh_notMem`, which is helpful for proofs using `aesop`
- add an existential theorem for `HasFresh` types, which is nice when you are in `Prop` and don't care about the computational content of your fresh variable

I couple of things that are completely flexible if you would like changed:

- I am not sure how strict style is here versus say Mathlib. Please point out anything egregious. (Do we want to enable some of the linters Mathlib has?)
- If any notations seem problematic, just let me know. They can easily be scoped, precedence changed, etc. 
- Since we would now have multiple representations, I moved `Computability/LambdaCalculus/Untyped/Basic.lean` into its own subdirectory and namespace `LambdaCalculus.Named`.

### DIFF
--- a/Cslib/Computability/LambdaCalculus/Untyped/LocallyNameless/Basic.lean
+++ b/Cslib/Computability/LambdaCalculus/Untyped/LocallyNameless/Basic.lean
@@ -1,0 +1,74 @@
+/-
+Copyright (c) 2025 Chris Henson. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Chris Henson
+-/
+
+import Cslib.Data.HasFresh
+
+variable {Fvar : Type} [HasFresh Fvar] [DecidableEq Fvar]
+
+namespace LambdaCalculus.LocallyNameless
+
+/-- Syntax of locally nameless lambda terms, with free variables over `Fvar`. -/
+inductive Term (Fvar : Type)
+| bvar : ℕ → Term Fvar
+| fvar : Fvar → Term Fvar
+| lam  : Term Fvar → Term Fvar
+| app  : Term Fvar → Term Fvar → Term Fvar
+
+/-- Variable opening of the ith bound variable. -/
+@[simp]
+def Term.open_rec (i : ℕ) (sub : Term Fvar) : Term Fvar → Term Fvar
+| bvar i' => if i = i' then sub else bvar i'
+| fvar x  => fvar x
+| app l r => app (open_rec i sub l) (open_rec i sub r)
+| lam M   => lam $ open_rec (i+1) sub M
+
+notation:68 e "⟦" i " ↝ " sub "⟧"=> Term.open_rec i sub e 
+
+/-- Variable opening of the closest binding. -/
+@[simp]
+def Term.open' {X} (e u):= @Term.open_rec X 0 u e
+
+infixr:80 " ^ " => Term.open'
+
+/-- Variable closing, replacing a free `fvar x` with `bvar k` -/
+@[simp]
+def Term.close_rec (k : ℕ) (x : Fvar) : Term Fvar → Term Fvar
+| fvar x' => if x = x' then bvar k else fvar x'
+| bvar i  => bvar i
+| app l r => app (close_rec k x l) (close_rec k x r)
+| lam t   => lam $ close_rec (k+1) x t
+
+notation:68 e "⟦" k " ↜ " x "⟧"=> Term.close_rec k x e 
+
+/-- Variable closing of the closest binding. -/
+@[simp]
+def Term.close {Fvar} [DecidableEq Fvar] (e u):= @Term.close_rec Fvar _ 0 u e
+
+infixr:80 " ^* " => Term.close
+
+/- Substitution of a free variable to a term. -/
+@[simp]
+def Term.subst (x : Fvar) (sub : Term Fvar) : Term Fvar → Term Fvar
+| bvar i  => bvar i
+| fvar x' => if x = x' then sub else fvar x'
+| app l r => app (subst x sub l) (subst x sub r)
+| lam M   => lam $ subst x sub M
+
+notation:67 e "[" x ":=" sub "]" => Term.subst x sub e 
+
+/-- Free variables of a term. -/
+@[simp]
+def Term.fv : Term Fvar → Finset Fvar
+| bvar _ => {}
+| fvar x => {x}
+| lam e1 => e1.fv
+| app l r => l.fv ∪ r.fv
+
+/-- Locally closed terms. -/
+inductive Term.LC : Term Fvar → Prop
+| fvar (x)  : LC (fvar x)
+| lam (L : Finset Fvar) (e : Term Fvar) : (∀ x : Fvar, x ∉ L → LC (e ^ fvar x)) → LC (lam e)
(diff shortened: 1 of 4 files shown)


Cslib/Computability/LambdaCalculus/Untyped/LocallyNameless/Basic.lean
```diff
@@ -0,0 +1,74 @@
+/-
+Copyright (c) 2025 Chris Henson. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Chris Henson
+-/
+
```
fmontesi, line 6: Could you add a doc header with a title and a ref to some paper or similar that explains the theory, for reference?

Cslib/Computability/LambdaCalculus/Untyped/LocallyNameless/Basic.lean
```diff
@@ -0,0 +1,74 @@
...
+import Cslib.Data.HasFresh
+
+variable {Fvar : Type} [HasFresh Fvar] [DecidableEq Fvar]
```
fmontesi: I guess just Var would be a good name, to be consistent with the untyped nomenclature?

Cslib/Computability/LambdaCalculus/Untyped/LocallyNameless/Basic.lean
```diff
@@ -0,0 +1,74 @@
...
+
+namespace LambdaCalculus.LocallyNameless
+
+/-- Syntax of locally nameless lambda terms, with free variables over `Fvar`. -/
+inductive Term (Fvar : Type)
```
fmontesi: Here and above, Type could be Type*? Or maybe declare a universe u and make these Type u.

Cslib/Computability/LambdaCalculus/Untyped/LocallyNameless/Basic.lean
```diff
@@ -0,0 +1,74 @@
...
+
+/-- Variable opening of the ith bound variable. -/
+@[simp]
+def Term.open_rec (i : ℕ) (sub : Term Fvar) : Term Fvar → Term Fvar
+| bvar i' => if i = i' then sub else bvar i'
+| fvar x  => fvar x
+| app l r => app (open_rec i sub l) (open_rec i sub r)
+| lam M   => lam $ open_rec (i+1) sub M
+
+notation:68 e "⟦" i " ↝ " sub "⟧"=> Term.open_rec i sub e
```
fmontesi: Probably a good idea to make all these notations locally scoped.

I'm not very familiar with them, so the reference would be helpful. :⁠-⁠)

Cslib/Computability/LambdaCalculus/Untyped/LocallyNameless/Basic.lean
```diff
@@ -0,0 +1,74 @@
...
+
+/-- Variable closing of the closest binding. -/
+@[simp]
+def Term.close {Fvar} [DecidableEq Fvar] (e u):= @Term.close_rec Fvar _ 0 u e
+
+infixr:80 " ^* " => Term.close
+
+/- Substitution of a free variable to a term. -/
+@[simp]
+def Term.subst (x : Fvar) (sub : Term Fvar) : Term Fvar → Term Fvar
```
fmontesi: Could you make the order of params consistent with Term.subst in the untyped version? It'll enable using the dot notation on the subject term.

fmontesi (changes requested):

Cslib/Computability/LambdaCalculus/Untyped/LocallyNameless/Basic.lean
chenson2018: References added.

Cslib/Computability/LambdaCalculus/Untyped/LocallyNameless/Basic.lean
chenson2018: I changed to `Type u` for consistency with other files.

Cslib/Computability/LambdaCalculus/Untyped/LocallyNameless/Basic.lean
```diff
@@ -0,0 +1,93 @@
...
+namespace LambdaCalculus.LocallyNameless
+
+/-- Syntax of locally nameless lambda terms, with free variables over `Var`. -/
+inductive Term (Var : Type u)
+/-- Bound variables that appear under a lambda abstraction, using a de-Bruijn index. -/
+| bvar : ℕ → Term Var
+/-- Free variables. -/
+| fvar : Var → Term Var
+/-- Lambda abstraction, introducing a new bound variable. -/
+| lam  : Term Var → Term Var
```
fmontesi: This is called `abs` both in the reference you give and in the Named variant, so I'd align it here too.

Cslib/Computability/LambdaCalculus/Untyped/LocallyNameless/Basic.lean
```diff
@@ -0,0 +1,93 @@
...
+/-- Function application. -/
+| app  : Term Var → Term Var → Term Var
+
+/-- Variable opening of the ith bound variable. -/
+@[simp]
+def Term.open_rec (i : ℕ) (sub : Term Var) : Term Var → Term Var
```
fmontesi: I might be wrong, but I think defs like this are supposed to be camelCased, so openRec.

Cslib/Computability/LambdaCalculus/Untyped/LocallyNameless/Basic.lean
```diff
@@ -0,0 +1,93 @@
...
+/- Substitution of a free variable to a term. -/
+@[simp]
+def Term.subst (m : Term Var) (x : Var) (sub : Term Var) : Term Var :=
+  match m with
+  | bvar i  => bvar i
+  | fvar x' => if x = x' then sub else fvar x'
+  | app l r => app (l.subst x sub) (r.subst x sub)
+  | lam M   => lam $ M.subst x sub
+
+scoped notation:67 e "[" x ":=" sub "]" => Term.subst e x sub
```
fmontesi: Now that you have the right order you can just instantiate HasSubstitution.

Cslib/Computability/LambdaCalculus/Untyped/LocallyNameless/Basic.lean
chenson2018: I think I did this at annoyance at clashing with `abs` (absolute value) being in the global namespace and the fact that `Lean.Expr` uses lam. (I don't mind changing for consistency though.)

fmontesi:
Wonderful! Ready to be merged for me. Could you have a look at the merge conflicts?

chenson2018:
> Wonderful! Ready to be merged for me. Could you have a look at the merge conflicts?

I have merged the recent PR that was conflicting.

PR #9, opened 2025-07-15
Reduction notation attribute

chenson2018:
This PR will add an attribute for relations that generates reduction notations. See the test file and docs for the proposed usage. If we identify other common notations that are "parameterized" in this way, we could repeat this pattern. 

The precedence and Unicode arrows I have used here are not meant to be definitive. Let me know what we've decided on and I will update.


CslibTests/ReductionNotation.lean
thomaskwaring, line 1: Very nice! Ideally we would also automatically generate `Transitive` instances, so eg the second example could just be something like
```
calc
  5 ⇢ₙ 4 := by simp
  4 ⇢ₙ 3 := by simp
```
using an instance of `Transitive PredReduction PredReduction (Relation.ReflTransGen PredReduction)`

chenson2018: This already works, I have updated to use this cleaner style.

thomaskwaring: Nice! with the right instances you could remove the "apply single" (and replace `↠ₙ` with  `⇢ₙ`), but that's only a minor golf I suppose

chenson2018: Oh I see, this would be nice. Are these instances that really should be in Mathlib, but are missing?

CslibTests/ReductionNotation.lean
```diff
@@ -0,0 +1,15 @@
+import Cslib.Syntax.ReductionNotation
+
+@[reduction rs "ₙ", simp]
+def PredReduction (a b : ℕ) : Prop := a = b + 1
+
+open Relation.ReflTransGen
+
+lemma single_step : 5 ⭢ₙ 4 := by
+  change PredReduction _ _
```
chenson2018: What is the preferred way to write this? It is a bit cumbersome to always have to rewrite from `rs.Red` to actually use the relation.

Cslib/Semantics/ReductionSystem/Basic.lean
```diff
@@ -27,7 +27,7 @@ section MultiStep
 /-! ## Multi-step reductions -/
 
 /-- Multi-step reduction relation. -/
-def ReductionSystem.MRed (rs : ReductionSystem Term) :=
+abbrev ReductionSystem.MRed (rs : ReductionSystem Term) :=
```
chenson2018, line 30: Having this as a `def` prevents typeclasses like `Trans` for using `calc` blocks.

CslibTests/ReductionNotation.lean
thomaskwaring: Yep exactly, the class I wanted was actually `Trans`, eg for that example you'd want something similar to this (from `CombinatoryLogic/Defs.lean`:
```
instance RedRedTrans :
    Trans Red Red MRed := by
  constructor
  intro a b c hab hbc
  replace hab := MRed.single _ _ hab
  replace hbc := MRed.single _ _ hbc
  exact Relation.ReflTransGen.trans hab hbc
```

Cslib/Semantics/ReductionSystem/Basic.lean
fmontesi: :+1:

CslibTests/ReductionNotation.lean
chenson2018: I think it would be nice if we have these three `Trans` instances:
```
Trans Red Red MRed
Trans MRed Red MRed
Trans Red MRed MRed
```

Then in a `calc ` block we can switch back and forth between single and multiple steps as needed.

chenson2018: (this is really polymorphic over a `ReductionSystem` value, I just wrote it that way for readability)

Cslib/Semantics/ReductionSystem/Basic.lean
```diff
@@ -40,4 +40,62 @@ theorem ReductionSystem.MRed.single (rs : ReductionSystem Term) (h : rs.Red a b)
   rs.MRed a b :=
   Relation.ReflTransGen.single h
 
+open Relation Relation.ReflTransGen
+
+-- these instance allow us to switch between single and multistep reductions in a `calc` block
+instance {α} (R : α → α → Prop) : Trans R R (ReflTransGen R) where
```
chenson2018, line 46: @thomaskwaring I think this is a little nicer way of getting the `Trans` instances.

PR #11, opened 2025-07-17
Locally Nameless Beta Confluence

chenson2018:
This PR would add a proof of beta reduction confluence for the locally nameless representation of the lambda calculus. I have used the new `reduction_sys` attribute, which seems to work well. There are a few relatively small issues I have marked with `TODO` comments. These could be addressed in a future PR, but I will mention them here:

- It is useful to have a lemma of the form `rs.Red = Rel`, where `rs` is the reduction system generated from `Rel`. I think this could automatically be generated with the reduction system.
- Occasionally the notation causes problems for automation like `aesop`, where you end up needing to carrying around two versions of a proof.
- My proof for the diamond property of parallel reduction is long. I tried using the Takahashi translation proof, but had trouble doing this for locally nameless terms.

Edit: I fixed the issue with `Trans`

### DIFF
--- a/Cslib/Computability/LambdaCalculus/Untyped/LocallyNameless/AesopRuleset.lean
+++ b/Cslib/Computability/LambdaCalculus/Untyped/LocallyNameless/AesopRuleset.lean
@@ -1,3 +1,3 @@
 import Aesop
 
-declare_aesop_rule_sets [ln] (default := true)
+declare_aesop_rule_sets [LambdaCalculus.LocallyNameless.ruleSet] (default := true)
--- a/Cslib/Computability/LambdaCalculus/Untyped/LocallyNameless/AesopRuleset.lean
+++ b/Cslib/Computability/LambdaCalculus/Untyped/LocallyNameless/AesopRuleset.lean
@@ -1,0 +1,3 @@
+import Aesop
+
+declare_aesop_rule_sets [ln] (default := true)
--- a/Cslib/Computability/LambdaCalculus/Untyped/LocallyNameless/BetaReduction.lean
+++ b/Cslib/Computability/LambdaCalculus/Untyped/LocallyNameless/BetaReduction.lean
@@ -6,6 +6,7 @@ Authors: Chris Henson
 
 import Cslib.Computability.LambdaCalculus.Untyped.LocallyNameless.Basic
 import Cslib.Computability.LambdaCalculus.Untyped.LocallyNameless.Properties
+import Cslib.Semantics.ReductionSystem.Basic
 
 /-! # β-reduction for the λ-calculus
 
@@ -24,6 +25,7 @@ variable {Var : Type u}
 namespace LambdaCalculus.LocallyNameless.Term
 
 /-- A single β-reduction step. -/
+@[reduction_sys beta_rs "β"]
 inductive Step : Term Var → Term Var → Prop
 /-- Reduce an application to a lambda term. -/
 | β : LC (abs M)→ LC N → Step (app (abs M) N) (M ^ N)
@@ -34,18 +36,12 @@ inductive Step : Term Var → Term Var → Prop
 /-- Congruence rule for lambda terms. -/
 | ξ (xs : Finset Var) : (∀ x ∉ xs, Step (M ^ fvar x) (N ^ fvar x)) → Step (abs M) (abs N) 
 
-/-- Notation for a single reduction step. -/
-notation:39 t " ⇢β " t' => Step t t'
-
-/-- Notation for multiple reduction steps (the reflexive transitive closure). -/
-notation:39 t " ↠β " t' => Relation.ReflTransGen Step t t'
-
 open Step
 
 variable {M M' N N' : Term Var}
 
 /-- The left side of a reduction is locally closed. -/
-lemma step_lc_l (step : M ⇢β M') : LC M := by
+lemma step_lc_l (step : M ⭢β M') : LC M := by
   induction step <;> constructor
   all_goals assumption
 
@@ -66,22 +62,22 @@ theorem redex_app_r_cong : (M ↠β M') → LC N → (app N M ↠β app N M') :=
 variable [HasFresh Var] [DecidableEq Var]
 
 /-- The right side of a reduction is locally closed. -/
-lemma step_lc_r (step : M ⇢β M') : LC M' := by
+lemma step_lc_r (step : M ⭢β M') : LC M' := by
   induction step
-  case β => apply beta_lc <;> assumption
+  case «β» => apply beta_lc <;> assumption
   all_goals try constructor <;> assumption 
 
 /-- Substitution respects a single reduction step. -/
-lemma redex_subst_cong (s s' : Term Var) (x y : Var) : (s ⇢β s') -> (s [ x := fvar y ]) ⇢β (s' [ x := fvar y ]) := by
+lemma redex_subst_cong (s s' : Term Var) (x y : Var) : (s ⭢β s') -> (s [ x := fvar y ]) ⭢β (s' [ x := fvar y ]) := by
   intros step
   induction step
   case ξₗ ih => exact ξₗ (subst_lc (by assumption) (by constructor)) ih 
   case ξᵣ ih => exact ξᵣ (subst_lc (by assumption) (by constructor)) ih  
-  case β m n abs_lc n_lc => 
+  case «β» m n abs_lc n_lc => 
     cases abs_lc with | abs xs _ mem =>
       simp only [open']
       rw [subst_open x (fvar y) 0 n m (by constructor)]
-      refine β ?_ (subst_lc n_lc (by constructor))
+      refine «β» ?_ (subst_lc n_lc (by constructor))
       exact subst_lc (LC.abs xs m mem) (LC.fvar y)
   case ξ m' m xs mem ih => 
     apply ξ ({x} ∪ xs)
@@ -97,7 +93,7 @@ lemma redex_subst_cong (s s' : Term Var) (x y : Var) : (s ⇢β s') -> (s [ x :=
     all_goals aesop
 
 /-- Abstracting then closing preserves a single reduction. -/
-lemma step_abs_close {x : Var} : (M ⇢β M') → (abs (M⟦0 ↜ x⟧) ⇢β abs (M'⟦0 ↜ x⟧)) := by
+lemma step_abs_close {x : Var} : (M ⭢β M') → (abs (M⟦0 ↜ x⟧) ⭢β abs (M'⟦0 ↜ x⟧)) := by
   intros step
   apply ξ ∅
   intros y _
--- a/Cslib/Computability/LambdaCalculus/Untyped/LocallyNameless/BetaReduction.lean
+++ b/Cslib/Computability/LambdaCalculus/Untyped/LocallyNameless/BetaReduction.lean
@@ -1,0 +1,126 @@
+/-
+Copyright (c) 2025 Chris Henson. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Chris Henson
+-/
+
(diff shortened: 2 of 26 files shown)


chenson2018:
Side comment: would we like to configure GitHub to merge PRs as a squash commit? I personally like this because it makes the history cleaner, but I know people have differing opinions.

fmontesi:
Yeah, squash and merge sounds good, especially for PRs with so many commits.

fmontesi:
Half-way through it, looking good so far. You still have quite a few definitions (also from the previous pr, I think) where the order of parameters is different from the corresponding notation. I'd consider switching, for example: in defs about terms for opening and closing, I'd make the first parameter the term being manipulated. This will allow for dot-notation, which is pretty nice.

Do you see counter-indications to what I'm saying?

(I think I understand the reasoning behind your current approach btw, it makes the signature at the end a clean Term -> Term in many cases.)

chenson2018:
> Half-way through it, looking good so far. You still have quite a few definitions (also from the previous pr, I think) where the order of parameters is different from the corresponding notation. I'd consider switching, for example: in defs about terms for opening and closing, I'd make the first parameter the term being manipulated. This will allow for dot-notation, which is pretty nice.
> 
> Do you see counter-indications to what I'm saying?
> 
> (I think I understand the reasoning behind your current approach btw, it makes the signature at the end a clean Term -> Term in many cases.)

This PR does not introduce any new notations (outside of using `reduction_sys`), so these would be previously reviewed definitions. There are exactly four notations defined in `LocallyNameless.Basic`: opening, closing, and their specializations to the closest binding. 

My opinion on the dot notation is similar to Yael's in the Zulip discussion the other day. In Mathlib, the idiomatic pattern seems to be treating notation as the cannonical form and building an "API" around them. We seem to be following this idiom with our notations for reduction, substitution, etc., and ideally we would never see the non-notation terms and dot notation at all (perhaps needing delaborators to realize this fully). I do like dot notation myself on occasion, but I think it is considered a bit of an anti-pattern to be concerned with argument layout in this way. I also find it a hassle for functions with multiple paramaters of the same type to have to think about the order in which they were used. And finally yes, I like the aesthetic of pattern matching on a `Term -> Term`, though of course this is not a reason to make a decision one way or another.

In short, I think the order of paramaters for a definition used as notation should be treated as irrelevant. It's no big deal at all to change it if you would like, but I think as we encounter situations where it seems we need to work with these terms directly we should be thinking about why the abstraction of using these notations as API has broken down and aim to fix this. 

(A very long message to conclude something doesn't matter 😂)

fmontesi:
Mmmh.... Ok, I think I'm sold! 😀

Cslib/Computability/LambdaCalculus/Untyped/LocallyNameless/BetaReduction.lean
```diff
@@ -0,0 +1,122 @@
...
+
+universe u
+
+variable {Var : Type u} 
+
+namespace LambdaCalculus.LocallyNameless.Term
+
+/-- A single β-reduction step. -/
+@[reduction_sys beta_rs "β"]
+inductive Step : Term Var → Term Var → Prop
```
fmontesi: What strategy is this? Full beta-reduction + reduction under abstraction?

chenson2018: I always have to look up the strategy names... yes this is full, as there are no side conditions about values.

fmontesi: Probably good to give it a descriptive name then. Shall we call it fullBetaRs (I'm not sure if reduction systems should be camel or snake cased, do you know what they do in mathlib perhaps?) and fullβ (or fβ) as notation suffix then?

Other strategy suffixes are probably gonna be cbv, cbn, cbf, no, ao, etc.

chenson2018: Following the style guide, I think the inductive should be `UpperCamelCase`, while an instance of a structure `lowerCamelCase`, so your suggested names sound good.

Cslib/Computability/LambdaCalculus/Untyped/LocallyNameless/FullBetaConfluence.lean
```diff
@@ -0,0 +1,248 @@
...
+/-! # β-confluence for the λ-calculus -/
+
+universe u
+
+variable {Var : Type u} 
+
+namespace LambdaCalculus.LocallyNameless.Term
+
+/-- A parallel β-reduction step. -/
+@[aesop safe [constructors], reduction_sys paraRs "ₚ"]
```
fmontesi: This might be ok because it's 'safe' (so faster than unsafe), but I've had bad experiences in another big project with aesop annotations without a specified ruleset, especially when it comes to cases and constructors. Search becomes just too slow. Grind seems better at this (without requiring rulesets) but I don't think they support constructors/cases.

Should we just introduce a ruleset for nameless lambda calculus? If so, this is probably a good time to figure out/discuss a name scheme for rule sets.

Cslib/Computability/LambdaCalculus/Untyped/LocallyNameless/FullBeta.lean
```diff
@@ -0,0 +1,122 @@
...
+
+variable {Var : Type u} 
+
+namespace LambdaCalculus.LocallyNameless.Term
+
+/-- A single β-reduction step. -/
+@[reduction_sys fullBetaRs "βᶠ"]
+inductive FullBeta : Term Var → Term Var → Prop
+/-- Reduce an application to a lambda term. -/
+| β : LC (abs M)→ LC N → FullBeta (app (abs M) N) (M ^ N)
```
fmontesi: Where do the rule names come from? I'm particularly curious about the contextual/congruence rules, which are labelled by the symbol typically used to range over contexts, but they don't actually (formally) use contexts. I would've expected names like appL, appR, etc. (but I can't find a reference, I'm making this up from experience!).

chenson2018: This is just misleading notation, your names are better and closer to the ones used in the [listed reference](https://chargueraud.org/research/2009/ln/main.pdf) (section 4.5). I think I accidentally mixed this in from Barendregt.

Cslib/Computability/LambdaCalculus/Untyped/LocallyNameless/FullBetaConfluence.lean
chenson2018: Yeah, I should do this now rather than later. I also want to revisit exactly what is marked as `simp`. Having calls to `aesop (config := {enableSimp := false})` is probably an indicator I'm unfold notations too eagerly. I think maybe I should remove some of the simp attributes on definitions used in notation and have a lemma for each match case, then use these in an `aesop` ruleset.

(Regarding naming, no strong opinions yet.)

chenson2018: @fmontesi I have added an aesop ruleset. I think this is a great improvement! This is much closer to the ideal of treating notation as the canonical form with API built around it. I was able to remove places where I was using `simp` to destruct the opening, closing, and substitution notations, and also greatly simplified some of the proofs in `Properties.lean`.

Cslib/Computability/LambdaCalculus/Untyped/LocallyNameless/FullBeta.lean
```diff
@@ -0,0 +1,122 @@
...
+variable [HasFresh Var] [DecidableEq Var]
+
+/-- The right side of a reduction is locally closed. -/
+lemma step_lc_r (step : M ⭢βᶠ M') : LC M' := by
+  induction step
+  case beta => apply beta_lc <;> assumption
+  all_goals try constructor <;> assumption 
+
+/-- Substitution respects a single reduction step. -/
+lemma redex_subst_cong (s s' : Term Var) (x y : Var) : (s ⭢βᶠ s') -> (s [ x := fvar y ]) ⭢βᶠ (s' [ x := fvar y ]) := by
```
fmontesi: Ascii arrow instead of \to

fmontesi: Ascii arrow instead of \to

Cslib/Computability/LambdaCalculus/Untyped/LocallyNameless/FullBetaConfluence.lean
```diff
@@ -0,0 +1,249 @@
...
+    Parallel n n' → 
+    Parallel (app (abs m) n) (m' ^ n')
+
+-- TODO: I think this could be generated along with `para_rs`
+lemma para_rs_Red_eq {α} : (@paraRs α).Red = Parallel := by rfl
+
+variable {M M' N N' : Term Var}
+
+/-- The left side of a parallel reduction is locally closed. -/
+@[aesop unsafe]
```
fmontesi: Probably good to have a ruleset here as well.

Cslib/Computability/LambdaCalculus/Untyped/LocallyNameless/AesopRuleset.lean
```diff
@@ -0,0 +1,3 @@
+import Aesop
+
+declare_aesop_rule_sets [ln] (default := true)
```
fmontesi: Proposal for a proto-convention: qualify the ruleset name with the namespace of the locally nameless lambda calculus?

Cslib/Utils/Relation.lean
```diff
@@ -1,10 +1,51 @@
...
+  revert C
+  induction AB using Relation.ReflTransGen.head_induction_on <;> intros C BC
+  case refl => exists C
+  case head _ _ A'_C' _ ih =>
+    obtain ⟨D, ⟨CD, C'_D⟩⟩ := diamond_extend h BC A'_C'
+    obtain ⟨D', ⟨B_D', D_D'⟩⟩ := ih C'_D
+    exact ⟨D', ⟨B_D', trans CD D_D'⟩⟩
+
+/-- Equivalence of relations preserves the diamond property. -/
+theorem diamond_bisim (sim : ∀ {M N : α}, R M N ↔ R' M N) (h : Diamond R) : Diamond R' := by
```
fmontesi: More than bisim, isn't this relational equality? (Can't you prove R = R' from sim?)

chenson2018: Oh yeah, this is silly  to write in Lean... I can remove this entirely.

chenson2018: In case it's not clear why I wrote something so weird: imagine we were instead in Type or a theory without proof irrelevance. Then I would really need a witness.

Cslib/Computability/LambdaCalculus/Untyped/LocallyNameless/AesopRuleset.lean
chenson2018: I don't understand, what would it be here for instance?

fmontesi: Something like:

Computability.LambdaCalculus.Nameless.aesopRS

Too much of a mouthful? 😅

fmontesi: Or just ruleSet instead of aesopRS

chenson2018: With the way our namespaces are set up, we don't include `Computability`, so I think `LambdaCalculus.LocallyNameless.ruleSet` seems fine. Unfortunately I don't think the aesop attributes check namespaces so you have to write it out each time, but I think this is still best.

fmontesi (approved):

PR #12, opened 2025-07-18
Add proofs of exponential equivalences

m-ow:
Hi! Added two classic linear logic equivalences:  
- `!⊤ ≡ 1` (`exp_top_eq_one`)  
- `?0 ≡ ⊥` (`exp_zero_eq_bot`)  

Inspired by Girard's right-sided sequents, I made this equivalence definition:  
`Proof [a.dual, b] ∧ Proof [b.dual, a]`

Would love your thoughts on this and the perm proofs too before tackling distributive laws next! 😊

### DIFF
--- a/Cslib/Logic/LinearLogic/CLL/Basic.lean
+++ b/Cslib/Logic/LinearLogic/CLL/Basic.lean
@@ -150,6 +150,18 @@ inductive Proof : @Sequent Atom → Prop where
 
 def Proposition.equiv (a b : @Proposition Atom) : Prop := Proof [a.dual, b] ∧ Proof [b.dual, a]
 
+namespace Proposition
+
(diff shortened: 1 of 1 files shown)


fmontesi:
This makes sense to me, since what you defined as equivalence is essentially linear bi-implication. 👍

I'd define a locally scoped notation ≡ for equiv and use in your statements (but I'm merging this in the meantime, it can be done later). In the future we might even make it into a typeclass (LogicalEquiv?), so that we can use the notation consistently across different logics.

fmontesi:
I've added the notation and applied some docstrings and name fixes. Please have a look. :-)

Since we're parameterised over `Atom`, the notation requires at least one of the constructors of `Proof` to be used in `@`-form, otherwise the type checker can't figure out what implicit parameter to use.

PR #15, opened 2025-07-21
Add proofs for tensor-zero and parr-top equivalences

m-ow:
This PR adds two logical equivalences:
1. `a ⊗ 0 ≡ 0` (theorem `tensor_zero_eqv_zero`)
2. `a ⅋ ⊤ ≡ ⊤` (theorem `parr_top_eqv_top`)

It's a bit tedious to handle all these explicit `List.Perm` swaps in the equivalence proofs. Maybe we should create a separate file with a general API for moving stuff around in lists? That way we could have lemmas or even a tactic to make these permutation steps smoother. 

What do you think? 🙌

### DIFF
--- a/Cslib/Logic/LinearLogic/CLL/Basic.lean
+++ b/Cslib/Logic/LinearLogic/CLL/Basic.lean
@@ -189,6 +189,17 @@ theorem tensor_zero_eqv_zero (a : @Proposition Atom) :
     exact Proof.top
   · exact Proof.top
 
+/-- a ⅋ ⊤ ≡ ⊤ -/
+theorem parr_top_eqv_top (a : @Proposition Atom) :
+    parr a top ≡ top := by
+  constructor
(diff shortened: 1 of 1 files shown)


fmontesi (commented):
I've recently instantiated some typeclasses that give access to the notations for 0, 1, \top, and \bot, so I suggest we use them in the theorem statements. What do you think?

I haven't added any notations yet for parr, tensor, etc. If there are no typeclasses, we should either create them or make some scoped notation I guess.

m-ow:
You're absolutely right about the existing typeclasses! I've updated the theorem statements to use the notations for 0 and⊤

Regarding parr and tensor: I searched for existing typeclasses but couldn't find any. I'd be happy to create scoped notations for these operators in the meantime if you'd like.

fmontesi (approved):

PR #16, opened 2025-07-22
Switch to the standard type for relations.

fmontesi:
This PR creates a new `Cslib.Data.Relation` file with some utilities and a union notation for relations, merging the previous files `Cslib.Utils.Rel` and `Cslib.Utils.Relation` (now deleted).

It removes all dependencies of the recently-changed `Rel` type from mathlib, in preparation for bumping our mathlib dependency.

I'm a bit in doubt as to the name styling for definitions such as 'Relation.inv' and 'Relation.union'. In mathlib, they are uppercased for `Relation` but lowercased for `Rel`, I guess because for `Relation` one knows that the final type is `Prop`. It is however a bit weird.


Cslib/Data/Relation.lean
```diff
@@ -16,9 +40,9 @@ abbrev Diamond (R : α → α → Prop) := ∀ {A B C : α}, R A B → R A C →
 abbrev Confluence (R : α → α → Prop) := Diamond (Relation.ReflTransGen R)
 
 /-- Extending a multistep reduction by a single step preserves multi-joinability. -/
-lemma Relation.ReflTransGen.diamond_extend (h : Diamond R) : 
-  Relation.ReflTransGen R A B → 
-  R A C → 
+lemma Relation.ReflTransGen.diamond_extend (h : Diamond R) :
```
chenson2018, line 43: Are these identified by a linter? If so, can we enable this by default/include it in a workflow check?

Cslib/Data/Relation.lean
```diff
@@ -1,13 +1,37 @@
...
+  fun x y => r x y ∨ s x y
+
+instance {α : Type u} {β : Type v} : Union (α → β → Prop) where
+  union := Relation.union
 
-variable {α : Type u} {R R' : α → α → Prop}
+/-- Inverse of a relation. -/
+def Relation.inv (r : α → β → Prop) : β → α → Prop := flip r
+
+-- /-- Composition of two relations. -/
```
chenson2018, line 25: Commented out definition?

Cslib/Semantics/LTS/Simulation.lean
```diff
@@ -74,8 +74,8 @@ theorem Similarity.refl (s : State) : s ≤[lts] s := by
 
 /-- The composition of two simulations is a simulation. -/
 theorem Simulation.comp
-  (r1 r2 : Rel State State) (h1 : Simulation lts r1) (h2 : Simulation lts r2) :
-  Simulation lts (r1.comp r2) := by
+  (r1 r2 : State → State → Prop) (h1 : Simulation lts r1) (h2 : Simulation lts r2) :
+  Simulation lts (Relation.Comp r1 r2) := by
```
chenson2018, line 78: Does `Relation.Comp` have a notation? If so, that would be preferred here.

Cslib/Semantics/LTS/Simulation.lean
```diff
@@ -88,24 +88,24 @@ theorem Simulation.comp
...
 /-- Similarity is transitive. -/
 theorem Similarity.trans (h1 : s1 ≤[lts] s2) (h2 : s2 ≤[lts] s3) : s1 ≤[lts] s3 := by
   obtain ⟨r1, hr1, hr1s⟩ := h1
   obtain ⟨r2, hr2, hr2s⟩ := h2
-  exists r1.comp r2
+  exists Relation.Comp r1 r2
   constructor
   case left =>
-    simp only [Rel.comp]
+    simp only [Relation.Comp]
```
chenson2018, line 101: All uses of `simp only [Relation.Comp]` in this file are not needed.

Cslib/Semantics/LTS/Bisimulation.lean
```diff
@@ -175,7 +174,7 @@ theorem Bisimulation.comp
     exists s2''
     constructor
     · exact h2'tr
-    · simp [Rel.comp]
+    · simp only [Relation.Comp]
```
chenson2018, line 177: All uses of `simp only [Relation.Comp]` in this file are not needed.

Cslib/Semantics/LTS/Simulation.lean
chenson2018: Same question regarding the usage of `Relation.inv`.

Cslib/Data/Relation.lean
```diff
@@ -1,13 +1,37 @@
...
+universe u v
+
+section Relation
+
+/-- Union of two relations. -/
+def Relation.union (r s : α → β → Prop) : α → β → Prop :=
+  fun x y => r x y ∨ s x y
+
+instance {α : Type u} {β : Type v} : Union (α → β → Prop) where
+  union := Relation.union
```
eric-wieser, line 20: Mathlib already provides this with `⊔` notation

Cslib/Data/Relation.lean
```diff
@@ -1,13 +1,37 @@
...
+-- /-- Composition of two relations. -/
+-- def Relation.comp (r : α → β → Prop) (s : β → γ → Prop) : α → γ → Prop :=
+--   fun x z => ∃ y, r x y ∧ s y z
+
+/-- The relation `r` 'up to' the relation `s`. -/
+def Relation.upTo (r s : α → α → Prop) : α → α → Prop := Relation.Comp s (Relation.Comp r s)
+
+/-- The identity relation. -/
+inductive Relation.Id : α → α → Prop where
+| id {x : α} : Id x x
```
eric-wieser, line 34: This is `Eq`

fmontesi: Oh, that's a surprising name from the textbooks I'm used to. Thanks for pointing it out, I'd been naively looking for Id.

fmontesi: Ah, it's because I was looking for it in mathlib as a `Rel` a long time ago.. the `Eq` from Lean is of course what I want. Adopting it is even simplifying some proofs!

PR #17, opened 2025-07-23
Locally nameless STLC

chenson2018:
This PR adds simple typing derivations for locally nameless terms, with proofs of progress and preservation. 

I have also reversed the directory structure of `LambdaCalculus` to be the representation then the system, e.g. `{LocallyNameless,Named}/{Untyped,STLC}`. This matches the namespaces we have been using and better groups together related code.


Cslib/Computability/LambdaCalculus/LocallyNameless/STLC/Basic.lean
```diff
@@ -0,0 +1,152 @@
...
+-/
+
+universe u
+
+variable {Var : Type u} [DecidableEq Var]
+
+namespace LambdaCalculus.LocallyNameless.STLC
+
+/-- Types of the simply typed lambda calculus. -/
+inductive Ty
```
fmontesi: Can't we write `Type` under a namespace, or maybe `StlcType`?

Cslib/Computability/LambdaCalculus/LocallyNameless/STLC/Basic.lean
```diff
@@ -0,0 +1,152 @@
+/-
```
fmontesi: Seeing stlc again triggered my memory: in many other libs, people spell acronyms CamelCase instead of all uppercase. So `Stlc` instead of `STLC`. What do you think?
If we choose to go for `Stlc` here, I should probably rename `LTS` to `Lts`, I guess.

What do they do in mathlib?

Cslib/Computability/LambdaCalculus/LocallyNameless/STLC/Basic.lean
```diff
@@ -0,0 +1,152 @@
...
+universe u
+
+variable {Var : Type u} [DecidableEq Var]
+
+namespace LambdaCalculus.LocallyNameless.STLC
+
+/-- Types of the simply typed lambda calculus. -/
+inductive Ty
+/-- A base type, from a typing context. -/
+| base : Ty
```
fmontesi: This should take a `BaseType` as parameter. (And `BaseType` should be a parameter of this inductive.)

Cslib/Computability/LambdaCalculus/LocallyNameless/STLC/Basic.lean
```diff
@@ -0,0 +1,152 @@
...
+/-- A function type. -/
+| arrow : Ty → Ty → Ty
+
+scoped infixr:70 " ⤳ " => Ty.arrow
```
fmontesi: I'd use the normal arrow.

Cslib/Computability/LambdaCalculus/LocallyNameless/STLC/Basic.lean
```diff
@@ -0,0 +1,152 @@
...
+scoped prefix:90 "~ " => Ty.base
```
fmontesi: Where does this notation come from?

Cslib/Computability/LambdaCalculus/LocallyNameless/STLC/Basic.lean
```diff
@@ -0,0 +1,152 @@
...
+
+open Term Ty
+
+/-- An extrinsic typing derivation for locally nameless terms. -/
+@[aesop unsafe [constructors (rule_sets := [LambdaCalculus.LocallyNameless.ruleSet])]]
+inductive Typing : Ctx Var Ty → Term Var → Ty → Prop
```
fmontesi: We're typically using the name `Context` for syntactic contexts (see CCS), and I was planning to do the same in lambda.
We could call Ctx:
- TContext
- TypingContext
- Environment

Or we could just manage this with namespaces, I guess? Stlc.Typing.Context?

Cslib/Computability/LambdaCalculus/LocallyNameless/STLC/Context.lean
```diff
@@ -0,0 +1,60 @@
...
+abbrev Ctx (Var : Type u) (Ty : Type v) := List (Var × Ty)
+
+namespace Ctx
+
+/-- The domain of a context is the finite set of free variables it uses. -/
+@[simp]
+def dom : Ctx Var Ty → Finset Var := λ Γ ↦ Γ.map Prod.fst |>.toFinset
+
+/-- A well-formed context. -/
+inductive Ok : Ctx Var Ty → Prop
```
fmontesi: Ooh, we do this a lot in other projects, I'm gonna intro a class for well-formedness and get back.

Cslib/Computability/LambdaCalculus/LocallyNameless/STLC/Basic.lean
chenson2018: This is similarly namespaced already, the qualified name is `LambdaCalculus.LocallyNameless.STLC.Ty`. 

While technically allowed, I have a strong preference to still not use the name `Type` under a namespace. It is confusing as there are situations where it will be picked up as a token that is awkward with namespaces, causing us to write code that deviates from anywhere else in the library.

Cslib/Computability/LambdaCalculus/LocallyNameless/STLC/Basic.lean
chenson2018: Joint response to this and your other comment about a well-formedness class:

This is currently namespaced under `LambdaCalculus.LocallyNameless.STLC`. It would be odd to also place it under `Typing`, because that inductive depends on this, plus we'd be writing into a namespace before its defining declaration. I am fine with `Context` if you feel it is a better name? I gravitate to short names, but don't feel strongly. I don't want to include anything about "type" because that will be confusing for other usages.

This is a general construction as mentioned in your other comment. Though the choice of pairs is motivated by the locally nameless syntax, it has no dependency on anything in the lambda calculus directory except for adding it to the Aesop ruleset (which could be separated). It could live higher up if it seems like it can be reused. For instance, I foresee using it when I get to System F for both typing and kinding contexts.

This style of API around `List.Perm` is something I came up with myself because I found it simplified some proofs. I spent a bit thinking about the spelling of `Ctx.Ok` as well. I settled on using that inductive because it seemed simpler than saying that the domain had no duplicates. As long as you don't have dependent types, it seems pretty generally applicable.

Cslib/Computability/LambdaCalculus/LocallyNameless/STLC/Basic.lean
chenson2018: Do you have a suggestion for what the precedence would be here? I have a pretty strong preference for avoiding a clash with a core Lean notation, especially this one.

Cslib/Computability/LambdaCalculus/LocallyNameless/STLC/Basic.lean
chenson2018: You mean for base types? It's arbitrary, I don't recall where I picked it up from. It's not even used, I could remove for now.

Cslib/Computability/LambdaCalculus/LocallyNameless/STLC/Basic.lean
chenson2018: From the style guide:
> Acronyms like LE are written upper-/lowercase as a group, depending on what the first character would be.

So I think you're right that it should be `Stlc`, I have made this change.

Cslib/Computability/LambdaCalculus/LocallyNameless/STLC/Basic.lean
fmontesi: I've been spinning on this for a while but couldn't reach a good suggestion yet. I'll open a discussion on Zulip. :-)

fmontesi: Ok, see there and let me know what you think. The idea could be to use term elaborators.

chenson2018: I am onboard with term elaborators and delaborators (pretty printers); this is something I've been considering already. However, my preference would still be to start initially with just this unambiguous notation, as I believe this will be a bit of effort that should happen in a separate PR. 

Some of the questions I have echo our previous development of the attributes that define reduction notation. It seems what will be desired is not a single term elaborator, but a family of them. We developed the notation attribute to unify how we wrote reductions and provide a simple interface that is invoked when defining a reduction. What would this look like for term elaborators? I don't have a clear idea of this yet.

fmontesi: I think the elaborator design is gonna take some thinking and that it should have its own scope, too.

Let's go forward with this notation for now then. I'm gonna review the rest ignoring the 'notation proliferation' issue: we definitely shouldn't get stuck because of it.

I'm gonna open a separate issue for talking about notation and elaborators in general.

Cslib/Computability/LambdaCalculus/LocallyNameless/Stlc/Context.lean
```diff
@@ -0,0 +1,60 @@
...
+abbrev Ctx (Var : Type u) (Ty : Type v) := List (Var × Ty)
+
+namespace Ctx
+
+/-- The domain of a context is the finite set of free variables it uses. -/
+@[simp]
+def dom : Ctx Var Ty → Finset Var := List.toFinset ∘ List.map Prod.fst
+
+/-- A well-formed context. -/
+inductive Ok : Ctx Var Ty → Prop
```
fmontesi: This looks like `List.NodupKeys`.

Cslib/Computability/LambdaCalculus/LocallyNameless/Stlc/Context.lean
```diff
@@ -0,0 +1,60 @@
...
+namespace LambdaCalculus.LocallyNameless.Stlc
+
+/-- A typing context is a list of free variables and corresponding types. -/
+abbrev Ctx (Var : Type u) (Ty : Type v) := List (Var × Ty)
+
+namespace Ctx
+
+/-- The domain of a context is the finite set of free variables it uses. -/
+@[simp]
+def dom : Ctx Var Ty → Finset Var := List.toFinset ∘ List.map Prod.fst
```
fmontesi: You can use `List.keys.toFinset`. (Might require editing the Ctx type a bit.)

Cslib/Computability/LambdaCalculus/LocallyNameless/Stlc/Context.lean
```diff
@@ -0,0 +1,60 @@
...
+-/
+
+universe u v
+
+variable {Var : Type u} {Ty : Type v} [DecidableEq Var]
+
+namespace LambdaCalculus.LocallyNameless.Stlc
+
+/-- A typing context is a list of free variables and corresponding types. -/
+abbrev Ctx (Var : Type u) (Ty : Type v) := List (Var × Ty)
```
fmontesi: Is there a particular reason for which you're not using the stuff from https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/List/Sigma.html ?
In particular, you could use the defs about lists and NodupKeys or the bundled `AList` (but that can get annoying if well-formedness is anything ever more than nodupkeys).
I give a couple of examples of useful defs from there below.

Cslib/Computability/LambdaCalculus/LocallyNameless/Stlc/Context.lean
```diff
@@ -0,0 +1,60 @@
...
+variable {Γ Δ : Ctx Var Ty}
+
+/-- Context membership is preserved on permuting a context. -/
+theorem dom_perm_mem_iff (h : Γ.Perm Δ) {x : Var} : 
+    x ∈ Γ.dom ↔ x ∈ Δ.dom := by
+  induction h <;> aesop
+
+/-- Context well-formedness is preserved on permuting a context. -/
+@[aesop safe forward (rule_sets := [LambdaCalculus.LocallyNameless.ruleSet])]
+theorem perm (h : Γ.Perm Δ) : Ok Γ → Ok Δ := by
```
fmontesi: This is https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/List/Sigma.html#List.perm_nodupKeys

Cslib/Computability/LambdaCalculus/LocallyNameless/Stlc/Context.lean
chenson2018: I just didn't know it existed! This is what I meant when I said earlier "...it seemed simpler than saying that the domain had no duplicates", but since it already exists this seems good. I have updated to your (unbundled) suggestion.

Cslib/Computability/LambdaCalculus/LocallyNameless/Stlc/Context.lean
chenson2018: Now using this, but since `List.perm_nodupKeys` uses ↔, I still define a theorem for `aesop` purposes.

Cslib/Computability/LambdaCalculus/LocallyNameless/STLC/Context.lean
chenson2018: I have updated to use the notation you added in #22.

Cslib/Computability/LambdaCalculus/LocallyNameless/STLC/Basic.lean
chenson2018: I have changed to your last suggestion of `Context`.

chenson2018:
After this is merged, while we don't have any other PRs pending, I'm going to address all the outstanding lints so we can start with a clean slate. 

Could you take a look at the YAML for the lints workflow? I think there is an issue as I'm getting emails that say `.github/workflows/lean_lint_suggest.yml: no jobs were run` and it's not appearing here on the PR.

Cslib/Computability/LambdaCalculus/LocallyNameless/Stlc/Basic.lean
```diff
@@ -0,0 +1,165 @@
...
+import Cslib.Computability.LambdaCalculus.LocallyNameless.Stlc.Context
+import Cslib.Computability.LambdaCalculus.LocallyNameless.Untyped.Properties
+
+/-! # λ-calculus
+
+The simply typed λ-calculus, with a locally nameless representation of syntax.
+
+## References
+
+* [A. Chargueraud, *The Locally Nameless Representation*] [Chargueraud2012]
```
fmontesi: No space between the two: [..][..]. Otherwise the ref doesn't work.

Cslib/Computability/LambdaCalculus/LocallyNameless/Stlc/Basic.lean
```diff
@@ -0,0 +1,165 @@
...
+* See also https://www.cis.upenn.edu/~plclub/popl08-tutorial/code/, from which
```
fmontesi: Link should be in <>.

Cslib/Computability/LambdaCalculus/LocallyNameless/Stlc/Basic.lean
```diff
@@ -0,0 +1,165 @@
...
+universe u v
+
+variable {Var : Type u} {Base : Type v} [DecidableEq Var]
+
+namespace LambdaCalculus.LocallyNameless.Stlc
+
+/-- Types of the simply typed lambda calculus. -/
+inductive Ty (Base : Type v)
+/-- A base type, from a typing context. -/
+| base : Base → Ty Base
```
fmontesi: Cases of inductives should be intended. (I've recently found out and I've violated this rule myself.. :sweat_smile:)

Cslib/Computability/LambdaCalculus/LocallyNameless/Stlc/Basic.lean
```diff
@@ -0,0 +1,165 @@
...
+
+scoped infixr:70 " ⤳ " => Ty.arrow
+
+open Term Ty
+
+/-- An extrinsic typing derivation for locally nameless terms. -/
+@[aesop unsafe [constructors (rule_sets := [LambdaCalculus.LocallyNameless.ruleSet])]]
+inductive Typing : Context Var (Ty Base) → Term Var → Ty Base → Prop
+/-- Free variables, from a context judgement. -/
+| var : Γ ✓ → ⟨x,σ⟩ ∈ Γ → Typing Γ (fvar x) σ
```
fmontesi: I personally wouldn't put a space between \Gamma and the checkmark, but I'm perfectly ok either way.

Cslib/Computability/LambdaCalculus/LocallyNameless/Stlc/Basic.lean
```diff
@@ -0,0 +1,165 @@
...
+/-- Lambda abstraction. -/
+| abs (L : Finset Var) : (∀ x ∉ L, Typing (⟨x,σ⟩ :: Γ) (t ^ fvar x) τ) → Typing Γ t.abs (σ ⤳ τ)
```
fmontesi: What's `L`'s role here? It's not used anywhere outside of checking that `x` is not in it.

Cslib/Computability/LambdaCalculus/LocallyNameless/Stlc/Basic.lean
```diff
@@ -0,0 +1,165 @@
...
+/-- Lambda abstraction. -/
+| abs (L : Finset Var) : (∀ x ∉ L, Typing (⟨x,σ⟩ :: Γ) (t ^ fvar x) τ) → Typing Γ t.abs (σ ⤳ τ) 
+/-- Function application. -/
+| app : Typing Γ t (σ ⤳ τ) → Typing Γ t' σ → Typing Γ (app t t') τ
+
+scoped notation:50 Γ " ⊢ " t " ∶" τ:arg => Typing Γ t τ
+
+namespace Typing
+
+variable {Γ Δ Φ : Context Var (Ty Base)}
```
fmontesi: In papers people typically use \Theta as third context variable, not \Phi.

Cslib/Computability/LambdaCalculus/LocallyNameless/Stlc/Basic.lean
```diff
@@ -0,0 +1,165 @@
...
+
+omit [DecidableEq Var] in
+/-- Typing is preserved on permuting a context. -/
+theorem perm : Γ.Perm Δ → Γ ⊢ t ∶ τ → Δ ⊢ t ∶ τ := by
```
fmontesi: Is there a more specific name we could use than just `perm`? Maybe `perm_typing`?

I haven't read the rest yet and maybe I'm wrong here, but people typically phrase these preservation results by stating the typing hypothesis first, e.g.

```
theorem typing_perm : Γ ⊢ t ∶ τ → Γ.Perm Δ → Δ ⊢ t ∶ τ := by
```

Another note: in mathlib there seems to be a convention to name hypothesis. I've found it convenient before, since you typically start every proof by intro-ing them anyway. That'd give something like:

```
theorem typing_perm (ht : Γ ⊢ t ∶ τ) (hperm : Γ.Perm Δ) : Δ ⊢ t ∶ τ := by
```

Cslib/Computability/LambdaCalculus/LocallyNameless/Stlc/Basic.lean
```diff
@@ -0,0 +1,165 @@
...
+  case var => 
+    have := @p.mem_iff
+    aesop
+  case abs ih => 
+    constructor
+    intros x mem
+    exact ih x mem (by aesop)
+
+/-- Weakening of a typing derivation with an appended context. -/
+lemma weakening_strengthened :
```
fmontesi: Is it called `_strengthened` just because it has a stronger hypothesis and it is used to prove the normal weakening result?

If so, we could just call it `weakening'`, `weakening_aux`, or similar. Do you know what people do in mathlib for this kind of stuff perhaps?

(Apologies if `_strengthened` is actually what is commonly done in Lean, I just find 'weaken..._strengthened' a bit weird to read.)

Cslib/Computability/LambdaCalculus/LocallyNameless/Stlc/Basic.lean
```diff
@@ -0,0 +1,165 @@
...
+    rw [
+      subst_def, 
+      subst_open_var _ _ _ _ _ der.lc,
+      show ⟨x, σ⟩ :: (Δ ++ Γ) = (⟨x, σ⟩ :: Δ) ++ Γ by aesop
+      ]
+    apply ih 
+    all_goals aesop  
+
+/-- Substitution for a context weakened by a single type. -/
+lemma typing_subst :
```
fmontesi: `typing_subst` made me expect a more general result where <x, \sigma> can appear anywhere (\in) and then it is removed at the end. Maybe `typing_subst_head`?

Cslib/Computability/LambdaCalculus/LocallyNameless/Stlc/Basic.lean
chenson2018: Yes, this is correct. This is the "cofinite quantification" associated with a locally nameless style. The idea is that instead of quantifying over our free variable universally or existentially, we have something in between.

See section 4.2 of https://chargueraud.org/research/2009/ln/main.pdf

Cslib/Computability/LambdaCalculus/LocallyNameless/Stlc/Basic.lean
chenson2018: No, I think I picked it up from a Rocq proof. I agree that `_aux` is more aligned with typical Lean style.

Cslib/Computability/LambdaCalculus/LocallyNameless/Stlc/Basic.lean
chenson2018: This is namespaced as `Typing.perm`, does that seem fine?

I think you are correct on the other style changes, I will try to adhere to these going forward (and clean them up while doing my PR to fix all lint issues).

chenson2018:
Thanks for fixing the lint YAML file, seems to be running now. Looking at the output however, it is not catching the lint errors I see locally. I see [these messages on the latest run](https://github.com/cs-lean/cslib/actions/runs/16626869915/job/47045617794?pr=17#step:2:267):
```
warning: nolints file could not be read; treating as empty: scripts/nolints-style.txt
error: `print-style-error.sh` exited with code 255
could not execute external process './scripts/print-style-errors.sh'
```

Cslib/Computability/LambdaCalculus/LocallyNameless/Stlc/Basic.lean
chenson2018: More specifically, there is a bit more than just checking `x ∉ L`.  We have that adding this fresh `x` with some type `σ` to our context and then replacing the first bound variable with this free `x` always results in a typing `τ`.  Intuitively this is saying that when we substitute a `σ`, we get a `τ`, i.e. a function type.

fmontesi: Thanks for the interesting read!
After reading (admittedly a bit rapidly) the PDF I'm left in doubt as to whether `L` should be required to be nonempty. Shouldn't it, to prevent being able to type things with vacuous checks?

Cslib/Computability/LambdaCalculus/LocallyNameless/Stlc/Basic.lean
fmontesi: I think that could work. For some reason it looks clearer to me with the named parameters. That way it sounds like an operation that could be used with the dot notation: given a typing and a proof that its context can be permuted into another, then I get a typing with the other context.

fmontesi:
> Thanks for fixing the lint YAML file, seems to be running now. Looking at the output however, it is not catching the lint errors I see locally. I see [these messages on the latest run](https://github.com/cs-lean/cslib/actions/runs/16626869915/job/47045617794?pr=17#step:2:267):
> 
> ```
> warning: nolints file could not be read; treating as empty: scripts/nolints-style.txt
> error: `print-style-error.sh` exited with code 255
> could not execute external process './scripts/print-style-errors.sh'
> ```

Maybe it's got something to do with the version? I'm now using the same they use in mathlib. Otherwise, it might be that we should provide that script?..

Cslib/Computability/LambdaCalculus/LocallyNameless/Stlc/Basic.lean
chenson2018: I don't think it is correct to make this nonempty. There's not an example here for typing, but in other uses of cofinite quantification (for instance `FullBeta.abs`) we do use an empty set. Because it's `x ∉ L` (as opposed to `x ∈ L`) it is not vacuous. We are saying that it is a typing that works regardless of the freeness of `x`.  Imagine for instance that we are working with a function that returns a constant, so the input `x` is unused. It would be strange to in turn require `L` to be nonempty.

fmontesi: Of course you're right, I'd read it wrong! Thanks for the clarification. It should definitely not be required to be nonempty. Also because using the empty set recovers the universally quantified rule. :-)

Cslib/Computability/LambdaCalculus/LocallyNameless/Stlc/Basic.lean
```diff
@@ -0,0 +1,164 @@
...
+/-- Substitution for a context weakened by a single type. -/
+lemma typing_subst_head : 
+    ⟨x, σ⟩ :: Γ ⊢ t ∶ τ  → Γ ⊢ s ∶ σ → Γ ⊢ (t [x := s]) ∶ τ := by
+  intros weak der
+  rw [←List.nil_append Γ]
+  exact subst_aux weak der
+
+/-- Typing preservation for opening. -/
+@[aesop safe forward (rule_sets := [LambdaCalculus.LocallyNameless.ruleSet])]
+theorem preservation_opening {xs : Finset Var} :
```
fmontesi: `opening` should just be `open` (they usually use the name of the operation in theorem names iiuc).

This makes me think back to `Typing.perm` though: we're inconsistent, since this theorem should be simply called `open` then. But I honestly don't have strong feelings about that `perm`, so we can leave it as is.

Cslib/Computability/LambdaCalculus/LocallyNameless/Stlc/Context.lean
```diff
@@ -0,0 +1,57 @@
...
+
+/-- Context membership is preserved on permuting a context. -/
+theorem dom_perm_mem_iff (h : Γ.Perm Δ) {x : Var} : 
+    x ∈ Γ.dom ↔ x ∈ Δ.dom := by
+  induction h <;> aesop
+
+omit [DecidableEq Var] in
+/-- Context well-formedness is preserved on permuting a context. -/
+@[aesop safe forward (rule_sets := [LambdaCalculus.LocallyNameless.ruleSet])]
+theorem perm (h : Γ.Perm Δ) : Γ✓ → Δ✓ := (List.perm_nodupKeys h).mp
```
fmontesi: I'd call this `wf_perm`.

I'd also personally call `Ok` `Wf`, but I see it's called ok in the paper, so it looks ok (!) anyway.

Cslib/Computability/LambdaCalculus/LocallyNameless/Stlc/Context.lean
```diff
@@ -0,0 +1,57 @@
...
+
+omit [DecidableEq Var] in
+@[aesop safe forward (rule_sets := [LambdaCalculus.LocallyNameless.ruleSet])]
+theorem strengthen : (Δ ++ ⟨x, σ⟩ :: Γ)✓ → (Δ ++ Γ)✓ := by
```
fmontesi: I'd call this `wf_strengthen`.

Also, maybe we should call `weakening` `weaken`, to be alike?

Cslib/Computability/LambdaCalculus/LocallyNameless/Stlc/Safety.lean
```diff
@@ -0,0 +1,92 @@
...
+
+universe u v
+
+namespace LambdaCalculus.LocallyNameless
+
+open Stlc Typing
+
+variable {Var : Type u} {Base : Type v} {R : Term Var → Term Var → Prop}
+
+def preserves (R : Term Var → Term Var → Prop) (Base : Type v) :=
```
fmontesi: Should probably be `Preserves`, or even better `PreservesTyping`.

fmontesi: Note for later: this might be worth generalising. I gotta think about it though, and it's orthogonal to this PR.

Cslib/Computability/LambdaCalculus/LocallyNameless/Stlc/Safety.lean
```diff
@@ -0,0 +1,92 @@
...
+open Stlc Typing
+
+variable {Var : Type u} {Base : Type v} {R : Term Var → Term Var → Prop}
+
+def preserves (R : Term Var → Term Var → Prop) (Base : Type v) := 
+  ∀ {Γ t t'} {τ : Ty Base}, Γ ⊢ t ∶ τ → R t t' → Γ ⊢ t' ∶ τ
+
+/-- If a reduction preserves types, so does its reflexive transitive closure. -/
+@[aesop safe forward (rule_sets := [LambdaCalculus.LocallyNameless.ruleSet])]
+theorem preservation_redex : preserves R Base → preserves (Relation.ReflTransGen R) Base := by
```
fmontesi: `redex_preservesTyping`

Cslib/Computability/LambdaCalculus/LocallyNameless/Stlc/Safety.lean
```diff
@@ -0,0 +1,92 @@
...
+  intros _ _ _ _ _ _ redex
+  induction redex <;> aesop
+
+open Relation in
+/-- Confluence preserves type preservation. -/
+theorem preservation_confluence {τ : Ty Base} :
```
fmontesi: `confluence_preservesTyping`

Cslib/Computability/LambdaCalculus/LocallyNameless/Stlc/Safety.lean
```diff
@@ -0,0 +1,92 @@
...
+  intros der
+  revert t'
+  induction der <;> intros t' step <;> cases step
+  case' abs.abs xs _ _ _ xs' _=> apply Typing.abs (xs ∪ xs')
+  case' app.beta der_l _ _ => cases der_l
+  all_goals aesop
+
+omit [HasFresh Var] [DecidableEq Var] in
+/-- A typed term either full beta reduces or is a value. -/
+theorem progress : ([] : Context Var (Ty Base)) ⊢ t ∶ τ → t.Value ∨ ∃ t', t ⭢βᶠ t' := by
```
fmontesi: In general here and above I'd name hyps as they do in mathlib. But that might just be taste.

fmontesi: Rest looks good. Great work! :-)

fmontesi (approved):

PR #22, opened 2025-07-27
Class for well-formedness.

fmontesi:


chenson2018 (approved):
Seems straightforward as a notation typeclass.

PR #25, opened 2025-08-02
use notation3 for Lts and ReductionSystem

chenson2018:
This is in reference to https://github.com/cs-lean/cslib/issues/23#issuecomment-3131138166. It turns out that Mathlib already supports deriving a delaborator for notations declared with `notation3`. This is even better than an unexpander, because it works with expressions rather than syntax and by default works with uses of `variable`. The end result is that any use of `{Lts,ReductionSystem}.{Red,MRed}` is printed as their corresponding arrow in the infoview. 

I also:
- cleaned up a few small details about whitespace in the notations.
- moved the precedence markers to the arguments so that you can for instance write `¬ a ↠β b` without parenthesis

### DIFF
--- a/Cslib/Semantics/Lts/Basic.lean
+++ b/Cslib/Semantics/Lts/Basic.lean
@@ -654,13 +654,13 @@ syntax "lts_transition_notation" ident (str)? : command
 macro_rules
   | `(lts_transition_notation $lts $sym) =>
     `(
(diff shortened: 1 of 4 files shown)


fmontesi (approved):

fmontesi:
Neat!
It says in the mathlib docs that notation3 has an uncertain future, so we should keep an eye on that. But for now this does improve our lives. :-)

PR #27, opened 2025-08-05
add wfail option to CI build

chenson2018:
This solves half of the linter issue, promoting syntax linter failures to an error in CI builds.


chenson2018:
@fmontesi I am done testing here, this is working for identifying syntax lints. Please note that this does not check syntax lints in the `CslibTests` directory, as the GitHub action does not have a configuration to pass arguments to `lake test` like it does for `lake build`. I have asked in https://github.com/leanprover/lean-action/issues/134 if this would be a welcome feature.

fmontesi (approved):

PR #29, opened 2025-08-14
feat(HasFresh): characterize

tristan-f-r:
Associated lemmas to `HasFresh` which characterize its behavior.


fmontesi (approved):
Very nice!

PR #32, opened 2025-08-14
term elaborator for selecting fresh variables

### LINKED ISSUE #30: tactic for selecting fresh variables
I had an idea for a tactic that, for a type that implements `HasFresh`, selects a fresh variable with respect to the local context, i.e. the hypotheses to a theorem. This doesn't seem too difficult. I wrote a term elaborator [here](https://github.com/chenson2018/cslib/blob/23f42587de1d80ffd0d8e6d0b3a969269af1b50a/Cslib/Data/HasFresh.lean#L39) that constructs the union of any variables and Finsets in the local context. There is an example usage below the definition. I imagine this would be wrapped in a tactic `fresh_select` in its final form instead of using the term elab directly.

I would like to add the ability to also pass a function for computing free variables, e.g. something like `fv : Term → Finset Var`, with the union also including `fv t` for any `t : Term` in the local context.

chenson2018:
Closes #30 

This PR adds a term elaborator that constructs a union independent of any variables, variable Finset, and optionally terms in the local context.


chenson2018:
@fmontesi I think this turned out well! I felt a tactic was overkill, there is just the term elaborator. Check out the proofs I have updated here, I think it shows how this cleans things up a bit.

fmontesi:
Woah, this looks very convenient!

The naming and doc of the elaborator seem a bit off though, probably because of design evolution. `fresh_union` actually has little to do with invoking `fresh` or the like. It's more like the union of all free variables of type `α`, or `Finset α`, etc. (Maybe we could even extend this in the future.)

Maybe one of `var_union`, `union_var`, `fv_union`, `free_union`, or `union_free`?

For the `free` parameter, do you expect to add other parameters? Because if not, we could just use square brackets, like `var_union [myFunction, foo, @fv Term] Nat`.
Sidenote: I found `free` a bit surprising as a name (thought it wanted the name of a variable or something like that, not a function).

fmontesi:
Ah, and an example in the docstring of the elaborator would probably be really good to have. Trying an example in the editor made immediately evident what the elab does for me, at least. :)

chenson2018:
@fmontesi Glad you think it looks useful! 

You are right about the naming being off because of the development history. I have gone with your `free_union` suggestion and added an example in the docs. In fact, this doesn't directly depend on `HasFresh`, but I do think it makes sense to have them in the same file. (Though if you gave a compelling argument to move to its own file, I wouldn't object too much)

Regarding the parameters, I at one point had a paramater `var` where you specified the variable type. Intuitively you should only need to specify one of `free` and `var`, as `var` can be inferred from the image of `free`. This worked in simple cases, but I hit a snag when used with `variable`, and thus the current design where it is a required argument. So I'd like to keep this door open, and I think it is nice in general to work with `optConfig` for extensibility. However, in your example you specify several functions for free variables. This could still be supported in the future if we find it is needed.

Regarding the option name `free`, I am open to suggestions. Though now that I use the name `free_union`, it does match up somewhat.

fmontesi:
Excellent, thanks!

The doc is still a bit weird, cutting a part at the beginning seems to make it more intuitive. From

```
Given a `DecidableEq Var` instance, this elaborator automatically constructs a term that is free with respect to variables in the local context. It creates a union of any variables, finite sets of variables, and optionally the results of a provided function for free variables.
```

to

```
Given a `DecidableEq Var` instance, this elaborator automatically constructs the union of any variables, finite sets of variables, and optionally the results of a provided function for free variables.
```

The last part saying `free variables` might not be very intuitive. It makes perfect sense in lambda-calculus, but here we're talking about Lean variables. So what does the function do, in general? It 'obtains' variables? It 'deconstructs' a (Lean) value to obtain variables? It 'intros' variables? That might help us figure out a more specific name for the `free` parameter.

(Or am I wrong, and free variables makes sense? :-))

I also think it should be in `HasFresh` for now, we can always move it later if it becomes relevant.

chenson2018:
I have adopted your docs change, I agree it sounds better.

What specifically happens when you pass `(free := free : α → Finset β)` is that the elaborator finds all `aᵢ : α` in the local context, and includes `free aᵢ : Finset β` in the union it is constructing (i.e. all it does is make a function application). If a better name stands out from that description, I am completely open to it.

Cslib/Data/HasFresh.lean
```diff
@@ -30,6 +30,86 @@ in proofs. -/
...
+      if (← isDefEq local_type var) then
+        let singleton := 
+          mkAppN 
+          (mkConst ``Singleton.singleton [dl, dl]) 
+          #[var, FinsetType, SingletonInst, ldecl.toExpr]
+        finsets := finsets.push singleton
+      else
+      -- free variables of terms
+      if (←isDefEq local_type free_dom) then
+        finsets := finsets.push (mkApp free ldecl.toExpr)
```
chenson2018: To be specific, this is the code that works with the `free` option. (Plus some elaboration at the very beginning)

fmontesi:
> I have adopted your docs change, I agree it sounds better.
> 
> What specifically happens when you pass `(free := free : α → Finset β)` is that the elaborator finds all `aᵢ : α` in the local context, and includes `free aᵢ : Finset β` in the union it is constructing (i.e. all it does is make a function application). If a better name stands out from that description, I am completely open to it.

Mmh, what about `map`, or `fmap`?

chenson2018:
`map` sounds good to me, I have pushed this change.

By the way, I'm finding this works well with `grind`! Casting the widest net for free variables helps for inferring some proofs completely.

fmontesi (approved):

fmontesi:
Very good to hear about grind, I'm starting to understand a bit more how to use it myself but I haven't had time to delve seriously into it yet. :⁠-⁠)

PR #34, opened 2025-08-16
feat(LinearLogic): logical equivalence is an equivalence relation

tristan-f-r:
This also adds some notation and removes some hampering implicits.


Cslib/Logic/LinearLogic/CLL/Basic.lean
```diff
@@ -22,46 +21,49 @@ import Mathlib.Order.Notation
...
-  bot := Proposition.bot
+@[inherit_doc] scoped infix:35 " ⊗ " => Proposition.tensor
+@[inherit_doc] scoped infix:35 " ⊕ " => Proposition.oplus
+@[inherit_doc] scoped infix:30 " ⅋ " => Proposition.parr
+@[inherit_doc] scoped infix:30 " & " => Proposition.with
 
-scoped infix:35 " ⊗ " => Proposition.tensor
-scoped infix:30 " ⅋ " => Proposition.parr
+@[inherit_doc] scoped prefix:95 "!" => Proposition.bang
+@[inherit_doc] scoped prefix:95 "¿" => Proposition.quest
```
tristan-f-r: I don't like this, but `?` is mostly reserved. Do we just want to use `.quest`?

Cslib/Logic/LinearLogic/CLL/Basic.lean
```diff
@@ -104,68 +102,55 @@ def Proposition.dual (a : @Proposition Atom) : @Proposition Atom :=
...
-abbrev Sequent := List (@Proposition Atom)
+abbrev Sequent (Atom) := List (Proposition Atom)
 
 /-- Checks that all propositions in `Γ` are question marks. -/
-def Sequent.allQuest (Γ : @Sequent Atom) :=
+def Sequent.allQuest (Γ : Sequent Atom) :=
   ∀ a ∈ Γ, ∃ b, a = Proposition.quest b
 
 open Proposition in
 /-- Sequent calculus for CLL. -/
```
tristan-f-r, line 137: General question, but is there a nice way to formalize the sequent calculus in such a way that we can take advantage of proof irrelevance? I don't like the idea of how future theorems will be proven for linear logic with the current formalization, and I hate the idea of using quotients. This might be something to ask about in Zulip.

Cslib/Logic/LinearLogic/CLL/Basic.lean
fmontesi: Right.
We could also use the small question mark (﹖) if it works, or the glottal stop symbol maybe ʔ.

The inverted question mark has another meaning (https://dl.acm.org/doi/10.1145/3473567), so I'd rather not use it here.

fmontesi:
Very nice, thanks!

Side question: Have you had a look at the `Provable` formulation in the cll branch that we're discussing on zulip perhaps?
In particular, I'm interested in gathering thoughts on its impact on proofs/theorems.

tristan-f-r:
I haven't! I'll take a look :+1:

tristan-f-r:
@fmontesi I'll play around with the `Provable` definition a little bit as well on a new branch based on this branch to see how helpful this is.

[(You can also try defining `Sequent` over `Finset`)](https://github.com/tristan-f-r/cslib/compare/more-ll...tristan-f-r:cslib:ll-provable)

fmontesi (approved):

fmontesi:
A sequent should be a multiset, since propositions can be repeated. I was thinking of defining `MSequent` and `MProof`, based on `Multiset`, and then prove that `MProof` and `Proof` are equivalent. Then we could add a `Provable.mproof` theorem that uses that characterisation to allow showing things using `MProof`, whenever it's more convenient.

tristan-f-r:
The `MSequent` and `MProof` pattern sounds nice.

Showing that `Sequent` and `Proof` are equivalent to some `FSequent` and `FProof` directly does still seem fine, as you also get the `contract` rule to not break equality, instead of just `exchange`. (While `Finset` is defined as the `Multiset` without duplicates, nearly all `Finset` operations automatically deduplicate Finsets, so you are still able to naturally write equalities like `({1, 1, 2} : Finset ℕ) = {1, 2}`).

fmontesi:
I've drafted the approach in the `cll-mproof` branch and done a few cases, looks doable.

Re `Finset`: contract applies only to `?`-propositions, how would we control that?

tristan-f-r:
We can't! That was a misunderstanding on my part.

PR #37, opened 2025-08-19
chore(Data): golf entire `FinFun.eq_char₁`. golf using `grind`. remove commented out code.

euprunin:


chenson2018:
I think it is good to simplify the proofs here. A lot more here can be made compact with `grind`. See https://github.com/leanprover/cslib/compare/main...chenson2018:cslib:finfun-grind for what I would further add. You are welcome to add that commit to your PR. I only used `grind` locally because I wasn't sure what should be exposed outside this file.

chenson2018:
While we're touching this file, can we also remove the commented out code? https://github.com/leanprover/cslib/blob/dd8c1e07e57a1c05180e63ead5dce19964fe9ada/Cslib/Data/FinFun.lean#L75-L80

euprunin:
@chenson2018 Done! Thanks for reviewing.

lakefile.toml
```diff
@@ -18,3 +18,8 @@ globs = "Cslib.*"
 [[lean_lib]]
 name = "CslibTests"
 globs = "CslibTests.+"
+
+[[require]]
+name = "tryAtEachStep"
+git = "https://github.com/dwrensha/tryAtEachStep"
+rev = "lean-v4.22.0"
```
chenson2018: Am I correct that this proof search tool is how you found some of the other simplifications? Very cool, but let's not add this as a dependency.

euprunin: Oh, sorry. I accidentally added those lines by doing `git commit -a` instead of `git commit`. But yes, it is very helpful when golfing.

chenson2018:
@fmontesi I don't have permission to approve, but this looks good to me.

fmontesi:
Agreed! Thanks for the nice simplifications!

PR #41, opened 2025-08-20
chore(Semantics/Lts): golf `Lts.deterministic_image_char`, `Lts.strN.trans_τ`, `Lts.strN.comp` and `Lts.divergent_drop` using `grind`, `aesop` and `simp_all`

euprunin:


Cslib/Semantics/Lts/Basic.lean
```diff
@@ -286,32 +286,11 @@ theorem Lts.deterministic_image_char (hDet : lts.Deterministic) :
...
     simp [Image]
-    simp [setOf]
-    simp [EmptyCollection.emptyCollection]
-    funext s''
-    by_contra hf
-    simp at hf
-    simp at hs'
-    specialize hs' s''
-    contradiction
+    simp_all
```
chenson2018: This could be written in one line as
```lean4
by_cases hs' : ∃ s', lts.Tr s μ s' <;> aesop (add simp [Image])
```

chenson2018:
There are several non-terminal `simp` in this file, but we can address that in a later PR.

Cslib/Semantics/Lts/Basic.lean
chenson2018: @fmontesi Do you have any thoughts about mixed use of `grind` and `aesop`? My loosely held opinion is that `grind` should be the default, but it is okay to use `aesop` for certain niches it still fills. It is still useful for case analysis in conjunction with `simp` and certain forward reasoning not supported by `grind`.

euprunin: @chenson2018 Thanks! I've now added.

euprunin: In general, I think `aesop` (which is relatively costly) should be avoided when another tactic can close the goal, but if it’s the only single-tactic option to do so, then using it is fine. Does that sound like a reasonable policy?

fmontesi:
Yeah, that's basically what I think.

But sometimes aesop makes things really slow. In those cases, I like to
paste what aesop? prints and put a comment right above the code like

```
-- Generated by: aesop (add safe cases Foo)
```

That way, when things break, I know where to start from.

And no, I don't have a precise measure for 'too slow'. But 5 seconds is
definitely too slow. :⁠-⁠)

Grind seems to be quite quick most of the time.

fmontesi (approved):

PR #43, opened 2025-08-22
chore(Semantics/Lts): golf `Bisimilarity.refl` and `Bisimulation.simulation_iff` using `grind` and `aesop`

euprunin:


Cslib/Semantics/Lts/Bisimulation.lean
```diff
@@ -867,15 +859,7 @@ theorem Bisimulation.simulation_iff (lts : Lts State Label) (r : State → State
...
-    intro s1 s2 hr μ
-    constructor
-    case left =>
-      intro s1' htr
-      simp only [Simulation] at hs
-      apply hs _ _ hr _ _ htr
-    case right =>
-      intro s2' htr
-      apply hsinv _ _ hr _ _ htr
+    aesop
```
chenson2018: I would prefer that we simplify proofs at least at the level of granularity of entire theorems. Here is what I came up with:
```lean
theorem Bisimulation.simulation_iff (lts : Lts State Label) (r : State → State → Prop) :
  Bisimulation lts r ↔ (Simulation lts r ∧ Simulation lts (flip r)) := by
  constructor
  case mp => grind [Bisimulation, Simulation, flip]
  case mpr => aesop (add simp [Bisimulation])
```

Cslib/Semantics/Lts/Bisimulation.lean
```diff
@@ -116,15 +116,7 @@ theorem Bisimilarity.refl (s : State) : s ~[lts] s := by
...
-    intro s1 s2 hr μ
-    cases hr
-    constructor
-    case left =>
-      intro s1' htr
-      exists s1'
-    case right =>
-      intro s1' htr
-      exists s1'
+    grind
```
chenson2018: I would make this:
```lean
theorem Bisimilarity.refl (s : State) : s ~[lts] s := by
  exists Eq
  grind [Bisimulation]
```

fmontesi: Agreed!

Cslib/Semantics/Lts/Bisimulation.lean
fmontesi: This also looks like a good proposal.

fmontesi (changes requested):

Cslib/Semantics/Lts/Bisimulation.lean
euprunin: Looks great! Thanks for reviewing. Patch applied.

Cslib/Semantics/Lts/Bisimulation.lean
euprunin: Looks great! Thanks for reviewing. Patch applied.

fmontesi (approved):
Very nice, thank you!

PR #46, opened 2025-08-24
Allow multiple maps in `free_union`

### LINKED ISSUE #44: Allow multiple maps in `free_union`
In #32 where I added the `free_union` term elaborator, @fmontesi suggested that we could take a bracketed list of map arguments. I'd like to implement this feature (currently only one map is supported) and switch to this suggested syntax. My motivation for the syntax change is both readability and the fact that I think my usage of `optConfig` was a bit hacky, as I don't believe it is meant to be used for terms to be elaborated.

This may also simplify the implementation a bit, as we currently have special cases for picking variables and finite sets from the local context. (These really do belong in `optConfig`, as they are simple booleans.)

This is of practical use for the upcoming locally nameless formalization of System F with subtyping, as there are different free variable functions on terms and types.

chenson2018:
This closes #44. As I suspected, it also simplifies the implementation, as singleton variables and finsets in the local context are just special cases. I also like your suggested syntax better.

You can't really see it at work here much yet (besides the testing) but this is very useful for polymorphism, where you can now concisely select a variable that is free from any type or term variables.


fmontesi:
Very nice, thank you!

PR #48, opened 2025-08-26
fix: `LambdaCalculus.Named.Term.subst` scoping problem

thelissimus:
## Counterexamples

Old:

```lean
def t : Term ℕ := .abs 1 (.var 0)
#eval t.subst 0 (.app (.app (.var 1) (.var 2)) (.var 3)) -- abs 2 (app (app 1 2) 3)
```

New:

```lean
def t : Term ℕ := .abs 1 (.var 0)
#eval t.subst 0 (.app (.app (.var 1) (.var 2)) (.var 3)) -- abs 4 (app (app 1 2) 3)
```

### DIFF
--- a/CslibTests/LambdaCalculus.lean
+++ b/CslibTests/LambdaCalculus.lean
@@ -17,4 +17,8 @@ example : (abs 0 (var 0)) =α (abs 1 (var 1)) := by
   constructor
   simp [Term.fv]
 
+example : (abs 1 (var 0)).subst 0 (app (var 1) (var 2)) = (abs 3 (app (var 1) (var 2))) := by
+  simp [Term.subst, Term.fv, Term.bv, Term.vars, Term.rename]
+  apply rfl
+
 -- example : (abs 0 (abs 1 (app (var 0) (var 1)))) =α (abs 1 (abs 0 (app (var 1) (var 0)))) := by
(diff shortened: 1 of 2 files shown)


CslibTests/LambdaCalculus.lean
```diff
@@ -17,4 +17,20 @@ example : (abs 0 (var 0)) =α (abs 1 (var 1)) := by
...
+def z := 2
+def w := 3
+
+attribute [simp] x y z w
+
+local instance coeNatTerm : Coe ℕ (Term ℕ) := ⟨Term.var⟩
+
+-- section 5.3.4 of TAPL
+example : (abs y (app x y))[x := (app y z : Term ℕ)] = (abs w (app (app y z) w)) := by 
+  simp [subst, fv, bv, vars, rename, instHasFreshNat, HasFresh.ofSucc, instHasSubstitutionTerm]
```
chenson2018, line 34: I added an additional test here.

chenson2018 (approved):
Thank you for catching this! As you can see this directory isn't quite as far as `LambdaCalculus.LocallyNameless`, so without many proofs it is easy for such things to slip by. Any further contributions would be welcome.

@fmontesi Can you confirm this also looks correct to you before I merge?

fmontesi (approved):
Lgtm too, thanks!

PR #49, opened 2025-08-26
CLL equivalences and substitution of equivalent formulas

jpyamamoto:
I added some theorems to the CLL namespace regarding equivalences and substitution of equivalent formulas.

First PR to this library (and first contribution to the Lean language), so any feedback is welcome and appreciated.

I would love to keep contributing to the Logic module, but I'm not sure where to go from here. I was thinking about trying to tackle the cut-elimination theorem, but I believe that's beyond my current abilities. If you think of any other theorems that might be useful to have in the library, please let me know.

### DIFF
--- a/Cslib/Logic/LinearLogic/CLL/Basic.lean
+++ b/Cslib/Logic/LinearLogic/CLL/Basic.lean
@@ -256,17 +256,21 @@ theorem tensor_distrib_oplus (a b c : Proposition Atom) :
       · apply Proof.oplus₂
         exact Proof.ax
 
-theorem subst_eqv_left {Γ : Sequent Atom} {a b : Proposition Atom} (heqv : a ≡ b) :
+/-- The proposition at the head of a proof can be substituted by an equivalent
+  proposition. -/
+theorem subst_eqv_head {Γ : Sequent Atom} {a b : Proposition Atom} (heqv : a ≡ b) :
   ⊢(a :: Γ) → ⊢(b :: Γ) :=
   fun h => Proof.exchange (List.perm_append_singleton b Γ) (Proof.cut h heqv.left)
 
+/-- Any proposition in a proof (regardless of its position) can be substituted by
+  an equivalent proposition. -/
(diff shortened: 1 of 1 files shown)


Cslib/Logic/LinearLogic/CLL/Basic.lean
```diff
@@ -221,6 +228,90 @@ theorem parr_top_eqv_top (a : Proposition Atom) :
...
+        exact Proof.ax
+    · apply Proof.parr
+      apply Proof.exchange
+        (List.Perm.trans (List.Perm.swap ..) (List.Perm.cons _ (List.Perm.swap ..)))
+      apply Proof.tensor (Γ := [a.dual])
+      · exact Proof.ax
+      · apply Proof.oplus₂
+        exact Proof.ax
+
+theorem subst_eqv_left {Γ : Sequent Atom} {a b : Proposition Atom} (heqv : a ≡ b) :
```
fmontesi: Could you add a docstring? Also, it could be called subst_eqv_head or something like that. (I'm fine with left as well though.. any thoughts, @m-ow and @chenson2018 ?)

Cslib/Logic/LinearLogic/CLL/Basic.lean
```diff
@@ -221,6 +228,90 @@ theorem parr_top_eqv_top (a : Proposition Atom) :
...
+  ⊢(a :: Γ) → ⊢(b :: Γ) :=
+  fun h => Proof.exchange (List.perm_append_singleton b Γ) (Proof.cut h heqv.left)
+
+theorem subst_eqv {Γ Δ : Sequent Atom} {a b : Proposition Atom} (heqv : a ≡ b) :
```
fmontesi, line 267: Very nice!! Please add a docstring here as well.

jpyamamoto:
@fmontesi Added docstrings to the substitution theorems and renamed `subst_eqv_left` to `subst_eqv_head`. I agree that it's a better name.

fmontesi:
Very nice, thank you! :+1:

PR #52, opened 2025-08-28
add `weak.linter.flexible = true` to lakefile.toml

chenson2018:
This PR enables Mathlib's flexible linter, a generalized version of checking for non-terminal `simp`. As part of this PR I have removed every non-terminal `simp` detected by the linter.

There is one minor category of  false positive I encountered, which is when `<;> simp` is followed by the rest of a proof. In this case the linter can be suppressed locally or you can slightly reorder a proof. I counted only five occurrences (and some of these didn't need `simp` anyway). I think that the proof stability is worth this uncommon and minor inconvenience.

Side note: `Bisimulation.traceEq_not_bisim` was painful to edit performance-wise! Vim consistently froze on me while trying to use `simp?` in the middle of it.


fmontesi (approved):

fmontesi:
Very nice! 👏

PR #57, opened 2025-09-05
Rice's theorem

thomaskwaring:
This pull request extends the development of SKI combinatory logic — adding notions of evaluation and normal forms, and proving a version of [Rice's theorem](https://en.wikipedia.org/wiki/Rice%27s_theorem). The results and definitions are contained in the new file `CombinatoryLogic/Evaluation.lean`, which is based heavily on the [following development](https://gist.github.com/b-mehta/e412c837818223b8f16ca0b4aa19b166) by Bhavik Mehta.

### Notes

The definitions and results of this PR allows us to define normal-order evaluation as a `PFun`, in [the sense of Mathlib](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/PFun.html). The missing piece to make this useful is the so-called _Standardisation theorem_, saying that normal-order reduction will find a normal form, if there is one. The proof of Theorem T9 of Section C in _A Mathematical Logic Without Variables_ (Rosser, 1935) should be amenable to formalization. This would be very interesting, as it ought to define a surjection from SKI terms to, for instance, [partial recursive functions](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Computability/Partrec.html) on the natural numbers, demonstrating that SKI terms are Turing-complete.

### DIFF
--- a/Cslib/Computability/CombinatoryLogic/Basic.lean
+++ b/Cslib/Computability/CombinatoryLogic/Basic.lean
@@ -61,7 +61,7 @@ def Polynomial.eval {n : Nat} (Γ : SKI.Polynomial n) (l : List SKI) (hl : List.
 def Polynomial.varFreeToSKI (Γ : SKI.Polynomial 0) : SKI := Γ.eval [] (by trivial)
 
 /-- Inductively define a polynomial `Γ'` so that (up to the fact that we haven't
-defined reduction on polynomials) `Γ' ⬝ t ⇒* Γ[xₙ ← t]`. -/
+defined reduction on polynomials) `Γ' ⬝ t ↠ Γ[xₙ ← t]`. -/
 def Polynomial.elimVar {n : Nat} : SKI.Polynomial (n+1) → SKI.Polynomial n
   /- The K-combinator leaves plain terms unchanged by substitution `K ⬝ x ⬝ t ⇒ x` -/
   | SKI.Polynomial.term x => K ⬝' x
@@ -83,7 +83,7 @@ for the inner variables.
 -/
 theorem Polynomial.elimVar_correct {n : Nat} (Γ : SKI.Polynomial (n+1)) {ys : List SKI}
     (hys : ys.length = n) (z : SKI) :
-    Γ.elimVar.eval ys hys ⬝ z ⇒* Γ.eval (ys ++ [z])
+    Γ.elimVar.eval ys hys ⬝ z ↠ Γ.eval (ys ++ [z])
       (by rw [List.length_append, hys, List.length_singleton])
     := by
   match n, Γ with
@@ -125,13 +125,13 @@ def Polynomial.toSKI {n : Nat} (Γ : SKI.Polynomial n) : SKI :=
 
 /-- Correctness for the toSKI (bracket abstraction) algorithm. -/
 theorem Polynomial.toSKI_correct {n : Nat} (Γ : SKI.Polynomial n) (xs : List SKI)
-    (hxs : xs.length = n) : Γ.toSKI.applyList xs ⇒* Γ.eval xs hxs := by
+    (hxs : xs.length = n) : Γ.toSKI.applyList xs ↠ Γ.eval xs hxs := by
   match n with
   | 0 =>
     unfold toSKI varFreeToSKI applyList
     rw [List.length_eq_zero_iff] at hxs
     simp_rw [hxs, List.foldl_nil]
-    apply MRed.refl
+    rfl
   | n+1 =>
     -- show that xs = ys + [z]
     have : xs ≠ [] := List.ne_nil_of_length_eq_add_one hxs
@@ -165,7 +165,7 @@ choose a descriptive name.
 def RPoly : SKI.Polynomial 2 := &1 ⬝' &0
 /-- A SKI term representing R -/
 def R : SKI := RPoly.toSKI
-theorem R_def (x y : SKI) : R ⬝ x ⬝ y ⇒* y ⬝ x :=
+theorem R_def (x y : SKI) : R ⬝ x ⬝ y ↠ y ⬝ x :=
   RPoly.toSKI_correct [x, y] (by simp)
 
 
@@ -173,7 +173,7 @@ theorem R_def (x y : SKI) : R ⬝ x ⬝ y ⇒* y ⬝ x :=
 def BPoly : SKI.Polynomial 3 := &0 ⬝' (&1 ⬝' &2)
 /-- A SKI term representing B -/
 def B : SKI := BPoly.toSKI
-theorem B_def (f g x : SKI) : B ⬝ f ⬝ g ⬝ x ⇒* f ⬝ (g ⬝ x) :=
+theorem B_def (f g x : SKI) : B ⬝ f ⬝ g ⬝ x ↠ f ⬝ (g ⬝ x) :=
   BPoly.toSKI_correct [f, g, x] (by simp)
 
 
@@ -181,7 +181,7 @@ theorem B_def (f g x : SKI) : B ⬝ f ⬝ g ⬝ x ⇒* f ⬝ (g ⬝ x) :=
 def CPoly : SKI.Polynomial 3 := &0 ⬝' &2 ⬝' &1
 /-- A SKI term representing C -/
 def C : SKI := CPoly.toSKI
-theorem C_def (f x y : SKI) : C ⬝ f ⬝ x ⬝ y ⇒* f ⬝ y ⬝ x :=
+theorem C_def (f x y : SKI) : C ⬝ f ⬝ x ⬝ y ↠ f ⬝ y ⬝ x :=
   CPoly.toSKI_correct [f, x, y] (by simp)
 
 
@@ -189,7 +189,7 @@ theorem C_def (f x y : SKI) : C ⬝ f ⬝ x ⬝ y ⇒* f ⬝ y ⬝ x :=
 def RotRPoly : SKI.Polynomial 3 := &2 ⬝' &0 ⬝' &1
 /-- A SKI term representing RotR -/
 def RotR : SKI := RotRPoly.toSKI
-theorem rotR_def (x y z : SKI) : RotR ⬝ x ⬝ y ⬝ z ⇒* z ⬝ x ⬝ y :=
+theorem rotR_def (x y z : SKI) : RotR ⬝ x ⬝ y ⬝ z ↠ z ⬝ x ⬝ y :=
   RotRPoly.toSKI_correct [x, y, z] (by simp)
 
 
@@ -197,7 +197,7 @@ theorem rotR_def (x y z : SKI) : RotR ⬝ x ⬝ y ⬝ z ⇒* z ⬝ x ⬝ y :=
 def RotLPoly : SKI.Polynomial 3 := &1 ⬝' &2 ⬝' &0
 /-- A SKI term representing RotL -/
 def RotL : SKI := RotLPoly.toSKI
-theorem rotL_def (x y z : SKI) : RotL ⬝ x ⬝ y ⬝ z ⇒* y ⬝ z ⬝ x :=
+theorem rotL_def (x y z : SKI) : RotL ⬝ x ⬝ y ⬝ z ↠ y ⬝ z ⬝ x :=
   RotLPoly.toSKI_correct [x, y, z] (by simp)
 
 
@@ -205,7 +205,7 @@ theorem rotL_def (x y z : SKI) : RotL ⬝ x ⬝ y ⬝ z ⇒* y ⬝ z ⬝ x :=
 def δPoly : SKI.Polynomial 1 := &0 ⬝' &0
 /-- A SKI term representing δ -/
 def δ : SKI := δPoly.toSKI
-theorem δ_def (x : SKI) : δ ⬝ x ⇒* x ⬝ x :=
+theorem δ_def (x : SKI) : δ ⬝ x ↠ x ⬝ x :=
   δPoly.toSKI_correct [x] (by simp)
 
 
@@ -213,7 +213,7 @@ theorem δ_def (x : SKI) : δ ⬝ x ⇒* x ⬝ x :=
 def HPoly : SKI.Polynomial 2 := &0 ⬝' (&1 ⬝' &1)
 /-- A SKI term representing H -/
 def H : SKI := HPoly.toSKI
-theorem H_def (f x : SKI) : H ⬝ f ⬝ x ⇒* f ⬝ (x ⬝ x) :=
+theorem H_def (f x : SKI) : H ⬝ f ⬝ x ↠ f ⬝ (x ⬝ x) :=
   HPoly.toSKI_correct [f, x] (by simp)
 
 
@@ -221,7 +221,7 @@ theorem H_def (f x : SKI) : H ⬝ f ⬝ x ⇒* f ⬝ (x ⬝ x) :=
 def YPoly : SKI.Polynomial 1 := H ⬝' &0 ⬝' (H ⬝' &0)
 /-- A SKI term representing Y -/
 def Y : SKI := YPoly.toSKI
-theorem Y_def (f : SKI) : Y ⬝ f ⇒* H ⬝ f ⬝ (H ⬝ f) :=
+theorem Y_def (f : SKI) : Y ⬝ f ↠ H ⬝ f ⬝ (H ⬝ f) :=
   YPoly.toSKI_correct [f] (by simp)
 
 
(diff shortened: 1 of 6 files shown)


fmontesi:
There seem to be conflicts?

thomaskwaring:
oops my mistake! i've fixed the conflicts but it doesn't want to build remotely (it goes through on my computer) i think because of the linter warnings — i'll come back to this soon but it seems to want no space after the transition arrows (ie `(P ⬝ Abs) ↠TT` instead of `(P ⬝ Abs) ↠ TT`) which seems strange

chenson2018:
> it seems to want no space after the transition arrows (ie `(P ⬝ Abs) ↠TT` instead of `(P ⬝ Abs) ↠ TT`) which seems strange

Depending on how your terminal displays certain Unicode, what appears as roughly a single space after an arrow may actually be multiple spaces

thomaskwaring:
Hmm I don't think that's it, it's definitely asking for no spaces. It's only complaining about hypotheses of the form "blank reduces to blank" (eg `hxz : x ↠z`), not for instance `hxt : ∃ x : SKI, (P ⬝ x) ↠ TT`, and not anywhere other than as hypotheses to a theorem. Anyway not a huge deal, just strange.

I expect it's because in the declaration
```notation3 t:39 " ⭢" $sym:str t':39 => (ReductionSystem.Red  $rs) t t'```
there's no space after the arrow, was that for any reason?

**Comment by chenson2018** [comment]
> I expect it's because in the declaration `notation3 t:39 " ⭢" $sym:str t':39 => (ReductionSystem.Red $rs) t t'` there's no space after the arrow, was that for any reason?

For this branch with a symbol after the arrow, I think this makes sense. I see though that for the notation without a symbol it is also like this, is that what is being linted for you? This should be changed, probably not noticed previously because we don't use it much.

**Comment by thomaskwaring** [comment]
> For this branch with a symbol after the arrow, I think this makes sense. I see though that for the notation without a symbol it is also like this, is that what is being linted for you? This should be changed, probably not noticed previously because we don't use it much.

Ahh yes you're absolutely right — I've changed it and replaced the spaces after the transitions now

File: Cslib/Languages/CombinatoryLogic/Evaluation.lean [hunk 1]
```diff
@@ -0,0 +1,312 @@
...
+
+## References
+
+This file draws heavily from <https://gist.github.com/b-mehta/e412c837818223b8f16ca0b4aa19b166>
+-/
+
+open SKI Red
+
+/-- The predicate that a term has no reducible sub-terms. -/
+def RedexFree : SKI → Prop
```
fmontesi, line 43: Maybe `SKI.RedexFree`, to enable dot notation?

Cslib/Languages/CombinatoryLogic/Evaluation.lean
```diff
@@ -0,0 +1,312 @@
...
+  | S ⬝ _ ⬝ _ ⬝ _ => False
+  | K ⬝ _ ⬝ _ ⬝ _ => False
+  | I ⬝ _ ⬝ _ ⬝ _ => False
+  | a ⬝ b ⬝ c ⬝ d ⬝ e => RedexFree (a ⬝ b ⬝ c ⬝ d) ∧ RedexFree e
+
+/--
+One-step evaluation as a function: either it returns a term that has been reduced by one step,
+or a proof that the term is redex free. Uses normal-order reduction.
+-/
+def evalStep : (x : SKI) → PLift (RedexFree x) ⊕ SKI
```
fmontesi: `SKI.evalStep`, and similar every time it makes sense to have dot notation.

Cslib/Languages/CombinatoryLogic/Evaluation.lean
```diff
@@ -0,0 +1,324 @@
...
+
+/--
+If `x` reduces to both `y` and `z`, and both `y` and `z` are in normal form, then they are equal.
+-/
+lemma unique_normal_form {x y z : SKI}
+    (hxy : x ↠ y) (hxz : x ↠ z) (hy : y.RedexFree) (hz : RedexFree z) : y = z :=
+  (redexFree_iff'.1 hy _).1 (confluent_redexFree hxy hxz hz)
+
+/-- If `x` and `y` are normal and have a common reduct, then they are equal. -/
+lemma unique_normal_form' {x y : SKI} (h : CommonReduct x y)
```
chenson2018: Is there a more descriptive name we could use here and below instead of a primed name?

thomaskwaring: good point, thanks

fmontesi:
Hi @thomaskwaring,
I think this is cooked enough now and would like to merge it, but I still get some weird diffs from old commits. Could I ask you to do a rebase, perhaps?

thomaskwaring:
> Hi @thomaskwaring, I think this is cooked enough now and would like to merge it, but I still get some weird diffs from old commits. Could I ask you to do a rebase, perhaps?

Thanks! I'm doing that now — sorry for the noise I'm not much up on my git-yoga

docs/lake-manifest.json
fmontesi, line 1: Why this change, and in the next lake-manifest.json as well?

Otherwise this looks ready to me. 👍

thomaskwaring: that is strange, probably some git chaos when i was learning how to rebase — i'm endeavouring to fix it now, thanks

thomaskwaring: hmm i've matched up those changes and it builds on my machine, but it would appear not remotely — any idea why that might be?

fmontesi: Could it have something to do with the introduction of the `Cslib` namespace?

thomaskwaring: ahh good catch, should be good to go now

chenson2018: I am still seeing the manifest in the file diff.

thomaskwaring: oh strange, thanks for fixing

chenson2018: No problem. For reference, what I did was run `git checkout main -- lake-manifest.json` to make sure I got exactly the file from the main branch.

chenson2018:
I find it mildly remarkable that there isn't a merge conflict here. Because @Shreyas4991 consistently added the namespace at the very top and bottom of each file, it's able to be merged automatically.

Cslib/Foundations/Semantics/ReductionSystem/Basic.lean
```diff
@@ -46,11 +46,15 @@ open Relation Relation.ReflTransGen
...
+instance (rs : ReductionSystem Term) : Transitive rs.MRed := transitive_of_trans rs.MRed
+instance (rs : ReductionSystem Term) : Trans rs.MRed rs.MRed rs.MRed := instTransOfIsTrans
+
 end MultiStep
 
 open Lean Elab Meta Command Term
 
--- thank you to Kyle Miller for this: 
+-- thank you to Kyle Miller for this:
 -- https://leanprover.zulipchat.com/#narrow/channel/239415-metaprogramming-.2F-tactics/topic/Working.20with.20variables.20in.20a.20command/near/529324084
```
Shreyas4991, line 58: I recommend using a single multi line comment and a proper markdown formatted link, in case this ever becomes part of a docstring in the future. I am also not sure it is a great idea to put reference links in comments to Zulip discussions. I understand the urge to cite but maybe the better to do this is in the GitHub pull request page as a comment.

Cslib/Foundations/Semantics/ReductionSystem/Basic.lean
```diff
@@ -94,19 +98,19 @@ elab "create_reduction_sys" rel:ident name:ident : command => do
...
 syntax "reduction_notation" ident (str)? : command
 macro_rules
-  | `(reduction_notation $rs $sym) => 
+  | `(reduction_notation $rs $sym) =>
     `(
       notation3 t:39 " ⭢" $sym:str t':39 => (ReductionSystem.Red  $rs) t t'
       notation3 t:39 " ↠" $sym:str t':39 => (ReductionSystem.MRed $rs) t t'
      )
-  | `(reduction_notation $rs) => 
+  | `(reduction_notation $rs) =>
```
Shreyas4991, line 106: I recommend generating and using scoped notation here. These symbols are used very often in various areas of CS. I also recommend using an enclosing namespace like SKI for all files in this folder which occurs just below the Cslib namespace.

Cslib/Languages/CombinatoryLogic/Basic.lean
```diff
@@ -165,63 +165,63 @@ choose a descriptive name.
 def RPoly : SKI.Polynomial 2 := &1 ⬝' &0
 /-- A SKI term representing R -/
 def R : SKI := RPoly.toSKI
-theorem R_def (x y : SKI) : R ⬝ x ⬝ y ⇒* y ⬝ x :=
+theorem R_def (x y : SKI) : (R ⬝ x ⬝ y) ↠ y ⬝ x :=
```
Shreyas4991, line 170: This bracket shouldn't be necessary. This could possibly be fixed by adapting the notation priority of the notation it is replacing. But I could be wrong. Worth checking it up.

Cslib/Languages/CombinatoryLogic/Basic.lean
```diff
@@ -239,29 +239,29 @@ rather than up to a common reduct. An alternative is to use Turing's fixed-point
 (defined below).
 -/
 def fixedPoint (f : SKI) : SKI := H ⬝ f ⬝ (H ⬝ f)
-theorem fixedPoint_correct (f : SKI) : f.fixedPoint ⇒* f ⬝ f.fixedPoint := H_def f (H ⬝ f)
+theorem fixedPoint_correct (f : SKI) : f.fixedPoint ↠ f ⬝ f.fixedPoint := H_def f (H ⬝ f)
 
 /-- Auxiliary definition for Turing's fixed-point combinator: ΘAux := λ x y. y (x x y) -/
 def ΘAuxPoly : SKI.Polynomial 2 := &1 ⬝' (&0 ⬝' &0 ⬝' &1)
```
Shreyas4991: I recommend against using Unicode characters in identifiers. This makes them hard to search on the docs page which doesn't Unicode input in the search bar. This is also true for the Ctrl + F and Ctrl + Shift + F find and replace  interfaces of VScode.

Cslib/Languages/CombinatoryLogic/Evaluation.lean
```diff
@@ -0,0 +1,328 @@
...
+
+lemma sk_nequiv : ¬ CommonReduct S K := by
+  intro ⟨z, hsz, hkz⟩
+  have hS : RedexFree S := by simp [RedexFree]
+  have hK : RedexFree K := by simp [RedexFree]
+  cases (redexFree_iff_mred_eq.1 hS z).1 hsz
+  cases (redexFree_iff_mred_eq.1 hK _).1 hkz
+
+/-- Injectivity for booleans. -/
+theorem isBool_injective (x y : SKI) (u v : Bool) (hx : IsBool u x) (hy : IsBool v y)
```
Shreyas4991, line 219: When you are referring to a definition `IsBool`, the theorem of this should also ideally be named `IsBool_injective`. Again this is beneficial in search.

Cslib/Languages/CombinatoryLogic/Evaluation.lean
```diff
@@ -0,0 +1,328 @@
...
+@[simp]
+lemma churchK_size : (n : Nat) → (churchK n).size = n+1
+  | 0 => rfl
+  | n + 1 => by rw [churchK, size, size, churchK_size, Nat.add_comm]
+
+lemma churchK_injective : Function.Injective churchK :=
+  fun n m h => by simpa using congrArg SKI.size h
+
+/-- Injectivity for Church numerals -/
+theorem isChurch_injective (x y : SKI) (n m : Nat) (hx : IsChurch n x) (hy : IsChurch m y)
```
Shreyas4991, line 271: Same comment as `isBool_injective`

Shreyas4991:
Some of my review comments might need discussion on Zulip and collective decisions. I recommend merging in the main branch into this branch, before merging this PR because namespace related issues for identifiers could still bite. 

EDIT: I highly recommend an 'SKI` or `CombinatoryLogic` namespace under the `Cslib` namespace to enable scoping all notation here.

Cslib/Languages/CombinatoryLogic/Evaluation.lean
thomaskwaring: On this and others I was following [mathlib conventions](https://leanprover-community.github.io/contribute/naming.html), see point 5 of general conventions "When something named with `UpperCamelCase` is part of something named with `snake_case`, it is referenced in `lowerCamelCase`.

Shreyas4991: Ok fair

Cslib/Foundations/Semantics/ReductionSystem/Basic.lean
chenson2018: The formatting suggestion is fine, but I disagree about not linking to Zulip discussions. Both Mathlib (and even core in tests) have links to the Zulip. Is there a particular harm you have in mind of including links that outweighs the ease of discoverability?

Cslib/Foundations/Semantics/ReductionSystem/Basic.lean
chenson2018: Yes, this is probably a good idea. There was a concern previously about if the attribute were used outside a namespace.

thomaskwaring: the diffs here are bc of my git errors — perhaps these edits outside of the `CombinatoryLogic/*` section could be a separate PR?

Cslib/Languages/CombinatoryLogic/Basic.lean
thomaskwaring: I agree, but the notation is used elsewhere so perhaps this is something to be fixed in a broader PR after this is merged?

Cslib/Languages/CombinatoryLogic/Basic.lean
thomaskwaring: good call, thanks! fixed now

chenson2018:
> EDIT: I highly recommend an 'SKI`or`CombinatoryLogic`namespace under the`Cslib` namespace to enable scoping all notation here.

There is an `SKI` namespace already, right? But yes, a pass to scope notations would be good.

Cslib/Foundations/Semantics/ReductionSystem/Basic.lean
Shreyas4991: It has been discouraged in the past. Also if you do wish to make citation style comments, it might help to put it as a note in the module docstring. A comment in the middle of the code is harder to discover.

Cslib/Languages/CombinatoryLogic/Basic.lean
Shreyas4991: If this notation is scoped then it should only be used in other places where the SKI namespace is opened. Surely this is fixable then?

thomaskwaring:
> There is an `SKI` namespace already, right?

Yes there is, but I was a bit inconsistent with it — fixed now & added the `Cslib` namespace to my branch

Cslib/Languages/CombinatoryLogic/Basic.lean
thomaskwaring: the notation `↠` is generated by the meta code in `Cslib/Foundations/Semantics/ReductionSystem/Basic`, which is used elsewhere

Cslib/Foundations/Semantics/ReductionSystem/Basic.lean
chenson2018: Yes, don't let any of this discussion outside the scope of your PR hold it up.

Cslib/Languages/CombinatoryLogic/Basic.lean
chenson2018: Yes, changing the precedence of notations for `ReductionSystem` and `LTS` has been on my TODO list. I just opened #109 to track, but as you say let's do this in a follow-up PR.

Cslib/Languages/CombinatoryLogic/Defs.lean
```diff
@@ -72,74 +82,75 @@ inductive Red : SKI → SKI → Prop where
...
 
-instance MRedRedTrans : Trans MRed Red MRed :=
-  ⟨fun hab hbc => Relation.ReflTransGen.trans hab (MRed.single hbc)⟩
+-- instance MRedRedTrans : Trans MRed Red MRed :=
+--   ⟨fun hab hbc => Relation.ReflTransGen.trans hab (MRed.single hbc)⟩
 
-instance RedMRedTrans : Trans Red Red MRed :=
-  ⟨fun hab hbc => Relation.ReflTransGen.trans (MRed.single hab) (MRed.single hbc)⟩
+-- instance RedMRedTrans : Trans Red Red MRed :=
+--   ⟨fun hab hbc => Relation.ReflTransGen.trans (MRed.single hab) (MRed.single hbc)⟩
```
chenson2018: Can you remove these commented out instances?

Cslib/Foundations/Semantics/ReductionSystem/Basic.lean
chenson2018: It's not really just a citation, I linked to the thread because it explains why there's complexity with the metaprogramming. I don't want it in the proper documentation, but I personally like having it next to the relevant elab.

Regarding Zulip links generally, I would need to see the argument against to respond properly. Let's continue any further discussion on Zulip, this was already out of scope for this PR (it only appears in the diff because of a whitespace fix)

Cslib/Languages/CombinatoryLogic/Defs.lean
thomaskwaring: done!

thomaskwaring:
gentle ping @fmontesi, I think I've addressed everything (correct me?) if you think this is merge-ready?

PR #58, opened 2025-09-10
Boole examples of quicksort and PRM

andrewmw94:
Implementations of quicksort and a simplified PRM in a Boogie-like language.

### DIFF
--- a/Cslib/Foundations/Data/Relation.lean
+++ b/Cslib/Foundations/Data/Relation.lean
@@ -13,7 +13,7 @@ universe u v
 section Relation
 
 /-- The relation `r` 'up to' the relation `s`. -/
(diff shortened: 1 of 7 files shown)


arademaker:
> Strata and Mathlib use different Lean versions, so we can't import Strata currently.

Considering this comment, this code does not fully integrate into the repository, nor can it be compiled. I would prefer not to accept this PR at this time. Keeping the PR open is an option, but over time it will likely become difficult to manage pending reviews. What do you think @fmontesi ?

barrettcw:
This PR is going to the Boole-sandbox branch.  Andrew - I checked and you should have permission to merge it directly.  No need for reviews on the Boole-sandbox branch.

arademaker:
So maybe we just need to adjust the place of the files. The other examples are in `Cslib/Languages/Boole/examples`!

barrettcw:
Ah - right - please put the files in the examples directory

PR #68, opened 2025-09-24
Locally nameless Fsub

chenson2018:
This PR adds a formalization of type safety for System F with subtyping, let bindings, and sums.


chenson2018:
I think I'm done bikeshedding which context lemmas are strictly needed, sorry for the noise. Still ready for review.

fmontesi:
Long but easy to read, thanks! :-)

PR #83, opened 2025-10-06
feat (Computability/Automata): Automata theory based on LTS, including DFA, NFA, EpsilonNFA, and translations between them, plus supporting theorems generalised to LTS

fmontesi:


Cslib/Computability/Automata/DA.lean
```diff
@@ -0,0 +1,19 @@
...
+Authors: Fabrizio Montesi
+-/
+
+import Cslib.Computability.Automata.NA
+
+/-! # Deterministic Automaton
+
+A Deterministic Automaton (DA) is an automaton defined by a transition function.
+-/
+structure DA (State : Type _) (Symbol : Type _) where
```
chenson2018, line 13:
```suggestion
structure DA (State Symbol : Type*) where
```

Cslib/Computability/Automata/DFA.lean
```diff
@@ -0,0 +1,59 @@
...
+
+## References
+
+* [J. E. Hopcroft, R. Motwani, J. D. Ullman, *Introduction to Automata Theory, Languages, and Computation*][Hopcroft2006]
+-/
+
+/-- A Deterministic Finite Automaton (DFA) consists of a labelled transition function
+`tr` over finite sets of states and labels (called symbols), a starting state `start` and a finite
+set of accepting states `accept`. -/
+structure DFA (State : Type _) (Symbol : Type _) extends DA State Symbol where
```
chenson2018, line 23:
```suggestion
structure DFA (State Symbol : Type*) extends DA State Symbol where
```

Cslib/Computability/Automata/DFA.lean
```diff
@@ -0,0 +1,59 @@
...
+def mtr (dfa : DFA State Symbol) (s : State) (xs : List Symbol) := xs.foldl dfa.tr s
+
+@[grind]
+theorem mtr_nil_eq {dfa : DFA State Symbol} : dfa.mtr s [] = s := by rfl
+
+/-- A DFA accepts a string if there is a multi-step accepting derivative with that trace from
+the start state. -/
+@[grind]
+def Accepts (dfa : DFA State Symbol) (xs : List Symbol) :=
+  (dfa.mtr dfa.start xs) ∈ dfa.accept
```
chenson2018:
```suggestion
  dfa.mtr dfa.start xs ∈ dfa.accept
```

Cslib/Computability/Automata/DFA.lean
```diff
@@ -0,0 +1,59 @@
...
+
+
+namespace DFA
+
+/-- Extended transition function.
+
+Implementation note: compared to [Hopcroft2006], the definition consumes the input list of symbols
+from the left (instead of the right), in order to match the way lists are constructed.
+-/
+@[grind]
```
chenson2018: Though these are simple `grind` annotations, I prefer explicitly marking them as `grind =`. I also have been scoping everything, as this also seems more explicit. (I'm not sure ultimately how much a difference it makes for performance).

Cslib/Computability/Automata/DFAToNFA.lean
```diff
@@ -0,0 +1,75 @@
...
+
+/-- `DFA` is a special case of `NFA`. -/
+@[grind =]
+def toNFA (dfa : DFA State Symbol) : NFA State Symbol := {
+  start := {dfa.start}
+  accept := dfa.accept
+  Tr := fun s1 μ s2 => dfa.tr s1 μ = s2
+  finite_state := dfa.finite_state
+  finite_symbol := dfa.finite_symbol
+}
```
chenson2018: I think using `where` for structure definitions is more readable, especially with code using `Set`.

```suggestion
def toNFA (dfa : DFA State Symbol) : NFA State Symbol where
  start := {dfa.start}
  accept := dfa.accept
  Tr := fun s1 μ s2 => dfa.tr s1 μ = s2
  finite_state := dfa.finite_state
  finite_symbol := dfa.finite_symbol
```

Cslib/Computability/Automata/EpsilonNFA.lean
```diff
@@ -0,0 +1,50 @@
...
+
+/-! # Nondeterministic automata with ε-transitions. -/
+
+/-- A nondeterministic finite automaton with ε-transitions (`εNFA`) is an `NA` with an `Option`
+symbol type. The special symbol ε is represented by the `Option.none` case.
+
+Internally, ε (`Option.none`) is treated as the `τ` label of the underlying transition system,
+allowing for reusing the definitions and results on saturated transitions of `LTS` to deal with
+ε-closure. -/
+structure εNFA (State : Type _) (Symbol : Type _) extends NA State (Option Symbol) where
```
chenson2018, line 17:
```suggestion
structure εNFA (State Symbol : Type*) extends NA State (Option Symbol) where
```

Cslib/Computability/Automata/EpsilonNFAToNFA.lean
```diff
@@ -0,0 +1,87 @@
...
+-/
+
+import Cslib.Computability.Automata.EpsilonNFA
+import Cslib.Computability.Automata.NFA
+
+/-! # Translation of εNFA into NFA -/
+
+/-- Converts an `LTS` with Option labels into an `LTS` on the carried label type, by removing all
+ε-transitions. -/
+@[grind]
```
chenson2018, line 14: Based on the fact that you are having to use `simp`  and `rw` wedged between `grind` is an indicator that something is preventing the annotations in this file from being useful.

Cslib/Computability/Automata/EpsilonNFAToNFA.lean
```diff
@@ -0,0 +1,87 @@
...
+private def LTS.noε (lts : LTS State (Option Label)) :
+  LTS State Label := {
+  Tr := fun s μ s' => lts.Tr s (some μ) s'
+}
```
chenson2018:
```suggestion
private def LTS.noε (lts : LTS State (Option Label)) : LTS State Label where
  Tr := fun s μ s' => lts.Tr s (some μ) s'
```

Cslib/Computability/Automata/EpsilonNFAToNFA.lean
```diff
@@ -0,0 +1,87 @@
...
+
+/-- Any εNFA can be converted into an NFA that does not use ε-transitions. -/
+@[grind]
+def toNFA (enfa : εNFA State Symbol) : NFA State Symbol := {
+  start := enfa.εClosure enfa.start
+  accept := enfa.accept
+  Tr := enfa.saturate.noε.Tr
+  finite_state := enfa.finite_state
+  finite_symbol := enfa.finite_symbol
+}
```
chenson2018:
```suggestion
def toNFA (enfa : εNFA State Symbol) : NFA State Symbol where
  start := enfa.εClosure enfa.start
  accept := enfa.accept
  Tr := enfa.saturate.noε.Tr
  finite_state := enfa.finite_state
  finite_symbol := enfa.finite_symbol
```

Cslib/Computability/Automata/NA.lean
```diff
@@ -0,0 +1,21 @@
...
+
+/-! # Nondeterministic Automaton
+
+A Nondeterministic Automaton (NA) is an automaton with a set of initial states.
+
+This definition extends `LTS` and thus stores the transition system as a relation `Tr` of the form
+`State → Symbol → State → Prop`. Note that you can also instantiate `Tr` by passing an argument of
+type `State → Symbol → Set State`; it gets automatically expanded to the former shape.
+-/
+structure NA (State : Type _) (Symbol : Type _) extends LTS State Symbol where
```
chenson2018, line 17:
```suggestion
structure NA (State Symbol : Type*) extends LTS State Symbol where
```

Cslib/Computability/Automata/NFA.lean
```diff
@@ -0,0 +1,35 @@
...
+Copyright (c) 2025 Fabrizio Montesi. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Fabrizio Montesi
+-/
+
+import Cslib.Computability.Automata.NA
+
+/-- A Nondeterministic Finite Automaton (NFA) is a nondeterministic automaton (NA)
+over finite sets of states and symbols. -/
+structure NFA (State : Type _) (Symbol : Type _) extends NA State Symbol where
```
chenson2018, line 11:
```suggestion
structure NFA (State Symbol : Type*) extends NA State Symbol where
```

Cslib/Computability/Automata/NFAToDFA.lean
```diff
@@ -0,0 +1,64 @@
...
+@[grind]
+theorem toDA_mtr_setImageMultistep {nfa : NFA State Symbol} :
+  nfa.toDFA.mtr = nfa.setImageMultistep := by grind
+
+/-- The `DFA` constructed from an `NFA` has the same language. -/
+@[grind]
+theorem toDA_language_eq {nfa : NFA State Symbol} :
+  nfa.toDFA.language = nfa.language := by
+  ext xs
+  rw [← DFA.accepts_mem_language, ← NFA.accepts_mem_language]
```
chenson2018: One of these works with `grind`
```suggestion
  rw [← DFA.accepts_mem_language]
```

Cslib/Computability/Automata/NFAToDFA.lean
```diff
@@ -0,0 +1,64 @@
...
+section SubsetConstruction
+
+/-- Converts an `NFA` into a `DFA` using the subset construction. -/
+@[grind]
+def toDFA (nfa : NFA State Symbol) : DFA (Set State) Symbol := {
+  start := nfa.start
+  accept := { S | ∃ s ∈ S, s ∈ nfa.accept }
+  tr := nfa.setImage
+  finite_state := by
+    haveI := nfa.finite_state
```
chenson2018, line 23: Shouldn't we be able to infer this by putting an instance on `NFA`? Also I don't think `have` vs `haveI` makes a difference here, does it?

Cslib/Computability/Automata/NFAToDFA.lean
```diff
@@ -0,0 +1,64 @@
...
+    apply Set.instFinite
```
chenson2018: Maybe more idiomatic?
```suggestion
    infer_instance
```

Cslib/Computability/Automata/EpsilonNFAToNFA.lean
```diff
@@ -0,0 +1,87 @@
...
+  case cons μ μs ih =>
+    apply Iff.intro <;> intro h
+    case mp =>
+      simp only [LTS.noε]
+      cases h
+      rename_i sb htr hmtr
+      apply LTS.MTr.stepL (s1 := s) (s2 := sb) (s3 := s') <;> grind
+    case mpr =>
+      simp only [LTS.noε] at h
+      grind
```
chenson2018:
```suggestion
  induction μs generalizing s <;> grind [<= LTS.MTr.stepL]
```

Cslib/Computability/Automata/EpsilonNFAToNFA.lean
```diff
@@ -0,0 +1,87 @@
...
+  case mpr =>
+    rw [← εNFA.accepts_mem_language, ← NFA.accepts_mem_language]
+    simp only [NFA.Accepts, toNFA, Accepts, forall_exists_index, and_imp]
+    intro s hs s' hs' hmtr
+    exists s
+    apply And.intro hs
+    exists s'
+    apply And.intro hs'
+    rw [← LTS.noε_saturate_mTr]
+    assumption
```
chenson2018: This makes it clear where `grind` fails. I think something to do with it not being able to see how to use `LTS.noε_saturate_mTr`, which is parameterized by an `LTS`, with an `εNFA`.
```suggestion
  have : ∀ s s', enfa.saturate.MTr s (xs.map (some ·)) s' = enfa.saturate.noε.MTr s xs s' := by
    simp [LTS.noε_saturate_mTr]
  grind
```

Cslib/Computability/Automata/DFAToNFA.lean
```diff
@@ -0,0 +1,75 @@
...
+@[grind =]
+lemma toNFA_start {dfa : DFA State Symbol} : dfa.toNFA.start = {dfa.start} := rfl
+
+instance : Coe (DFA State Symbol) (NFA State Symbol) where
+  coe := toNFA
+
+/-- `DA.toNA` correctly characterises transitions. -/
+@[grind]
+theorem toNA_tr {dfa : DFA State Symbol} :
+  dfa.toNFA.Tr s1 μ s2 ↔ dfa.tr s1 μ = s2 := by
```
chenson2018, line 33: Short enough for one line:
```suggestion
theorem toNA_tr {dfa : DFA State Symbol} : dfa.toNFA.Tr s1 μ s2 ↔ dfa.tr s1 μ = s2 := by
```

Cslib/Computability/Automata/DFAToNFA.lean
```diff
@@ -0,0 +1,75 @@
...
+  dfa.toNFA.language = dfa.language := by
+  ext xs
+  rw [← DFA.accepts_mem_language, ← NFA.accepts_mem_language]
+  simp only [NFA.Accepts, DFA.Accepts]
+  apply Iff.intro <;> intro h
+  case mp =>
+    grind
+  case mpr =>
+    exists dfa.start
+    grind
```
chenson2018:
```suggestion
  refine ⟨?_, fun _ => ⟨dfa.start, ?_⟩⟩ <;> grind
```

Cslib/Computability/Automata/DFAToNFA.lean
```diff
@@ -0,0 +1,75 @@
...
+theorem toNFA_mtr {dfa : DFA State Symbol} :
+  dfa.toNFA.MTr s1 xs s2 ↔ dfa.mtr s1 xs = s2 := by
+  constructor <;> intro h
+  case mp =>
+    induction h <;> grind
+  case mpr =>
+    induction xs generalizing s1
+    case nil => grind
+    case cons x xs ih =>
+      apply LTS.MTr.stepL (s2 := dfa.tr s1 x) <;> grind
```
chenson2018:
```suggestion
  have : ∀ x, dfa.toNFA.Tr s1 x (dfa.tr s1 x) := by grind
  constructor <;> intro h
  case mp => induction h <;> grind
  case mpr => induction xs generalizing s1 <;> grind
```

chenson2018 (changes requested):
I think we should settle early on a style preference for explicit and scoped `grind`.

I made several `grind` suggestions, but I am still bothered by a few places where `grind` fails unexpectedly, as noted in your TODO. As I note in a comment, I think at least some of it it is because of failure to how structures extend each other.

Cslib/Computability/Automata/DA.lean
```diff
@@ -0,0 +1,19 @@
...
+  /-- The initial state of the automaton. -/
+  start : State
+  /-- The set of accepting states of the automaton. -/
+  accept : Set State
```
ctchou: Could you move the accept condition to DFA?  I want to define (for example) deterministic Muller automata by extending DA, but such automata have an acceptance condition of type: accept : Set (Set State).

Cslib/Computability/Automata/NA.lean
```diff
@@ -0,0 +1,21 @@
...
+  /-- The set of initial states of the automaton. -/
+  start : Set State
+  /-- The set of accepting states of the automaton. -/
```
ctchou: Could you move the accept condition to NFA and \epsilonNFA?  Automata on infinite words have acceptance conditions of types like `Set (Set State)` (Muller condition) and `Set (Set State × Set State))` (Rabin condition).

fmontesi:
> I think we should settle early on a style preference for explicit and scoped `grind`.
> 
> I made several `grind` suggestions, but I am still bothered by a few places where `grind` fails unexpectedly, as noted in your TODO. As I note in a comment, I think at least some of it it is because of failure to how structures extend each other.

Yeah, I've been putting `grind` annotations in many places because we haven't got a convention yet. Let's settle on explicit and `scoped` then, I've been using that in another project and seems to work well.

Cslib/Computability/Automata/DA.lean
fmontesi: I still don't get what the advantage of `Type*` is given that `Type _` (even by listing the identifiers before it) does the same thing and seems native?

Cslib/Computability/Automata/DFA.lean
fmontesi: I'll switch to explicit and scoped wherever I see it, we should plan for a future where cslib gets big in size.

Cslib/Computability/Automata/EpsilonNFAToNFA.lean
fmontesi: Nice one. :-)

Cslib/Computability/Automata/EpsilonNFAToNFA.lean
fmontesi: Good comment, it might be a problem related to structure extensions... Might be worth notifying the grind people about. @kim-em, what do you think?

Cslib/Computability/Automata/DFAToNFA.lean
fmontesi: For the record, we're also a bit inconsistent about when a parameter should be implicit or explicit in this kind of theorems. (Here dfa is implicit.) We might wanna set a convention about this.

Cslib/Computability/Automata/NFAToDFA.lean
```diff
@@ -0,0 +1,64 @@
...
+import Mathlib.Data.Fintype.Powerset
+
+/-! # Translation of NFA into DFA (subset construction) -/
+
+namespace NFA
+section SubsetConstruction
+
+/-- Converts an `NFA` into a `DFA` using the subset construction. -/
+@[grind]
+def toDFA (nfa : NFA State Symbol) : DFA (Set State) Symbol := {
```
ctchou, line 18: Perhaps you can separate out the `start` and `tr` components to a `NAtoDA.lean`, so that other extensions of NA and DA for omega-automata can also use the results?  Note that all the heavy-lifting in the proof of the subset construction involve only the `start` and `tr` components anyway.

Cslib/Computability/Automata/EpsilonNFA.lean
ctchou, line 17: Instead of extending NA to obtain εNFA, it may be cleaner to have an εNA (namely, an NA that supports ε-transitions) and extend εNA to obtain εNFA.  My comments about separating out the main part of the subset construction from NFA to NA also applies, mutatis mutandis, here.  I think such a design would be more orthogonal than the current one.

Cslib/Computability/Automata/DA.lean
chenson2018: My opinion is that it is more readable and as a side benefit more compact. As is, you have to twice interpret an annotation and `_`, whereas with `Type*` you get one piece of syntax that says "all these types are in arbitrary universes". As this is a very stable piece of Mathlib syntax, I do not find an advantage in the native spelling. 

Not a deal breaker for me, this can be considered personal preference.

Cslib/Computability/Automata/NFAToDFA.lean
fmontesi: Not sure what you mean, could you elaborate?
(Re have vs haveI, nah, I don't think it makes a difference.)

Cslib/Computability/Automata/DA.lean
fmontesi: I see what you mean, although I still find `(State : Type _) (Symbol : Type _)` to be less ambiguous. I'll mull over it.

fmontesi:
Thanks @chenson2018, I think I'm done with all the changes that make immediate sense. :-)

Cslib/Computability/Automata/NA.lean
fmontesi: Done!

Cslib/Computability/Automata/EpsilonNFA.lean
fmontesi: I agree with both comments, though I'd have to test `grind` (seems to be obstructed a bit by structure extensions sometimes). I'll implement them in a separate PR when I have some time for this again, since it seems like a backwards compatible extension to what is here. (Or, if you like, feel free to submit one.)

Cslib/Computability/Automata/NFAToDFA.lean
chenson2018: I meant that if `NFA.{finite_state,finite_symbol}` were made instances, the corresponding `DFA` fields could be inferred completely here. That seems to not quite work though...

Cslib/Computability/Automata/DA.lean
eric-wieser: This is a false dichotomy, `(State Symbol : Type _)` and `(State : Type _) (Symbol : Type _)` are identical, and the choice to combine is  orthogonal to the choice of `Type _` vs `Type*`.

PR #86, opened 2025-10-07
feat: customize Cslib/Foundations/Data/OmegaList/*

ctchou:
In the previous PR:
https://github.com/leanprover/cslib/pull/80
`ωList` is just a copy of `Stream'` with suitable name changes.  Now we make three major modifications:
1. We make it clear that `ωList α` is really an abstraction of `ℕ → α`.  In particular, for `s : ωList` and `n : ℕ`, the `n`-th elements of `s` is now obtained by `s n`, not `s.get n` or `get s n`.  Indeed, we have got rid of the API function `get` altogether.
2. The infix notation for `appendωList` is changed from`++ₛ` to `++ₗ`.  (The last letter is a lowercase "L" subscript.). This is to avoid conflict with `Stream'`.
3. A new API function `extract`, which is the analogue of `List.extract`, and a number of theorems involving it are added.


ctchou:
Rebase on the current "main".

ctchou:
This PR Is superseded by https://github.com/leanprover/cslib/pull/90.

PR #89, opened 2025-10-08
feat(Logics/Propositional): definitions

thomaskwaring:
This PR defines propositions for propositional logic, including notation for the logical connectives. A `Theory` is a set of `Proposition`. Also defines minimal, intuitionistic and classical theories, and extensions of maps `Atom → Atom'` to maps `Proposition Atom → Proposition Atom'` and `Theory Atom → Theory Atom'`.


Cslib/Logics/Propositional/Defs.lean
```diff
@@ -0,0 +1,155 @@
...
+
+We introduce notation for the logical connectives: `⊥ ⊤ ⋏ ⋎ ⟶ ~` for, respectively, falsum, verum,
+conjunction, disjunction, implication and negation.
+-/
+
+universe u
+
+variable {Atom : Type u} [DecidableEq Atom]
+
+namespace PL
```
fmontesi: I'd call this either `Cslib.Logic.PL` or `Cslib.Logic.Propositional`

Cslib/Logics/Propositional/Defs.lean
```diff
@@ -0,0 +1,155 @@
...
+
+/-- Propositions. -/
+inductive Proposition (Atom : Type u) : Type u where
+  /-- Propositional atoms -/
+  | atom (x : Atom)
+  /-- Conjunction -/
+  | conj (a b : Proposition Atom)
```
fmontesi: `and`

Cslib/Logics/Propositional/Defs.lean
```diff
@@ -0,0 +1,155 @@
...
+  /-- Disjunction -/
+  | disj (a b : Proposition Atom)
```
fmontesi: `or`

Cslib/Logics/Propositional/Defs.lean
```diff
@@ -0,0 +1,155 @@
...
+abbrev Proposition.neg [Bot Atom] : Proposition Atom → Proposition Atom := (Proposition.impl · ⊥)
+
+/-- A fixed choice of a derivable proposition (of course any two are equivalent). -/
+abbrev Proposition.top [Inhabited Atom] : Proposition Atom := impl (.atom default) (.atom default)
+
+instance instTopProposition [Inhabited Atom] : Top (Proposition Atom) := ⟨.top⟩
+
+example [Bot Atom] : (⊤ : Proposition Atom) = Proposition.impl ⊥ ⊥ := rfl
+
+@[inherit_doc] scoped infix:35 " ⋏ " => Proposition.conj
```
fmontesi: This is purely aesthetic, so not important for merging:

I know that we should avoid overlapping notation with Lean's core, but this all makes me a bit sad. This is *the* logic of ∧, →, etc. :-)
Have you tried using the standard symbols? They'd be in a different kind, so maybe it'd work.

Cslib/Logics/Propositional/Defs.lean
```diff
@@ -0,0 +1,155 @@
...
+instance {Atom Atom' : Type u} : FunLike (Atom → Atom') (Theory Atom) (Theory Atom') where
+  coe := Theory.map
+  coe_injective' f f' h := by
+    ext x
+    have : Theory.map f {Proposition.atom x} = Theory.map f' {Proposition.atom x} :=
+      congrFun h {Proposition.atom x}
+    simpa [Theory.map, Proposition.map] using this
+
+/-- The base theory is minimal propositional logic. -/
+abbrev MPL : Theory (Atom) := ∅
```
fmontesi, line 106: Would it make sense to namespace this under Theory? (Theory.Minimal or Theory.MPL)

Cslib/Logics/Propositional/Defs.lean
```diff
@@ -0,0 +1,155 @@
...
+
+instance {Atom Atom' : Type u} : FunLike (Atom → Atom') (Theory Atom) (Theory Atom') where
+  coe := Theory.map
+  coe_injective' f f' h := by
+    ext x
+    have : Theory.map f {Proposition.atom x} = Theory.map f' {Proposition.atom x} :=
+      congrFun h {Proposition.atom x}
+    simpa [Theory.map, Proposition.map] using this
+
+/-- The base theory is minimal propositional logic. -/
```
fmontesi: I'm a bit unsure if saying 'is' is correct here. Thoughts?

fmontesi (changes requested):

Cslib/Logics/Propositional/Defs.lean
thomaskwaring: it seems to work okay actually, i've changed it now & if it breaks later we can revert

Cslib/Logics/Propositional/Defs.lean
thomaskwaring: good point, thanks

Cslib/Logics/Propositional/Defs.lean
thomaskwaring: how about now?

fmontesi (approved):

PR #101, opened 2025-10-14
feat: add Cslib/Foundations/Data/Nat/Segment.lean

ctchou:
This is a theory of "segments" which will be used in defining the flatten and limit operations of OmegaSequence.


ctchou:
Close this PR now.  On 2nd thought, I think I'll include it as part of the OmegaSequence/Flatten PR, which I'm working on.

PR #102, opened 2025-10-14
chore: don't check docs toolchain unless in a PR or main branch

chenson2018:
Normally we check that our library and documentation are using the same `lean-toolchain`. However the [nightly testing workflow](.github/workflows/merge_main_into_nightly-testing.yml) by design is meant to test a toolchain ahead of our current pinned version. To accommodate this, this PR changes the doc toolchain check to only happen during pull requests and commits to the `main` branch.

### DIFF
--- a/.github/workflows/lean_action_ci.yml
+++ b/.github/workflows/lean_action_ci.yml
@@ -10,7 +10,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - run: |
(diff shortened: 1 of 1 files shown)


arademaker:
I guess @chenson2018 wants to test toolchains in other branches without dealing with the docs, but in the future, I would prefer a minimal explanation in the PR description.

chenson2018:
> I guess @chenson2018 wants to test toolchains in other branches without dealing with the docs, but in the future, I would prefer a minimal explanation in the PR description.

Sorry about that, description added.

PR #110, opened 2025-10-15
feat(CI): add workflow to check for commit convention

### LINKED ISSUE #108: enforce commit conventions
I suggested the following [on Zulip](https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/commit.20conventions/near/544862054) regarding using the Mathlib/Lean commit message format: 

> Looking at the last couple of weeks of commits, it seems we have *partially* converged on [commit conventions], so I say yes and that we should copy over the workflow lean4 has to enforce this.

I was referring to [this workflow](https://github.com/leanprover/lean4/blob/master/.github/workflows/pr-title.yml).

chenson2018:
closes #108


chenson2018:
Note: I allow anything in the parentheses. I tested here by changing the title a few times.

PR #118, opened 2025-10-22
feat: Support extensional equality in FinFun

fmontesi:
This PR changes `FinFun` to support extensional equality, making it essentially equivalent to `Finsupp` in Mathlib, with the important difference that definitions and theories about `FinFun` are computable. It also makes FinFun adopt the same notation as Finsupp, so to facilitate a future replacement with the latter in the future if it ever changes in Mathlib to have computable defs/theorems (this is relegated to the future since FinFun is used downstream in at least one big project I know of).

### DIFF
--- a/Cslib/Foundations/Data/FinFun.lean
+++ b/Cslib/Foundations/Data/FinFun.lean
@@ -43,7 +43,7 @@ private def fromFun {α β : Type _} [Zero β] [DecidableEq α]
   mem_support_fn := by grind
 
 @[inherit_doc]
-scoped notation f:25 "↾" support:51 => FinFun.fromFun f support
+scoped notation f:25 "↾₀" support:51 => FinFun.fromFun f support
 
 instance instFunLike [Zero β] : FunLike (α →₀ β) α β where
   coe f := f.fn
@@ -89,42 +89,42 @@ theorem congrFinFun [DecidableEq α] [Zero β] {f g : α →₀ β} (h : f = g) 
 @[scoped grind =]
 theorem fromFun_eq [Zero β] [DecidableEq α] [∀ y : β, Decidable (y = 0)]
     (f : α → β) (support : Finset α) (h : ∀ a, a ∉ support → f a = 0) :
-    (f ↾ support) = f := by grind
+    (f ↾₀ support) = f := by grind
 
 @[scoped grind =]
 theorem fromFun_fn [Zero β] [DecidableEq α] [∀ y : β, Decidable (y = 0)]
     (f : α → β) (support : Finset α) :
-    (f ↾ support).fn = (fun a => if a ∈ support then f a else 0) := by
+    (f ↾₀ support).fn = (fun a => if a ∈ support then f a else 0) := by
   grind
 
 @[scoped grind =]
 theorem fromFun_support [Zero β] [DecidableEq α] [∀ y : β, Decidable (y = 0)]
     (f : α → β) (support : Finset α) :
-    (f ↾ support).support = support.filter (f · ≠ 0) := by
+    (f ↾₀ support).support = support.filter (f · ≠ 0) := by
   grind
 
 /-- Restricting a function twice to the same support is idempotent. -/
 @[scoped grind =]
 theorem fromFun_idem [Zero β] [DecidableEq α]
     [∀ y : β, Decidable (y = 0)] {f : α → β} {support : Finset α} :
-    (f ↾ support) ↾ support = f ↾ support := by grind
+    (f ↾₀ support) ↾₀ support = f ↾₀ support := by grind
 
 /-- Restricting a `FinFun` to its own support is the identity. -/
 @[scoped grind =]
 theorem coe_fromFun_id [Zero β] [DecidableEq α] [∀ y : β, Decidable (y = 0)] {f : α →₀ β} :
-    (f ↾ f.support) = f := by grind
+    (f ↾₀ f.support) = f := by grind
 
 /-- Restricting a function twice to two supports is equal to restricting to their intersection. -/
 @[scoped grind =]
 theorem fromFun_inter [Zero β] [DecidableEq α]
     [∀ y : β, Decidable (y = 0)] {f : α → β} {support1 support2 : Finset α} :
-    (f ↾ support1) ↾ support2 = f ↾ (support1 ∩ support2) := by grind
+    (f ↾₀ support1) ↾₀ support2 = f ↾₀ (support1 ∩ support2) := by grind
 
 /-- Restricting a function is commutative. -/
 @[scoped grind =]
 theorem fromFun_comm [Zero β] [DecidableEq α]
     [∀ y : β, Decidable (y = 0)] {f : α → β} {support1 support2 : Finset α} :
-    (f ↾ support1) ↾ support2 = (f ↾ support2) ↾ support1 := by grind
+    (f ↾₀ support1) ↾₀ support2 = (f ↾₀ support2) ↾₀ support1 := by grind
 
 end FinFun
 


Cslib/Foundations/Data/FinFun.lean
```diff
@@ -6,108 +6,106 @@ Authors: Fabrizio Montesi, Xueying Qin
...
-    f = g := by
-  cases f
-  cases g
+/-- A `FinFun` is a function `fn` with a finite `support`.
+
+This is similar to `Finsupp` in Mathlib, but definitions are computable. -/
+structure FinFun (α : Type _) (β : Type _) [Zero β] where
+  fn : α → β
+  support : Finset α
+  mem_support_fn {a : α} : a ∈ support ↔ fn a ≠ 0
```
chenson2018, line 27: It wouldn't hurt to document these fields.

Cslib/Foundations/Data/FinFun.lean
```diff
@@ -6,108 +6,106 @@ Authors: Fabrizio Montesi, Xueying Qin
...
+
+attribute [scoped grind _=_] FinFun.mem_support_fn
+
+namespace FinFun
+
+scoped notation:50 α " →₀ " β => FinFun α β
```
chenson2018: Compare with the Finsupp notation:
```lean
@[inherit_doc]
infixr:25 " →₀ " => Finsupp
```

Cslib/Foundations/Data/FinFun.lean
```diff
@@ -6,108 +6,106 @@ Authors: Fabrizio Montesi, Xueying Qin
...
+
+/-- Constructs a `FinFun` from any function and a given support, filtering out all elements not
+mapped to 0 in the support. -/
+@[scoped grind]
+private def mkRestrictFun {α β : Type _} [Zero β] [DecidableEq α]
+  [hdec : ∀ y : β, Decidable (y = 0)] (fn : α → β) (support : Finset α) : α →₀ β where
```
chenson2018: No need to name this instance?

Cslib/Foundations/Data/FinFun.lean
```diff
@@ -6,108 +6,106 @@ Authors: Fabrizio Montesi, Xueying Qin
...
+    ext a
+    grind
+
+@[scoped grind =]
+theorem coe_fn [Zero β] {f : α →₀ β} : (f : α → β) = f.fn := by simp [DFunLike.coe]
+
+@[scoped grind =]
+theorem coe_eq_fn [Zero β] {f : α →₀ β} : f a = f.fn a := by
+  rcases f with ⟨f, support, mem_support_fn⟩
+  simp [DFunLike.coe]
```
chenson2018:
```suggestion
  simp [DFunLike.coe]
```

Cslib/Foundations/Data/FinFun.lean
```diff
@@ -6,108 +6,106 @@ Authors: Fabrizio Montesi, Xueying Qin
...
+
+/-- Extensional equality for `FinFun`. -/
+@[scoped grind ←=]
+theorem ext [Zero β] {f g : α →₀ β} (h : ∀ (a : α), f a = g a) :
+  f = g := by
+  apply DFunLike.ext
+  exact h
```
chenson2018: The typeclass synthesis works if you help with the implict function args:
```suggestion
theorem ext [Zero β] {f g : α →₀ β} (h : ∀ (a : α), f a = g a) : f = g :=
  DFunLike.ext (f := f) (g := g) h
```

Cslib/Foundations/Data/FinFun.lean
```diff
@@ -6,108 +6,106 @@ Authors: Fabrizio Montesi, Xueying Qin
...
+    simp only [coe_eq_fn]
+    rw [← FinFun.mem_support_fn]
+    simp only [Membership.mem] at h
+    simp only [Membership.mem]
+    exact h
+  case mpr =>
+    have hmem := f.mem_support_fn (a := a)
+    simp only [Membership.mem] at hmem
+    simp only [Membership.mem]
+    grind
```
chenson2018:
```suggestion
  grind
```

Cslib/Foundations/Data/FinFun.lean
```diff
@@ -6,108 +6,106 @@ Authors: Fabrizio Montesi, Xueying Qin
...
+theorem eq_char [DecidableEq α] [Zero β] {f g : α →₀ β} :
+  f = g ↔ (f.fn = g.fn ∧ f.support = g.support) := by
+  apply Iff.intro <;> intro h
+  · grind
+  · obtain ⟨h1, h2⟩ := h
+    apply DFunLike.ext
+    intro x
+    rw [coe_fn]
+    rw [coe_fn]
+    rw [h1]
```
chenson2018:
```suggestion
theorem eq_char [DecidableEq α] [Zero β] {f g : α →₀ β} :
    f = g ↔ f.fn = g.fn ∧ f.support = g.support := by
  have (h : f.fn = g.fn) : f = g := ext (congrFun h)
  grind
```

Cslib/Foundations/Data/FinFun.lean
```diff
@@ -6,108 +6,106 @@ Authors: Fabrizio Montesi, Xueying Qin
...
+
+@[scoped grind =>]
+theorem congrFinFun [DecidableEq α] [Zero β] {f g : α →₀ β} (h : f = g) (a : α) :
+    f a = g a := by grind
```
chenson2018, line 87: The signature fits on one line:
```suggestion
theorem congrFinFun [DecidableEq α] [Zero β] {f g : α →₀ β} (h : f = g) (a : α) : f a = g a :=
  by grind
```

Cslib/Foundations/Data/FinFun.lean
```diff
@@ -6,108 +6,106 @@ Authors: Fabrizio Montesi, Xueying Qin
...
-
-notation:50 α " ⇀ " β => FinFun α β
-notation:50 f "↾" dom => FinFun.mk f dom
-
-abbrev CompleteDom [Zero β] (f : α ⇀ β) := ∀ x, x ∉ f.dom → f.f x = 0
-
-def FinFun.defined (f : α ⇀ β) (x : α) : Prop := x ∈ f.dom
-
-@[simp]
-abbrev FinFun.apply (f : α ⇀ β) (x : α) : β := f.f x
```
XYUnknown, line 31: It would be convenient if we keep this

Cslib/Foundations/Data/FinFun.lean
```diff
@@ -6,108 +6,106 @@ Authors: Fabrizio Montesi, Xueying Qin
...
+
+instance instFunLike [Zero β] : FunLike (α →₀ β) α β where
+  coe f := f.fn
+  coe_injective' := by
+    rintro ⟨f1, support1, mem_support_fn1⟩ ⟨f2, support2, mem_support_fn2⟩
+    simp only
+    intro heq
+    simp only [heq, mk.injEq, true_and]
+    ext a
+    grind
```
chenson2018:
```suggestion
    rintro ⟨_, _⟩ ⟨_, _⟩
    simp_all [Finset.ext_iff]
```

Cslib/Foundations/Data/FinFun.lean
```diff
@@ -6,108 +6,106 @@ Authors: Fabrizio Montesi, Xueying Qin
...
-  <;> rcases h with ⟨_, _, _, _⟩
-  <;> grind
-
-theorem FinFun.mapBin_char₂ [DecidableEq α] (f g : α ⇀ β)
-    (op : Option β → Option β → Option β) (hdom : f.dom = g.dom)
-    (hop : ∀ x ∈ f.dom, (op (some (f.f x)) (some (g.f x))).isSome)
-    : (FinFun.mapBin f g op).isSome := by grind [mapBin]
-
--- Fun to FinFun
-def Function.toFinFun [DecidableEq α] (f : α → β) (dom : Finset α) : α ⇀ β := FinFun.mk f dom
```
XYUnknown, line 84: Similarly it would also be convenient if we keep this Function to FinFun conversion.

XYUnknown (commented):
It should be compatible with my code base. 
Two small syntactic things: it would be convenient if we keep the abbreviation `FinFun.apply` and the definition`Function.toFinFun`.

Cslib/Foundations/Data/FinFun.lean
chenson2018: The reverse direction works without the support hypothesis, this is intentional?

Cslib/Foundations/Data/FinFun.lean
fmontesi: Why? You can now just write `f x` and it'll do the same thing, because of the new `FunLike` instance. Making that an `abbrev` reveals the internal details of `FinFun` a bit.

XYUnknown: Ah I see, I was not aware of `FunLike` allows me to write `f x` directly; I thought without `FinFun.apply` I would need to write `f.fn x` which would reveal the internal details of `FinFun`.

Cslib/Foundations/Data/FinFun.lean
fmontesi: Not sure what you mean?

Cslib/Foundations/Data/FinFun.lean
fmontesi: Looks like I forgot to backtrack and transform some proofs into `grind` calls. Gonna call it `grind` debt. ;-)

Cslib/Foundations/Data/FinFun.lean
fmontesi: This is now `FinFun.mkRestrictFun`, accessible through the uprighthook notation: `f ↾ support`.

I made `mkRestrictFun` private for now because I'm not in particular love with the name...... :-)

XYUnknown: How does the notation work with `simp`?

Cslib/Foundations/Data/FinFun.lean
chenson2018: I meant removing `hdec` fom `[hdec : ∀ y : β, Decidable (y = 0)]`. Not sure the official style guidance,  but you usually only see named instances is they are needed explicitly in the signature/proof.

Cslib/Foundations/Data/FinFun.lean
```diff
@@ -6,108 +6,90 @@ Authors: Fabrizio Montesi, Xueying Qin
...
+  support : Finset α
+  /-- Proof that `support` is the support of the underlying function. -/
+  mem_support_fn {a : α} : a ∈ support ↔ fn a ≠ 0
+
+attribute [scoped grind _=_] FinFun.mem_support_fn
+
+namespace FinFun
+
+@[inherit_doc]
+scoped infix:25 " →₀ " => FinFun
```
chenson2018: Just making sure this is not a typo: you made this `infix` versus `Finsupp` using `infixr`?

Cslib/Foundations/Data/FinFun.lean
fmontesi: That's correct. I was convinced to be using it, must've been an old proof. Will fix, thx.

Cslib/Foundations/Data/FinFun.lean
fmontesi: typo :-)

fmontesi:
@XYUnknown I've added some results about restriction, which has pretty nice properties, and renamed it to `fromFun`.

@XYUnknown @chenson2018 I am though in doubt whether I should use the `↾` symbol for the 'restriction to a finite support' operator. In general, that operator might make sense for any Set. Maybe I should rename it to `↾₀` ?

chenson2018:
I don't have a strong opinion on the notation. One comparison is that in Mathlib [Matroid.«term_↾_»](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Combinatorics/Matroid/Minor/Restrict.html#Matroid.%C2%ABterm_%E2%86%BE_%C2%BB) is used as a scoped notation for restricting a matroid to a set.

XYUnknown:
> @XYUnknown I've added some results about restriction, which has pretty nice properties, and renamed it to `fromFun`.

It is indeed a more reasonable name compared to `mkRestrictFun`

> @XYUnknown @chenson2018 I am though in doubt whether I should use the `↾` symbol for the 'restriction to a finite support' operator. In general, that operator might make sense for any Set. Maybe I should rename it to `↾₀` ?

Since the symbol of the `FinFun` is `→₀`, it does make sense to me to use `↾₀`.

fmontesi:
Thanks, both! I've adopted the `↾₀` notation. This will leave us with the possibility of using `↾` for the version that doesn't require a `Finset`.

PR #127, opened 2025-10-25
chore: more environment linting

chenson2018:
This PR adds almost all the lints that were left commented out in #123. All that remains is documentation linting and `explicitVarsOfIff` which I felt might be too opinionated.

### DIFF
--- a/Cslib/Foundations/Data/OmegaSequence/Init.lean
+++ b/Cslib/Foundations/Data/OmegaSequence/Init.lean
@@ -484,12 +484,12 @@ theorem extract_lu_extract_lu {xs : ωSequence α} {m n i j : ℕ} (h : j ≤ n 
   ext k
   cases Decidable.em (k < j - i) <;> grind [extract_eq_ofFn]
 
-@[nolint simpNF, simp, scoped grind =]
+@[scoped grind =]
 theorem extract_0u_extract_lu {xs : ωSequence α} {n i j : ℕ} (h : j ≤ n) :
     (xs.extract 0 n).extract i j = xs.extract i j := by
   grind
 
-@[nolint simpNF, simp, scoped grind =]
+@[scoped grind =]
 theorem extract_0u_extract_l {xs : ωSequence α} {n i : ℕ} :
     (xs.extract 0 n).extract i = xs.extract i n := by
   grind


chenson2018:
@ctchou There are a couple of small changes to `OmegaSequence` here:
- I remove a theorem that was both duplicated and unused.
- I marked a couple of theorems as `nolint simpNF`. They aren't actually used by `simp` anywhere currently, but the way their types overlap with other theorems is what makes them problematic for the linter.

ctchou:
It's OK to remove take_one'.  I've double-checked that it is not used in my PR.

Any idea what other theorems the `nolint simpNF` theorems overlap?

chenson2018:
This is the linter output from `lake test` without the `nolint`:
<details>

```
error: CslibTests/Lint.lean:21:0: /- The `simpNF` linter reports:
SOME SIMP LEMMAS ARE NOT IN SIMP-NORMAL FORM.
see note [simp-normal form] for tips how to debug this.
https://leanprover-community.github.io/mathlib_docs/notes.html#simp-normal%20form -/
-- Cslib.Foundations.Data.OmegaSequence.Init
#check @Cslib.ωSequence.extract_0u_extract_lu /- simp can prove this:
  by simp only [*, @List.extract_eq_drop_take, @tsub_zero, @Cslib.ωSequence.extract_lu_extract_lu, @zero_add]
One of the lemmas above could be a duplicate.
If that's not the case try reordering lemmas or adding @[priority]. -/
#check @Cslib.ωSequence.extract_0u_extract_l /- simp can prove this:
  by simp only [*,
 @Cslib.ωSequence.length_extract,
 @tsub_zero,
 @List.extract_eq_drop_take,
 @le_refl,
 @Cslib.ωSequence.extract_lu_extract_lu,
 @zero_add]
One of the lemmas above could be a duplicate.
If that's not the case try reordering lemmas or adding @[priority]. -/
error: Lean exited with code 1
Some required targets logged failures:
- CslibTests.Lint
error: build failed
```

</details>

I find the message a little hard to interpret, but my guess from it was `extract_lu_extract_lu`.

ctchou:
What exactly is a simp normal form?  I find the following explanation not very illuminating:
https://lean-lang.org/doc/reference/latest/The-Simplifier/Simp-Normal-Forms/
In any case, please feel free to remove the simp attributes from the extract_??_extract_?? theorems.  I think we can live with invoking them explicitly.

chenson2018:
Okay, I have removed them for now. Please let me know if you find yourself specifying them a lot, we can revisit and put some more thought into it.

The idea of simp normal forms is that they are the preferred way to write certain terms. `simp` rewrites the left-hand side into the right, so we place the preferred form on the right. As a simple example, consider:
```
@[simp] theorem add_eq : Nat.add x y = x + y := rfl
```
The idea is that we never really want to work with `Nat.add`, so the normal form here is the plus notation. `simp` will rewrite into this form, and in general we state our lemmas in terms of the normal form. Ideally we would like a confluent rewrite system such that `simp` converges into a unique representative of that term, but this is not always true in practice.

fmontesi (approved):

PR #92, opened 2025-10-11
feat: add Cslib/Computability/Languages/OmegaLanguage.lean

ctchou:
This is the first version of OmegaLanguage, which is an analogue of Mathlib.Computability.Language, but for infinite sequences.  We now have a more satisfactory definition of omegaPow, which enables more algebraic and much shorter proofs of the harder results (see the "main theorems" section of the main comment in OmegaLanguage.lean) than in:
https://github.com/ctchou/AutomataTheory/blob/main/AutomataTheory/Languages/Basic.lean
The results proved so far are sufficient for starting the development of the theory of omega-automata.  The main TODOs are results about omegaPow and map, which are not needed for now.

As part of this work, we added several new definitions and theorems about Language and OmegaSequence.flatten.  We also took this opportunity to change some infelicitous theorem names in Flatten.lean and Segment.lean and to use `namespace Nat` to remove the prefix `Nat.` from many places in Segment.lean.

PR #129, opened 2025-10-26
chore: prefer use to refine

chenson2018:
A few proofs that become more readable when using `use` versus `refine`. We should in general encourage this style, especially in proofs using `grind`, `omega`, etc.


ctchou:
@chenson2018, I did a lot of renaming in Segment.lean in PR https://github.com/leanprover/cslib/pull/92, which may conflict with this PR.

chenson2018:
I think it is fine, locally merging with `--no-commit` worked for me.

Issue #95, opened 2025-10-13
add `mergeWithGrind` linter


chenson2018:
I don't want to commit to the entire new `weeklyLintSet`, but this linter that is currently the only one included seems like a good addition. This is a small change, it only lints three proofs currently (and all seem like improvements). We would add this on the next Mathlib release.

PR #130, opened 2025-10-26
chore: golfing in `BehaviouralTheory`

chenson2018:
I separated this from #129 because it is a bit wider scope. The unfolding required to make `LTS` work with `grind` remains painful, but I found a way to somewhat make it work locally. I'm still not sure if/how there's a way to state a theorem that could be `grind` annotated for the same effect.

### DIFF
--- a/Cslib/Languages/CCS/BehaviouralTheory.lean
+++ b/Cslib/Languages/CCS/BehaviouralTheory.lean
@@ -33,7 +33,7 @@ namespace CCS
 
 attribute [local grind] Tr
 
-@[scoped grind]
+@[local grind]
 private inductive ParNil : Process Name Constant → Process Name Constant → Prop where
 | parNil : ParNil (par p nil) p
 
@@ -50,7 +50,7 @@ theorem bisimilarity_par_nil : (par p nil) ~[lts (defs := defs)] p := by
     exists par s2' nil
     grind
 
-@[scoped grind]
+@[local grind]
 private inductive ParComm : Process Name Constant → Process Name Constant → Prop where
 | parComm : ParComm (par p q) (par q p)
 
@@ -72,7 +72,7 @@ theorem bisimilarity_nil_par : (par nil p) ~[lts (defs := defs)] p :=
     (par nil p) ~[lts (defs := defs)] (par p nil) := by grind
     _ ~[lts (defs := defs)] p := by simp
 
-@[scoped grind]
+@[local grind]
 private inductive ParAssoc : Process Name Constant → Process Name Constant → Prop where
   | assoc : ParAssoc (par p (par q r)) (par (par p q) r)
   | id : ParAssoc p p
@@ -160,7 +160,7 @@ theorem bisimilarity_choice_nil : (choice p nil) ~[lts (defs := defs)] p := by
     · exact ChoiceNil.id
   all_goals grind [ChoiceNil]
 
-@[scoped grind]
+@[local grind]
 private inductive ChoiceIdem : Process Name Constant → Process Name Constant → Prop where
   | idem : ChoiceIdem (choice p p) p
   | id : ChoiceIdem p p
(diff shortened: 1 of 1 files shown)


Cslib/Languages/CCS/BehaviouralTheory.lean
```diff
@@ -31,7 +31,9 @@ open CCS CCS.Process CCS.Act
 
 namespace CCS
 
-@[grind]
+attribute [local grind] Tr
```
chenson2018, line 34: I made this local because I wasn't sure if it would be problematic elsewhere.

Cslib/Languages/CCS/BehaviouralTheory.lean
```diff
@@ -58,50 +61,37 @@ private inductive ParComm : (Process Name Constant) → (Process Name Constant)
 theorem bisimilarity_par_comm : (par p q) ~[lts (defs := defs)] (par q p) := by
```
chenson2018, line 59: I feel like for these proofs there should be a way to cleverly separate data and proof, but I couldn't figure it out. Something like a function that gives exactly the `exist` lines and then a proof that is just a call to `grind`.

chenson2018: Ah, I think I have it! Look at this: https://github.com/leanprover/cslib/pull/130/commits/b20c5771acd982e211a8370b2735ce53934c5640. I'll see if it generalizes to the other cases.

chenson2018: Unfortunately, this seems like a special case that didn't depend on `Tr`. I think I am done.

chenson2018:
(I realized this also closes #95, so I've added pushed that here as well)

Cslib/Languages/CCS/BehaviouralTheory.lean
```diff
@@ -25,83 +25,47 @@ namespace Cslib
...
 
 open CCS CCS.Process CCS.Act
 
 namespace CCS
 
-@[grind]
-private inductive ParNil : (Process Name Constant) → (Process Name Constant) → Prop where
+attribute [local grind] Tr
+
+@[scoped grind]
```
fmontesi: Shouldn't this be `local` too? It's used only for the proof. In fact, if I could've written as a local def in the proof, I would've done so. Is it possible?

Cslib/Languages/CCS/BehaviouralTheory.lean
```diff
@@ -25,83 +25,47 @@ namespace Cslib
...
+    exists par s2' nil
+    grind
 
-private inductive ParComm : (Process Name Constant) → (Process Name Constant) → Prop where
+@[scoped grind]
+private inductive ParComm : Process Name Constant → Process Name Constant → Prop where
 | parComm : ParComm (par p q) (par q p)
 
+@[scoped grind]
+def defParComm : Process Name Constant → Process Name Constant
```
fmontesi: This looks like it could just be inlined in the proof, I guess. Otherwise, we should make this private at least I guess?

Cslib/Languages/CCS/BehaviouralTheory.lean
```diff
@@ -110,15 +74,15 @@ theorem bisimilarity_nil_par : (par nil p) ~[lts (defs := defs)] p :=
     (par nil p) ~[lts (defs := defs)] (par p nil) := by grind
     _ ~[lts (defs := defs)] p := by simp
 
-
-private inductive ParAssoc : (Process Name Constant) → (Process Name Constant) → Prop where
+@[scoped grind]
```
fmontesi: same comment as for the above (make local?).

Cslib/Languages/CCS/BehaviouralTheory.lean
chenson2018: I was able to inline this.

Cslib/Languages/CCS/BehaviouralTheory.lean
chenson2018: I don't  think you can inline an `inductive`? I have made the annotations local for all the private inductives.

fmontesi (approved):

Issue #125, opened 2025-10-24
check that all directories are imported in `Cslib.lean`


chenson2018:
While I think [this action](https://github.com/leanprover/cslib/blob/main/.github/workflows/lean_lint_suggest.yml) checks for this and leaves a comment, an actual standalone test would be nice.

jessealama:
You can run the check locally with:
```bash
lake exe mk_all --lib Cslib --check
```

To make this more visible/discoverable, we could:
1. Add a shell script in `scripts/check-imports.sh`
2. Add a lake script in `lakefile.toml` (runnable via `lake script run check-imports`)
3. Note somewhere that you can run GitHub Actions locally via `act`, like so:
   ```bash
   act -W .github/workflows/lean_action_ci.yml
   ```

I'm not sure which approach we should take. I'm inclined to tweak the lakefile. What do you think?

chenson2018:
(that last PR mention was a typo, please ignore)

PR #131, opened 2025-10-26
feat: add Cslib.Init module and checkInitImports linter

### LINKED ISSUE #122: create Cslib.Init and remove `weak` options in lakefile
From discussion at [**CSLib>weak in lakefile** ](https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/weak.20in.20lakefile/with/546654795).

jessealama:
Closes #122

This PR implements `Cslib.Init` following Mathlib's pattern, along with a `checkInitImports` linter to ensure modules are connected to the library's import graph.

## Changes

- Created `Cslib/Init.lean` importing `Mathlib.Init` and `Cslib.Foundations.Lint.Basic`
- Added `checkInitImports` environment linter
- Integrated validation into `lake test` via new test scripts in `scripts/`
- Added `import Cslib.Init` to 10 foundational modules

## Files excluded from import

Some files could not import `Cslib.Init` due to technical constraints:

- `Cslib.Foundations.Lint.Basic`: Circular dependency (imported by `Cslib.Init`)
- `Cslib.Foundations.Data.FinFun`: Notation conflict with `Mathlib.Finsupp` (both use `→₀`)
- `Cslib.Foundations.Semantics.LTS.Basic`: Causes type elaboration issues in downstream Automata files
- `Cslib.Logics.LinearLogic.CLL.Basic`: Syntax elaboration conflicts

TODO comments have been added to document these issues for future investigation.

### DIFF
--- a/Cslib/Computability/Languages/OmegaLanguage.lean
+++ b/Cslib/Computability/Languages/OmegaLanguage.lean
@@ -256,11 +256,11 @@ theorem le_omegaPow_congr [Inhabited α] {l1 l2 : Language α} (h : l1 ≤ l2) :
 theorem omegaPow_of_sub_one [Inhabited α] : (l - 1)^ω = l^ω := by
   ext s ; simp
 
-@[simp]
+@[simp, nolint simpNF]
 theorem zero_omegaPow [Inhabited α] : (0 : Language α)^ω = ⊥ := by
   ext s ; simp
 
-@[simp]
+@[simp, nolint simpNF]
 theorem one_omegaPow [Inhabited α] : (1 : Language α)^ω = ⊥ := by
   rw [← omegaPow_of_sub_one, tsub_self, zero_omegaPow]
 
--- a/Cslib/Foundations/Data/OmegaSequence/Flatten.lean
+++ b/Cslib/Foundations/Data/OmegaSequence/Flatten.lean
@@ -121,8 +121,8 @@ theorem append_flatten [Inhabited α] {ls : ωSequence (List α)} (h_ls : ∀ k,
   induction n generalizing ls <;> grind [tail_eq_drop, take_succ]
 
 /-- The length of `(ls.take n).flatten` is `ls.cumLen n`. -/
-@[simp, scoped grind =]
-theorem length_flatten_take [Inhabited α] {ls : ωSequence (List α)} (h_ls : ∀ k, (ls k).length > 0)
+@[simp, nolint simpNF, scoped grind =]
+theorem length_flatten_take {ls : ωSequence (List α)} (h_ls : ∀ k, (ls k).length > 0)
     (n : ℕ) : (ls.take n).flatten.length = ls.cumLen n := by
   induction n <;> grind [take_succ']
 
@@ -169,7 +169,7 @@ theorem toSegs_def (s : ωSequence α) (f : ℕ → ℕ) (n : ℕ) :
 
 /-- `(s.toSegs f).cumLen` is `f` itself. -/
 @[simp]
-theorem segment_toSegs_cumLen [Inhabited α] {f : ℕ → ℕ}
+theorem segment_toSegs_cumLen {f : ℕ → ℕ}
     (hm : StrictMono f) (h0 : f 0 = 0) (s : ωSequence α) :
     (s.toSegs f).cumLen = f := by
   ext n


Cslib/Foundations/Lint/Basic.lean
```diff
@@ -35,4 +35,29 @@ def topNamespace : Batteries.Tactic.Lint.Linter where
       else
         return m!"{declName} is not namespaced."
 
+/-- A linter that checks if the current module imports at least one `Cslib` module.
+This ensures all files are connected to the library's import graph and receive `Cslib.Init`. -/
+@[env_linter]
+def checkInitImports : Batteries.Tactic.Lint.Linter where
```
chenson2018: Are there other environment linters that ignore the declaration in this this way? If this fails, is it going to duplicate the error for every declaration in the import? My inclination is that this is not the right way to format this test.

Can you also explain how the logic ensures `Cslib.Init` is imported? I am not understanding how this prevents a portion of the import graph being isolated from it.

scripts/check-init-imports.lean
```diff
@@ -0,0 +1,62 @@
...
+This script checks that all CSLib modules import at least one module from the Cslib namespace,
+ensuring they are connected to the library's import graph and receive Cslib.Init.
+
+Based on Mathlib's lint-style.lean script.
+-/
+
+open System
+
+/-- Parse import statements from a file using simple text parsing -/
+def findImports (path : FilePath) : IO (Array String) := do
```
chenson2018: This seems brittle. There is already [findImports](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Tactic/Linter/DirectoryDependency.html#findImports) which I think captures your intent better.

scripts/CheckInitImports.lean
```diff
@@ -0,0 +1,62 @@
...
+    IO.println "✓ All modules import at least one Cslib module"
+    return 0
+  else
+    IO.eprintln s!"error: the following {modulesWithoutCslibImports.size} module(s) do not import any Cslib module:"
+    for module in modulesWithoutCslibImports do
+      IO.eprintln s!"  {module}"
+    IO.eprintln "\nThese modules should import Cslib.Init or another Cslib module."
+    return modulesWithoutCslibImports.size.toUInt32
+
+def main : IO UInt32 := checkInitImports
```
chenson2018: I think if we make this into a test that this works fine and we don't need the environment linter? Again, please clarify the logic of the test first.

lakefile.toml
```diff
@@ -6,8 +6,10 @@ testDriver = "CslibTests"
 [leanOptions]
 weak.linter.mathlibStandardSet = true
 weak.linter.flexible = true
-# The next three linters are provided by Mathlib, but don't work here.
-# This should be fixed, see discussion at https://leanprover.zulipchat.com/#narrow/channel/113488-general/topic/weak.2Elinter.2EmathlibStandardSet.20and.20lint-style-action/with/544726745
+# The next three linters are provided by Mathlib, but don't work with CSLib.
+# pythonStyle and allScriptsDocumented are Mathlib-specific, so we keep them disabled.
+# checkInitImports: we now have our own CSLib-specific implementation in
+# Cslib.Foundations.Lint.Basic (marked with @[env_linter], so it runs automatically).
```
chenson2018: Just to clarify, it is not the case that `@[env_linter]` makes the test run. The environment linters that run are those specified in [this file](https://github.com/leanprover/cslib/blob/main/CslibTests/Lint.lean).

chenson2018 (changes requested):

Cslib/Foundations/Lint/Basic.lean
jessealama: My thinking was to mimic how Mathlib does it, where `Mathlib.Init` is placed at certain high-level points in its hierarchy, in particular in the linters. This environment linter by itself does't guarantee that a module isn't detached from CSLib; we need the text-based linter (`scripts/check-init-imports.lean`), too.

jessealama: Ah, I think I misunderstood what the setup is in Mathlib. We don't need (or want!) the environment linter. I'll remove it.

scripts/check-init-imports.lean
jessealama: Ah, yes,I didn't know about that one. Let's use it.

scripts/CheckInitImports.lean
jessealama: Agree; I misunderstood the Mathlib setup. We don't need the environment linter.

jessealama:
Thanks! I think I've addressed the feedback. Can you take another look?

scripts/check-init-imports.lean
```diff
@@ -0,0 +1,62 @@
...
+    -- Convert module name to file path (replace . with /)
+    let pathParts := module.components.map toString
+    let path := mkFilePath pathParts |>.addExtension "lean"
+
+    -- Get imports for this module using Mathlib's robust parser
+    let imports ← findImports path
+
+    -- Check if any import starts with Cslib
+    let hasCslibImport := imports.any fun imp =>
+      match imp.components with
```
chenson2018: small nit, you can use [Lean.Name.getRoot](https://leanprover-community.github.io/mathlib4_docs/Init/Meta/Defs.html#Lean.Name.getRoot) instead of pattern matching

chenson2018:
Thanks for the work, I think this is moving in the right direction! A couple of follow up questions.

Regarding the logic for checking imports, let me make sure I have it straight:
- first you grab all the imports from `Cslib.lean`
- you then check that the only of these without a `Cslib` import are the current exceptions for Mathlib compatibility
- this implies that they all import `Cslib.Init` (because this is the only file excluded from `Cslib.lean`??)

With the exceptions existing though, do you need to change `hasCslibImport` to check for an import that is in the `Cslib` namespace *and* isn't on the exception list? Does this still work if you add `Cslib.Init` to `Cslib.lean`?

I also don't think you've made `scripts/check-init-imports.lean` run for `lake test`?

It would also be nice to go ahead and address #125 here since it affects the logic of checking `Cslib.Init` dependency. This should be much simpler in comparison. Edit: To answer your question from over there, I think maybe it is easiest if you write a short Lean function that checks this?

scripts/check-init-imports.lean
jessealama: Ah, good spot, I didn't know about that one.

jessealama:
> Thanks for the work, I think this is moving in the right direction! A couple of follow up questions.
> 
> Regarding the logic for checking imports, let me make sure I have it straight:
> 
> * first you grab all the imports from `Cslib.lean`
> * you then check that the only of these without a `Cslib` import are the current exceptions for Mathlib compatibility
> * this implies that they all import `Cslib.Init` (because this is the only file excluded from `Cslib.lean`??)
> 
> With the exceptions existing though, do you need to change `hasCslibImport` to check for an import that is in the `Cslib` namespace _and_ isn't on the exception list? Does this still work if you add `Cslib.Init` to `Cslib.lean`?
> 
> I also don't think you've made `scripts/check-init-imports.lean` run for `lake test`?
> 
> It would also be nice to go ahead and address #125 here since it affects the logic of checking `Cslib.Init` dependency. This should be much simpler in comparison. Edit: To answer your question from over there, I think maybe it is easiest if you write a short Lean function that checks this?

Thanks! I've refactored the import checking logic to filter out exceptions upfront, and added that to `lake test` (with a new test driver). I think I've addressed the issue in #125, too. Can you take another look when you have a moment?

chenson2018:
I just pushed a commit that made some alterations to `scripts/CheckInitImports.lean` to ensure it checks transitive imports. (I used the `ImportGraph` module, which I think simplifies what we were doing by hand).

I will take a look at `scripts/CheckCslibComplete.lean` and the test runner next.

chenson2018:
@jessealama Okay, I think we are nearly done here. I would appreciate you taking a look and testing. Here are the highlights of what I just pushed:

- I realized your initial suggestion of using `lake exe mk_all --check` was the right thing to do, and in fact there is already an option for this in `lean-action`, so I've enabled this and removed `scripts/CheckCslibComplete.lean`. Sorry for the churn, I should have realized this earlier!
- following suit with Mathlib, I do require `Cslib.Init` to be imported in `Cslib.lean`. 
- The test runner you added was missing the original tests from `CslibTests`, I have added this back.

I am pretty happy with everything here. My only hesitation is being unsure about the setup of the test runner, which feels a bit nonstandard in how it includes a `lean_exe` target and uses `IO.Process.spawn`. I have asked a related question [on Zulip](https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/MathlibTest.2EMathlibTestExecutable/with/547674520). Let's wait a bit to see if that generates any useful suggestions.

chenson2018:
One final consideration: do we actually remove the `weak` from lakefile options? Either way is not great for beginners IMO:
- if we keep `weak` and someone doesn't import Init, they find out in CI and potentially get slammed with lint errors
- if we remove them, it maybe gets caught earlier, but with a confusing error message

chenson2018:
@fmontesi Can you please review this before merging anything else? Because a change to testing merged at the same time as some other PRs, there are some failures on `main`. Since this also changes testing, I'd like to fix this here (and ask that folks merge main on other open PRs)

PR #135, opened 2025-10-28
chore: make various automata accept `Language _` rather than `Set (List _)`

ctchou:
This patch makes various automata (on finite words) accept `Language _` rather than `Set (List _)`.  Note that this requires us to explicitly declare `Symbol` as a type, because `Symbol` is currently a constant in the namespace `Language` in mathlib.  Just to be complete, I went ahead and provided explicit declarations for all `{...}` parameters of theorems.

Also, I reversed the direction of each `accepts_mem_language` theorem and renamed them `men_language`, since this seems more idiomatic in Lean.

### DIFF
--- a/CslibTests/DFA.lean
+++ b/CslibTests/DFA.lean
@@ -43,7 +43,7 @@ def tr (floor : Floor) (dir : Direction) : Floor :=
   | .two, .up => .two
   | .two, .down => .one
 
-def elevator : DFA Floor Direction := {
+def elevator : Cslib.DFA Floor Direction := {
   tr := tr
   start := .one
   accept := { Floor.one }


Cslib/Computability/Automata/DFA.lean
```diff
@@ -52,13 +54,13 @@ def Accepts (dfa : DFA State Symbol) (xs : List Symbol) :=
...
-def language (dfa : DFA State Symbol) : Set (List Symbol) :=
+def language (dfa : DFA State Symbol) : Language Symbol :=
   { xs | dfa.Accepts xs }
 
-/-- A string is accepted by a DFA iff it is in the language of the DFA. -/
+/-- A string is in the language of a DFA iff it is accepted by the DFA. -/
 @[scoped grind _=_]
-theorem accepts_mem_language (dfa : DFA State Symbol) (xs : List Symbol) :
-  dfa.Accepts xs ↔ xs ∈ dfa.language := by rfl
+theorem mem_language (dfa : DFA State Symbol) (xs : List Symbol) :
```
fmontesi, line 62: I'm ok with reversing the iff, but shouldn't it be called `mem_language_accepts` then? That seems to be what the mathlib guidelines imply.

fmontesi: (same for the other automata)

ctchou: My impression from reading mathlib is that `men_...` theorems mention only the LHS of the iff.  For example:
https://leanprover-community.github.io/mathlib4_docs/Mathlib/Computability/DFA.html#DFA.mem_acceptsFrom
https://leanprover-community.github.io/mathlib4_docs/Mathlib/Computability/DFA.html#DFA.mem_accepts
Note that the latter is the exact analogue of our DFA.mem_language.  Also see all the `mem_...` theorems in:

chenson2018:
Can you check if any of the `grind` annotations can switch from `_=_` to just `=` after reversing directions?

ctchou:
Will do.  Is `=` to be preferred over `_ = _`?

ctchou:
@chenson2018, just checked.  Switching from `_=_` to just `=` in grind annotations works fine after reversing direction.  Actually, only DFA.lean had `grind `_=_` .

chenson2018:
The CI failure should be resolved by #131. Please do not merge this until that is done.

chenson2018:
> Will do. Is `=` to be preferred over `_ = _`?

Sometimes `_ = _` can be unnecessarily expensive in checking both directions.

Cslib/Computability/Automata/DFA.lean
fmontesi: Huh.. that's so weird to me: what if something can be characterised in two different ways?

But I'll accept this, ok.

PR #137, opened 2025-10-29
chore: add PR write permissions to lint workflow

chenson2018:
WIP: I'm testing to see if this permission is the reason for the silent failure for this workflow.


chenson2018:
This doesn't seem to be the right permission, closing this.

PR #143, opened 2025-11-04
feat(Cslib/Logics/LinearLogic/CLL/PhaseSemantics/Basic.lean): add properties of additive connectives

m-ow:
This PR adds several basic properties for the additive connectives (`oplus` and `withh`), such as commutativity, associativity, and relations to `0` and `⊤`.

### TODO 
Distributivity Properties

### DIFF
--- a/Cslib/Logics/LinearLogic/CLL/PhaseSemantics/Basic.lean
+++ b/Cslib/Logics/LinearLogic/CLL/PhaseSemantics/Basic.lean
@@ -528,16 +528,16 @@ def quest (X : Fact P) : Fact P := dualFact (X⫠ ∩ I)
 
 /-! ### Properties of Additives -/
 
-lemma plus_of_with : (G ⊕ H : Fact P) = (Gᗮ & Hᗮ)ᗮ := by
+lemma plus_eq_with_dual : (G ⊕ H : Fact P) = (Gᗮ & Hᗮ)ᗮ := by
   apply SetLike.coe_injective
   rw [oplus, withh]
   aesop
 
-lemma with_of_plus : (G & H : Fact P) = (Gᗮ ⊕ Hᗮ)ᗮ := by simp [plus_of_with]
+lemma with_eq_plus_dual : (G & H : Fact P) = (Gᗮ ⊕ Hᗮ)ᗮ := by simp [plus_eq_with_dual]
 
-lemma neg_plus {G H : Fact P} : (G ⊕ H)ᗮ = Gᗮ & Hᗮ := by rw [plus_of_with, neg_neg]
+lemma neg_plus {G H : Fact P} : (G ⊕ H)ᗮ = Gᗮ & Hᗮ := by rw [plus_eq_with_dual, neg_neg]
 
-lemma neg_with {G H : Fact P} : (G & H)ᗮ = Gᗮ ⊕ Hᗮ := by rw [with_of_plus, neg_neg]
+lemma neg_with {G H : Fact P} : (G & H)ᗮ = Gᗮ ⊕ Hᗮ := by rw [with_eq_plus_dual, neg_neg]
 
 lemma with_comm : (G & H : Fact P) = H & G :=
   SetLike.coe_injective <| by simp [withh, Set.inter_comm]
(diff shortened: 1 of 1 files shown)


Cslib/Logics/LinearLogic/CLL/PhaseSemantics/Basic.lean
```diff
@@ -524,6 +526,34 @@ defined as the dual of the intersection of the orthogonal with the idempotents.
...
 @[inherit_doc] prefix:100 " ʔ " => quest
 
+/-! ### Properties of Additives -/
+
+lemma plus_of_with {G H : Fact P} : oplus G H = (Gᗮ & Hᗮ)ᗮ := by
+  refine SetLike.coe_injective ?_
+  rw [oplus, withh]
+  ext m; constructor
+  · simp; grind
+  · simp; grind
```
tannerduve:
```suggestion
  ext m; constructor <;> aesop
```
`aesop` closes both of these, in general avoid [non-terminal simp](https://leanprover-community.github.io/extras/simp.html#non-terminal-simps). otherwise this LGTM but @chenson2018 @fmontesi are codeowners. 

in general instead of non terminal simp you can use `simp?` to see a list `[h1, h2,..., hn]` of simp lemmas that would be used if you called simp, and then just call `simp only [h1, h2,..., hn]`. but aesop happens to work here.

chenson2018: This can be just `aesop`, I have commited this change.

Cslib/Logics/LinearLogic/CLL/PhaseSemantics/Basic.lean
```diff
@@ -524,6 +526,40 @@ defined as the dual of the intersection of the orthogonal with the idempotents.
...
+
+lemma neg_with {G H : Fact P} : (G & H)ᗮ = Gᗮ ⊕ Hᗮ := by rw [with_of_plus, neg_neg]
+
+lemma with_comm {G H : Fact P} : withh G H = H & G :=
+  SetLike.coe_injective <| by simp [withh, Set.inter_comm]
+
+@[simp] lemma with_assoc {G H K : Fact P} : withh (G & H) K = G & (H & K) := by
+  apply SetLike.coe_injective
+  rw [withh, coe_min]
+  exact Set.inter_assoc ..
```
chenson2018: To explain why I changed this as it may not be immediately obvious. You originally had `simp only [withh, coe_min, Set.inter_assoc]` here. The convention as Tanner mentioned is to avoid "non-terminal simp". So `simp only` should be used in the middle of proofs, but at the end of a branch like this, we should just use `simp`. However, making that change here breaks the proof, likely because some other `simp` rule is applied first. This seems brittle, so I changed to proof to be explicit.

Cslib/Logics/LinearLogic/CLL/PhaseSemantics/Basic.lean
```diff
@@ -524,6 +526,40 @@ defined as the dual of the intersection of the orthogonal with the idempotents.
 def quest (X : Fact P) : Fact P := dualFact (X⫠ ∩ I)
 @[inherit_doc] prefix:100 " ʔ " => quest
 
+/-! ### Properties of Additives -/
+
+lemma plus_of_with {G H : Fact P} : oplus G H = (Gᗮ & Hᗮ)ᗮ := by
```
chenson2018: I think that in the signature of all of these lemmas that you should be using the notation, not the names directly, e.g. `(G ⊕ H : Fact P) = (Gᗮ & Hᗮ)ᗮ` here.

chenson2018: Re: your question of naming for unicode, what you did in using the identifiers is standard. I'm not sure about the naming convention for these `a_of_b` lemmas. I'll defer to @fmontesi.

Cslib/Logics/LinearLogic/CLL/PhaseSemantics/Basic.lean
m-ow: Thanks! I appreciate your help

Cslib/Logics/LinearLogic/CLL/PhaseSemantics/Basic.lean
m-ow: I've just pushed a commit to address this. Thanks!

chenson2018 (approved):

Cslib/Logics/LinearLogic/CLL/PhaseSemantics/Basic.lean
fmontesi: `plus_of_with` sounds a bit weird, maybe `plus_eq_with_dual` or just `plus_eq_with`?

m-ow: Thanks @fmontesi and @chenson2018!

- `plus_of_with` is now `plus_eq_with_dual`
- `with_of_plus` is now `with_eq_plus_dual`

I also updated all the `simp` proofs that were using the old names.

PR #142, opened 2025-11-03
feat: unbundled design for automata theory

ctchou:
This patch ports automata theory to an unbundled design, building on top of https://github.com/leanprover/cslib/pull/141.  The main changes are:
* Finite-ness assumptions are removed.  There are no more DFA, NFA, and EpsilonNFA.  In their stead, the theory works with DA, NA, and EpsilonNA directly.
* Acceptance conditions are separated out into the file Accept.lean and are made orthogonal to automata proper (DA and NA).
* Buchi and Muller acceptance conditions are introduced to illustrate the last point, though no theorems about them have been proved.
* All previous results in automata theory (the equivalence between DA and NA on finite words, the closure properties of regular languages under boolean operations, etc) are ported to the new design.

PR #141, opened 2025-10-31
feat: add Cslib/Computability/Languages/RegularLanguage.lean and associated changes

ctchou:
This patch proves the closure properties of regular languages under boolean operations.  We first prove that `Language.IsRegular`, which is defined in mathlib, can also be characterized using cslib's DFA and NFA.  We then prove those closure properties using `Cslib.DFA`.  The key idea is a product construction for `DA`, which allows us to prove that regular languages are closed under both union and intersection (called `add` and `inf` in Language).  The definition of `DFA.mtr` is moved to `DA.mtr`, so that it can be applied to `DA` in general.  Some miscellaneous properties about `Language` are also proved.

The closure properties under concatenation and Kleene star will be proved later.  They require some nontrivial work.

This PR depends on https://github.com/leanprover/cslib/pull/135.

PR #144, opened 2025-11-07
feat: new design for automata theory, results on regular languages, and deterministic labelled transition systems

fmontesi:
This PR merges the developments that @ctchou and I carried out in the `automata` branch, #141, and #142, as discussed on Zulip ([#CSLib > Question Lean structure and extends](https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/Question.20Lean.20structure.20and.20extends/with/554287177)). If this is merged, the `automata` branch and the two mentioned PRs should be erased.

The design based on `Acceptor` and structure extension works pretty well. Many definitions could be removed because of the `Acceptor` typeclass, and some proofs got easier thanks to the combination of structure extension, record updates, and grind.

Co-authored with @ctchou.

### DIFF
--- a/Cslib/Computability/Automata/Acceptor.lean
+++ b/Cslib/Computability/Automata/Acceptor.lean
@@ -9,9 +9,9 @@ import Cslib.Computability.Languages.Language
 namespace Cslib.Automata
 
 /-- An `Acceptor` is a machine that recognises strings (lists of symbols in an alphabet). -/
-class Acceptor (α : Type _) (Symbol : outParam (Type _)) where
+class Acceptor (A : Type _) (Symbol : outParam (Type _)) where
   /-- Predicate that establishes whether a string `xs` is accepted. -/
-  Accepts (a : α) (xs : List Symbol) : Prop
+  Accepts (a : A) (xs : List Symbol) : Prop
 
 namespace Acceptor
 
@@ -19,12 +19,12 @@ variable {Symbol : Type _}
 
 /-- The language of an `Acceptor` is the set of strings it `Accepts`. -/
 @[scoped grind .]
-def language [Acceptor α Symbol] (a : α) : Language Symbol :=
+def language [Acceptor A Symbol] (a : A) : Language Symbol :=
   { xs | Accepts a xs }
 
 /-- A string is in the language of an acceptor iff the acceptor accepts it. -/
 @[scoped grind =]
-theorem mem_language [Acceptor α Symbol] (a : α) (xs : List Symbol) :
+theorem mem_language [Acceptor A Symbol] (a : A) (xs : List Symbol) :
   xs ∈ language a ↔ Accepts a xs := Iff.rfl
 
 end Acceptor
--- a/Cslib/Computability/Automata/DAToNA.lean
+++ b/Cslib/Computability/Automata/DAToNA.lean
@@ -33,17 +33,18 @@ namespace FinAcc
 def toNAFinAcc (a : DA.FinAcc State Symbol) : NA.FinAcc State Symbol :=
   { a.toNA with accept := a.accept }
 
-open scoped Acceptor FLTS NA.FinAcc in
+open Acceptor in
+open scoped FLTS NA.FinAcc in
 /-- The `NA` constructed from a `DA` has the same language. -/
-@[scoped grind =]
+@[scoped grind _=_]
 theorem toNAFinAcc_language_eq {a : DA.FinAcc State Symbol} :
-    Acceptor.language a = Acceptor.language a.toNAFinAcc := by
+    language a.toNAFinAcc = language a := by
   ext xs
   constructor
+  · grind
   · intro _
     use a.start
     grind
-  · grind
 
 end FinAcc
 
--- a/Cslib/Computability/Automata/EpsilonNAToNA.lean
+++ b/Cslib/Computability/Automata/EpsilonNAToNA.lean
@@ -43,7 +43,7 @@ def toNAFinAcc (a : εNA.FinAcc State Symbol) : NA.FinAcc State Symbol where
 open Acceptor in
 open scoped NA.FinAcc in
 /-- Correctness of `toNAFinAcc`. -/
-@[scoped grind =]
+@[scoped grind _=_]
 theorem toNAFinAcc_language_eq {ena : εNA.FinAcc State Symbol} :
     language ena.toNAFinAcc = language ena := by
   ext xs
--- a/Cslib/Computability/Automata/EpsilonNA.lean
+++ b/Cslib/Computability/Automata/EpsilonNA.lean
@@ -34,6 +34,8 @@ namespace εNA
 structure FinAcc (State Symbol : Type*) extends εNA State Symbol where
   accept : Set State
 
+namespace FinAcc
+
 /-- An `εNA.FinAcc` accepts a string if there is a saturated multistep accepting derivative with
 that trace from the start state. -/
 @[scoped grind =]
@@ -41,6 +43,8 @@ instance : Acceptor (FinAcc State Symbol) Symbol where
   Accepts (a : FinAcc State Symbol) (xs : List Symbol) :=
     ∃ s ∈ a.εClosure a.start, ∃ s' ∈ a.accept,
     a.saturate.MTr s (xs.map (some ·)) s'
+
+end FinAcc
 
 end εNA
 
--- a/Cslib/Computability/Automata/NAToDA.lean
+++ b/Cslib/Computability/Automata/NAToDA.lean
@@ -32,8 +32,9 @@ def toDAFinAcc (a : NA.FinAcc State Symbol) : DA.FinAcc (Set State) Symbol :=
   { a.toDA with accept := { S | ∃ s ∈ S, s ∈ a.accept } }
 
 open Acceptor in
+open scoped DA.FinAcc LTS in
 /-- The `DA` constructed from an `NA` has the same language. -/
-@[scoped grind =]
+@[scoped grind _=_]
 theorem toDAFinAcc_language_eq {na : NA.FinAcc State Symbol} :
   language na.toDAFinAcc = language na := by
   ext xs
@@ -42,7 +43,7 @@ theorem toDAFinAcc_language_eq {na : NA.FinAcc State Symbol} :
   Moving from `nightly-2025-09-15` to `nightly-2025-10-19` required
   increasing the number of allowed splits.
   -/
-  open DA.FinAcc LTS in grind (splits := 11)
+  grind (splits := 11)
 
 end FinAcc
 
--- a/Cslib/Computability/Automata/OmegaAcceptor.lean
+++ b/Cslib/Computability/Automata/OmegaAcceptor.lean
@@ -9,9 +9,9 @@ import Cslib.Computability.Languages.OmegaLanguage
 namespace Cslib.Automata
 
 /-- An `ωAcceptor` is a machine that recognises infinite sequences of symbols. -/
-class ωAcceptor (α : Type _) (Symbol : outParam (Type _)) where
+class ωAcceptor (A : Type _) (Symbol : outParam (Type _)) where
   /-- Predicate that establishes whether a string `xs` is accepted. -/
-  Accepts (a : α) (xs : ωSequence Symbol) : Prop
+  Accepts (a : A) (xs : ωSequence Symbol) : Prop
 
 namespace ωAcceptor
 
@@ -19,12 +19,12 @@ variable {Symbol : Type _}
 
 /-- The language of an `ωAcceptor` is the set of sequences it `Accepts`. -/
 @[scoped grind .]
-def language [ωAcceptor α Symbol] (a : α) : ωLanguage Symbol :=
+def language [ωAcceptor A Symbol] (a : A) : ωLanguage Symbol :=
   { xs | Accepts a xs }
 
 /-- A string is in the language of an acceptor iff the acceptor accepts it. -/
 @[scoped grind =]
-theorem mem_language [ωAcceptor α Symbol] (a : α) (xs : ωSequence Symbol) :
+theorem mem_language [ωAcceptor A Symbol] (a : A) (xs : ωSequence Symbol) :
   xs ∈ language a ↔ Accepts a xs := Iff.rfl
 
 end ωAcceptor


fmontesi:
bot fix style

chenson2018:
> bot fix style

Does this work??? Very cool if so. I was surprised to see the bot commenting at all actually, it had been silently failing to post recently... did you do something to fix it?

Cslib/Computability/Automata/Acceptor.lean
```diff
@@ -0,0 +1,32 @@
...
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Fabrizio Montesi
+-/
+
+import Cslib.Computability.Languages.Language
+
+namespace Cslib.Automata
+
+/-- An `Acceptor` is a machine that recognises strings (lists of symbols in an alphabet). -/
+class Acceptor (α : Type _) (Symbol : outParam (Type _)) where
```
chenson2018: It is _mildly_ jarring to half follow the Mathlib convention of lowercase Greek for type variables.

Cslib/Computability/Automata/DAToNA.lean
```diff
@@ -0,0 +1,50 @@
...
+
+/-- The `NA` constructed from a `DA` has the same language. -/
+@[scoped grind =]
+theorem toNAFinite_language_eq {a : DA.Finite State Symbol} :
+    Acceptor.language a = Acceptor.language a.toNAFinite := by
+  ext xs
+  refine ⟨?_, ?_⟩
+  · refine fun h => ⟨a.start, ?_⟩
+    open Acceptor FLTS in grind
+  · open NA.Finite Acceptor FLTS in grind
```
chenson2018: For reasons of readability:
```suggestion
open scoped Acceptor FLTS NA.Finite in
/-- The `NA` constructed from a `DA` has the same language. -/
@[scoped grind =]
theorem toNAFinite_language_eq {a : DA.Finite State Symbol} :
    Acceptor.language a = Acceptor.language a.toNAFinite := by
  ext xs
  constructor
  · intro _
    use a.start
    grind
  · grind
```

Cslib/Computability/Automata/EpsilonNAToNA.lean
```diff
@@ -0,0 +1,54 @@
...
+
+/-! # Translation of εNA into NA -/
+
+namespace Cslib
+
+/-- Converts an `LTS` with Option labels into an `LTS` on the carried label type, by removing all
+ε-transitions. -/
+@[local grind =]
+private def LTS.noε (lts : LTS State (Option Label)) : LTS State Label where
+  Tr := fun s μ s' => lts.Tr s (some μ) s'
```
chenson2018:
```suggestion
  Tr s μ s' := lts.Tr s (some μ) s'
```

Cslib/Computability/Automata/EpsilonNAToNA.lean
```diff
@@ -0,0 +1,54 @@
...
+@[local grind .]
+private lemma LTS.noε_saturate_tr
+  {lts : LTS State (Option Label)} {h : μ = some μ'} :
+  lts.saturate.Tr s μ s' ↔ lts.saturate.noε.Tr s μ' s' := by
+  simp only [LTS.noε]
+  grind
+
+@[scoped grind =]
+lemma LTS.noε_saturate_mTr {lts : LTS State (Option Label)} :
+  lts.saturate.MTr s (μs.map (some ·)) = lts.saturate.noε.MTr s μs := by
```
chenson2018: Writing `some ·` is redundant since it is already the function you want. You could either write

```suggestion
  lts.saturate.MTr s (μs.map some) = lts.saturate.noε.MTr s μs := by
```
or

```suggestion
  lts.saturate.MTr s (μs.map (↑)) = lts.saturate.noε.MTr s μs := by
```

Cslib/Computability/Automata/EpsilonNAToNA.lean
```diff
@@ -0,0 +1,54 @@
...
+  Tr := a.saturate.noε.Tr
+
+/-- Correctness of `toNAFinite`. -/
+@[scoped grind =]
+theorem toNAFinite_language_eq {ena : εNA.Finite State Symbol} :
+    Acceptor.language ena.toNAFinite = Acceptor.language ena := by
+  ext xs
+  have : ∀ s s', ena.saturate.MTr s (xs.map (some ·)) s' = ena.saturate.noε.MTr s xs s' := by
+    simp [LTS.noε_saturate_mTr]
+  open NA.Finite Acceptor in grind
```
chenson2018: I'd somewhat prefer we do an `open scoped` for the whole theorem, because this gets quickly cluttered.

Cslib/Computability/Languages/RegularLanguage.lean
```diff
@@ -0,0 +1,128 @@
...
+theorem IsRegular.iff_cslib_nfa {l : Language Symbol} :
+    l.IsRegular ↔ ∃ State : Type, ∃ _ : Finite State,
+      ∃ nfa : Cslib.Automata.NA.Finite State Symbol, Cslib.Automata.Acceptor.language nfa = l := by
+  rw [IsRegular.iff_cslib_dfa]; constructor
+  · rintro ⟨State, h_fin, ⟨da, acc⟩, rfl⟩
+    use State, h_fin, ⟨da.toNA, acc⟩
+    grind
+  · rintro ⟨State, _, na, rfl⟩
+    use Set State, inferInstance, na.toDAFinite
+    exact Cslib.Automata.NA.Finite.toDAFinite_language_eq
```
chenson2018:
```suggestion
    grind
```

Cslib/Computability/Languages/RegularLanguage.lean
```diff
@@ -0,0 +1,128 @@
...
+
+-- From this point onward we will use only automata from Cslib in the proofs.
+open Cslib
+
+@[simp]
+theorem IsRegular.compl {l : Language Symbol} (h : l.IsRegular) : (lᶜ).IsRegular := by
+  rw [IsRegular.iff_cslib_dfa] at h ⊢
+  obtain ⟨State, _, ⟨da, acc⟩, rfl⟩ := h
+  use State, inferInstance, ⟨da, accᶜ⟩
+  ext ; grind
```
chenson2018:
```suggestion
  grind
```

Cslib/Foundations/Semantics/LTS/FLTSToLTS.lean
```diff
@@ -0,0 +1,46 @@
...
+import Cslib.Foundations.Semantics.LTS.Basic
+
+variable {State : Type _} {Label : Type _}
+
+namespace Cslib.FLTS
+
+/-- `FLTS` is a special case of `LTS`. -/
+@[scoped grind =]
+def toLTS (flts : FLTS State Label) : LTS State Label where
+  Tr := fun s1 μ s2 => flts.tr s1 μ = s2
```
chenson2018:
```suggestion
  Tr s1 μ s2 := flts.tr s1 μ = s2
```

ctchou:
Can we not use "Finite" in `DA.Finite` and `NA.Finite`?  This conflicts with the use of "Finite" in `Finite T` (where `T` is a type), which also appears in our code.  The "Finite" in `{DA,NA}.Finite` refers to the finite-ness of the accepted run, while the "Finite" in `Finite T` refers to the finite-ness of the cardinality of `T`.  Now we end up having expressions like this:

`
theorem IsRegular.iff_cslib_dfa {l : Language Symbol} :
    l.IsRegular ↔ ∃ State : Type, ∃ _ : Finite State,
      ∃ dfa : Cslib.Automata.DA.Finite State Symbol, Cslib.Automata.Acceptor.language dfa = l := by
`

in which the two "Finite"s mean totally different things.  Even more confusing is that "finite automata" is a well-established terminology where the "finite" means that at least the state space is finite and perhaps the alphabet is finite as well.

That is why I chose the name `FinAcc` in my PR.  `FinRun` is another possible name.

See the 1st commit of https://github.com/leanprover/cslib/pull/145:
https://github.com/leanprover/cslib/pull/145/commits/99fb99e48980c8efed9b985b77e5f5c018059e01
which renames {DA,NA}.Finite to {DA,NA}.FinAcc.

ctchou:
I suggest we merge Acceptor.lean and OmegaAcceptor.lean into a single `Accept.lean`.  They are imported together by both DA.lean and NA.lean, which are at the bottom of the import hierarchy. anyway  There is no point splitting them.

fmontesi:
Re FinAcc: agreed, I'll switch.

Re Accept: I need to think about it a bit. My thinking was that they're conceptually separate. Maybe we'll use Acceptor somewhere else, or maybe the different substructures of DA and NA will get bigger and merit their own files.

Re all the patches by @chenson2018 : will integrate.

ctchou:
For the `Finite` -> `FinAcc` renaming, I've already made a patch in cslib#145 at commit: https://github.com/leanprover/cslib/commit/99fb99e48980c8efed9b985b77e5f5c018059e01

fmontesi:
> > bot fix style
> 
> Does this work??? Very cool if so. I was surprised to see the bot commenting at all actually, it had been silently failing to post recently... did you do something to fix it?

It didn't work. It was too good not to try... :smile: 
I haven't done anything to fix it. :shrug:

Cslib/Computability/Automata/Acceptor.lean
fmontesi: I was in doubt myself, it gives some goosebumps. But I couldn't come up with a better name than alpha.. maybe `A` for automaton or `M` for machine?

fmontesi:
> For the `Finite` -> `FinAcc` renaming, I've already made a patch in cslib#145 at commit: [99fb99e](https://github.com/leanprover/cslib/commit/99fb99e48980c8efed9b985b77e5f5c018059e01)

That's great, thanks. I'll merge it as soon as I'm done applying all the other suggestions.

ctchou:
In the commit:
https://github.com/leanprover/cslib/pull/145/commits/4c1d61f6b172b335f661b1a97fb3654930be4428
I added the attributes `language` and `mem_language` to `{DA,NA}.FinAcc` so that we can write `{da,na}.language` rather than having to explicitly mention Acceptor.  For example, RegularLanguage.lean is improved as follows:

`
@@ -26,7 +25,7 @@ variable {Symbol : Type*}
 /-- A characterization of Language.IsRegular using Cslib.DA -/
 theorem IsRegular.iff_cslib_dfa {l : Language Symbol} :
     l.IsRegular ↔ ∃ State : Type, ∃ _ : Finite State,
-      ∃ dfa : Cslib.Automata.DA.FinAcc State Symbol, Cslib.Automata.Acceptor.language dfa = l := by
+      ∃ dfa : Cslib.Automata.DA.FinAcc State Symbol, dfa.language = l := by
   constructor
   · rintro ⟨State, h_fin, ⟨tr, start, acc⟩, rfl⟩
     let dfa := Cslib.Automata.DA.FinAcc.mk {tr, start} acc
@@ -40,7 +39,7 @@ theorem IsRegular.iff_cslib_dfa {l : Language Symbol} :
 /-- A characterization of Language.IsRegular using Cslib.NA -/
 theorem IsRegular.iff_cslib_nfa {l : Language Symbol} :
     l.IsRegular ↔ ∃ State : Type, ∃ _ : Finite State,
-      ∃ nfa : Cslib.Automata.NA.FinAcc State Symbol, Cslib.Automata.Acceptor.language nfa = l := by
+      ∃ nfa : Cslib.Automata.NA.FinAcc State Symbol, nfa.language = l := by
   rw [IsRegular.iff_cslib_dfa]; constructor
   · rintro ⟨State, h_fin, ⟨da, acc⟩, rfl⟩
     use State, h_fin, ⟨da.toNA, acc⟩
`

If you like this, I can make similar changes to {DA,NA}.{Buchi,Muller} as well.

fmontesi:
I thought and tried to do the same. My first impression was that we get better ergonomics with making them `abbrev`.

But I chose to avoid it for now because it adds redundant definitions to every automaton, which kinda kills the point of having a typeclass. There were discussions about supporting dot notation for instances of a class, which would solve our problem. I'd like to take this discussion up with the Lean developers first. (I'll do that now.)

So I'm just gonna cherry-pick https://github.com/leanprover/cslib/commit/99fb99e48980c8efed9b985b77e5f5c018059e01 for now (the `FinAcc` patch). We can always revisit this later.

fmontesi:
All done.

I've been thinking and searching more about the dot notation thing from the typeclass: if we really want to use dot notation for `language`, `Accepts`, etc., couldn't we make a macro to make all these thin abbreviations?

I'm not sure we should though.. we should ask the mathlibbers about their experience with this.

ctchou:
In my opinion, such "redundant" definitions and theorems are very much in the style of Lean mathlib.  For example, take a look at:
One could argue that all the `mem_...` theorems there are "redundant", for they are nothing but a slight re-statement of the definitions.  In fact, all the `..._def`s are actually theorems, not the real definitions, which are mostly instances very much like in our case.

More importantly, don't you find an expression like:
`
∃ dfa : Cslib.Automata.DA.FinAcc State Symbol, Cslib.Automata.Acceptor.language dfa = l 
`
just intolerable?  One advantage of bundling the accepting states in the acceptor is that we don't have to mention and quantifying over them explicitly in a theorem like the one above.  If we have to write `Cslib.Automata.Acceptor.language dfs`, what advantage have we got?

ctchou:
Here's an example from mathlib's `Language`:
`
instance : Mul (Language α) :=
  ⟨image2 (· ++ ·)⟩

theorem mul_def (l m : Language α) : l * m = image2 (· ++ ·) l m :=
  rfl

theorem mem_mul : x ∈ l * m ↔ ∃ a ∈ l, ∃ b ∈ m, a ++ b = x :=
  mem_image2
`
The {DA.NA}.FinAcc.{language,mem_language} follow exactly the same pattern.

fmontesi:
Not all of these examples are direct copies of the definitions/theorems in the class, but that's beyond the point: as I wrote, my problem is not with having the copied definitions for dot notation (btw, these definitions should always just be a reference to the class definitions/theorems), but more with doing so manually. I would just much rather explore a systematic way, e.g., supplying a simple annotation when one creates an instance that produces all the copies needed for dot notation automatically. I'm now discussing this on Zulip.

fmontesi:
Link to the Zulip discussion: [#Is there code for X? > Deriving dot notation from class instances](https://leanprover.zulipchat.com/#narrow/channel/217875-Is-there-code-for-X.3F/topic/Deriving.20dot.20notation.20from.20class.20instances/with/554475581)

ctchou:
Another response to the problem we are facing in this PR is to ask: Why is having an "acceptor" class preferable to the last design in https://github.com/leanprover/cslib/pull/142, which doesn't have an "acceptor" class?  What exactly is being gained from having an "acceptor" class?

BTW, this brings up another question: Why do you need two acceptor classes?  `Acceptor` and `OmegaAcceptor` are identical, except for the some types.  If you want to abstract the notion of "accepting" into an "acceptor" class, why not go all the way and have a single "acceptor" class and two different instantiations of it?

fmontesi:
1. Code deduplication.
2. Right you are, I could define

```
/-- An `Acceptor` is a machine that recognises strings (lists of symbols in an alphabet). -/
class Acceptor (α : Type _) (β : outParam (Type _)) where
  /-- Predicate that establishes whether a string `xs` is accepted. -/
  Accepts (a : α) (b : β) : Prop
```

Then we'd have things like:

```
instance : Acceptor (Buchi State Symbol) (ωSequence Symbol) where
  Accepts (a : Buchi State Symbol) (xs : ωSequence Symbol) := ∃ᶠ k in atTop, a.run xs k ∈ a.accept
```

which looks totally fine to me. I'll get to it, that's even more deduplication. :-)

ctchou:
Do we really get code deduplication?  If I still have to explicitly make instances of `language` and `mem_language` in the style of the `Mul` example from `Language`, what code deduplication do I get?  I asked the question on Zulip too:
https://leanprover.zulipchat.com/#narrow/channel/217875-Is-there-code-for-X.3F/topic/Deriving.20dot.20notation.20from.20class.20instances/near/554501460

ctchou:
Note that the `Mul` example from `Language` actually gets code deduplication, because there are many theorems already proved about `image2` which can now be instantiated to produce theorems about `Mul`.  (This actually happens in `Language`.)  But do we have any analogues here?  Are there nontrivial abstract theorems about `Acceptor` that can be proved and re-used by its instances?  This seems to me rather unlikely, because `Acceptor` contains only a predicate, a set defined by the predicate, and a theorem relating the predicate and the set.  What sophisticated theorems can be proved about them that are not already in mathlib?

fmontesi:
> 1. Code deduplication.
> 2. Right you are, I could define
> 
> ```
> /-- An `Acceptor` is a machine that recognises strings (lists of symbols in an alphabet). -/
> class Acceptor (α : Type _) (β : outParam (Type _)) where
>   /-- Predicate that establishes whether a string `xs` is accepted. -/
>   Accepts (a : α) (b : β) : Prop
> ```
> 
> Then we'd have things like:
> 
> ```
> instance : Acceptor (Buchi State Symbol) (ωSequence Symbol) where
>   Accepts (a : Buchi State Symbol) (xs : ωSequence Symbol) := ∃ᶠ k in atTop, a.run xs k ∈ a.accept
> ```
> 
> which looks totally fine to me. I'll get to it, that's even more deduplication. :-)

Mmh actually doesn't work quite as expected, as Acceptor.language returns a Language whereas OmegaAcceptor.language returns an OmegaLanguage... I'll keep them separate for the time being.

chenson2018 (approved):

ctchou:
Zulip thread: https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/deriving.20dot.20notation/with/554560824

PR #154, opened 2025-11-14
chore: Add helper lemma to address grind limitation in LTS.toFLTS_mem_mtr

jessealama:
This PR addresses issue #147 by adding a helper lemma that allows the `grind` tactic to complete the proof of `toFLTS_mem_mtr` without manual intervention.

## Changes

1. **Added `LTS.mem_foldl_setImage` to `Basic.lean`**: A new theorem that directly characterizes membership in `List.foldl lts.setImage` in terms of `MTr`, with the `@[grind =]` annotation.

2. **Simplified `toFLTS_mem_mtr` in `LTSToFLTS.lean`**: Removed the manual `rw` step and TODO comment.

This contributes to #147 but doesn't close it. The need for this helper lemma suggests that `grind`'s E-matching could potentially be enhanced to apply function extensionality under congruence automatically. It may be worth opening an issue in the Lean repository to discuss whether theorems like `LTS.mem_foldl_setImage` ought to be derived automatically from `setImageMultistep_foldl_setImage` and `mem_setImageMultistep`.

### DIFF
--- a/Cslib/Foundations/Semantics/LTS/LTSToFLTS.lean
+++ b/Cslib/Foundations/Semantics/LTS/LTSToFLTS.lean
@@ -28,8 +28,7 @@ original `LTS`. -/
 @[scoped grind =]
 theorem toFLTS_mem_mtr {lts : LTS State Label} {S : Set State} {s' : State} {μs : List Label} :
   s' ∈ lts.toFLTS.mtr S μs ↔ ∃ s ∈ S, lts.MTr s μs s' := by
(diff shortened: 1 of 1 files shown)


fmontesi:
Ah-ha! Thank you. I second @chenson2018's suggestion, it'd be useful to identify the general issue with an ME.

jessealama:
> Ah-ha! Thank you. I second @chenson2018's suggestion, it'd be useful to identify the general issue with an ME.

Good idea -- I'll see if I can massage this case into an ME. We can continue the conversation on Zulip.

PR #157, opened 2025-11-16
chore: change FinFun notation to not conflict with Finsupp

### LINKED ISSUE #138: Notation conflict with Mathlib.Finsupp (both use →₀)
import Cslib.Init

https://github.com/leanprover/cslib/blob/9dd5453bf85ddb0a05253a8a39dea5c9ed67083c/Cslib/Foundations/Data/FinFun.lean#L7-L8

chenson2018:
Closes #138.


chenson2018:
@fmontesi Please let me know if there is another non-conflicting arrow you'd prefer here.

fmontesi:
Mmh I don't really know yet, mainly because FinFun does exactly the same job as Finsupp but using computable definitions (so, very annoyingly, the signatures in the API are slightly different sometimes...), so I was using the same notation on purpose as a 'drop-in replacement'. But it's really not, at least not 100%, so a different symbol would make sense... will think about it.

PR #158, opened 2025-11-16
chore(CLL): notation conflict between Proposition.Equiv and ZMOD

### LINKED ISSUE #140: Syntax elaboration conflicts
import Cslib.Init

https://github.com/leanprover/cslib/blob/9dd5453bf85ddb0a05253a8a39dea5c9ed67083c/Cslib/Logics/LinearLogic/CLL/Basic.lean#L7-L8

chenson2018:
Closes #140.


fmontesi:
Shouldn't it really be `≡` though? My plan was to even introduce a typeclass for that symbol in any logic. It's a symbol also used a lot in semantics...

chenson2018:
To be clear, I am fine with any non-conflicting symbol. Is there any other available Unicode you'd be willing to compromise to use? The problem is that is when we have a conflict with a non-scoped Mathlib notation, it makes it impossible to import `Cslib.Init` so we lose all of our linting downstream of this file and have to think much more about imports for classes, tactics, etc.

It would be ideal if these were scoped notations in Mathlib, but because of the current limitation of documentation being unable to display scoped notations, this is unlikely to change.

(This all equally applies to #157)

chenson2018:
There is a unused symbol `≣` (U+2263, "Strictly Equivalent To") that is close. Besides being non-standard, I don't believe there is another meaning that would cause confusion?

fmontesi:
I guess this is because we are importing `Mathlib` from `Cslib.Init`? What's the reason for importing `Mathlib` there?

chenson2018:
> I guess this is because we are importing `Mathlib` from `Cslib.Init`? What's the reason for importing `Mathlib` there?

We import `Mathlib.Init` here primarily as a prerequisite to having the standard Mathlib linters.[^1] In CI I check that everywhere (transitively) imports `Cslib.Init`, with the only exceptions being files that have conflicts with `≡` and `→₀` notations.

[^1]: as [suggested on Zulip](https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/weak.20in.20lakefile/near/546654795) by you :)

The alternative is a headache. First, it means that files are free to silently not adhere to the style guide. More troublesome is what happens as the library grows and these files with syntax conflicts are no longer in leaf nodes of our import graph. Imagine that we leave these as is, compiling fine on their own. But if we then want to import these files into other parts of CSLib that are dependent on Mathlib, we can suddenly find ourselves unable to use notation. 

While it is unfortunate that it means sometimes deviating from standard notations, I think the alternative of inconsistently having compatibility with Mathlib is much worse. I have [asked on Zulip](https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/unscoped.20notations/with/553789290) about scoping more notations, but this is currently blocked by a core RFC. (Which I would love to try to work on myself, but there's a steep learning curve that will take time as this would be my first core contribution)

PR #160, opened 2025-11-17
chore: minimize all imports

chenson2018:
This PR does the following:

- remove a full `import Mathlib` from `Cslib.Foundations.Lint.Basic` that was then transitively imported everywhere
- add `import Mathlib.Tactic.Common` to `Cslib.Init.lean` so that we don't need to manually import these tactics each time while also reducing the number of build jobs
- when appropriate, change the imports of every file to match what is given by `lake exe shake` or record an exception in `scripts/noshake.json`
- Add `lake exe shake` in CI to maintain this going forward.

### DIFF
--- a/Cslib/Logics/LinearLogic/CLL/Basic.lean
+++ b/Cslib/Logics/LinearLogic/CLL/Basic.lean
@@ -5,8 +5,6 @@ Authors: Fabrizio Montesi
 -/
 
 import Cslib.Init
-import Aesop
-import Mathlib.Tactic.ApplyAt
 import Mathlib.Order.Notation
 import Mathlib.Order.Defs.Unbundled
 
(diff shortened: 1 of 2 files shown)


ctchou:
Presumably you used a script to identify the changes in this PR?  If so, perhaps that script can be added to the repo so that everyone can run it.  I for one often forget to run `#min_imports`.

chenson2018:
I actually did this by hand because it wasn't that long... but now that you mention it, I think `lake exe shake Cslib` works here. Let me try to set it up...

chenson2018:
Okay, that was pretty simple, great idea @ctchou!

Now if you run `lake exe shake Cslib`, it will list what imports should be changed and I have added this as a check to CI. Any false positives due to typeclasses and tactics can be excluded by listing them in `scripts/noshake.json` (which has a couple of examples already).

This slightly conflicts with imports in #157 and #158. Probably easier to review/merge those first.

chenson2018:
On second thought, if the notation changes in those other PRs are problematic maybe we should merge this first to minimize merge conflicts.

chenson2018:
At Kim's suggestion I have changed `Mathlib.Tactic` to `Mathlib.Tactic.Common`. Altogether this drastically reduces the number of build jobs from ~7000 to ~1600!

@fmontesi As an added bonus, this removes the syntax conflicts in #157 and #158! I still slightly worry that it could cause trouble down the line, but we are not forced to make that decision now (and it is relatively simple to change) so I am closing those PRs for now.

kim-em (approved):

PR #161, opened 2025-11-17
feat(Automata): remove specialised versions of (Omega)Acceptor lemmas from automata structures

fmontesi:


chenson2018:
(just an FYI, the PR title check is not expecting a space after `feat`)

fmontesi:
> (just an FYI, the PR title check is not expecting a space after `feat`)

*That* is what was bothering it then, thanks! :-)

ctchou (approved):
I did some cursory testing by manually doing `simp` at various places (even if it is not needed in the final proof).  The patch seems to work.

chenson2018 (approved):

PR #163, opened 2025-11-17
chore: update LICENSE with copyright holder

arademaker:
The license file didn't filled the copyright year and holder. I am just using the same information I used in the LICENSE of the website source code.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
+   Copyright 2025 CSLib Steering Committee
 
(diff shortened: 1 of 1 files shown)


fmontesi:
That line is not supposed to be filled, it's a template for headers. (Side note: when we refer to the group holding copyright over CSLib, it should be CSLib Contributors, not CSLib Steering Committee.)

chenson2018:
This isn't meant to be filled out, it is a template given in the explanation.

chenson2018:
(We managed to make the same comment within a second or so, sorry! I saw it pop up as I hit the comment button...)

arademaker:
My fault. Sorry for the mistake.

PR #165, opened 2025-11-18
feat: time monad + MergeSort analysis

sorrachai:
Introducing time monad and analysis of mergeSort, including correctness and running time.

### DIFF
--- a/Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
+++ b/Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
@@ -8,7 +8,6 @@ import Cslib.Algorithms.Lean.TimeM
 import Mathlib.Data.Nat.Cast.Order.Ring
 import Mathlib.Data.Nat.Lattice
 import Mathlib.Data.Nat.Log
-import Mathlib.Tactic.Bound
 
 
 /-!
@@ -67,49 +66,45 @@ abbrev IsSorted (l : List α) : Prop := Sorted (· ≤ ·) l
 /-- `x` is a minimum element of list `l` if `x ≤ b` for all `b ∈ l`. -/
 abbrev MinOfList (x : α) (l : List α) : Prop := ∀ b ∈ l, x ≤ b
 
-theorem mem_either_merge (xs ys : List α) (z : α)
-  (hz : z ∈ ⟪merge xs ys⟫) : z ∈ xs ∨ z ∈ ys := by
+@[grind →]
+theorem mem_either_merge (xs ys : List α) (z : α) (hz : z ∈ ⟪merge xs ys⟫) : z ∈ xs ∨ z ∈ ys := by
   fun_induction merge
-  · simp_all only [Pure.pure, pure, not_mem_nil, or_true]
-  · simp_all only [imp_false, Pure.pure, pure, not_mem_nil, or_false]
-  simp_all only [mem_cons, Bind.bind, tick, Pure.pure, pure, ret_bind,
-    decide_eq_true_eq]
-  split_ifs at hz <;> (simp_all ; grind)
+  · exact mem_reverseAux.mp hz
+  · left
+    exact hz
+  · simp_all only [Bind.bind, Pure.pure]
+    grind
 
-theorem min_all_merge (x : α) (xs ys : List α)
-(hxs : MinOfList x xs) (hys : MinOfList x ys) : MinOfList x ⟪merge xs ys⟫ := by
-  simp_all only [MinOfList]
-  intro a ha
-  apply mem_either_merge at ha
+theorem min_all_merge (x : α) (xs ys : List α) (hxs : MinOfList x xs) (hys : MinOfList x ys) :
+    MinOfList x ⟪merge xs ys⟫ := by
   grind
 
-theorem sorted_merge {l1 l2 : List α} (hxs : IsSorted l1)
-  (hys : IsSorted l2) : IsSorted ⟪merge l1 l2⟫ := by
-  fun_induction merge l1 l2 <;> try simpa
-  simp only [Bind.bind, tick, ret_bind, decide_eq_true_eq,IsSorted]
-  split_ifs
-  all_goals
-  simp_all only [forall_const, sorted_cons, Pure.pure, pure, ret_bind, and_true]
-  apply min_all_merge <;> grind
+theorem sorted_merge {l1 l2 : List α} (hxs : IsSorted l1) (hys : IsSorted l2) :
+    IsSorted ⟪merge l1 l2⟫ := by
+  fun_induction merge l1 l2 with
+  | case3 =>
+    simp only [Bind.bind, Pure.pure]
+    grind [sorted_cons]
+  | _ => simpa
 
 theorem mergeSort_sorted (xs : List α) : IsSorted ⟪mergeSort xs⟫ := by
   fun_induction mergeSort xs with
   | case1 x =>
-    simp only [IsSorted, Pure.pure, pure]
-    rcases x with _ | ⟨a, _ | ⟨b, rest⟩⟩ <;> simp [sorted_nil,sorted_singleton]; grind
-  | case2 _ _ _ _ _ ih2 ih1 =>
-    simp only [IsSorted, Bind.bind, ret_bind]
-    exact sorted_merge ih2 ih1
+    simp only [Pure.pure]
+    rcases x with _ | ⟨a, _ | ⟨b, rest⟩⟩ <;> grind [sorted_nil, sorted_singleton]
+  | case2 _ _ _ _ _ ih2 ih1 => exact sorted_merge ih2 ih1
 
 lemma merge_perm (l₁ l₂ : List α) : ⟪merge l₁ l₂⟫ ~ l₁ ++ l₂ := by
-  fun_induction merge <;> try simp
-  split_ifs with h
-  · simp_all only [cons_append, perm_cons]
-  · grind
+  fun_induction merge with
+  | case1 => simp
+  | case2 => simp
+  | case3 =>
+    simp only [Bind.bind, Pure.pure]
+    grind
 
 theorem mergeSort_perm (xs : List α) : ⟪mergeSort xs⟫ ~ xs := by
   fun_induction mergeSort xs with
-  | case1 => simp only [Pure.pure, pure, Perm.refl]
+  | case1 => simp
   | case2 x _ _ left right ih2 ih1 =>
     simp only [Bind.bind, ret_bind]
     calc
@@ -119,11 +114,8 @@ theorem mergeSort_perm (xs : List α) : ⟪mergeSort xs⟫ ~ xs := by
       _ ~ x := by simp only [take_append_drop, Perm.refl, left, right]
 
 /-- MergeSort is functionally correct. -/
-theorem mergeSort_correct (xs : List α) :
-  IsSorted ⟪mergeSort xs⟫ ∧ ⟪mergeSort xs⟫ ~ xs  := by
-  constructor
-  · apply mergeSort_sorted
-  · apply mergeSort_perm
+theorem mergeSort_correct (xs : List α) : IsSorted ⟪mergeSort xs⟫ ∧ ⟪mergeSort xs⟫ ~ xs :=
+  ⟨mergeSort_sorted xs, mergeSort_perm xs⟩
 
 end Correctness
 
@@ -141,33 +133,26 @@ def timeMergeSortRec : ℕ → ℕ
 | n@(_+2) => timeMergeSortRec (n/2) + timeMergeSortRec ((n-1)/2 + 1) + n
 
 /-- The ceiling of Nat.log 2 -/
+@[grind =]
 def clog2 (n : ℕ) : ℕ :=
   if n ≤ 1 then 0 else Nat.log 2 (n - 1) + 1
 
 /-- Key Lemma: ⌈log2 ⌈n/2⌉⌉ ≤ ⌈log2 n⌉ - 1 for n > 1 -/
+@[grind →]
 lemma clog2_half_le (n : ℕ) (h : n > 1) : clog2 ((n + 1) / 2) ≤ clog2 n - 1 := by
-  unfold clog2
-  split_ifs with h_small
-  · omega
-  · simp only [add_tsub_cancel_right, zero_le]
-  · omega
-  · simp only [add_tsub_cancel_right]
-    rw [← Nat.sub_mul_div (n + 1) 2 1]
-    simp only [mul_one, Nat.reduceSubDiff, Nat.log_div_base]
-    rw [Nat.sub_one_add_one ?_]
-    simp only [ne_eq, Nat.log_eq_zero_iff, Nat.not_ofNat_le_one, or_false, not_lt]
-    omega
+  have := Nat.sub_mul_div (n + 1) 2 1
+  grind [Nat.log_eq_zero_iff, Nat.log_div_base]
 
 /-- Same logic for the floor half: ⌈log2 ⌊n/2⌋⌉ ≤ ⌈log2 n⌉ - 1 -/
+@[grind →]
 lemma clog2_floor_half_le (n : ℕ) (h : n > 1) : clog2 (n / 2) ≤ clog2 n - 1 := by
   apply Nat.le_trans _ (clog2_half_le n h)
   simp only [clog2]
-  split_ifs with h_small
-  · rfl
-  · omega
-  · omega
-  · expose_names
-    grw [Nat.log_mono_right]
+  split_ifs
+  · grind
+  · grind
+  · grind
+  · grw [Nat.log_mono_right]
     grind
 
 private lemma some_algebra (n : ℕ) :
@@ -180,16 +165,12 @@ private lemma some_algebra (n : ℕ) :
   have t1 : n / 2 + 1 = N / 2 := by omega
   have t2 : (n + 1) / 2 + 1 = (N + 1) / 2 := by omega
   have t3 : n + 1 + 1 = N := by omega
-  rw [t1, t2, t3]
   let k := clog2 N
   have h_bound_l : clog2 (N / 2) ≤ k - 1 := clog2_floor_half_le N hN
   have h_bound_r : clog2 ((N + 1) / 2) ≤ k - 1 := clog2_half_le N hN
-  grw [h_bound_l,h_bound_r]
-  rw [←Nat.add_mul]
   have h_split : N / 2 + (N + 1) / 2 = N := by omega
-  rw [h_split]
-  bound
-
+  grw [t1, t2, t3, h_bound_l, h_bound_r, ←Nat.add_mul, h_split]
+  exact Nat.le_refl (N * (k - 1) + N)
 
 /-- Upper bound function for merge sort time complexity: `T(n) = n * ⌈log₂ n⌉` -/
 abbrev T (n : ℕ) : ℕ := n * clog2 n
@@ -197,39 +178,40 @@ abbrev T (n : ℕ) : ℕ := n * clog2 n
 /-- Solve the recurrence -/
 theorem timeMergeSortRec_le (n : ℕ) : timeMergeSortRec n ≤ T n := by
   fun_induction timeMergeSortRec with
-  | case1 => simp [T]
-  | case2 => simp
+  | case1 => grind
+  | case2 => grind
   | case3 n ih2 ih1 =>
     grw [ih1,ih2]
-    simp only [Nat.ofNat_pos, Nat.add_div_right, Nat.add_one_sub_one, Nat.succ_eq_add_one]
-    exact some_algebra n
-
+    have := some_algebra n
+    grind [Nat.add_div_right]
 
 @[simp] theorem merge_ret_length_eq_sum (xs ys : List α) :
-  ⟪merge xs ys⟫.length = xs.length + ys.length := by
-  fun_induction merge <;> simp only [Pure.pure, pure, List.length_nil, add_zero,zero_add]
-  simp only [Bind.bind, tick, ret_bind, decide_eq_true_eq, List.length_cons]
-  split_ifs <;> (simp only [ret_bind, List.length_cons];grind)
-
+    ⟪merge xs ys⟫.length = xs.length + ys.length := by
+  fun_induction merge with
+  | case3 =>
+    simp only [Pure.pure, Bind.bind]
+    grind
+  | _ => simp
 
 @[simp] theorem mergeSort_same_length (xs : List α) :
   ⟪mergeSort xs⟫.length = xs.length := by
   fun_induction mergeSort
-  · simp only [Pure.pure, pure]
-  · simp only [Bind.bind, ret_bind, merge_ret_length_eq_sum]
-    grind
+  · simp
+  · simp only [Bind.bind]
+    grind [merge_ret_length_eq_sum]
 
-@[simp] theorem merge_time (xs ys : List α) :
-  (merge xs ys).time ≤ xs.length + ys.length := by
-  fun_induction merge <;> simp only [Pure.pure, pure, List.length_nil, add_zero, zero_le,zero_add]
-  simp only [Bind.bind, tick, time_of_bind, decide_eq_true_eq, List.length_cons]
-  split_ifs <;> all_goals (simp_all only [List.length_cons, not_le, time_of_bind, add_zero]; grind)
+@[simp] theorem merge_time (xs ys : List α) : (merge xs ys).time ≤ xs.length + ys.length := by
+  fun_induction merge with
+  | case3 =>
+    simp only [Pure.pure, Bind.bind]
+    grind
+  | _ => simp
 
 theorem mergeSort_time_le (xs : List α) :
   (mergeSort xs).time ≤ timeMergeSortRec xs.length := by
   fun_induction mergeSort with
   | case1 =>
-    simp only [Pure.pure, pure]
+    simp only [Pure.pure]
     grind
   | case2 _ _ _ _ _ ih2 ih1 =>
     simp only [Bind.bind, time_of_bind]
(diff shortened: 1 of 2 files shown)


Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
```diff
@@ -0,0 +1,190 @@
...
+theorem sorted_merge {l1 l2 : List ℕ} (hxs : IsSorted l1)
+  (hys : IsSorted l2) : IsSorted ((merge l1 l2).ret) := by
+  fun_induction merge.go l1 l2 <;> all_goals (first| simpa [merge,merge.go,tick]|
+    simp only [IsSorted, merge, merge.go, h, ↓reduceIte, Bind.bind, tick, ret_bind, sorted_cons]
+    simp [merge] at ih1
+    simp_all only [IsSorted, sorted_cons, implies_true, forall_const, and_true]
+    apply min_all_merge <;> all_goals grind
+  )
+
+theorem MSMCorrect (xs : List ℕ) : IsSorted (mergeSort xs).ret := by
```
jt0202: I think it is not enough to show that mergeSort returns something that is sorted in order to be correct. One would also need to show that it is actually a permutation of the original list.

Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
```diff
@@ -0,0 +1,190 @@
...
+  let r := f m.ret
+  ⟨r.ret, m.time + r.time⟩
+
+instance : Monad TimeM where
+  pure := pure
+  bind := bind
+
+-- Increment time
+@[simp] def tick {α : Type} (a : α) (c : ℕ := 1) : TimeM α :=
+  ⟨a, c⟩
```
jt0202: If we use this for now to prove runtimes of algorithms, it deserves a single file to be reuseable.

fmontesi: Yes, in general TimeM should have its own file, containing its notation as well (which should be scoped).

Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
```diff
@@ -0,0 +1,190 @@
+import Mathlib.Tactic
```
fmontesi: Missing header.

Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
```diff
@@ -0,0 +1,190 @@
...
+
+set_option tactic.hygienic false
+set_option autoImplicit false
+
+structure TimeM (α : Type) where
```
fmontesi: When you get to making a file about this (see comments below), please document the structure and its defs as well.

Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
```diff
@@ -0,0 +1,190 @@
...
+
+@[grind =] def MS_REC : ℕ → ℕ
+| 0 => 0
+| 1 => 0
+| n@(_+2) => MS_REC (n/2) + MS_REC ((n-1)/2 + 1) + n
+
+#eval mergeSort [4,1,3,4,5,6]
+#eval MS_REC 6
+
+theorem mergeSort_time_eq_MS_REC (xs : List ℕ) :
```
fmontesi: See https://github.com/leanprover/cslib?tab=contributing-ov-file#style-and-documentation for a link to the style documentation.

Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
```diff
@@ -0,0 +1,190 @@
...
+ (TimeM.bind m f).time = m.time + (f m.ret).time := rfl
+@[grind, simp] theorem time_of_tick {α} (a : α) (c : ℕ) : (tick a c).time = c := rfl
+@[grind, simp] theorem ret_bind {α β} (m : TimeM α) (f : α → TimeM β) :
+  (TimeM.bind m f).ret = (f m.ret).ret := rfl
+
+-- allow us to simplify the chain of compositions
+attribute [simp] Bind.bind Pure.pure TimeM.pure
+
+-- Analysis of Merge Sort
+def merge (xs ys : List ℕ) : TimeM (List ℕ) :=
```
fmontesi: We should abstract over the type of elements of the list, assuming only that the necessary ordering operator is defined.

Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
```diff
@@ -0,0 +1,190 @@
...
+@[simp] def tick {α : Type} (a : α) (c : ℕ := 1) : TimeM α :=
+  ⟨a, c⟩
+
+notation "✓" a:arg ", " c:arg => tick a c
+notation "✓" a:arg => tick a  -- Default case with only one argument
+
+def tickUnit : TimeM Unit :=
+  ✓ () -- This uses the default time increment of 1
+
+@[grind, simp] theorem time_of_pure {α} (a : α) : (pure a).time = 0 := rfl
```
fmontesi: Probably a good idea to make all these `grind` scoped.

Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
```diff
@@ -0,0 +1,190 @@
...
+  else
+    let n := xs.length
+    let half := n / 2
+    let left :=  xs.take half
+    let right := xs.drop half
+    let sortedLeft ← mergeSort left
+    let sortedRight ← mergeSort right
+    merge sortedLeft sortedRight
+
+#check mergeSort
```
fmontesi: These checks and evals should be removed. If you like, you can add tests to the `CslibTests` directory.

chenson2018:
I am holding off on reviewing style until discussion on the content is settled. There was a [previous Zulip thread](https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/Proposal.20on.20Time.20Complexity/with/541407869) that raised some issues that I would like to see addressed with respect to this PR.

Also, where does this fit in with CSLib at large? Boole is going to also cover this ground, yes? Will there be two parallel developments of time complexity?

Shreyas4991:
Personally I think this model leaves too much room for tinkering in the specification. Users manually annotate their function calls with tick. A good spec method should minimise the extent to which a user can determine where complexity is counted, for example by specifying a query model upfront and having all queries counted. There have been meetings within the community going as far back as March 2023 where we discussed ways to impose a model, where the user did not get to pick and choose when an operation is counted inside the algorithms def.

sorrachai:
> Personally I think this model leaves too much room for tinkering in the specification. Users manually annotate their function calls with tick. A good spec method should minimise the extent to which a user can determine where complexity is counted, for example by specifying a query model upfront and having all queries counted. There have been meetings within the community going as far back as March 2023 where we discussed ways to impose a model, where the user did not get to pick and choose when an operation is counted inside the algorithms def.

That's a valuable perspective, and I appreciate the concern about specification rigor. You're right that there's an inherent trade-off between flexibility and formal control in complexity analysis. I see both approaches as complementary rather than mutually exclusive. The manual annotation model gives users more discretion, with both drawbacks and benefits. From my experience with TCS research, this flexibility enables more proofs to be developed, especially when working with existing libraries. The lightweight tracking approach functions somewhat like a "soft sorry"—the algorithmic statement requires human verification. Still, the advantage is that we can directly leverage whatever Mathlib provides and annotate the time complexity of operations.

A fully automated system with a predefined query model would be the ideal end goal. However, there's real value in having algorithms with manual runtime tracking as intermediate artifacts. These can serve as concrete examples and test cases to inform the development of more rigorous automation—similar to how early formalization work often guides better tooling.

Given the discussions since March 2023 about imposing a model where users don't control what gets counted, I'd suggest two constructive paths forward: we could continue monitoring projects like Boole as they develop automated approaches, or if you'd like to implement the stricter specification model you're envisioning, I'd be very interested in comparing both approaches. Having concrete implementations of both paradigms would provide valuable data on their respective trade-offs.

What do you think about this?

Shreyas4991:
I think the Boole model has its issues too. In TCS papers what we are typically counting are calls to black box procedures i.e. queries and Boole likely won't capture that. So I agree with your point that actually we need some control over what operations we count. 

My point is more concretely that this can still be done in a different way that limits the user to specifying which operations to count by listing them out. This is much easier to review than a spec style that allows users to insert the ticks in the algorithmic functions directly. The monadic computation should then collect and sum up these operations internally without user spec inside the algorithm spec. And then an interpretation function should take the monadic computation and run the operations. This way as long as one can see the basic operations you are counting, all in one place and can trust how the complexity operations compose with monadic bind, one can review your algorithm spec more easily. 

In the array case, this means you write a monadic composition spec which simply collects operation tokens for array reads, compares, and swaps. Then the interpretation function takes this function and runs the operations corresponding to these tokens. The complexity spec simply counts these operation tokens.

To summarise: I am saying that the monadic approach can (and for reliability purposes, should) be much more robust. 

EDIT: The deepmind debate formalisation (in a non-main branch) shows the comparison based sorting lower bound. I think that's a good example to build on.

fmontesi:
@Shreyas4991 do I understand correctly that you're suggesting to provide annotations/attributes about what should 'tick' externally from the algorithm definition? Do you have an example of this?

Cslib/Algorithms/Lean/TimeMonad.lean
```diff
@@ -0,0 +1,64 @@
+/-
+Copyright (c) 2025 Sorrachai Yingchareonthawornhcai. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Sorrachai Yingchareonthawornhcai
+-/
+
+import Mathlib.Tactic
+
+/-!
+# TimeMonad
```
fmontesi: Time Monad

Cslib/Algorithms/Lean/TimeM.lean
```diff
@@ -0,0 +1,64 @@
+/-
```
fmontesi, line 1: We typically call files with the same name as the main type, in this case it'd be `TimeM.lean`.

Cslib/Algorithms/Lean/TimeM.lean
```diff
@@ -0,0 +1,64 @@
...
+
+@[simp] theorem time_of_pure {α} (a : α) : (pure a).time = 0 := rfl
+@[simp] theorem time_of_bind {α β} (m : TimeM α) (f : α → TimeM β) :
+ (TimeM.bind m f).time = m.time + (f m.ret).time := rfl
+@[simp] theorem time_of_tick {α} (a : α) (c : ℕ) : (tick a c).time = c := rfl
+@[simp] theorem ret_bind {α β} (m : TimeM α) (f : α → TimeM β) :
+  (TimeM.bind m f).ret = (f m.ret).ret := rfl
+
+-- this allow us to simplify the chain of compositions
+attribute [simp] Bind.bind Pure.pure TimeM.pure
```
fmontesi: Not sure what to think of adding simp to these things, mmmh... @chenson2018 any thoughts?

Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
```diff
@@ -0,0 +1,171 @@
...
+  apply mem_either_merge at ha
+  grind
+
+theorem sorted_merge {l1 l2 : List α} (hxs : IsSorted l1)
+  (hys : IsSorted l2) : IsSorted ((merge l1 l2).ret) := by
+  fun_induction merge l1 l2 <;> try simpa
+  all_goals (simp_all only [IsSorted, sorted_cons, and_imp, Bind.bind, tick, Nat.cast_one, ret_bind,
+      implies_true, and_true, forall_const]; apply min_all_merge <;> grind)
+
+theorem MSCorrect_sorted (xs : List α) : IsSorted (mergeSort xs).ret := by
```
fmontesi: `mergeSort_sorted`

Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
```diff
@@ -0,0 +1,171 @@
...
+  fun_induction mergeSort xs
+  · simp only [IsSorted, Pure.pure, pure]
+    rcases x with _ | ⟨a, _ | ⟨b, rest⟩⟩ <;> simp [sorted_nil,sorted_singleton]; grind
+  · simp only [IsSorted, Bind.bind, ret_bind]
+    exact sorted_merge ih2 ih1
+
+lemma merge_perm (l₁ l₂ : List α) : (merge l₁ l₂).ret ~ l₁ ++ l₂ := by
+  fun_induction merge <;> simp only [Bind.bind, tick, ret_bind, cons_append, perm_cons] <;> grind
+
+theorem MSCorrect_perm (xs : List α) : (mergeSort xs).ret ~ xs := by
```
fmontesi: `mergeSort_perm`, I guess?
I'd also add a docstring.

Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
```diff
@@ -0,0 +1,171 @@
...
+  fun_induction mergeSort xs
+  · simp only [Pure.pure, pure, Perm.refl]
+  · simp only [Bind.bind, ret_bind]
+    calc
+      (merge (mergeSort left).ret (mergeSort right).ret).ret ~
+      (mergeSort left).ret ++ (mergeSort right).ret  := by apply merge_perm
+      _ ~ left++right := Perm.append ih2 ih1
+      _ ~ x := by simp only [take_append_drop, Perm.refl, left, right]
+
+theorem MSCorrect (xs : List α) : IsSorted (mergeSort xs).ret ∧ (mergeSort xs).ret ~ xs  := by
```
fmontesi: `mergeSort_correct`. I'd also add a docstring like 'MergeSort is functionally correct'.

Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
```diff
@@ -0,0 +1,171 @@
...
+| 1 => 0
+| n@(_+2) => MS_REC (n/2) + MS_REC ((n-1)/2 + 1) + n
+
+@[simp] theorem merge_ret_length_eq_sum (xs ys : List α) :
+  (merge xs ys).ret.length = xs.length + ys.length := by
+  fun_induction merge <;> all_goals
+  simp_all only [length_cons, Bind.bind, tick, ret_bind]
+  grind
+
+@[simp] theorem mergeSort_same_length (xs : List α) :
```
fmontesi, line 196: Not sure how to follow the Mathlib convention here. `mergeSort_length_eq`? `mergeSort_length`?
cc @chenson2018

Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
```diff
@@ -0,0 +1,171 @@
...
+    unfold MS_REC
+    norm_cast
+    grind
+  · simp only [Bind.bind, time_of_bind, merge_time, mergeSort_same_length]
+    unfold MS_REC
+    split <;> all_goals (simp_all)
+    norm_cast
+    grind
+
+@[grind =] def MS_REC_SIMP : ℕ → ℤ
```
fmontesi: In general IIRC defs should be camelCased and theorems snake_camelCased

Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
```diff
@@ -0,0 +1,171 @@
...
+  · unfold MS_REC_SIMP MS_REC
+    split <;> grind
+
+theorem MS_REC_SIMP_EQ_CLOSED (n : ℕ) : MS_REC_SIMP (2^n) = 2^n * n := by
+  induction n
+  · grind
+  · unfold MS_REC_SIMP
+    split <;> grind
+
+-- The time is written assuming the length of the list is a power of two (for simplicity).
```
fmontesi: `mergeSort_time`?

Cslib/Algorithms/Lean/TimeM.lean
sorrachai: These simp lemmas crucially allow the users to state algorithms via do-notation (which is close to a pen-and-paper version) and prove time and correctness without knowledge of monadic composition of pure and bind operations at all.

Shreyas4991: Do you have the simpNF linter enabled?

Cslib/Algorithms/Lean/TimeMonad.lean
Shreyas4991: I also recommend adding the very explicit warning here in the module doc string. about using explicit tick annotations. Ideally we want a link to the discussion thread and an explanation of why we are allowing explicit annotations for now and what the future plan is (such as the metaprogram, use of blocks indicating that these explicit functions are not the "right" way for new contributors to submit PRs etc.)

Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
sorrachai: Yes, I have added the permutation proof.

arademaker:
The Lean Action CI is now not working. The error is 

```
The file 'Cslib.lean' is out of date: run `lake exe mk_all` to update it
  Error:  lake exe mk_all failed
  Error: Process completed with exit code 1.
```

Maybe this branch needs to be updated?

Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
sorrachai: Yes, done.

Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
sorrachai: Sure

sorrachai:
> I am holding off on reviewing style until discussion on the content is settled. There was a [previous Zulip thread](https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/Proposal.20on.20Time.20Complexity/with/541407869) that raised some issues that I would like to see addressed with respect to this PR.
> 
> Also, where does this fit in with CSLib at large? Boole is going to also cover this ground, yes? Will there be two parallel developments of time complexity?

This will mainly be used for analyzing theoretical algorithms that often appear in theoretical computer science papers. In TCS papers, it is rare for mistracking time to be the correctness concern, since most algorithms in theory papers are easy to track. The main advantage is the flexibility in expressing algorithms and running time across many models. This will serve as a good example for the future development of Boole languages as well.  As Boole matures, these Lean developments may be ported to it. Until then, nothing is holding us back from proving the analysis of algorithms in Lean. 

I am aware of the heated discussion about time complexity. There is no real harm in accepting multiple proposals. It is healthy to have many options to choose from, depending on the context.  While there may be many proposals at the start, it is too premature to rule out good ones before seeing them grow. If someone believes in their design, we should let them shine and do their thing. In the end, good design will stand the test of time.

Shreyas4991:
I maintain that the free monad based query approach (Tanner Duve's WIP PR) is a far more systematic and better suited approach for TCS papers than explicitly marking operations to count inside an algorithm

1. It makes the basic operations explicit and upfront. 

2. It allows compositional reasoning about complexity 

3. It cleanly separates the complexity relevant code from its interpretation. Nothing can be sneaked in. 

4. It allows compositional translation of algorithms from one query model to another. 

5. Compiling these queries  to canonical Boole style models will be more systematic.

6. It can actually multi model multiple sequential models systematically with the same complexity theorems. 


The plain tick approach is far too arbitrary to allow this level of compositionality of complexity results and clean abstractions of models 


Note that I support the lightweight approach for formalisation of time complexity and have argued for this since March 2023. I think it is the only feasible way to modularly formalise frontier results in algorithms theory. I just think such an effort needs to be systematic for a library of CSLib's proposed scale, to avoid mistakes creeping into the spec.

Shreyas4991:
Also I think we agreed (Fabrizio stated) to push this approach by noting that explicit ticks is an unsafe approach in the docs. This hasn't been done yet in this PR.

Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
```diff
@@ -0,0 +1,230 @@
...
+
+- `mergeSort_correct`: `mergeSort` returns a sorted list which is a permutation of the input list.
+- `mergeSort_time`:  `mergeSort` time is at most `n*⌈log₂ n⌉` where `n` is the list size.
+
+-/
+
+set_option tactic.hygienic false
+set_option autoImplicit false
+
+namespace TimeM
```
fmontesi: Everything should be under some `Cslib` namespace. It might also be a good time to introduce a namespace for algorithms written in this 'manual tracking' style. How about `Cslib.Algorithms.Lean.TimeM` for now? (Bit of a mouthful...)

CC @swaratchaudhuri

Cslib/Algorithms/Lean/TimeM.lean
```diff
@@ -0,0 +1,71 @@
...
+
+## Notation
+- **`✓`** : a tick of time, see `tick`.
+- **`⟪tm⟫`** : Extract the pure value from a `TimeM` computation (notation for `tm.ret`)
+
+## References
+
+See [Danielsson'08] for the discussion.
+-/
+
```
fmontesi, line 35: namespace `Cslib`

Cslib/Algorithms/Lean/TimeM.lean
```diff
@@ -0,0 +1,71 @@
...
+- **What is free?** (variable lookup, pattern matching, etc.)
+- **Recursive calls:** Do you charge for the call itself?
+
+## Notation
+- **`✓`** : a tick of time, see `tick`.
+- **`⟪tm⟫`** : Extract the pure value from a `TimeM` computation (notation for `tm.ret`)
+
+## References
+
+See [Danielsson'08] for the discussion.
```
fmontesi: Please add this reference to `references.bib` in the root directory.
Also, it should probably read `[Danielsson2008]`.

Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
sorrachai: Fine by me.

Cslib/Algorithms/Lean/TimeM.lean
sorrachai: Yes, added.

Cslib/Algorithms/Lean/TimeM.lean
sorrachai: I am using namespace Cslib.Algorithms.Lean for now.

sorrachai:
I got the following Lean Action CI error message.

/home/runner/work/cslib/cslib/Cslib/Algorithms/Lean/MergeSort/MergeSort.lean:
  remove #[Mathlib.Tactic.Bound]

However, I cannot remove this as I use this tactic in my proof. How do I proceed? @chenson2018, do you have advice?

Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
```diff
@@ -0,0 +1,247 @@
...
+Copyright (c) 2025 Sorrachai Yingchareonthawornhcai. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Sorrachai Yingchareonthawornhcai
+-/
+
+import Cslib.Algorithms.Lean.TimeM
+import Mathlib.Data.Nat.Cast.Order.Ring
+import Mathlib.Data.Nat.Lattice
+import Mathlib.Data.Nat.Log
+import Mathlib.Tactic
```
chenson2018: We try to avoid `Mathlib.Tactic` as it is a very large of which we use only a small portion. Can you please change this back to `Mathlib.Tactic.Bound`?

chenson2018 (changes requested):

Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
sorrachai: Yes, reverted.

fmontesi (approved):
This looks pretty good now, thanks for all the changes!

fmontesi:
@chenson2018 are you happy with the latest changes? Github still shows me a cross for your review, but I think it's the comment about the tactic import?

chenson2018:
Let me do a pass for style, I had held off while more conceptual discussion was ongoing.

Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
```diff
@@ -0,0 +1,247 @@
...
+
+--
+## Main results
+
+- `mergeSort_correct`: `mergeSort` permutes the list into a sorted one.
+- `mergeSort_time`:  The number of comparisons of `mergeSort` is at most `n*⌈log₂ n⌉`.
+
+-/
+
+set_option tactic.hygienic false
```
chenson2018: Can you please rewrite as needed to not unset hygiene? We should avoid this sort of technical debt.

Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
```diff
@@ -0,0 +1,232 @@
+/-
+Copyright (c) 2025 Sorrachai Yingchareonthawornhcai. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Sorrachai Yingchareonthawornhcai
+-/
+
+import Cslib.Algorithms.Lean.TimeM
```
chenson2018, line 7: Are we sure about this directory structure? I probably would have gone with maybe `Cslib/Algorithms/TimeM` and had `Basic.lean` and `MergeSort.lean` within.

fmontesi: For now we're pretty sure. We discussed this a long time ago in the SG, the motivation being that there will be a directory per language/approach, e.g., Cslib/Algorithms/Boole.

fmontesi:
Roger, thanks! (I plan on merging and will wait for the release.)

ctchou:
This patch is generating a lot of warning messages during `lake build` in the current `main` branch (commit: 2ee7c691667a90531229dd99bc17c22e1264b494):

https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/Warning.20messages.20generated.20by.20cslib.23165/near/563906945

PR #166, opened 2025-11-18
chore: cleanup imports

kim-em:
This avoids `import Mathlib` anywhere, and replaces `import Mathlib.Tactic` with `import Mathlib.Tactic.Common`. (We really should rename `Mathlib.Tactic` to `Mathlib.Tactic.KitchenSink`.) This reduce the number of build jobs in `lake build` from about 7000 to about 1000. I've added a test in CslibTests that ensures we can import Mathlib and Cslib in their entirety together.

We should later install `shake` (not yet, it is being replaced soon), to maintain this state.


chenson2018:
@kim-em , did you see #160? I had addressed the `import Mathlib` there and added shake. Can we just add the change to use `Mathlib.Tactic.Common` there?

kim-em:
Closing in favour of #160, sorry, I had missed that.

PR #168, opened 2025-11-19
feat: prove that omega-regularity is closed under union and is universe-polymorphic

ctchou:
This patch proves that:
(1) omega-regular languages are closed under finite union, and
(2) the definition of omega-regular languages is universe-polymorphic.
The `iSum` construction used to prove (1) actually works over an arbitrary indexed family of NA.Buchi automata.

### DIFF
--- a/Cslib/Computability/Automata/NA.lean
+++ b/Cslib/Computability/Automata/NA.lean
@@ -1,7 +1,7 @@
 /-
 Copyright (c) 2025 Fabrizio Montesi. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
-Authors: Fabrizio Montesi, Ching-Tsun Chou
+Authors: Fabrizio Montesi, Ching-Tsun Chou, Chris Henson.
 -/
 
 import Cslib.Computability.Automata.Acceptor


Cslib/Computability/Automata/NABuchiEquiv.lean
```diff
@@ -0,0 +1,66 @@
...
+theorem reindex_run_iff {f : State ≃ State'} {nba : Buchi State Symbol}
+    {xs : ωSequence Symbol} {ss' : ωSequence State'} :
+    (nba.reindex f).Run xs ss' ↔ nba.Run xs (ss'.map f.symm) := by
+  constructor <;>
+  { rintro ⟨h_init, h_next⟩
+    constructor
+    · grind
+    · intro n
+      specialize h_next n
+      simp_all [reindex] }
```
chenson2018: I find it clearer to write this term explicitly:
```suggestion
    · exact fun n => h_next n }
```

Cslib/Computability/Automata/NABuchiEquiv.lean
```diff
@@ -0,0 +1,66 @@
...
+@[simp, scoped grind =]
+theorem reindex_language_eq {f : State ≃ State'} {nba : Buchi State Symbol} :
+    language (nba.reindex f) = language nba := by
+  ext xs
+  constructor
+  · rintro ⟨ss', h_run', h_acc'⟩
+    grind [reindex_run_iff]
+  · rintro ⟨ss, h_run, h_acc⟩
+    use ss.map f
+    constructor <;> grind [reindex_run_iff']
```
chenson2018, line 62: Needing to explicitly split this conjunction a single time to make `grind` work seems like an indicator that the annotations need some refinement.

I think the problem is that the `=` annotation on `NA.Run` eagerly splits it into the underlying conjunction, so then further API around `NA.Run` is not useful. I'm not sure, but could you try removing that annotation and seeing if that's an improvement?

A theorem (does this exist already?)
```lean
theorem foo (na : NA State Symbol) (xs : ωSequence Symbol) (ss : ωSequence State)
    (h₁ : ss 0 ∈ na.start) (h₂ : ∀ n, na.Tr (ss n) (xs n) (ss (n + 1))) : Run na xs ss := ⟨h₁, h₂⟩
```
with maybe annotation `[grind =>]` or `[grind <=]`  could be helpful.

Cslib/Computability/Automata/Sum.lean
```diff
@@ -0,0 +1,81 @@
...
+open scoped Cslib.LTS
+
+namespace Cslib.Automata.NA
+
+variable {Symbol I : Type*} {State : I → Type*}
+
+@[scoped grind =]
+def iSum (na : (i : I) → NA (State i) Symbol) : NA (Σ i, State i) Symbol where
+  start := ⋃ i, Sigma.mk i '' (na i).start
+  Tr s x t := ∃ i s_i t_i, (na i).Tr s_i x t_i ∧ Sigma.mk i s_i = s ∧ Sigma.mk i t_i = t
```
chenson2018: In cases where not using partial application, definitely use the syntax for `Sigma.mk`:

```suggestion
  Tr s x t := ∃ i s_i t_i, (na i).Tr s_i x t_i ∧ ⟨i, s_i⟩ = s ∧ ⟨i, t_i⟩ = t
```

Cslib/Computability/Automata/Sum.lean
```diff
@@ -0,0 +1,81 @@
...
+open Set Function Filter Cslib.ωSequence
+open scoped Cslib.LTS
+
+namespace Cslib.Automata.NA
+
+variable {Symbol I : Type*} {State : I → Type*}
+
+@[scoped grind =]
+def iSum (na : (i : I) → NA (State i) Symbol) : NA (Σ i, State i) Symbol where
+  start := ⋃ i, Sigma.mk i '' (na i).start
```
chenson2018, line 21: You could write
```suggestion
  start := ⋃ i, (⟨i, ·⟩) '' (na i).start
```
I could be convinced to leave as-is if we feel it helps readability in the case of partial application.

Cslib/Computability/Automata/Sum.lean
```diff
@@ -0,0 +1,81 @@
...
+  Tr s x t := ∃ i s_i t_i, (na i).Tr s_i x t_i ∧ Sigma.mk i s_i = s ∧ Sigma.mk i t_i = t
+
+@[simp, scoped grind =]
+theorem iSum_run_iff {na : (i : I) → NA (State i) Symbol}
+    {xs : ωSequence Symbol} {ss : ωSequence (Σ i, State i)} :
+    (iSum na).Run xs ss ↔ ∃ i ss_i, (na i).Run xs ss_i ∧ ss_i.map (Sigma.mk i) = ss := by
+  constructor
+  · rintro ⟨h_init, h_next⟩
+    obtain ⟨i, s0, h_s0, h_ss0⟩ := mem_iUnion.mp h_init
+    have h_exists n : ∃ s_i, Sigma.mk i s_i = ss n := by
```
chenson2018:
```suggestion
    have h_exists n : ∃ s_i, ⟨i, s_i⟩ = ss n := by
```

Cslib/Computability/Languages/OmegaRegularLanguage.lean
```diff
@@ -17,15 +19,32 @@ This file defines ω-regular languages and proves some properties of them.
...
+variable {Symbol : Type u}
 
 /-- An ω-language is ω-regular iff it is accepted by a
 finite-state nondeterministic Buchi automaton. -/
 def IsRegular (p : ωLanguage Symbol) :=
   ∃ State : Type, ∃ _ : Finite State, ∃ na : NA.Buchi State Symbol, language na = p
 
+/-- Helper lemma for `isRegular_iff` below. -/
+private lemma isRegular_iff.helper.{v'} {Symbol : Type u} {p : ωLanguage Symbol}
+    (h : ∃ σ : Type v, ∃ _ : Finite σ, ∃ na : NA.Buchi σ Symbol, language na = p) :
```
chenson2018: You can rewrite these consecutive existentials here and below more readably like
```suggestion
    (h : ∃ (σ : Type v) (_ : Finite σ) (na : NA.Buchi σ Symbol), language na = p) :
```
(should probably also be done in the definition of `IsRegular`.)

Cslib/Computability/Languages/OmegaRegularLanguage.lean
```diff
@@ -44,11 +63,57 @@ theorem IsRegular.not_da_buchi :
...
+  generalize h_n : s.ncard = n
+  induction n generalizing s
+  case zero =>
+    obtain ⟨rfl⟩ := (ncard_eq_zero (s := s)).mp h_n
+    simp only [mem_empty_iff_false, not_false_eq_true, iSup_neg, iSup_bot]
+    exact IsRegular.bot
+  case succ n h_ind =>
+    obtain ⟨i, t, h_i, rfl, rfl⟩ := (ncard_eq_succ (s := s)).mp h_n
+    rw [iSup_insert]
+    apply IsRegular.sup <;> grind
```
chenson2018:
```suggestion
  case zero =>
    have := ncard_eq_zero (s := s)
    grind [IsRegular.bot, iSup_bot]
  case succ n h_ind =>
    obtain ⟨i, t, h_i, rfl, rfl⟩ := (ncard_eq_succ (s := s)).mp h_n
    rw [iSup_insert]
    grind [IsRegular.sup]
```

ctchou:
Merge in the new upstream/main and fix an import error caused by it.

ctchou:
@chenson2018  About your question about NABuchiEquiv above: I basically copied the code, mutatis mutandis, from:
https://leanprover-community.github.io/mathlib4_docs/Mathlib/Computability/DFA.html
See the last part of the file.  The issue here is that if you change the definition of `ωLanguage.IsRegular` to quantify over a polymorphic `State` type, many things will break.  (You can give it a try in `OmegaRegularLanguage.lean`.). So the definition of `IsRegular` quantifies over only `Type 0` and a separate theorem `isRegular_iff` is proved to show that there is no loss of generality.  The issue was discussed here:
https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/Language.2EIsRegular.20in.20mathlib/with/548165505

Cslib/Computability/Automata/NABuchiEquiv.lean
ctchou: Fixed.

Cslib/Computability/Automata/Sum.lean
ctchou: Fixed.

Cslib/Computability/Automata/Sum.lean
ctchou: I think I'll keep `Sigma.mk i` in cases of partial application, because `(⟨i, ·⟩)` requires the surrounding parentheses and personally I don't find that an improvement in readability.

Cslib/Computability/Languages/OmegaRegularLanguage.lean
ctchou: Fixed.

Cslib/Computability/Automata/Sum.lean
ctchou: Fixed.

Cslib/Computability/Languages/OmegaRegularLanguage.lean
ctchou: Fixed.

Cslib/Computability/Automata/NABuchiEquiv.lean
chenson2018: Could you take a look at the commit I just pushed? (Not meant to be the final word, just experimenting). Regardless of `grind`, I think that if you are defining a type `Run`, it is good to have API theorems that construct and deconstruct it.

Adding `grind` annotations to these happens to help a little bit as a replacement for the annotation on `NA.Run`, but is still not perfect. The tradeoff is there are a few places where you need to know to specify `grind [NA.Run]`, and I don't understand yet why this is the case.

ctchou: I reorganized the code a little bit: renaming `Run.trace` to `Run.trans` and putting the grind lemmas in a `section` with a comment.  We should get more data as I develop the theory of NA.Buchi more.

ctchou: I also did some comparisons of the old and the new grind annotations.  It seems that the new annotations sped up `Sum.lean` by quite a bit, although both runs are on the new version of `Sum.lean`.

chenson2018: Hmm interesting. Soon we should have some benchmarking in place that makes these performance aspects easier to identify.

fmontesi:
Very nice!!

PR #173, opened 2025-11-20
chore: consolidate lint workflow and change mode to check

chenson2018:
Having `lint-style-action` use the `suggest` mode, which has a bot make PR comments suggesting lint fixes, has proven to be flaky in CI. This PR changes the action to the `check` mode and consolidates it into the existing workflow that already runs on PRs and pushes.


chenson2018:
The last commit is all of the lint errors that have accumulated because of the bot silently failing to post to the PR. The alternative here is that you instead run `lake exe lint-style --fix` once locally.

fmontesi:
:clap: :clap: :clap:

PR #187, opened 2025-11-24
feat(CLL): port CLL sequents to Multiset and make CLL equivalences proof-relevant

fmontesi:
This PR ports CLL sequents from List to Multiset and makes CLL equivalences proof-relevant.

### DIFF
--- a/Cslib/Logics/LinearLogic/CLL/Basic.lean
+++ b/Cslib/Logics/LinearLogic/CLL/Basic.lean
@@ -131,7 +131,7 @@ theorem Proposition.dual.involution (a : Proposition Atom) : a⫠⫠ = a := by
   induction a <;> grind [dual]
 
 /-- Linear implication. -/
-@[scoped grind .]
+@[scoped grind =]
 def Proposition.linImpl (a b : Proposition Atom) : Proposition Atom := a⫠ ⅋ b
 
 @[inherit_doc] scoped infix:25 " ⊸ " => Proposition.linImpl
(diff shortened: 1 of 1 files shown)


Cslib/Logics/LinearLogic/CLL/CutElimination.lean
```diff
@@ -40,8 +38,8 @@ abbrev CutFreeProof (Γ : Sequent Atom) := { q : ⇓Γ // q.cutFree }
 /- Cut admissibility: given two proofs with dual propositions, returns a cut-free proof of their
```
chenson2018, line 38: Some commented out code here.

chenson2018:
The commits I just pushed remove all uses of tactics for data in `Cslib/Logics/LinearLogic/CLL/Basic.lean` and a few other small golfs. Technically these still use tactics for `grind`, but only with `show` to isolate them from the rest of the term. All `rw` should be removed in favor of terms using `▸` or `Proof.sequent_rw`

I do understand why it would be nice to use tactics, as what I've written here would be annoying for a first pass at these definitions. Maybe worth discussion on Zulip in one of the general channels to see if others have worked with this sort of "proof-relevant" derivations before. (It's a topic that's relevant to other areas like lambda calculus, because when I get around to some categorical semantics we will likely want an intrinsically typed representation because of universe restrictions)

I also noted some overuse of `grind .`, which I would discourage when a theorem could have a more specific pattern or in the case of `def` when only `grind =` makes sense anyway.

chenson2018:
Also, could you add a description to the PR? Sorry to be a stickler, but it can come in handy later.

fmontesi:
Super, thank you very much! Description added.

chenson2018 (approved):

PR #188, opened 2025-11-25
feat: prove that omega-regular languages are closed under finite intersection

ctchou:
Proving that omega-regular languages are closed under finite intersection is harder than doing so for regular languages and requires the following ingredients:
- NAProd.lean: product construction for NA.
- NAHist.lean: adding history state to NA.
- NABuchiInter: the intersection construction that uses both NAProd and NAHist constructions.
- Temporal.lean: temporal reasoning over infinite sequences that is needed to show that NABuchiInter actually works.

I also renamed {Sum,Prod}.lean to {NASum,DAProd}.lean to keep the file naming scheme consistent.

### DIFF
--- a/Cslib/Computability/Automata/Acceptor.lean
+++ b/Cslib/Computability/Automata/Acceptor.lean
@@ -10,13 +10,13 @@ import Mathlib.Computability.Language
 namespace Cslib.Automata
 
 /-- An `Acceptor` is a machine that recognises strings (lists of symbols in an alphabet). -/
-class Acceptor (A : Type _) (Symbol : outParam (Type _)) where
+class Acceptor (A : Type u) (Symbol : outParam (Type v)) where
   /-- Predicate that establishes whether a string `xs` is accepted. -/
   Accepts (a : A) (xs : List Symbol) : Prop
 
 namespace Acceptor
 
-variable {Symbol : Type _}
+variable {Symbol : Type v}
 
 /-- The language of an `Acceptor` is the set of strings it `Accepts`. -/
 @[scoped grind .]
--- a/Cslib/Computability/Automata/DA.lean
+++ b/Cslib/Computability/Automata/DA.lean
@@ -36,7 +36,7 @@ structure DA (State Symbol : Type*) extends FLTS State Symbol where
 
 namespace DA
 
-variable {State : Type _} {Symbol : Type _}
+variable {State Symbol : Type*}
 
 /-- Helper function for defining `run` below. -/
 @[scoped grind =]
--- a/Cslib/Computability/Automata/DABuchi.lean
+++ b/Cslib/Computability/Automata/DABuchi.lean
@@ -15,7 +15,7 @@ namespace Cslib.Automata.DA
 
 open scoped FinAcc Buchi
 
-variable {State : Type _} {Symbol : Type _}
+variable {State Symbol : Type*}
 
 open Acceptor ωAcceptor in
 /-- The ω-language accepted by a deterministic Buchi automaton is the ω-limit
--- a/Cslib/Computability/Automata/DAToNA.lean
+++ b/Cslib/Computability/Automata/DAToNA.lean
@@ -14,7 +14,7 @@ This is the general version of the standard translation of DFAs into NFAs. -/
 
 namespace Cslib.Automata.DA
 
-variable {State : Type _} {Symbol : Type _}
+variable {State Symbol : Type*}
 
 section NA
 
--- a/Cslib/Computability/Automata/NA.lean
+++ b/Cslib/Computability/Automata/NA.lean
@@ -38,7 +38,7 @@ structure NA (State Symbol : Type*) extends LTS State Symbol where
 
 namespace NA
 
-variable {State : Type _} {Symbol : Type _}
+variable {State Symbol : Type*}
 
 /-- Infinite run. -/
 def Run (na : NA State Symbol) (xs : ωSequence Symbol) (ss : ωSequence State) :=
--- a/Cslib/Computability/Automata/NAToDA.lean
+++ b/Cslib/Computability/Automata/NAToDA.lean
@@ -15,7 +15,7 @@ This file implements the standard subset construction.
 
 namespace Cslib.Automata.NA
 
-variable {State : Type _} {Symbol : Type _}
+variable {State Symbol : Type*}
 
 section SubsetConstruction
 
--- a/Cslib/Computability/Automata/OmegaAcceptor.lean
+++ b/Cslib/Computability/Automata/OmegaAcceptor.lean
@@ -9,13 +9,13 @@ import Cslib.Computability.Languages.OmegaLanguage
 namespace Cslib.Automata
 
 /-- An `ωAcceptor` is a machine that recognises infinite sequences of symbols. -/
-class ωAcceptor (A : Type _) (Symbol : outParam (Type _)) where
+class ωAcceptor (A : Type u) (Symbol : outParam (Type v)) where
   /-- Predicate that establishes whether a string `xs` is accepted. -/
   Accepts (a : A) (xs : ωSequence Symbol) : Prop
 
 namespace ωAcceptor
 
-variable {Symbol : Type _}
+variable {Symbol : Type v}
 
 /-- The language of an `ωAcceptor` is the set of sequences it `Accepts`. -/
 @[scoped grind .]
--- a/Cslib/Foundations/Data/FinFun.lean
+++ b/Cslib/Foundations/Data/FinFun.lean
@@ -20,7 +20,7 @@ namespace Cslib
 /-- A `FinFun` is a function `fn` with a finite `support`.
 
 This is similar to `Finsupp` in Mathlib, but definitions are computable. -/
-structure FinFun (α : Type _) (β : Type _) [Zero β] where
+structure FinFun (α β : Type*) [Zero β] where
   /-- The underlying function. -/
   fn : α → β
   /-- The finite support of the function. -/
@@ -38,7 +38,7 @@ scoped infixr:25 " →₀ " => FinFun
 /-- Constructs a `FinFun` by restricting a function to a given support, filtering out all elements
 not mapped to 0 in the support. -/
 @[scoped grind .]
-private def fromFun {α β : Type _} [Zero β] [DecidableEq α]
+private def fromFun {α β : Type*} [Zero β] [DecidableEq α]
   [∀ y : β, Decidable (y = 0)] (fn : α → β) (support : Finset α) : α →₀ β where
   fn := (fun a => if a ∈ support then fn a else 0)
   support := support.filter (fn · ≠ 0)
--- a/Cslib/Foundations/Data/Relation.lean
+++ b/Cslib/Foundations/Data/Relation.lean
@@ -53,23 +53,23 @@ theorem Relation.ReflTransGen.diamond_confluence (h : Diamond R) : Confluence R 
 
 -- not sure why this doesn't compile as an "instance" but oh well
 /-- A pair of subrelations lifts to transitivity on the relation. -/
-def trans_of_subrelation {α : Type _} (s s' r : α → α → Prop) (hr : Transitive r)
+def trans_of_subrelation {α : Type*} (s s' r : α → α → Prop) (hr : Transitive r)
     (h : ∀ a b : α, s a b → r a b) (h' : ∀ a b : α, s' a b → r a b) : Trans s s' r where
   trans hab hbc := hr (h _ _ hab) (h' _ _ hbc)
 
 /-- A subrelation lifts to transitivity on the left of the relation. -/
-def trans_of_subrelation_left {α : Type _} (s r : α → α → Prop) (hr : Transitive r)
+def trans_of_subrelation_left {α : Type*} (s r : α → α → Prop) (hr : Transitive r)
     (h : ∀ a b : α, s a b → r a b) : Trans s r r where
   trans hab hbc := hr (h _ _ hab) hbc
 
 /-- A subrelation lifts to transitivity on the right of the relation. -/
-def trans_of_subrelation_right {α : Type _} (s r : α → α → Prop) (hr : Transitive r)
+def trans_of_subrelation_right {α : Type*} (s r : α → α → Prop) (hr : Transitive r)
     (h : ∀ a b : α, s a b → r a b) : Trans r s r where
   trans hab hbc := hr hab (h _ _ hbc)
 
 /-- This is a straightforward but useful specialisation of a more general result in
 `Mathlib.Logic.Relation`. -/
-theorem church_rosser_of_diamond {α : Type _} {r : α → α → Prop}
+theorem church_rosser_of_diamond {α : Type*} {r : α → α → Prop}
     (h : ∀ a b c, r a b → r a c → Relation.Join r b c) :
     Equivalence (Relation.Join (Relation.ReflTransGen r)) := by
   apply Relation.equivalence_join_reflTransGen
--- a/Cslib/Foundations/Semantics/LTS/FLTS.lean
+++ b/Cslib/Foundations/Semantics/LTS/FLTS.lean
@@ -25,13 +25,13 @@ namespace Cslib
 A Functional Labelled Transition System (`FLTS`) for a type of states (`State`) and a type of
 transition labels (`Label`) consists of a labelled transition function (`tr`).
 -/
-structure FLTS (State : Type _) (Label : Type _) where
+structure FLTS (State Label : Type*) where
   /-- The transition function. -/
   tr : State → Label → State
 
 namespace FLTS
 
-variable {State : Type _} {Label : Type _}
+variable {State Label : Type*}
 
 /-- Extended transition function.
 
--- a/Cslib/Foundations/Semantics/LTS/FLTSToLTS.lean
+++ b/Cslib/Foundations/Semantics/LTS/FLTSToLTS.lean
@@ -7,7 +7,7 @@ Authors: Fabrizio Montesi
 import Cslib.Foundations.Semantics.LTS.Basic
 import Cslib.Foundations.Semantics.LTS.FLTS
 
-variable {State : Type _} {Label : Type _}
+variable {State Label : Type*}
 
 namespace Cslib.FLTS
 
--- a/Cslib/Foundations/Semantics/LTS/LTSToFLTS.lean
+++ b/Cslib/Foundations/Semantics/LTS/LTSToFLTS.lean
@@ -7,7 +7,7 @@ Authors: Fabrizio Montesi
 import Cslib.Foundations.Semantics.LTS.Basic
 import Cslib.Foundations.Semantics.LTS.FLTS
 
-variable {State : Type _} {Label : Type _}
+variable {State Label : Type*}
 
 namespace Cslib.LTS
 
--- a/Cslib/Foundations/Syntax/HasWellFormed.lean
+++ b/Cslib/Foundations/Syntax/HasWellFormed.lean
@@ -9,7 +9,7 @@ import Cslib.Init
 namespace Cslib
 
 /-- Typeclass for types equipped with a well-formedness predicate. -/
-class HasWellFormed (α : Type _) where
+class HasWellFormed (α : Type u) where
   /-- Establishes whether `x` is well-formed. -/
   wf (x : α) : Prop
 
--- a/Cslib/Languages/CombinatoryLogic/Basic.lean
+++ b/Cslib/Languages/CombinatoryLogic/Basic.lean
@@ -38,7 +38,7 @@ open Red MRed
 /-! ### Polynomials and the bracket astraction algorithm -/
 
 /-- A polynomial is an SKI terms with free variables. -/
-protected inductive Polynomial (n : Nat) : Type _ where
+protected inductive Polynomial (n : Nat) : Type where
   | term : SKI → SKI.Polynomial n
   | var : Fin n → SKI.Polynomial n
   | app : SKI.Polynomial n → SKI.Polynomial n → SKI.Polynomial n


chenson2018:
Before I review, I've been meaning to ask about the naming scheme. As you add more files, I think it would be easier to navigate if we organized into subdirectories. What do you think about something like the structure below?
```
├── Automata
    ├── Acceptor.lean
    ├── DA
    │   ├── Basic.lean
    │   ├── Buchi.lean
    │   ├── Prod.lean
    │   └── ToNA.lean
    ├── NA
    │   ├── Basic.lean
    │   ├── Buchi
    │   │   ├── Equiv.lean
    │   │   └── Inter.lean
    │   ├── Epsilon
    │   │   ├── Basic.lean
    │   │   └── ToNA.lean
    │   ├── Hist.lean
    │   ├── Prod.lean
    │   ├── Sum.lean
    │   └── ToDA.lean
    └── OmegaAcceptor.lean
```
The details of exactly what directories can be changed as you see fit, I just think in general it could be nicer than the flat structure.

ctchou:
Actually I like a flat directory better.  I really hate the hassle of having to climb up and down in a directory tree when I need to open a file.  I'm also not sure a tree structure is any more "rational" than the current naming scheme, which already groups the files in a natural way by the alphabetical order.  For example, NABuchiIInter.lean contains a construction that is a composition of the constructions in NAProd.lean and NAHist.lean.  There seems to me no good reason why the first file should be placed in a subdirectory while the latter two are not.  Also note that if you pursue your logic to its conclusion, you probably should split NA.lean into NA/Basic.lean, NA/FinAcc/Basic.lean, NA/Buchi/Basic.lean, and NA/Muller/Basic.lean.  I'm really not sure what that would buy you.

I think we will end up with around 20 files in this directory, which I don't think is excessive.

chenson2018:
This is not critical at the moment, we can always revisit later. I don't want to introduce too much churn while you are still actively developing the library. Let's table this and I'll just review the PR.

Just to write it down for future reference however, the benefits I see and response to your concerns:
- subdirectories make easier to read names, especially with capitalization e.g. `NAHist.lean` versus `NA/Hist.lean`
- subdirectories make it easier to see which files are related to each other. 20 files is not inherently excessive, but it is harder at a glance to see what the overarching areas there are versus just a few things at the top level
- I find your closing argument a bit of a strawman, nothing requires the directory structure to be overly granular
- I don't understand the "hassle of having to climb up and down in a directory tree" argument, as you are still free to work from the top level directory

Cslib/Computability/Automata/NAHist.lean
```diff
@@ -0,0 +1,58 @@
...
+
+variable {na : NA State Symbol} {start' : State → Hist}
+  {tr' : State × Hist → Symbol → State → Hist}
+
+/-- Every run of the history automaton projects onto a run of the original automaton. -/
+theorem hist_run_proj {xs : ωSequence Symbol} {ss : ωSequence (State × Hist)}
+    (h_run : (na.addHist start' tr').Run xs ss) : na.Run xs (ss.map fst) := by
+  obtain ⟨h_start, h_trans⟩ := h_run
+  simp only [addHist] at h_trans
+  grind
```
chenson2018, line 39: Needing to `simp [addHist]` here is a bit strange because the `grind` below also uses `addHist`. Maybe some lemma needed for `grind`?

(Side note, this is successfully using the `Run.mk` lemma, so I think that is proving a useful `grind` annotation.)

Cslib/Computability/Automata/NAHist.lean
```diff
@@ -0,0 +1,58 @@
...
+  | 0 => start' (ss 0)
+  | n + 1 => tr' (ss n, makeHist start' tr' xs ss n) (xs n) (ss (n + 1))
+
+/-- For every run `ss` of the original automaton, there exists a run `ss'` of the history automaton
+which projects back onto `ss`. -/
+theorem hist_run_exists {xs : ωSequence Symbol} {ss : ωSequence State}
+    (h_run : na.Run xs ss) : ∃ ss', (na.addHist start' tr').Run xs ss' ∧ ss'.map fst = ss := by
+  use ⟨fun n ↦ (ss n, makeHist start' tr' xs ss n)⟩
+  constructor
+  · simp only [addHist]
```
chenson2018, line 54: Same thing here, we're missing some lemmas. Also for the second branch of the proof I think we should be able make `grind` pick up that `ext`, let me investigate.

Cslib/Computability/Automata/NAProd.lean
```diff
@@ -0,0 +1,46 @@
...
+    · simp only [mem_iInter] at h_start
+      grind
+    · grind
+  · intro h
+    constructor
+    · simp only [mem_iInter]
+      grind
+    · intro n i
+      have := (h i).2 n
+      grind
```
chenson2018: Personal preference I think, but for readability I like using the descriptive names (`fst`/`snd`, `left`/`right`, etc.) versus the numeric projection.

Less subjectively, there is no need to invoke grind:
```suggestion
      exact (h i).right n
```

Cslib/Computability/Automata/NAProd.lean
```diff
@@ -0,0 +1,46 @@
...
+    {xs : ωSequence Symbol} {ss : ωSequence (Π i, State i)} :
+    (iProd na).Run xs ss ↔ ∀ i, (na i).Run xs (ss.map (· i)) := by
+  rw [iProd]
+  constructor
+  · rintro ⟨h_start, h_trans⟩
+    intro i
+    constructor
+    · simp only [mem_iInter] at h_start
+      grind
+    · grind
```
chenson2018:
```suggestion
  · rintro ⟨h_start, h_trans⟩
    simp only [mem_iInter] at h_start
    grind
```

Cslib/Computability/Languages/OmegaRegularLanguage.lean
```diff
@@ -81,6 +82,20 @@ theorem IsRegular.bot : (⊥ : ωLanguage Symbol).IsRegular := by
...
+  let na : NA.Buchi Unit Symbol := {
+    Tr _ _ _ := True
+    start := univ
+    accept := univ }
+  use Unit, inferInstance, na
+  ext xs
+  simp only [na, NA.Buchi.instωAcceptor, mem_language, mem_univ, frequently_true_iff_neBot,
+    atTop_neBot, and_true, mem_top, iff_true]
+  use const ()
+  constructor <;> grind
```
chenson2018: If you open the `NA.Run` scope, this can be just `grind`. If that scope being so far nested makes it less usable, we could move it up.

Cslib/Computability/Languages/OmegaRegularLanguage.lean
```diff
@@ -114,6 +149,20 @@ theorem IsRegular.iSup {I : Type*} [Finite I] {s : Set I} {p : I → ωLanguage
...
+@[simp]
+theorem IsRegular.iInf {I : Type*} [Finite I] {s : Set I} {p : I → ωLanguage Symbol}
+    (h : ∀ i ∈ s, (p i).IsRegular) : (⨅ i ∈ s, p i).IsRegular := by
+  generalize h_n : s.ncard = n
+  induction n generalizing s
+  case zero =>
+    have := ncard_eq_zero (s := s)
+    grind [IsRegular.top, iInf_top]
+  case succ n h_ind =>
+    obtain ⟨i, t, h_i, rfl, rfl⟩ := (ncard_eq_succ (s := s)).mp h_n
```
chenson2018: `s` can be inferred here:

```suggestion
    obtain ⟨i, t, h_i, rfl, rfl⟩ := (ncard_eq_succ).mp h_n
```

Cslib/Foundations/Data/OmegaSequence/Temporal.lean
```diff
@@ -0,0 +1,103 @@
...
+/-- "`p` leads to `q`" means that whenever `p` holds at a position in `xs`, `q` holds at the same
+or a later position in `xs`. -/
+def LeadsTo (xs : ωSequence α) (p q : Set α) : Prop :=
+  ∀ k, xs k ∈ p → ∃ k' ≥ k, xs k' ∈ q
+
+variable {xs : ωSequence α}
+
+/-- `Step` implies `LeadsTo`. -/
+theorem step_leadsTo {p q : Set α}
+    (h : xs.Step p q) : xs.LeadsTo p q := by
```
chenson2018: This fits on one line

```suggestion
theorem step_leadsTo {p q : Set α} (h : xs.Step p q) : xs.LeadsTo p q := by
```

Cslib/Foundations/Data/OmegaSequence/Temporal.lean
```diff
@@ -0,0 +1,103 @@
...
+  intro k h_p
+  obtain ⟨k', _, h_q⟩ := h1 k h_p
+  have := h2 k' h_q
+  grind
+
+/-- If `p ∩ q` leads to `r` and `p ∩ qᶜ` leads to `s`, then `p` leads to `r ∪ s`. -/
+theorem leadsTo_cases_or {p q r s : Set α}
+    (h1 : xs.LeadsTo (p ∩ q) r) (h2 : xs.LeadsTo (p ∩ qᶜ) s) : xs.LeadsTo p (r ∪ s) := by
+  intro k h_p
+  rcases Classical.em (xs k ∈ q) with h_q | h_not_q
```
chenson2018:
```suggestion
  by_cases xs k ∈ q
```

Cslib/Foundations/Data/OmegaSequence/Temporal.lean
```diff
@@ -0,0 +1,103 @@
...
+  · have h_pq : xs k ∈ p ∩ q := by grind
+    have := h1 k h_pq
+    grind
+  · have h_pq : xs k ∈ p ∩ qᶜ := by grind
+    have := h2 k h_pq
+    grind
+
+private lemma ge_lemma {k k' : ℕ} (h : k' ≥ k) : ∃ n, k' = k + n :=
```
chenson2018: `≥` is an abbreviation for `≤`, so it is often easier to use and search for theorems using `≤`.

In this case, your lemma is exactly `Nat.exists_eq_add_of_le`, please use this instead.

Cslib/Foundations/Data/OmegaSequence/Temporal.lean
```diff
@@ -0,0 +1,103 @@
...
+  suffices ∀ k' ≥ k, xs k' ∈ p by
+    have := frequently_atTop.mp h2 k
+    grind
+  intro k' h_k'
+  obtain ⟨n, rfl⟩ := ge_lemma h_k'
+  induction n
+  case zero => grind
+  case succ n h_ind =>
+    have h_pq_n : xs (k + n) ∈ p ∩ qᶜ := by grind
+    exact h1 (k + n) h_pq_n
```
chenson2018:
```suggestion
  case succ n h_ind => exact h1 (k + n) (by grind)
```

Cslib/Foundations/Data/OmegaSequence/Temporal.lean
```diff
@@ -0,0 +1,103 @@
...
+
+/-- `Step p q` means that whenever `p` holds at a position in `xs`, `q` holds at the next position
+in `xs`. -/
+def Step (xs : ωSequence α) (p q : Set α) : Prop :=
+  ∀ k, xs k ∈ p → xs (k + 1) ∈ q
+
+/-- "`p` leads to `q`" means that whenever `p` holds at a position in `xs`, `q` holds at the same
+or a later position in `xs`. -/
+def LeadsTo (xs : ωSequence α) (p q : Set α) : Prop :=
+  ∀ k, xs k ∈ p → ∃ k' ≥ k, xs k' ∈ q
```
chenson2018: Please use `≤` in this definition.

Cslib/Foundations/Data/OmegaSequence/Temporal.lean
```diff
@@ -0,0 +1,103 @@
+/-
```
chenson2018, line 1: Will you have some need later for indexed propositions as used in separation logic? Something like
```lean
class iProp (ϕ : ωSequence Prop) where
  downward_closed (k : ℕ) (hk : ϕ k) (j : ℕ) (j_le_k : j ≤ k) : ϕ j
```
seems like it fits in with what you're doing here.

Cslib/Foundations/Data/OmegaSequence/Temporal.lean
```diff
@@ -0,0 +1,103 @@
...
+variable {α : Type*}
+
+/-- `Step p q` means that whenever `p` holds at a position in `xs`, `q` holds at the next position
+in `xs`. -/
+def Step (xs : ωSequence α) (p q : Set α) : Prop :=
+  ∀ k, xs k ∈ p → xs (k + 1) ∈ q
+
+/-- "`p` leads to `q`" means that whenever `p` holds at a position in `xs`, `q` holds at the same
+or a later position in `xs`. -/
+def LeadsTo (xs : ωSequence α) (p q : Set α) : Prop :=
```
chenson2018, line 34: I think making this a typeclass and defining a notation could be helpful.

Cslib/Computability/Automata/NABuchiInter.lean
```diff
@@ -0,0 +1,146 @@
...
+variable {Symbol : Type*} {State : Bool → Type*}
+
+/-- The initial history state. -/
+@[scoped grind =, nolint unusedArguments]
+def histStart (_ : Π i, State i) : Bool := false
+
+/-- The two accepting conditions of the intersection automaton. -/
+@[scoped grind =]
+def interAcc (j : Bool) (acc : (i : Bool) → Set (State i)) : Set ((Π i, State i) × Bool) :=
+  { s | s.fst j ∈ acc j ∧ s.snd = j }
```
chenson2018: You can use patterns on the lhs, like `{⟨foo, bar⟩ | ...}`. I would do this throughout this file for readability.

Cslib/Computability/Automata/NABuchiInter.lean
```diff
@@ -0,0 +1,146 @@
...
+
+/-- The transition function of the history state. -/
+@[scoped grind =, nolint unusedArguments]
+noncomputable def histTrans (acc : (i : Bool) → Set (State i))
+    (s : (Π i, State i) × Bool) (_ : Symbol) (_ : Π i, State i) : Bool :=
+  open scoped Classical in
```
chenson2018: I still get confused by `open scoped Classical`, but doesn't it usually go outside the definition?

Cslib/Computability/Automata/NABuchiInter.lean
```diff
@@ -0,0 +1,146 @@
...
+/-- The overall accepting conditon of the intersection automaton. -/
+@[scoped grind =]
+def interAccept (acc : (i : Bool) → Set (State i)) : Set ((Π i, State i) × Bool) :=
+  interAcc false acc ∪ interAcc true acc
+
+variable {na : (i : Bool) → NA (State i) Symbol} {acc : (i : Bool) → Set (State i)}
+
+/-- If the intersection automaton sees one accepting condtion infinitely many times,
+then it sees the other accepting condition infinitely many times as well. -/
+lemma inter_lemma1 {xs : ωSequence Symbol} {ss : ωSequence ((Π i, State i) × Bool)} {i : Bool}
```
chenson2018: A couple of things here. First, I would again use the descriptive names, and we even have one for `NA.Run`. At the beginning of the proof you can have
```lean
have run_trans (k) := (h_run.trans k).right
```
generalizing over k so you don't need to repeat the `have` twice.

Still this is a bit awkward, as each time you are doing `intro k` here, it is really just so `grind` can see through the definition of `Step`. Can you instead add a lemma that `grind` can use?

chenson2018: Please use a more descriptive name for these lemmas. We should almost never be adding declarations that have "lemma" or numbers in their name.

Cslib/Computability/Automata/NABuchiInter.lean
```diff
@@ -0,0 +1,146 @@
...
+  · apply until_frequently_not_leadsTo
+    · intro k
+      have := (h_run.2 k).2
+      grind
+    · apply Frequently.mono h_inf
+      grind
+
+/-- If the intersection automaton sees the accepting condtions of both component automata
+infinitely many times, then its own accepting condition also happens infinitely many times. -/
+lemma inter_lemma2 {xs : ωSequence Symbol} {ss : ωSequence ((Π i, State i) × Bool)}
```
chenson2018: Same issues here as the first lemma.

Cslib/Computability/Automata/NABuchiInter.lean
```diff
@@ -0,0 +1,146 @@
...
+    rw [h1, h2]
+    apply until_frequently_leadsTo_and (h2 := h_inf_t)
+    intro k
+    have := (h_run.2 k).2
+    grind
+
+/-- The language accepted by the intersection automaton is the intersection of
+the languages accepted by the two component automata. -/
+@[simp, scoped grind =]
+theorem inter_language_eq :
```
chenson2018, line 95: Maybe we can figure out a helper `grind` lemma that works with `Frequently.mono`? I'll hold off looking at this proof closely until the changes I mention above.

Cslib/Computability/Automata/NAHist.lean
ctchou: I don't know what `grind` lemma to add here.  I tried:
`
@[scoped grind ·]
theorem hist_run_tr {xs : ωSequence Symbol} {ss : ωSequence (State × Hist)}
    (h_run : (na.addHist start' tr').Run xs ss) (n : ℕ) :
    na.Tr (ss n).fst (xs n) (ss (n + 1)).fst ∧ tr' (ss n) (xs n) (ss (n + 1)).fst = (ss (n + 1)).snd := by
`
with different `grind` annotations (`=>`, `<=`, and `.`) and also with dropping the 2nd conjunct in the consequent.  None worked.  In any case, I don't think it is worthwhile to sweat this `simp`, because the theorems `hist_run_proj` and `hist_run_exists` should form a complete API that any use of the `NA.addHist` construction will ever need.  It doesn't really matter if the proofs of these two theorems are a little ugly.

Cslib/Computability/Automata/NAHist.lean
ctchou: See my comment above.

Cslib/Computability/Automata/NAProd.lean
ctchou: Fixed.

Cslib/Computability/Automata/NAProd.lean
ctchou: Fixed.

Cslib/Computability/Languages/OmegaRegularLanguage.lean
ctchou: I opened `NA.Run`.  But it can get rid of only the `constructor <;>`.  Note that the none of the arguments of the `simp only` actually has anything to do with `NA.Run`.  They are there to get rid of everything other than the `NA.Run` proof obligation, which can now be solved by `use` followed by `grind` with `NA.Run` opened.

Cslib/Computability/Languages/OmegaRegularLanguage.lean
ctchou: Fixed.  Also fixed the same pattern in `IsRegular.iSup` above.

I didn't know this trick.  Why does the parentheses around `ncard_eq_succ` make a difference?  With them it doesn't work.

Cslib/Foundations/Data/OmegaSequence/Temporal.lean
ctchou: Fixed.

Cslib/Foundations/Data/OmegaSequence/Temporal.lean
ctchou: Fixed.

Cslib/Foundations/Data/OmegaSequence/Temporal.lean
ctchou: Fixed.

Cslib/Foundations/Data/OmegaSequence/Temporal.lean
ctchou: Fixed.

Cslib/Foundations/Data/OmegaSequence/Temporal.lean
ctchou: Fixed.

Cslib/Foundations/Data/OmegaSequence/Temporal.lean
ctchou: No, I have not needed that definition even in the automata theory I have formalized so far in my personal project (https://github.com/ctchou/AutomataTheory).  This file `Temporal.lean` really should be part of a formalization of linear-time temporal logic.  If we get that in Cslib in the future, this file should be moved there.

Cslib/Foundations/Data/OmegaSequence/Temporal.lean
ctchou: I think we should await a proper development of temporal logic in Cslib.  Introducing a notation now will create a legacy that that development will need to deal with.  Currently I'm just developing enough temporal logic for me to do the proofs in `NABuchiInter.lean`.

Cslib/Computability/Automata/NABuchiInter.lean
ctchou: Fixed.

Cslib/Computability/Automata/NABuchiInter.lean
ctchou: First, I added `@[scoped grind =]` to the definition of `Step` in `Temporal.lean`, which greatly simplified the proofs in that file.  Then using the `have run_trans` you proposed, I can get rid of all the `intro k` and the subsequent `have` except one `intro k`.  But Lean does not allow me to add the following `grind` lemma` (the error is "invalid pattern"):

`
@[scoped grind =]
lemma foo {xs : ωSequence Symbol} {ss : ωSequence ((Π i, State i) × Bool)}
    (h_run : (interNA na acc).Run xs ss) (k : ℕ) :
    histTrans acc (ss k) (xs k) (ss (k + 1)).fst = (ss (k + 1)).snd := by
  exact (h_run.2 k).2
`

The patterns `=>` and `<=` are allowed, but they do not reproduce the effects of the `have run_trans`.  So I'll keep the `have run_trans` in both lemmas' proofs (though it is not necessary to give the `have` lemma a name).

The lemma names are fixed.

Cslib/Computability/Automata/NABuchiInter.lean
ctchou: Fixed.  New name: `inter_freq_comp_acc_freq_acc`.

Cslib/Computability/Automata/NABuchiInter.lean
ctchou: Fixed.  New name: `inter_freq_acc_freq_acc`.

Cslib/Computability/Automata/NABuchiInter.lean
ctchou: Fixed.

Cslib/Computability/Automata/NABuchiInter.lean
ctchou: If you take a look at:
https://leanprover-community.github.io/mathlib4_docs/Mathlib/Order/Filter/Basic.html#Filter.Frequently.mono
you'll see that I'm basically using the proof context to determine the predicates `p` and `q` and then dispatching the implication (`hpq`) using `grind`.  There are quite a few different `p`s and `q`s in this file and I doubt there is a common pattern.  I've tried giving `Frequent.mono` to `grind` as an argument, but Lean says `Frequent.mono` is not a valid `grind` lemma.

ctchou:
I'll do a PR after this one to add four subdirectories under `Automata`:

- Acceptors
- DA
- NA
- EpsilonNA

and spread the current files among the four.  I don't think adding more levels makes sense.

Cslib/Computability/Automata/NABuchiInter.lean
chenson2018: I just meant lemmas that work similar to `Run.mk`. I have added these.

ctchou: Thanks!  Any reason you used `attribute` rather than attaching the annotations to the theorems themselves?

Cslib/Computability/Automata/NABuchiInter.lean
chenson2018: After trying for a bit, I can't seem to do much better. I made one small change, otherwise this optimal currently.

chenson2018 (approved):

PR #192, opened 2025-12-01
feat: add initial definitions for complexity classes

BoltonBailey:
This is a draft PR for adding some complexity class definitions to CSLib.

This includes

- [x] P
- [x] The Polynomial Hierarchy, (following the ["quantified Boolean formulae" definition on Wikipedia](https://en.wikipedia.org/wiki/Polynomial_hierarchy#Quantified_Boolean_formulae_definition)) including NP and co-NP
- [ ] Various probabilistic classes

---

- [x] depends on: #269
- [ ] depends on: #726


Cslib/Complexity/Class.lean
```diff
@@ -0,0 +1,192 @@
...
+TODO: Bool or Prop?
+-/
+structure BitstringDecisionProblem where
+  toFun : List Bool → Bool
+
+instance : CoeFun BitstringDecisionProblem (fun _ ↦ List Bool → Bool) :=
+  ⟨BitstringDecisionProblem.toFun⟩
+
+instance : Membership (List Bool) BitstringDecisionProblem :=
+  ⟨fun X x ↦ X x⟩
```
eric-wieser: Supporting both `p x` and ` x ∈ p` sounds like a mistake to me, in terms of not having a simp-normal form. I think you should pick just one, and if you pick `∈`, consider just using `carrier : Set (List Bool)`

Cslib/Complexity/Class.lean
```diff
@@ -0,0 +1,192 @@
...
+
+/--
+The class P is the set of decision problems
+decidable in polynomial time by a deterministic Turing machine.
+-/
+def P : ComplexityClass :=
+  { L | isComputableInPolyTime finEncodingListBool finEncodingBoolBool L.toFun }
+
+/--
+The list of all bitstrings of length `n` or less.
```
eric-wieser: Can you either comment on the order, or return a Finset / Multiset?

ctchou:
Reading `ORGANISATION.md`, it seems to me that `Encoding.lean` should be placed under `Cslib/Foundations/` and `Class.lean` under `Cslib/Computability/`.

Cslib/Complexity/Class.lean
BoltonBailey: In the formal-conjectures PR that has some of the same content, I was advised to just use an abbrev, so I'll do that here.

Cslib/Computability/Languages/ComplexityClass.lean
```diff
@@ -0,0 +1,199 @@
...
+
+/--
+The type of decision problems on bitstrings.
+
+We define these as functions from lists of booleans to booleans,
+implictly assuming the usual encodings.
+
+TODO: An Decision Problem type over arbitrary types.
+-/
+abbrev BitstringDecisionProblem := List Bool → Bool
```
eric-wieser:
```suggestion
abbrev BitstringDecisionProblem : Type := List Bool → Bool
```
etc

BoltonBailey:
A few people seem to be commenting on/reviewing this PR, so be aware: I think some of these sorries are going to involve further development of TM2ComputableInPolyTime, so I don't expect to finish them soon, and I don't plan to move this PR out of draft mode until that's done.

PR #195, opened 2025-12-01
chore: add weekly linting workflow

### LINKED ISSUE #185: move `mergeWithGrind` linter to cron workflow
See #184. We should do something similar to [this Mathlib workflow](https://github.com/leanprover-community/mathlib4/blob/1f31ce1aba09b4a93a30abc2ad7248955f10a427/.github/workflows/weekly-lints.yml) if we'd like to enable slower or noisy syntax linters to run weekly.

jessealama:
This adds a weekly linting workflow mimicking Mathlib's approach. It re-enables the mergeWithGrind linter on a weekly cron schedule (Mondays at 05:00 UTC) and posts results to the CSLib Zulip stream. The workflow can also be triggered manually.

Feedback welcome on enriching the report script, for example by emitting the number of grind instances encountered during the build.

Closes #185

### DIFF
--- a/.github/workflows/weekly-lints.yml
+++ b/.github/workflows/weekly-lints.yml
@@ -1,0 +1,57 @@
+name: Weekly linting report
+
+on:
+  schedule:
+    - cron: '0 5 * * 1'  # Run at 05:00 UTC every Monday
+  workflow_dispatch:  # Allow manual triggering
+
+jobs:
+  weekly-lints:
+    name: Weekly Linting
+    runs-on: ubuntu-latest
+    if: github.repository == 'leanprover/cslib'
+    steps:
+    - uses: actions/checkout@v4
+      with:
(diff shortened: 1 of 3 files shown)


fmontesi:
LGTM but I can't test it. @kim-em @chenson2018 what do you think?

chenson2018:
At a glance this looks fine and analogous to what Mathlib has, but I've not tested either. If @jessealama has at least run once from his fork to make sure there's no simple syntax problems, I say merge it and then trigger a run manually to check that it works.

Also, can you add a short description in `scripts/README.md`?

jessealama:
Yes, I've run this locally; there aren't any syntax errors. If you like (and you're on macOS), you can check out this branch and run the action locally using the `act` tool.

chenson2018 (approved):

PR #198, opened 2025-12-04
feat: eta expansion for CLL

fmontesi:
Adds eta-expansion for CLL, i.e., a def that given a proposition `a` returns a `Proof` of `{a, a.dual}` that applies the axiom only to atomic propositions.

TODO for the future: prove that it is correct. :-)

### DIFF
--- a/CslibTests/GrindLint.lean
+++ b/CslibTests/GrindLint.lean
@@ -51,6 +51,8 @@ open_scoped_all Cslib
 #grind_lint skip Cslib.CLL.Proof.with_inversion₁.eq_1
 #grind_lint skip Cslib.CLL.Proof.with_inversion₂.eq_1
 #grind_lint skip Cslib.CLL.Proposition.Equiv.trans
+#grind_lint skip Cslib.CLL.Proposition.bang_top_eqv_one.eq_1
+#grind_lint skip Cslib.CLL.Proposition.parr_top_eqv_top.eq_1
 #grind_lint skip Cslib.CLL.Proposition.tensor_assoc.eq_1
 #grind_lint skip Cslib.CLL.Proposition.tensor_distrib_oplus.eq_1
 #grind_lint skip Cslib.CLL.Proposition.tensor_symm.eq_1


Cslib/Logics/LinearLogic/CLL/EtaExpansion.lean
```diff
@@ -0,0 +1,81 @@
...
+def Proposition.expand (a : Proposition Atom) : ⇓{a, a⫠} :=
+  match a with
+  | atom x
+    | atomDual x
+    | 1
+    | ⊥
+    | ⊤ => Proof.ax
+  | 0 =>
+    Proof.top (Γ := {0})
+    |> .rwConclusion (by simp only [Proposition.dual]; rfl : ⊤ ::ₘ {0} = 0⫠ ::ₘ {0})
```
chenson2018: You probably should have a `grind` annotation
```lean
@[grind =]
theorem Proposition.top_eq_zero_dual : ⊤ = (0⫠ : Proposition Atom) := by rfl
```
where this notation is defined, then this becomes just:

```suggestion
    |> .rwConclusion (by grind : ⊤ ::ₘ {0} = 0⫠ ::ₘ {0})
```

Cslib/Logics/LinearLogic/CLL/EtaExpansion.lean
```diff
@@ -0,0 +1,81 @@
...
+attribute [local grind _=_] Multiset.coe_eq_coe
+attribute [local grind _=_] Multiset.cons_coe
+attribute [local grind _=_] Multiset.singleton_add
+attribute [local grind =] Multiset.add_comm
+attribute [local grind =] Multiset.add_assoc
+attribute [local grind =] Multiset.insert_eq_cons
+
+/-- The η-expansion of a proposition `a` is a `Proof` of `{a, a⫠}` that applies the axiom
+only to atomic propositions. -/
+def Proposition.expand (a : Proposition Atom) : ⇓{a, a⫠} :=
```
chenson2018, line 26: Several of the type annotations you have for `grind` proofs can be omitted, but maybe this is intentional for clarity?

Cslib/Logics/LinearLogic/CLL/EtaExpansion.lean
```diff
@@ -0,0 +1,81 @@
...
+    |> .rwConclusion (by grind : (a⫠ ⊗ b⫠) ::ₘ ({a} + {b}) = a ::ₘ b ::ₘ {(a⫠.tensor b⫠)})
+    |> Proof.parr
+  | oplus a b =>
+    Proof.with
+      (a.expand.oplus₁ (b := b) |> .rwConclusion (by grind : (a ⊕ b) ::ₘ {a⫠} = a⫠ ::ₘ {a ⊕ b}))
+      (b.expand.oplus₂ (a := a) |> .rwConclusion (by grind : (a ⊕ b) ::ₘ {b⫠} = b⫠ ::ₘ {a ⊕ b}))
+    |> .rwConclusion (by grind : (a⫠ & b⫠) ::ₘ {a ⊕ b} = {(a ⊕ b), (a⫠ & b⫠)})
+  | .with a b =>
+    Proof.with
+      (a⫠.expand.oplus₁ (b := b⫠) |> .rwConclusion (by grind [Proposition.dual_involution] :
```
chenson2018: You do not need to manually pass `Proposition.dual_involution` to `grind` here or below, as it is already annotated.

Cslib/Logics/LinearLogic/CLL/EtaExpansion.lean
```diff
@@ -0,0 +1,81 @@
...
+      (a.expand.quest.rwConclusion (by grind : ʔa ::ₘ {a⫠} = a⫠ ::ₘ {ʔa}))
+    |> Proof.rwConclusion (by grind : ((a⫠).bang ::ₘ {ʔa}) = {ʔa, (ʔa)⫠})
+  | bang a =>
+    Proof.bang
+      (Γ := {ʔ(a⫠)})
+      rfl
+      (a⫠.expand.quest.rwConclusion (by grind : ʔa⫠ ::ₘ {a⫠⫠} = a⫠⫠ ::ₘ {ʔa⫠}))
+    |> Proof.rwConclusion (by grind : ((a⫠⫠).bang ::ₘ {ʔa⫠}) = {ʔa⫠, (ʔa⫠)⫠})
+    |> Proof.rwConclusion (by grind : {ʔa⫠, (ʔa⫠)⫠} = {ʔa⫠, !(a⫠⫠)})
+    |> Proof.rwConclusion (by rw [Proposition.dual_involution] : {ʔa⫠, !(a⫠⫠)} = {ʔa⫠, !a})
```
chenson2018: I'd use `grind` here for style consistency.

Cslib/Logics/LinearLogic/CLL/Basic.lean
```diff
@@ -114,6 +114,13 @@ def Proposition.dual : Proposition Atom → Proposition Atom
 
 @[inherit_doc] scoped postfix:max "⫠" => Proposition.dual
 
+/-- Duality preserves size. -/
+@[scoped grind .]
+theorem Proposition.dual_sizeOf (a : Proposition Atom) : sizeOf a = sizeOf a⫠ := by
+  induction a
+  any_goals simp [Proposition.dual]
+  all_goals grind
```
chenson2018: Is there a reason I'm missing that using `{any,all}_goals` is better than this?
```suggestion
  induction a <;> simp [dual] <;> grind
```

Cslib/Logics/LinearLogic/CLL/EtaExpansion.lean
```diff
@@ -0,0 +1,81 @@
...
+    |> Proof.rwConclusion (by grind : {ʔa⫠, !a} = {!a, ʔa⫠})
+termination_by sizeOf a
```
chenson2018: I think it is more usual to write
```suggestion
termination_by a
```
which gives the same goals.

Cslib/Logics/LinearLogic/CLL/EtaExpansion.lean
chenson2018: (and as a side note if they were needed you could omit `Proposition` for brevity since we are in that namespace already)

Cslib/Logics/LinearLogic/CLL/Basic.lean
fmontesi: Does that make sure that simp is used as final tactic?

fmontesi: Mmh mine doesn't either though... so yes, I'll switch to yours. :-)

chenson2018: You mean for the purposes of avoiding non-terminal `simp`? `grind` is a tactic that is meant to be allowed after `simp`, as checked by the flexible linter. (You can see `Mathlib/Tactic/Linter/FlexibleLinter.lean` for details)

I personally still try to avoid mixing too much `simp` and `grind`, but I think it makes sense here for `sizeOf`. I was surprised actually `grind` didn't already handle those.

Cslib/Logics/LinearLogic/CLL/EtaExpansion.lean
fmontesi: It's more a remnant of my thinking in doing the proofs. Sometimes it's needed to make sure I get the right rewriting. I've eliminated the unnecessary ones.

fmontesi:
Should be ok now, thanks!

I was actually toying with the idea of making a `Proof.rwByGrind` def taking the target wished sequent as an implicit parameter or something like that, but for now `.rwConclusion (by grind)` looks good enough.

chenson2018:
Your last commit caused some `grind` instantiations past the threshold that `CslibTests.GrindLint` checks for. I have added these as additional exceptions. It is a little unusual for this to happen to equational lemmas of an inductive, but it makes sense because of the nature of this recording a proof non-propositionally.

In general, there is some new tooling for `grind_pattern` (see docs added in https://github.com/leanprover/lean4/pull/11429) that allows specifying additional constraints on how `grind` is applied. I've been working on seeing where these can be used to remove some of the current exceptions.

PR #199, opened 2025-12-04
chore: only build docs in leanprover/cslib repo

chenson2018:
Adds a condition to `.github/workflows/docs.yml` to only run in the `leanprover/cslib` repo.


fmontesi (approved):

arademaker:
I guess the goal is to avoid building docs in forks?

PR #201, opened 2025-12-05
feat: query complexity via free monads

tannerduve:
This PR introduces a generic query-based computation model, and reuses the `TimeM` monad from #165 to give it a simple cost semantics


Cslib/Algorithms/MergeSort/MergeSort.lean
```diff
@@ -0,0 +1,95 @@
...
+
+/-- Lift a comparison into the query language at the top level. -/
+def cmpVal (x y : Nat) : Prog Bool :=
+  FreeM.lift (QueryF.cmp x y)
+
+/-- Concrete semantics for primitive queries, used to run programs. -/
+def evalQuery : {ι : Type} → QueryF ι → ι
+  | _, .read _      => 0
+  | _, .write _ _   => PUnit.unit
+  | _, .cmp x y     => x ≤ y
```
eric-wieser, line 48: Shouldn't this be more like
```lean
def evalQuery [LE α] [Inhabited α] : {ι : Type} → QueryF ι → StateT (Array α) ι
  | _, .read i      => return (← get)[i]!
  | _, .write i x   => do modify fun a => a.set! i x
  | _, .cmp i j     => do let a ← get; return a[i]! ≤ a[j]!
```
You need a real model if you want to prove correctness

Cslib/Algorithms/MergeSort/MergeSort.lean
```diff
@@ -0,0 +1,95 @@
...
+
+open Cslib.Algorithms
+
+
+/-- The Model for comparison sorting natural-number registers.-/
+inductive QueryF : Type → Type where
+  /-- Read the value stored at index `i`. -/
+  | read  : Nat → QueryF Nat
+  /-- Write value `v` at index `i`. -/
+  | write : Nat → Nat → QueryF PUnit
```
eric-wieser, line 36: If you refer to names in the docstring they should exist in the code
```lean
  /-- Write value `v` at index `i`. -/
  | write (i : Nat) (v : Nat) : QueryF PUnit
```
etc

Cslib/Algorithms/MergeSort/MergeSort.lean
```diff
@@ -0,0 +1,95 @@
...
+-/
+
+open Cslib
+
+namespace Cslib.Algorithms.MergeSort.QueryBased
+
+open Cslib.Algorithms
+
+
+/-- The Model for comparison sorting natural-number registers.-/
```
eric-wieser, line 31: Why not generalize to arbitrary registers?

Cslib/Algorithms/MergeSort/MergeSort.lean
eric-wieser: The `Inhabited` here flags that this query API is not great, as it allows illegal queries. You might want to parametrize `QueryF` by the array length.

Cslib/Algorithms/MergeSort/MergeSort.lean
Shreyas4991: This is still WIP. I will make this change once I have generalized the query type, and I am instantiating it.

Shreyas4991: The PR started from a proof-of-concept in the zulip, posted by Tanner. It has some way to go.

Cslib/Algorithms/MergeSort/MergeSort.lean
Shreyas4991: thanks. That's a valid point. Fwiw, this entire example has several points where it is still possible to sneak in flawed code. Currently I am trying to get the typeclasses and composition theorems right, then I can redo this example.

Cslib/Algorithms/MergeSort/MergeSort.lean
eric-wieser, line 36: The fact your sorting implementation doesn't use these at all suggests that they're not really useful as query operators. Even if you did force the API to manually `read` and `write`, I don't think you would be able to prevent an implementation that:
* reads every entry in sequence into a `List`
* does a mergesort
* writes the outputs back

which makes counting reads and writes pretty uninteresting.

Cslib/Algorithms/QueryModel.lean
```diff
@@ -0,0 +1,110 @@
...
+def forLoop {QType} (n : Nat) (body : Nat → Prog QType PUnit) : Prog QType PUnit :=
+  go n
+where
+  /-- Auxiliary recursive worker for `forLoop`. -/
+  go : Nat → Prog QType PUnit
+    | 0       => pure ()
+    | i + 1   =>
+      body i >>= fun _ => go i
+
+end Prog
```
eric-wieser, line 65: As discussed on zulip, this is `do if` and `do for`, no need for new defs

Cslib/Algorithms/MergeSort/MergeSort.lean
Shreyas4991: Do note that `evalQuery` is meant to execute the computations. The time complexity evaluation will be done by translation to the time monad with ticks

Cslib/Algorithms/MergeSort/MergeSort.lean
Shreyas4991: The correct model would have `compare` and `swap` operations.

PR #207, opened 2025-12-09
CLRS chapter 6

cs14b052:
Adding CLRS Chapter 6 -- Heapsort in Lean


arademaker:
This is going  to Boole-sandbox, so we don't really do PR review in this branch.

PR #208, opened 2025-12-10
NFA empty language check

luc-edixhoven:
Added a predicate stating whether an `Acceptor`'s language is empty, and the proof to a statement in Hopcroft: an NFA's (`NA.FinAcc`) language is empty iff there are no reachable accepting states.

### DIFF
--- a/Cslib/Computability/Automata/Acceptors/Acceptor.lean
+++ b/Cslib/Computability/Automata/Acceptors/Acceptor.lean
@@ -28,6 +28,10 @@ def language [Acceptor A Symbol] (a : A) : Language Symbol :=
 theorem mem_language [Acceptor A Symbol] (a : A) (xs : List Symbol) :
   xs ∈ language a ↔ Accepts a xs := Iff.rfl
 
+/-- The language of an `Acceptor` is empty iff it does not accept any string. -/
+def language.IsEmpty [Acceptor A Symbol] (a : A) : Prop :=
+  ¬∃ xs, xs ∈ language a
+
 end Acceptor
 
 end Cslib.Automata
--- a/Cslib/Computability/Automata/NA/Basic.lean
+++ b/Cslib/Computability/Automata/NA/Basic.lean
@@ -79,6 +79,69 @@ instance : Acceptor (FinAcc State Symbol) Symbol where
   Accepts (a : FinAcc State Symbol) (xs : List Symbol) :=
     ∃ s ∈ a.start, ∃ s' ∈ a.accept, a.MTr s xs s'
 
+/-- The language of an NFA is empty iff there are no final states
+reachable from the initial state [Hopcroft2006]. -/
+theorem is_empty (a : FinAcc State Symbol) :
+    -- a.IsEmpty ↔ Acceptor.language.IsEmpty a := by
+    Acceptor.language.IsEmpty a ↔
+    ¬∃ s1 s2, a.CanReach s1 s2 ∧ s1 ∈ a.start ∧ s2 ∈ a.accept
+    := by
+  apply Iff.intro <;> intro h1
+  -- If language empty, then no reachable accepting states.
+  case mp =>
+    rw [Acceptor.language.IsEmpty, not_exists] at h1
+    -- Assume that there is a reachable accepting state.
+    apply by_contradiction
+    intro h2
+    rw [not_not] at h2
+    -- Then concrete start and accepting states s1 and s2, resp.
+    cases h2 with
+    | intro s1 h2
+    cases h2 with
+    | intro s2 h2
+    rw [LTS.CanReach] at h2
+    cases h2 with
+    | intro h2 _
+    -- And concrete trace xs from s1 to s2.
+    cases h2 with
+    | intro xs h2
+    -- Then xs must be in the language.
+    have h3 : Acceptor.Accepts a xs := by
+      simp only [instAcceptor]
+      grind
+    rw [← Acceptor.mem_language] at h3
+    -- But also not in the language.
+    have h4 : xs ∉ Acceptor.language a := by exact h1 xs
+    -- Which is contradictory.
+    exact h4 h3
+  -- If no reachable accepting states, then language empty.
+  case mpr =>
+    -- rw [IsEmpty] at h1
+    simp only [not_exists, not_and] at h1
+    -- Assume that the language is not empty.
+    apply by_contradiction
+    intro h2
+    rw [Acceptor.language.IsEmpty, not_not] at h2
+    -- Then concrete accepted trace xs.
+    cases h2 with
+    | intro xs h2
+    -- Then concrete start and accepting states s1 and s2, resp.,
+    -- and s2 reachable from s1.
+    have h3 : ∃ s1 s2, s1 ∈ a.start ∧ s2 ∈ a.accept ∧ a.MTr s1 xs s2 := by
+      simp at h2
+      grind
+    cases h3 with
+    | intro s1 h3
+    cases h3 with
+    | intro s2 h3
+    have h4 : a.CanReach s1 s2 := by
+      rw [LTS.CanReach]
+      grind
+    -- Premise implies that s2 should not be an accepting state.
+    have h5 := h1 s1 s2 h4 h3.left
+    -- Which is contradictory.
+    exact h5 h3.right.left
+
 end FinAcc
 
 /-- Nondeterministic Buchi automaton. -/


Cslib/Computability/Automata/Acceptors/Acceptor.lean
```diff
@@ -28,6 +28,10 @@ def language [Acceptor A Symbol] (a : A) : Language Symbol :=
 theorem mem_language [Acceptor A Symbol] (a : A) (xs : List Symbol) :
   xs ∈ language a ↔ Accepts a xs := Iff.rfl
 
+/-- The language of an `Acceptor` is empty iff it does not accept any string. -/
+def language.IsEmpty [Acceptor A Symbol] (a : A) : Prop :=
```
fmontesi, line 32: I'd make this a def of `Language`. Surprisingly, I cannot find such a def in mathlib already. @chenson2018 @ctchou Are you aware of an existing def for 'empty Language'?
We can put this in Computability/Languages/Language.lean.

Cslib/Computability/Automata/NA/Basic.lean
```diff
@@ -79,6 +79,69 @@ instance : Acceptor (FinAcc State Symbol) Symbol where
   Accepts (a : FinAcc State Symbol) (xs : List Symbol) :=
     ∃ s ∈ a.start, ∃ s' ∈ a.accept, a.MTr s xs s'
 
+/-- The language of an NFA is empty iff there are no final states
+reachable from the initial state [Hopcroft2006]. -/
+theorem is_empty (a : FinAcc State Symbol) :
```
fmontesi, line 84: We should probably call this `isEmpty_canReach` to follow coding conventions.

Cslib/Computability/Automata/NA/Basic.lean
```diff
@@ -79,6 +79,69 @@ instance : Acceptor (FinAcc State Symbol) Symbol where
...
+    -- a.IsEmpty ↔ Acceptor.language.IsEmpty a := by
```
fmontesi, line 85: Erase the comment.

Cslib/Computability/Automata/Acceptors/Acceptor.lean
ctchou: I'm not sure we need this definition.  Given `a : NA State Symbol`, you can just say `language a = 0` to state that `a`'s language is empty.  Mathlib views `Language` as a Kleene algebra, so the empty language is simply 0.  For `ωLanguage`, use `⊥` instead of 0.

ctchou:
First, a general comment:`NA/Basic.lean` is meant to contain only the basic definitions about NA.  It is imported (directly or indirectly) by every file that want to talk about NA and hence should be kept small.  If you want to prove properties about NA, please create a separate file (or files) in which to do so.

Second, a lot of properties about NFA have been proved in mathlib:
https://leanprover-community.github.io/mathlib4_docs/Mathlib/Computability/NFA.html
In particular, it seems to me that the mathematical content of the theorem `is_empty` is essentially that of:
https://leanprover-community.github.io/mathlib4_docs/Mathlib/Computability/NFA.html#NFA.mem_evalFrom_iff_nonempty_path
Do we really want to re-state and re-prove such results in cslib?

Third, is there a "main theorem" or a set of results about NFA that @luc-edixhoven aims to prove?  Is there a context or future plan about this PR?

Some of the issues I raised above should probably be discussed in Zulip.

fmontesi:
I pointed this problem to @luc-edixhoven myself as a possible warm-up exercise to get used to our automata development.

In general, I agree that we should indeed ask ourselves this kind of questions. I'd missed this because I tried to find a result about language, and didn't notice the one about paths in NFA in mathlib.

I'd like to merge this when it's ready simply because it offers us a direct result on our API that I cannot find directly in mathlib, but I'll open a thread on Zulip to have a larger discussion on this topic.

luc-edixhoven:
Thank you both for your comments (both general and detailed)!

As mentioned by Fabrizio, he suggested this to me as an exercise to acquaint myself a little with Lean/CSLib by getting some hands-on experience. I'm not aware of any particular future plan regarding this PR.

I wasn't aware of the NFA results in mathlib either. The two theorems look similar to me, but different in scope: the one in mathlib is a statement about a single trace, while `is_empty` is a statement about the language. I'd say `is_empty` would be a corollary of some sorts of the other. In any case, I'm not sure how much my participation would add to the larger discussion on whether to re-do or duplicate results from mathlib. If it's alright with you, I'll just await your conclusion on that front.

A couple of thoughts, should they be helpful:
- Would it be helpful to include some comment in `NA/Basic` (in the section about `FinAcc`) mentioning or referring to `Mathlib/Computability/NFA`, e.g., an implementation note stating that the latter also exists?
- If I understand correctly that `NA/Basic.FinAcc` and `Mathlib/Computability/NFA` are disjoint implementations of NFAs, then what guarantees do the theorems in the latter give for the properties of the former?

Cslib/Computability/Automata/Acceptors/Acceptor.lean
luc-edixhoven: I agree. I initially tried `Acceptor.language a = \empty`, which didn't work --- if I understand it correctly, because the lhs is a language and the rhs is a set. I then had a look at https://leanprover-community.github.io/mathlib4_docs/Mathlib/Computability/Language.html#Language.instZero. I couldn't figure out how to use it, but perhaps that's what you're looking for.

ctchou:
@luc-edixhoven, I do want to insist that please create a new file under `Automata/NA` in which to put you code.

Considering that `LTS.CanReach` is almost `LTS.MTr`:
```
def LTS.CanReach (s1 s2 : State) : Prop := ∃ μs, lts.MTr s1 μs s2
```
and the acceptance condition is defined in terms of `LTS.MTr`, the theorem you want should be provable much more simply.  Indeed, it is:

```
namespace NA.FinAcc

open Acceptor

variable {State Symbol : Type*}

example (a : FinAcc State Symbol) :
    language a = 0 ↔ ¬∃ s1 s2, a.CanReach s1 s2 ∧ s1 ∈ a.start ∧ s2 ∈ a.accept := by
  simp only [LTS.CanReach]
  constructor
  · rintro h ⟨s1, s2, ⟨xs, _⟩, _⟩
    have : xs ∈ language a := by grind
    grind
  · intro h
    by_contra h_contra
    obtain ⟨xs, _⟩ := Set.nonempty_iff_ne_empty.mpr h_contra
    grind

end NA.FinAcc
```

Note that there is no need for `IsEmpty`.

ctchou:
As to your second question, we do prove in the following file:
https://github.com/leanprover/cslib/blob/main/Cslib/Computability/Languages/RegularLanguage.lean
that the notion of regular languages is captured by cslib's NFA and DFA.  In general, the notion of regular languages is much more robust than those of NFA and DFA.  You can make variations in the latter without impacting the former.  For example, some authors define NFA to have a single start state, which doesn't change the notion of regular languages but would render some constructions cumbersome to do.  For example, it would be ugly to do a sum construction on NFA if each NFA has a single start state:
https://github.com/leanprover/cslib/blob/main/Cslib/Computability/Automata/NA/Sum.lean

luc-edixhoven:
> I do want to insist that please create a new file under Automata/NA in which to put you code.

@ctchou Ah! Sorry, I didn't mean to ignore your comment. I intended to wait for the discussion on whether the code should be included at all before making further changes; it would've been clearer if I'd said so explicitly. (The same goes for Fabrizio's comments.)

> As to your second question, we do prove in the following file:

Thanks! That's interesting to know. I'm familiar with the various equivalent notions of finite automata, but wasn't aware of the two theorems linking regular languages to the cslib automata :)
If I understand correctly, then this theorem does not link the mathlib NFA to the cslib NFA beyond language equivalence, so I don't directly see how the mathlib NFA results would translate to cslib NFAs (i.e., not generic regular language results but specifically ones about NFAs, such as the earlier https://leanprover-community.github.io/mathlib4_docs/Mathlib/Computability/NFA.html#NFA.mem_evalFrom_iff_nonempty_path). Having said that, this is just my curiosity speaking; it's not relevant for this PR.

> the theorem you want should be provable much more simply. Indeed, it is:

Many thanks for the simplification! I'm unfamiliar with a couple of the constructs used, but I'll gratefully study it!

> Note that there is no need for IsEmpty.

I see now, yes! I initially tried `\empty` for the empty language, but I didn't know that `0` would work instead. Thanks! :)

Incidentally, I'm not sure with what intention you used `example` rather than `theorem` in your simplified proof, but I like it. Perhaps that'd be the better way to include it (pending the outcome of the aforementioned larger discussion): as a (simple) demonstration, rather than a formal result. Would you agree?
Additionally, assuming that simpler code is better, the final code would likely be the simplified code you posted, which I had no (or little) hand in. Would it still be appropriate to keep that in this PR? (I'm not familiar with PR etiquette and am not sure whether it matters, but I don't wish to claim inappropriate credit.)

Cslib/Computability/Automata/Acceptors/Acceptor.lean
luc-edixhoven: `0` works for the empty language; see https://github.com/leanprover/cslib/pull/208#issuecomment-3643254211 (thanks to ctchou).

ctchou:
You use `example` when you don't want to think about what name to give to a theorem.  Of course, you won't be able to refer to that theorem later.

I think you should join the following Zulip discussions:

[#mathlib4 > Move automata theory in mathlib to CSLib?](https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/Move.20automata.20theory.20in.20mathlib.20to.20CSLib.3F/with/563273770)

[#mathlib4 > Is there a plan to offload Mathlib.Computability to Cslib?](https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/Is.20there.20a.20plan.20to.20offload.20Mathlib.2EComputability.20to.20Cslib.3F/with/563288104)

fmontesi:
@luc-edixhoven you could, for example, create the new file and put both you and [ctchou](https://github.com/ctchou) to the list of 'Authors'.

As for the filename.. mmh.. `NA/Empty.lean`? `NA/Zero.lean`? @ctchou, any ideas?

ctchou:
It depends on what @luc-edixhoven wants to prove and whether there is a coherent theme for those results.

PR #212, opened 2025-12-13
chore: bump toolchain to v4.26.0

kim-em:


chenson2018:
Kim, I'm boarding a flight at the moment, but if you look at nightly testing I think you'll see where I turn off the currently failing `linter.unusedDecidableInType` linter that triggers in some files because we have a term elaborator it isn't aware of.

chenson2018:
The further failures I see are again things I have addressed in nightly testing.

PR #215, opened 2025-12-14
feat: equivalents of Church-Rosser, some normal form properties

chenson2018:
This is work from the ItaLean conference by myself, @XYUnknown, @a-dangelo, and @Aristotle-Harmonic. I made the decision here to move this file into only extending the existing `Relation` namespace which I think is much clearer. I also cleaned up some names for style consistency.


chenson2018:
(The excess of merges is because I used `git commit --amend` to ensure everyone shows up properly as coauthors in the git history)

chenson2018:
> This looks good to me! Is it time to start adding some grind annotations to our Relation stuff, maybe?

Yes, I agree. A few of them are already pending in a Mathlib PR, and we can add whatever additional ones we want here.

PR #216, opened 2025-12-14
feat: prove that omega-regular languages are closed under left-concatenation by regular languages

ctchou:
This patch proves that the concatenation of a regular language and an omega-regular language is omega-regular.

The construction in `NA/Concat.lean` (perhaps with some modification) should work for proving that the concatenation of two regular languages is regular as well.  But that is for future work.

### DIFF
--- a/Cslib/Computability/Languages/OmegaRegularLanguage.lean
+++ b/Cslib/Computability/Languages/OmegaRegularLanguage.lean
@@ -169,7 +169,7 @@ theorem IsRegular.iInf {I : Type*} [Finite I] {s : Set I} {p : I → ωLanguage 
 @[simp]
 theorem IsRegular.hmul {l : Language Symbol} {p : ωLanguage Symbol}
     (h1 : l.IsRegular) (h2 : p.IsRegular) : (l * p).IsRegular := by
-  obtain ⟨State1, h_fin1, ⟨na1, acc1⟩, rfl⟩ := Language.IsRegular.iff_cslib_nfa.mp h1
+  obtain ⟨State1, h_fin1, ⟨na1, acc1⟩, rfl⟩ := Language.IsRegular.iff_nfa.mp h1
   obtain ⟨State2, h_fin1, ⟨na2, acc2⟩, rfl⟩ := h2
   let State := State1 ⊕ State2
   let na := NA.concat ⟨na1, acc1⟩ na2


Cslib/Foundations/Semantics/LTS/Basic.lean
```diff
@@ -127,6 +127,39 @@ theorem LTS.MTr.nil_eq (h : lts.MTr s1 [] s2) : s1 = s2 := by
   cases h
   rfl
 
+/-- A generalized `MTr` is a `MTr` augmented with intermediate states. -/
+@[scoped grind =]
+def LTS.GMTr (lts : LTS State Label)
```
fmontesi: I have some thoughts on the design of this but I think that having a proof using this (as in your PR) is very important to explore them. Have you considered them already perhaps?
If not, I think the best would be that we merge what you have here (so that we progress), adding to its docstring something like 'Warning: this API is currently under active exploration, for internal use only', and then we/I can play with the design in another PR (in an informed way, that is, by making sure your proofs still work).

Specifically, we could also define this with any of these other signatures:
- `(s : State) (steps : List (Label x State))`, where `s` is the first state of the execution and then each element of the list is the ordered sequence of steps, ending in the state of the last element of the list. This guarantees that the labels and states are correctly paired and of the right lengths by construction. It might give a simpler, inductive definition (similar to `MTr`). But it might make the relation to `MTr` a bit more complicated, I haven't tried that yet. It's the definition I'd expect some category theorists to like. (See, e.g., https://coalg.org/cmcs12/papers/00010169.pdf)
- `(e : Execution)`, where execution bundles the start state and the list above. The pro of bundling is that then we can define appropriate projection functions, like `trace` (returning the list of labels) and `states` (returning the list of states). I'm not sure how useful this could be but worth keeping in mind for the future.
- `(steps : List (State x Label x State))`, the idea being that each step's third component must be the same as the next step's first component. This is what sometimes people do on pen and paper but I find it  unwieldy in formalisations. Also, there's something nice about seeing the start state explicitly in the signature (as in your option and the previous option here), which I can't pinpoint precisely right now but merits attention.

I'd also like to find a more specific name than 'generalized'. Maybe `IsExecution` -- at least, execution is what I called a sequence of transitions in [Montesi2023], partially inspired by the literature and also to avoid overlap with 'run' (which usually is a sequence of states).

chenson2018:
Sounds like Fabrizio has the theory covered, but one more small point in the "merge what you have category". I saw that the way you fixed the `grind` linting issues was to manually specify `NA.concat` when calling grind. If you think it could eventually be a reasonable rule, I would be okay if you add it as a linting exception. I almost prefer this if it doesn't cause other problems, as it'll then be a little easier for me to find later. (I'll be working on adding `grind_pattern` constraints now that they are released and encourage anyone else with time to try them out too.)

ctchou:
Regarding the issues that @fmontesi raised:

Yes, I have thought about these issues, even during my previous personal automata theory project.  First, using lists of pairs or triples will involve a lot of packing and unpacking of those pairs or triples in and out a list, which seems an unnecessary hassle to me.  For example, in my proof of `concat_run_exists`, I use the finite state sequence `ss1` asserted to exist by `LTS.MTr.existsGMTR` by concatenating it with `ss2`.  If I had used pairs, I would have had to extract those states from the list of pairs and to construct `ss1`.  If you think about it, you'll realize that the construction is not just an application of `List.map` or `List.unzip`, because the initial state is special and not in any pairs.  In the case of triples of the form "(state, label, state')", there is the additional constraints that the "state'" of each triple needs to equal the "state" of the next triple.  Those constraints have to be carried around, applied, and proved.  That again seems an unnecessary hassle.

In automata theory, sequences of symbols are concatenated with or partitioned into sequences of symbols, and sequences of states are concatenated with or partitioned into sequences of states.  So it is natural to keep sequences of symbols separate from sequences of states and not mix them.  The two types of sequences are related by constraints induced by `LTS.Tr` which ultimately boil to reasoning about sequence elements with different indices of type `Nat`.  Lean's `omega` and `grind` tactics are very good at reasoning about such indices and you can discharge most proof obligations automatically.  The definition of `LTS.GMTr` is designed specifically to take advantage of that strength of Lean. 

I don't have a strong opinion about the name `GMTr`.  We can call it `IsExecution` if you like.

Regarding the issues that @chenson2018 raised:

I think it is OK to not have a `grind` annotations on `NA.concat` and `concat_languages_eq`.  I don't expect `NA.concat` to be used in any other files.  Its only purpose is to prove ``concat_languages_eq`, which will probably only be used in the proof of `IsRegular.hmul` based on the experience from my personal automata theory project.  Both are technical in the sense that there are alternative definitions of automata concatenation which can also be used to prove `IsRegular.hmul`.  For example, I used a slightly different one in my personal automata theory project:
https://github.com/ctchou/AutomataTheory/blob/main/AutomataTheory/Automata/Concat.lean

fmontesi:
@ctchou Thanks! I agree the 'flat' approach is better. Though I'd have expected an inductive for something like `IsExecution`/`GMTr`. It seems to me that what you really need is just `LTS.MTr.exists_GMTr` as a lemma to extract the intermediate states in proofs, by expanding the definition of `GMTr` in its statement, e.g., (I'm not reporting the implicit params)

```
theorem LTS.MTr.exists_states (h : lts.MTr s1 μs s2) :
    ∃ ss, ∃ _ : ss.length = μs.length + 1, ss[0] = s1 ∧ ss[μs.length] = s2 ∧
      ∀ k, ∀ _ : k < μs.length, lts.Tr ss[k] μs[k] ss[k + 1]
```

Could we state that and postpone a bit the discussion on what the definition of an execution is perhaps? Would that suffice for your development for now?

ctchou:
@fmontesi As you suggested, I got rid of `LTS.GMTr` and its associated theorems and proved `LTS.MTr.exists_state`, which is indeed all I need.

BTW, it may interest you that in my personal automata theory project, I didn't have any inductive notion of multi-step transitions at all and both finite and infinite executions are defined on infinite sequences:
https://github.com/ctchou/AutomataTheory/blob/main/AutomataTheory/Automata/Basic.lean
Everything works (including theorems about languages on finite words), except that occasionally I had to assume that the alphabet type is inhabited (because I needed to pad a finite sequence of symbols into an infinite one in proofs).  It eliminates all the hassles with proving that every list index is within list length bound and reduces every induction to induction on `Nat`.  I was led to this approach because in automata theory on infinite words, there is no escape from reasoning about infinite executions as infinite sequences.  (In contrast, in automata theory on finite words, only the initial state and the final state of an execution matter.). Then I realized that finite executions can be treated in the same manner as well, as long as you view an infinite execution as a representative of the equivalent class of infinite executions that share a common finite prefix.

ctchou:
The last commit refactors the definition of `NA.concat` into a more readable form using simultaneous matching.  Because of the use of `grind`, none of the proofs needs to be changed.

Cslib/Computability/Automata/NA/Concat.lean
```diff
@@ -0,0 +1,158 @@
...
+  let n := Nat.find hr
+  have h1 (k) (h_k : k < n) : ∃ s1, ss (k) = inl s1 :=
+    isLeft_iff.mp <| not_isRight.mp <| Nat.find_min hr h_k
+  have h2 (k) : ∃ s2, ss (n + k) = inr s2 := by
+    induction k
+    case zero => exact isRight_iff.mp <| Nat.find_spec hr
+    case succ k h_ind =>
+      obtain ⟨s2, _⟩ := h_ind
+      have := concat.eq_1 na1 na2 ▸ hc.right (n + k)
+      grind
```
chenson2018: Please do not use equational theorems like `concat.eq_1` directly. You can write

```suggestion
    case succ k h_ind => grind [concat, hc.trans (n + k)]
```
(Being able to pass an expression like `hc.trans (n + k)` is new in the latest release, where before we would use `have` to add it to the environment.)

Cslib/Computability/Automata/NA/Concat.lean
```diff
@@ -0,0 +1,158 @@
...
+          have := LTS.MTr.stepR na1.toLTS (h_ind (by grind)) h_tr
+          grind
+      obtain ⟨t1, _⟩ : ∃ t1, t1 ∈ na1.accept ∧ na1.MTr (ss1 0 h_0) (xs.take n) t1 := by
+        obtain ⟨t1, h_tr, _⟩ :
+            ∃ t1, na1.Tr (ss1 (n - 1) (by grind)) (xs (n - 1)) t1 ∧ t1 ∈ na1.accept := by
+          have := concat.eq_1 na1 na2 ▸ hc.right (n - 1)
+          grind
+        use t1, by grind
+        have := LTS.MTr.stepR na1.toLTS (h_mtr (n - 1) (by grind)) h_tr
+        grind
```
chenson2018: These nested `obtain` are very hard to follow. For simplicity, I'm just going to push a commit that shows what I'd do here. (As always, still open to discussion.)

Cslib/Computability/Automata/NA/Concat.lean
```diff
@@ -0,0 +1,147 @@
...
+      | inl s1, inr t2 => ∃ t1, na1.Tr s1 x t1 ∧ t1 ∈ na1.accept ∧ t2 ∈ na2.start
+      | inr s2, inr t2 => na2.Tr s2 x t2
+      | inr _, inl _ => False
+    start := inl '' na1.start ∪ if [] ∈ language na1 then inr '' na2.start else ∅
+
+variable {na1 : FinAcc State1 Symbol} {na2 : NA State2 Symbol}
+
+/-- A run of `concat na1 na2` containing at least one `na2` state is the concatenation of
+an accepting finite run of `na1` followed by a run of `na2`. -/
+theorem concat_run_proj {xs : ωSequence Symbol} {ss : ωSequence (State1 ⊕ State2)}
```
chenson2018, line 36: For simplicity, I have pushed a commit that shows what I'd do here. (As always, still open to discussion.)

Here is what I addressed:

- use the API we have defined for `Run` instead of using `.right` and `.left`
- pass expressions directly to `grind` (a helpful new feature of the latest release, where before we would have used `have`)
- unnesting repeated obtain

Cslib/Computability/Automata/NA/Concat.lean
```diff
@@ -0,0 +1,147 @@
...
+the concatenation of the language of `na1` and the ω-language of `na2`. -/
+@[simp]
+theorem concat_language_eq {acc2 : Set State2} :
+    language (Buchi.mk (concat na1 na2) (inr '' acc2)) =
+    language na1 * language (Buchi.mk na2 acc2) := by
+  ext xs
+  constructor
+  · rintro ⟨ss, h_run, h_acc⟩
+    have h_ex2 : ∃ k, (ss k).isRight := by
+      obtain ⟨k, h_k⟩ := Frequently.exists h_acc
```
chenson2018: This `obtain` does not need to be nested here.

Cslib/Foundations/Data/OmegaSequence/Temporal.lean
```diff
@@ -87,4 +87,16 @@ theorem frequently_leadsTo_frequently {p q : Set α}
   have := h1 k0
   grind
 
+theorem drop_frequently_iff_frequently {p : Set α} (n : ℕ) :
+    (∃ᶠ k in atTop, (xs.drop n) k ∈ p) ↔ (∃ᶠ k in atTop, xs k ∈ p) := by
+  simp only [frequently_atTop, get_drop]
+  constructor
+  · intro h m
+    have := h m
+    grind
```
chenson2018:
```suggestion
    grind [h m]
```

Cslib/Foundations/Semantics/LTS/Basic.lean
```diff
@@ -127,6 +127,22 @@ theorem LTS.MTr.nil_eq (h : lts.MTr s1 [] s2) : s1 = s2 := by
...
+  induction h
+  case refl t =>
+    use [t]
+    grind
+  case stepL t1 μ t2 μs t3 h_tr h_mtr h_ind =>
+    obtain ⟨ss', _, _, _, _⟩ := h_ind
+    use [t1] ++ ss'
+    refine ⟨by grind, by grind, by grind, ?_⟩
+    intro k
+    by_cases k = 0 <;> grind
```
chenson2018:
```suggestion
    grind
```

Cslib/Computability/Automata/NA/Concat.lean
```diff
@@ -0,0 +1,147 @@
...
+    · by_cases h_n : n = 0
+      · grind [concat]
+      · obtain ⟨s1, _⟩ := h1 (n - 1) (by grind)
+        grind [concat, hc.trans (n - 1)]
+    · intro k
+      grind [concat, hc.trans (n + k)]
+
+/-- Given an accepting finite run of `na1` and a run of `na2`, there exists a run of
+`concat na1 na2` that is the concatenation of the two runs. -/
+theorem concat_run_exists {xs1 : List Symbol} {xs2 : ωSequence Symbol} {ss2 : ωSequence State2}
```
chenson2018, line 73: I don't understand the complexity of using `grind` as a `simp` discharger and all the extra `have`. This works:
```lean
theorem concat_run_exists {xs1 : List Symbol} {xs2 : ωSequence Symbol} {ss2 : ωSequence State2}
    (h1 : xs1 ∈ language na1) (h2 : na2.Run xs2 ss2) :
    ∃ ss, (concat na1 na2).Run (xs1 ++ω xs2) ss ∧ ss.drop xs1.length = ss2.map inr := by
  by_cases h_xs1 : xs1.length = 0
  · refine ⟨ss2.map inr, ⟨?_, ?_⟩, ?_⟩ <;> grind [concat, List.eq_nil_iff_length_eq_zero]
  · obtain ⟨s0, h_s0, t1, h_t1, h_mtr⟩ := h1
    obtain ⟨ss1, _, _, _, _⟩ := LTS.MTr.exists_states h_mtr
    let ss := (ss1.map inl).take xs1.length ++ω ss2.map inr
    use ss, ⟨?_, ?_⟩, ?_
    · grind [concat, get_append_left]
    · have (k) (h_k : ¬ k < xs1.length) : k + 1 - xs1.length = k - xs1.length + 1 := by grind
      grind [concat, get_append_right', get_append_left]
    · grind [drop_append_of_le_length]
```

chenson2018 (changes requested):
The biggest thing I want to stress here is the idiomatic way of working with a definition like `NA.Run` is to treat it as opaque, for instance not using `refine` to split it into the conjunction it is defined as. This is what is meant by "defeq abuse".

(We could enforce this by changing its reducibility, but I felt that was a little strict to start with.)

Cslib/Computability/Automata/NA/Concat.lean
chenson2018: (the `use` here for instance is defeq abuse, but it wasn't immediately apparent if there was an applicable API lemma)

chenson2018:
I pushed one more little change for readability. (See [this thread](https://leanprover.zulipchat.com/#narrow/channel/113488-general/topic/choose.20and.20autoparams/with/564211379) for an explanation if you are curious how it works)

Cslib/Computability/Automata/NA/Concat.lean
ctchou: @chenson2018  This is interesting.  My old proof was close to the one you showed above.  But then it was broken by the v4.26.0 -> v4.27.0-rc1 move (see: [#CSLib > Excessive grind?](https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/Excessive.20grind.3F/with/563725444)).  That's why I added all the extra `have`s to provide the intermediate results: 276702fc4429024b1598690dd41b962d9f8587b8

ctchou: Fixed now.

Cslib/Computability/Automata/NA/Concat.lean
ctchou: Fixed.

Cslib/Computability/Automata/NA/Concat.lean
ctchou: Fixed.

Cslib/Foundations/Data/OmegaSequence/Temporal.lean
ctchou: Fixed.

Cslib/Foundations/Semantics/LTS/Basic.lean

ctchou:
`git merge` in the current `main` and fix an issue necessitated by cslib#227 which is in `main` now.

PR #220, opened 2025-12-15
feat: tauSTr

PieterCuijpers:
Separating the series of tau steps from the saturation relation to benefit the definition of branching bisimulation later.

Not sure if this counts as a feat: or refactor:, since it is a relatively small change and should not touch anything outside the scope of the file at the moment.

### DIFF
--- a/.github/workflows/lean_action_ci.yml
+++ b/.github/workflows/lean_action_ci.yml
@@ -14,6 +14,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+      - name: Set TEST_ARGS manually
+        run: |
+          echo "TEST_ARGS='--wfail --iofail'" >> $GITHUB_ENV
+        shell: bash
       - uses: leanprover/lean-action@v1
         with:
           build-args: "--wfail --iofail"
--- a/Cslib.lean
+++ b/Cslib.lean
@@ -89,6 +89,7 @@ public import Cslib.Languages.LambdaCalculus.LocallyNameless.Fsub.Typing
 public import Cslib.Languages.LambdaCalculus.LocallyNameless.Fsub.WellFormed
 public import Cslib.Languages.LambdaCalculus.LocallyNameless.Stlc.Basic
 public import Cslib.Languages.LambdaCalculus.LocallyNameless.Stlc.Safety
+public import Cslib.Languages.LambdaCalculus.LocallyNameless.Stlc.StrongNorm
 public import Cslib.Languages.LambdaCalculus.LocallyNameless.Untyped.Basic
 public import Cslib.Languages.LambdaCalculus.LocallyNameless.Untyped.FullBeta
 public import Cslib.Languages.LambdaCalculus.LocallyNameless.Untyped.FullBetaConfluence
@@ -98,6 +99,7 @@ public import Cslib.Languages.LambdaCalculus.LocallyNameless.Untyped.LcAt
 public import Cslib.Languages.LambdaCalculus.LocallyNameless.Untyped.MultiApp
 public import Cslib.Languages.LambdaCalculus.LocallyNameless.Untyped.MultiSubst
 public import Cslib.Languages.LambdaCalculus.LocallyNameless.Untyped.Properties
+public import Cslib.Languages.LambdaCalculus.LocallyNameless.Untyped.StrongNorm
 public import Cslib.Languages.LambdaCalculus.Named.Untyped.Basic
 public import Cslib.Logics.HML.Basic
 public import Cslib.Logics.HML.LogicalEquivalence
--- a/Cslib/Computability/Automata/NA/Concat.lean
+++ b/Cslib/Computability/Automata/NA/Concat.lean
@@ -99,7 +99,7 @@ theorem concat_run_exists {xs1 : List Symbol} {xs2 : ωSequence Symbol} {ss2 : �
   · obtain ⟨rfl⟩ : xs1 = [] := List.eq_nil_iff_length_eq_zero.mpr h_xs1
     refine ⟨ss2.map inr, by simp only [concat]; grind [Run, LTS.ωTr], by simp⟩
   · obtain ⟨s0, _, _, _, h_mtr⟩ := h1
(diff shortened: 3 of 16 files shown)


PieterCuijpers:
I've updated the file to use #231
Still in doubt a bit on how to use @[grind] correctly.
When do you use "scoped" and which modifier (_=_, =, ->, .) should I use when?

Also, the naming of theorems is a bit confusing wrt the STr relation.
Do I name a theorem foo_sTr or foo_STr?
I'm personally inclided towards the latter, and tried to use it consistently, but since theorems generally prefer small letters I'm not sure if that is correct.

chenson2018:
> When do you use "scoped"

In CSLib we typically use `scoped` for almost all `grind` rules. Note that if this would put something in too deep of a namespace with a regular annotation, you can add it afterwards like `attribute [grind] foo`. (Soon there will be "`grind` sets" that make this easier)

> which modifier (_=_, =, ->, .) should I use when?
 
See [this page](https://lean-lang.org/doc/reference/latest/The--grind--tactic/E___matching/#e-matching) for a full description of patterns. 

> Also, the naming of theorems is a bit confusing wrt the STr relation. Do I name a theorem foo_sTr or foo_STr? I'm personally inclided towards the latter, and tried to use it consistently, but since theorems generally prefer small letters I'm not sure if that is correct.

We try to follow the [Mathlib naming conventions](https://leanprover-community.github.io/contribute/naming.html) but I will readily admit these don't come naturally to me for these oddly cased names. I think the relevant rule is:

> 5. When something named with UpperCamelCase is part of something named with snake_case, it is referenced in lowerCamelCase.

PieterCuijpers:
> See [this page](https://lean-lang.org/doc/reference/latest/The--grind--tactic/E___matching/#e-matching) for a full description of patterns.

Hm... maybe it is better if I leave out @[grind] altogether for the time being?
I have some incling of the choices I should make, but I also realize that the wrong choice might mess things up thoroughly.

What is better for this PR, just leaving it out, or making an educated guess and hoping you will catch it if its wrong?

fmontesi:
You can test what grind is doing by using the `?` modifier, which will tell you the grind pattern being created. So you can try `@[scoped grind? =>]`, `@[scoped grind .]`, etc. and check what you prefer.

fmontesi:
What's the status of this? I see some conflicts and I'm afraid of messing it up. :)

PieterCuijpers:
I got into trouble making the changes, because it turned out I overlooked one of the bisimulation theorems when I thought the "metric" version was not being used anymore. And after that, I ran out of time and got swamped in teaching preparations. So it is indeed a bit of a mess now.

If you have ideas on how to clean it up and have time, please feel free to do so!

Next week Thursday, I hope to get four students who want to work with Lean4.
I have asked them to introduce themselves on Zulip and find a theorem they would like to work on.
I would like to see if we can get them to make useful contributions to the library.
So if you have suggestions for good (beginner) contributions, please suggest those to them.
(Since I'm also running my masters course in that same period, my own contributions will likely be limited to monitoring and reviewing the work of those students.)

PieterCuijpers:
Fixed it! I hope... (Awaiting automatic checks to finish.)

I have now removed everything related to "induction over N tau-steps" and replaced this with a straightforward induction over tauSTr.

PieterCuijpers:
It seems that I'm dependent on maintainers to run automatic workflows to check my code.
Is that on purpose, or is there something I can do to fix that?
If not, what can I do to verify commits on my own machine, apart from running "lake build"?

chenson2018:
> It seems that I'm dependent on maintainers to run automatic workflows to check my code. Is that on purpose, or is there something I can do to fix that?

This is a security measure that GitHub implements. Once your first PR is merged, this requirement is removed. I would do so now, but I'm not sure if/how this can be done ahead of time.

> If not, what can I do to verify commits on my own machine, apart from running "lake build"?

[CONTRIBUTING.md](https://github.com/leanprover/cslib?tab=contributing-ov-file#continuous-integration) is where I have documented how to run the checks CI does locally.

The failure you are seeing is from `lake test`. Specifically these `grind` lints only slightly over the threshold I think can be registered as exceptions for the time being. (The annotations for `LTS` need some work in general)

PieterCuijpers:
> [CONTRIBUTING.md](https://github.com/leanprover/cslib?tab=contributing-ov-file#continuous-integration) is where I have documented how to run the checks CI does locally.
> 
> The failure you are seeing is from `lake test`. Specifically these `grind` lints only slightly over the threshold I think can be registered as exceptions for the time being. (The annotations for `LTS` need some work in general)

Thanks, that helps.
Except that I do not know how to register something as an exception... or where to find that?

chenson2018:
Yes, I should document this... you add a `#grind_lint skip ...` line in `CslibTests/GrindLint.lean`.

PieterCuijpers:
Thanks. For now, I managed to get all checks to pass by iterating on the test output.
I'm not sure if this is now already ready enough for the PR to go ahead, or whether I should add more @[grind] annotations.
Please let me know.

@fmontesi I have now removed all of the no-longer-needed theorems regarding STrN. I hope the changes are now a bit more readable for you.

fmontesi:
Looks pretty good to me now, thanks!

Last doubt: do I understand correctly that some 'scoped grind' are becoming global 'grind' now? Perhaps it's just an effect of merging with up-to-date main. Or is it intentional?

chenson2018:
> Last doubt: do I understand correctly that some 'scoped grind' are becoming global 'grind' now? Perhaps it's just an effect of merging with up-to-date main. Or is it intentional?

Yes, please keep `grind` scoped, especially for `LTS` which can be problematic for performance at the moment.

PieterCuijpers:
@fmontesi I think I may have messed up the commit history.
Last version I was happy with was the "added two necessary grinds" one.
But those grinds needed to be made local.
I did that, but when I updated today a lot got messed up.
Not sure what happened, but I think I accidentally deleted some things you may have added?

I need help here... too much of a git newby to fix this...

PieterCuijpers:
I've restarted this PR as #444 because I've lost an overview on which things have changed in main in the meantime.
I thought it was better to simply start a PR from scratch, since my attempts to fix the .git were... amateurish.

PR #221, opened 2025-12-15
feat: improve weekly lint report to use table format

jessealama:
Update the weekly lint report script to match Mathlib's format, grouping warnings/errors by description and showing counts in a table inside Zulip spoiler blocks.

This addresses [feedback on Zulip](https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/Weekly.20linting.20log/near/563737643) to use a more standard format for the report.

Example output:

```
CSLib weekly lint run [completed](...) ([d3d8a37](...)).
 Warnings: 3

```spoiler Warning counts
|    | Warning description                        |
|----|--------------------------------------------|
|  2 | 'ext; grind' can be replaced with 'grind'  |
|  1 | 'simp; grind' can be replaced with 'grind' |
```
```

### DIFF
--- a/scripts/weekly_lint_report.sh
+++ b/scripts/weekly_lint_report.sh
@@ -87,4 +87,4 @@ else
     fi
 fi
 
-echo "${delimiter}" >> "$GITHUB_OUTPUT"
+echo "${delimiter}"

### CODE REVIEW

**Comment by chenson2018** [comment]
I'm sorry, I think I misunderstood the situation and thought it was just a bug a table wasn't showing. I'm not particularly keen to adding a Python dependency. Mathlib uses [this shell script](https://github.com/leanprover-community/mathlib4/blob/master/scripts/zulip_build_report.sh), could we just do the same? Thanks!

**Comment by jessealama** [comment]
> I'm sorry, I think I misunderstood the situation and thought it was just a bug a table wasn't showing. I'm not particularly keen to adding a Python dependency. Mathlib uses [this shell script](https://github.com/leanprover-community/mathlib4/blob/master/scripts/zulip_build_report.sh?rgh-link-date=2025-12-15T17%3A52%3A01Z), could we just do the same? Thanks!

Sounds good. I've gone back to the bash script approach.

**Comment by chenson2018** [comment]
I tried pushing your branch to the `leanprover/cslib` repo and [manually triggering a run](https://github.com/leanprover/cslib/actions/runs/20291984051). The table still appears to be missing. I have not closely inspected your bash script, but I notice the [Mathlib version](https://github.com/leanprover-community/mathlib4/blob/master/scripts/zulip_build_report.sh) ends with 
```bash
echo "${delimiter}" >> "$GITHUB_OUTPUT"
```
while yours just has
```bash
echo "${delimiter}"
```
maybe this is related?

jessealama:
> I tried pushing your branch to the `leanprover/cslib` repo and [manually triggering a run](https://github.com/leanprover/cslib/actions/runs/20291984051). The table still appears to be missing. I have not closely inspected your bash script, but I notice the [Mathlib version](https://github.com/leanprover-community/mathlib4/blob/master/scripts/zulip_build_report.sh?rgh-link-date=2025-12-17T05%3A15%3A04Z) ends with
> 
> ```shell
> echo "${delimiter}" >> "$GITHUB_OUTPUT"
> ```
> 
> while yours just has
> 
> ```shell
> echo "${delimiter}"
> ```
> 
> maybe this is related?

Ah, yes, of course. I've pushed a fix. If you like, you can give this another shot.

chenson2018:
This is still not working. Are you sure `GITHUB_OUTPUT` is only needed in the workflow, and not the shell script like Mathlib? It might be easier if it you start with exactly what Mathlib has and make the minimal changes for CSLib.

jessealama:
Thanks for testing this! I investigated the missing table and found the cause.

Looking at the [GitHub Actions run logs](https://github.com/leanprover/cslib/actions/runs/20298052544), the workflow checked out `main` (commit `61c9ae2`), not the PR branch:

```
[command]/usr/bin/git checkout --progress --force -B main refs/remotes/origin/main
...
[command]/usr/bin/git log -1 --format=%H
61c9ae2f5df480f935fa7d91a797ed583704fa95
```

This is because the workflow has `ref: main` in its checkout step. So the script that ran was main's version (which doesn't have the table), while `${{ github.sha }}` in the Zulip message came from the trigger context (our branch).

I tested the PR's script locally and it produces the expected output with the spoiler table. Once merged, the table should appear in subsequent runs.

chenson2018:
Oh, I see, sorry for the confusion! I'm going to just merge and try one more time.

PR #224, opened 2025-12-15
test: iofail and merge_group trigger

chenson2018:
This PR makes two small changes to CI:
- Adds a `merge_group` trigger 
- Passes the `--iofail` option to `build-args`

The motivation of having the `merge_group` trigger is to ensure that there is a final build that happens before a PR is merged into `main`, preventing situations where a release happens while a PR is pending and then only failing after being merged. 

For this to take effect, I need someone with adequate permissions to require merge queues for the `main` branch. These should still be configured as squash commits. I would be fine setting the merge limit low (maybe even 1) for simplicity since we don't yet have high enough volume of commits for this to matter.

The `--iofail` option is straightforward, mirroring what Mathlib already does.


chenson2018:
I think changing an action works, it'll run testing for the merge queue on the next PR merged.

I also am not sure why it was waiting for Kim's approval to automerge. Maybe because you are not explicitly a codeowner of the workflows directory.

PR #227, opened 2025-12-16
chore: restore closure properties of regular languages under boolean operations

ctchou:
In view of the plan to develop a self-contained automata theory in cslib, the patch restores the closure properties of regular languages under boolean operations that were removed in https://github.com/leanprover/cslib/pull/209.

Note that in order to avoid conflicts with the equivalent theorems in mathlib, the `_cslib` suffix is added to theorem names.

### DIFF
--- a/Cslib/Computability/Languages/OmegaRegularLanguage.lean
+++ b/Cslib/Computability/Languages/OmegaRegularLanguage.lean
@@ -68,7 +68,7 @@ theorem IsRegular.not_da_buchi :
 @[simp]
 theorem IsRegular.regular_omegaLim {l : Language Symbol}
     (h : l.IsRegular) : (l↗ω).IsRegular := by
-  obtain ⟨State, _, ⟨da, acc⟩, rfl⟩ := Language.IsRegular.iff_cslib_dfa.mp h
+  obtain ⟨State, _, ⟨da, acc⟩, rfl⟩ := Language.IsRegular.iff_dfa.mp h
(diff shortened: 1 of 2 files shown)


chenson2018:
Can you clarify why you need the `_cslib` suffix? I expected any naming conflicts to be handled by being under the `Cslib` namespace.

ctchou:
This is because it seems to me a good idea to put all theorems about regular languages in the namespace `Language.IsRegular`.

chenson2018:
I agree but think we should find some middle ground, because having a namespace will probably make this easier to track than a suffix. What about `Language.IsRegular.Cslib` or `Cslib.Language.IsRegular`?

ctchou:
One alternative is to put all theorems we prove about regular languages in cslib in the namespace `Cslib.Language.IsRegular`.  Then if one day mathlib's automata theory is replaced by cslib's, we just have to strip away the topmost namespace.

ctchou:
In the last commit I put everything in `RegularLanguage.lean` in the `Cslib.Language` namespace and removed `_cslib` from all theorem names everywhere.

Lean's namespace mechanism works really well.  Now outside the `Cslib` namespace,  both `Language.IsRegular` and `Language.IsRegular.compl` refer to mathlib's definition and theorem.  But inside the `Cslib` namespace, the former still refers to mathlib's definition but the latter refers to cslib's equivalent theorem, which is exactly the behavior we want.

fmontesi:
Thanks for converging on a good namespacing usage, that's very nice indeed.

Issue #202, opened 2025-12-06
shellcheck warnings in some bash scripts


jessealama:
Running `shellcheck` on `scripts/create-adaptation-pr.sh` reveals several warnings:

| Code | Count | Description |
|------|-------|-------------|
| SC2086 | 12 | Unquoted variables can cause globbing/word splitting |
| SC2006 | 3 | Legacy backtick syntax (should use `$(...)`) |
| SC2162 | 4 | `read` without `-r` can mangle backslashes |

All of these are straightforward to fix. Happy to submit a PR if that would be helpful.

jessealama:
Would it be worth adding a CI check that runs `shellcheck` on PRs that add or modify bash scripts?

fmontesi:
That sounds like a good idea, yes. :)

jessealama:
> That sounds like a good idea, yes. :)

Sounds good: #229

PR #229, opened 2025-12-18
feat(CI): add shellcheck workflow for shell scripts

jessealama:
Adds a CI workflow that runs shellcheck on shell scripts in `scripts/`.

Related to #202.

### DIFF
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -17,6 +17,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          sparse-checkout: |
+            scripts
 
       - name: Run shellcheck
         run: find scripts -name '*.sh' -type f -print0 | xargs -0 shellcheck --severity=warning


arademaker:
It seems good to  me, can we accept it?

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,22 @@
...
+
+jobs:
+  shellcheck:
+    name: Check shell scripts
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Run shellcheck
+        run: find scripts -name '*.sh' -type f -print0 | xargs -0 shellcheck --severity=warning
```
chenson2018, line 25: Does this just leave a warning? Assuming the test is reliable, I usually prefer to use `set -e` to escalate to an error, otherwise it is too easy to ignore.

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,22 @@
...
+    paths:
+      - 'scripts/**/*.sh'
+  workflow_dispatch:
+
+jobs:
+  shellcheck:
+    name: Check shell scripts
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
```
chenson2018, line 19: You could use `sparse-checkout` here to just have the `scripts` directory.

.github/workflows/shellcheck.yml
jessealama: shellcheck will exit uncleanly if it detects any problems, so I think adding `set -e` wouldn't make any difference in this case.

.github/workflows/shellcheck.yml
jessealama: Good catch (I didn't know about `sparse-checkout`). I've added that, thanks!

PR #230, opened 2025-12-18
feat: Newman's lemma, commutation

chenson2018:
This adds Newman's lemma (terminating local confluence implies confluence) and commuting relations.


Cslib/Foundations/Data/Relation.lean
```diff
@@ -140,11 +141,138 @@ def trans_of_subrelation_right (s r : α → α → Prop) (hr : Transitive r)
...
+/-- Confluence implies that multi-step joinability is an equivalence. -/
+theorem Confluent.equivalence_join_reflTransGen (h : Confluent r) :
     Equivalence (Join (ReflTransGen r)) := by
-  apply Relation.equivalence_join reflexive_reflTransGen transitive_reflTransGen
-  intro a b c hab hac
-  exact h.toConfluent hab hac
+  grind [equivalence_join, reflexive_reflTransGen, transitive_reflTransGen]
+
+/-- A relation is terminating when the inverse of its transitive closure is well-founded. -/
+abbrev Terminating (r : α → α → Prop) := WellFounded (fun a b => TransGen r b a)
```
fmontesi: Isn't there a definition of inverse somewhere?

Cslib/Foundations/Data/Relation.lean
```diff
@@ -140,11 +141,138 @@ def trans_of_subrelation_right (s r : α → α → Prop) (hr : Transitive r)
...
+
+/-- A relation is locally confluent when all reductions with a common origin are multi-joinable -/
+abbrev LocallyConfluent (r : α → α → Prop) :=
```
fmontesi, line 181: It appears the file is inconsistent in the casing of variables for a relation (A, B, C here, some other times a, b, c).

fmontesi: (Since we're consistently using lowercase r for relations, I suggest we use lowercase letters for the elements as well.)

Cslib/Foundations/Data/Relation.lean
```diff
@@ -140,11 +141,138 @@ def trans_of_subrelation_right (s r : α → α → Prop) (hr : Transitive r)
...
+  ∀ {A B C : α}, r A B → r A C → Join (ReflTransGen r) B C
+
+theorem Confluent.toLocallyConfluent (h : Confluent r) : LocallyConfluent r := by
+  intro _ _ _ AB AC
+  exact h (.single AB) (.single AC)
+
+/-- Newman's lemma: a terminating, locally confluent relation is confluent. -/
+theorem LocallyConfluent.Terminating_toConfluent (h : LocallyConfluent r) (wf : Terminating r) :
```
fmontesi: `wf` is a new form of 'defeq' abuse: in variable naming. hehe. :-) I'd just call `h` `hlc` and `wf` `ht`.

ctchou:
I have to say that the convention of using lowercase letters for relations but uppercase letters for the objects being related looks very strange to me.  This is not a common convention and, in particular, is not the one used in `Mathlib.Logic.Relation`.  I think lowercase letters should be used throughout.

I also have a pet peeve: I really don't think subscripts \1, \2, etc actually help readability.  Depending on the font type and size one happens to be using, it can be hard to tell them apart (an example is shown below).  Full-size digits 1, 2, etc are much clearer.
image

Cslib/Foundations/Data/Relation.lean
```diff
@@ -31,7 +32,7 @@ attribute [scoped grind →] ReflGen.to_eqvGen TransGen.to_eqvGen ReflTransGen.t
 def UpTo (r s : α → α → Prop) : α → α → Prop := Comp s (Comp r s)
 
 /-- A relation has the diamond property when all reductions with a common origin are joinable -/
-abbrev Diamond (r : α → α → Prop) := ∀ {A B C : α}, r A B → r A C → Join r B C
+abbrev Diamond (r : α → α → Prop) := ∀ {a b c : α}, r a b → r a c → Join r b c
```
ctchou: Why is `abbrev` rather than `def` used here?  Similar questions for other uses of `abbrev` below.

chenson2018:
> I have to say that the convention of using lowercase letters for relations but uppercase letters for the objects being related looks very strange to me. This is not a common convention and, in particular, is not the one used in `Mathlib.Logic.Relation`. I think lowercase letters should be used throughout.

This was just inconsistency on my part, I have already pushed a change fixing this per Fabrizio's comment.

> I also have a pet peeve: I really don't think subscripts \1, \2, etc actually help readability. Depending on the font type and size one happens to be using, it can be hard to tell them apart (an example is shown below). Full-size digits 1, 2, etc are much clearer. 

I have a pet peeve in the opposite direction. I have refrained from nitpicking in reviews because I felt it wasn't the most pressing issue, but using subscripts is standard per the Mathlib naming conventions.

ctchou:
Mathlib is inconsistent.  For example, `Mathlib.Logic.Relation` does not use subscripts \1, \2, etc.

Cslib/Foundations/Data/Relation.lean
chenson2018: I considered `Function.swap` or `flip`, but decided against this. `Function.swap` would make the most sense as this already has
```lean4
Relation.transGen_swap : Relation.TransGen (Function.swap r) a b ↔ Relation.TransGen r b a
```
and is an `abbrev`, but I felt this was an unnecessary obfuscation for something that will appear in every induction hypothesis.

chenson2018:
> Mathlib is inconsistent. For example, `Mathlib.Logic.Relation` does not use subscripts \1, \2, etc.

I don't think this is true, you can grep and see the usage of subscripts. It is consistent in not using plain digits such as `h1`, `h2`, etc. Either a more descriptive name such as `hab : r a b` or subscripts are used. 

Again I am not keen to strictly enforce all nuances of Mathlib naming conventions at this point in CSLib's development (and the more complex theorem naming rules *are* much more inconsistent) but this point seems rather clear to me.

ctchou:
I think I was confused by the web listing:
https://leanprover-community.github.io/mathlib4_docs/Mathlib/Logic/Relation.html
which generates full-sized digits 1 and 2 for the anonymous dots `·` in the Lean source file.

chenson2018:
I have changed the definition of termination to the equivalent
```lean
abbrev Terminating (r : α → α → Prop) := WellFounded (fun a b => r b a)
```
along with proofs that allow you to easily switch between the corresponding induction principles. I still prefer to leave `Function.swap` out of the definition for readability, but could be convinced if you feel strongly.

Interestingly the proof of Newman's lemma did not change at all, because the way I structured the proofs with `grind` works in either case!

chenson2018:
Just a friendly reminder: before merging this, could we try the "status checks" setting Sebastian mentioned to hopefully get the merge queue testing working?

fmontesi:
Could you merge from main? It's gonna test if the merge queue with status checks works now. :)

PR #232, opened 2025-12-20
feat: add LTS.omegaTr for infinite executions

fmontesi:
Introduces infinite executions on LTSs as LTS.ωTr and refactors automata theory accordingly.

### DIFF
--- a/Cslib/Foundations/Semantics/LTS/Basic.lean
+++ b/Cslib/Foundations/Semantics/LTS/Basic.lean
@@ -6,7 +6,6 @@ Authors: Fabrizio Montesi
 
 import Cslib.Init
 import Mathlib.Data.Set.Finite.Basic
-import Mathlib.Data.Stream.Defs
 import Mathlib.Order.ConditionallyCompleteLattice.Basic
 import Cslib.Foundations.Data.OmegaSequence.Init
 


Cslib/Foundations/Semantics/LTS/Basic.lean
```diff
@@ -160,6 +164,36 @@ def LTS.generatedBy (s : State) : LTS {s' : State // lts.CanReach s s'} Label wh
...
+
+/-! ## Infinite sequences of transitions
+
+This section treats infinite executions as ω-sequences of transitions.
+-/
+
+/-- Definition of an infinite execution, or ω-sequence of transitions. -/
+@[scoped grind]
+def LTS.ωTr (lts : LTS State Label) (ss : ωSequence State) (μs : ωSequence Label) :
+    Prop := ∀ n, lts.Tr (ss n) (μs n) (ss (n + 1))
```
ctchou: Stylistic nitpick: it looks very strange to see `Prop := ` appears on a separate line from the `def`.  For a moment I thought you are defining a `structure` with a field named `Prop`.

Cslib/Computability/Automata/NA/Basic.lean
```diff
@@ -41,27 +41,9 @@ namespace NA
...
-  run.left
-
-@[scoped grind =>]
-lemma Run.trans (run : Run na xs ss) : ∀ n, na.Tr (ss n) (xs n) (ss (n + 1)) :=
-  run.right
-
-end NARunGrind
+structure Run (na : NA State Symbol) (xs : ωSequence Symbol) (ss : ωSequence State) where
+  start : ss 0 ∈ na.start
+  trans : na.ωTr ss xs
```
ctchou, line 46: I would like @chenson2018 to take a look at this change in the `grind` annotations.  Previously we had to `open scoped Run` in various files, but we never had to explicitly supply `Run` to `grind`.    Should we restore that behavior and (if yes) how?

ctchou:
In the following commit:
https://github.com/ctchou/cslib/commit/b3562f3d4cdf65017aac856b22995482e22b433c
I added/restored `open scoped LTS` in several files to eliminate the need for explicitly supplying `LTS.ωTr` to `grind`.  The only file where I could not make this work is `NA/Concat.lean`, where restoring `open scoped LTS` causes some instances of `grind` to time-out.

Cslib/Computability/Automata/NA/Basic.lean
```diff
@@ -41,27 +41,9 @@ namespace NA
...
-@[scoped grind →]
-lemma Run.start (run : Run na xs ss) : ss 0 ∈ na.start :=
-  run.left
-
-@[scoped grind =>]
-lemma Run.trans (run : Run na xs ss) : ∀ n, na.Tr (ss n) (xs n) (ss (n + 1)) :=
-  run.right
-
-end NARunGrind
+structure Run (na : NA State Symbol) (xs : ωSequence Symbol) (ss : ωSequence State) where
```
chenson2018, line 44: Using structures instead of `Prop`-valued `def`s is probably good practice in general, but this needs some work to recover the `grind` behavior we had previously.

Cslib/Foundations/Semantics/LTS/Basic.lean
```diff
@@ -160,6 +164,36 @@ def LTS.generatedBy (s : State) : LTS {s' : State // lts.CanReach s s'} Label wh
...
+section ωMultiStep
+
+/-! ## Infinite sequences of transitions
+
+This section treats infinite executions as ω-sequences of transitions.
+-/
+
+/-- Definition of an infinite execution, or ω-sequence of transitions. -/
+@[scoped grind]
+def LTS.ωTr (lts : LTS State Label) (ss : ωSequence State) (μs : ωSequence Label) :
```
chenson2018, line 235: If we prefer structures for defeq/`grind` reasons, we should do the same here.

Cslib/Foundations/Semantics/LTS/Basic.lean
```diff
@@ -160,6 +164,36 @@ def LTS.generatedBy (s : State) : LTS {s' : State // lts.CanReach s s'} Label wh
...
+  case pos => grind
+  case neg =>
+    cases m
+    case zero => grind
+    case succ m =>
+      rw [show μs.extract n (m + 1) = (μs.extract n m) ++ [μs m] by grind]
+      apply LTS.MTr.comp (s2 := ss m)
+      · apply LTS.ωTr_mTr h
+        grind
+      · grind
```
chenson2018: Using a proof recursively like this is a bit problematic for `grind`, but you can at least do: 
```suggestion
    case succ m =>
      have : lts.MTr (ss n) (μs.extract n m) (ss m) := ωTr_mTr (hnm := by grind) h
      grind [MTr.comp]
```

fmontesi: Very nice!

Cslib/Foundations/Semantics/LTS/Basic.lean
fmontesi: Don't we achieve the same with a normal def that is not an abbrev?

chenson2018: There are still differences between a `def` (with default reducibility) and a one-field structure with the same type.

ctchou:
In the following commit:
https://github.com/ctchou/cslib/commit/58429cb04c499d8e33617f06b040054cd75147d3
I broke up the proof of `NA.concat_run_proj` into four lemmas which I think are easier to understand and handle.  I also touched up the proof of `NA.concat_run_exists` a little bit.

fmontesi:
> In the following commit: [ctchou@b3562f3](https://github.com/ctchou/cslib/commit/b3562f3d4cdf65017aac856b22995482e22b433c) I added/restored `open scoped LTS` in several files to eliminate the need for explicitly supplying `LTS.ωTr` to `grind`. The only file where I could not make this work is `NA/Concat.lean`, where restoring `open scoped LTS` causes some instances of `grind` to time-out.

Thanks for that. I'd tried that option as well, but then got stuck exactly in Concat.lean as you describe. Let's go with what you have.

Cslib/Foundations/Semantics/LTS/Basic.lean
fmontesi: It bugs me as well, but the problem is that it's exactly `Prop` that sits at the char limit:

Screenshot_20251222_091920

I could either

```
def LTS.ωTr (lts : LTS State Label) (ss : ωSequence State) (μs : ωSequence Label) :=
  ∀ n, lts.Tr (ss n) (μs n) (ss (n + 1))
```

or

```
def LTS.ωTr (lts : LTS State Label)
  (ss : ωSequence State) (μs : ωSequence Label) : Prop := ∀ n, lts.Tr (ss n) (μs n) (ss (n + 1))
```

Cslib/Foundations/Semantics/LTS/Basic.lean
fmontesi: What are the relevant differences that make you prefer a structure? This sounds valuable to discuss as a general point.

Cslib/Computability/Automata/NA/Basic.lean
fmontesi: Adding a grind annotation to the `Run` structure seems to do the trick:
```
@[scoped grind]
structure Run ...
```

which essentially adds a grind pattern for Run.mk (which seems to be the logical one to me).

I didn't include it because it made the last line of BuchiEquiv timeout.

chenson2018:
There is a shake failure. Please also add a description now that this is no longer a draft PR.

Cslib/Computability/Automata/NA/Basic.lean
chenson2018: If you want to keep the same behavior, you can add
```
namespace Run
attribute [scoped grind <=] Run.mk
attribute [scoped grind →] Run.start
attribute [scoped grind =>] Run.trans
end Run
```
and should be able to remove all the manual specification of `Run` without any timeout issues.

Cslib/Foundations/Semantics/LTS/Basic.lean
chenson2018: The most straightforward difference is that `a : α` and the structure `{ a : α }` are not definitionally equal. Sometimes one-field structures are used for this reason. You can also change the reducibility of a `def` however, and it is not completely clear to me if one is preferable for `grind`.

fmontesi:
Since we seem to agree it's good enough and on Zulip it appears to be a blocker for some automata development, I'm merging this now. Let's keep the discussion up on design in the future.

PR #236, opened 2025-12-24
feat: prove that the omega-power of a regular language is an omega-regular language

ctchou:
This patch proves that the omega-power of a regular language is an omega-regular language, using a looping construction `NA.loop`.

The same construction should also be applicable to proving that regular languages are closed under Kleene-star, but that is left to future work.

### DIFF
--- a/CslibTests/FreeMonad.lean
+++ b/CslibTests/FreeMonad.lean
@@ -43,7 +43,7 @@ clean separation between syntax and semantics, enabling multiple interpretations
 
 ## References
 
-Based on "[Tutorial: A Verified Interpreter with Side Effects](https://tannerduve.github.io/blog/freer-monad/part4/)" demonstrating practical
+Based on "[Tutorial: A Verified Interpreter with Side Effects](https://tannerduve.github.io/blog/2025/freer-monad-part4/)" demonstrating practical
 applications of the mathematical theory developed in the core Free monad files.
 -/
 


Cslib/Computability/Automata/NA/Loop.lean
```diff
@@ -0,0 +1,163 @@
...
+
+lemma loop_run_left_left {xs : ωSequence Symbol} {ss : ωSequence (Unit ⊕ State)}
+    (h : (loop na).Run xs ss) (h1 : (ss 1).isLeft) : [xs 0] ∈ language na := by
+  have h_init := h.start
+  simp only [loop, Set.mem_singleton_iff] at h_init
+  have h_step := h.trans 0
+  obtain ⟨t, h_t⟩ := isLeft_iff.mp h1
+  simp only [loop, h_init, h_t] at h_step
+  obtain ⟨s, t, _⟩ := h_step
+  use s, by grind, t, by grind, by grind
```
chenson2018: With interleaved data and proofs, I think `refine` is clearer:

```suggestion
  refine ⟨s, ?_, t, ?_⟩ <;> grind
```

Cslib/Computability/Automata/NA/Loop.lean
```diff
@@ -0,0 +1,163 @@
...
+open Nat List Sum ωSequence Acceptor Language
+open scoped Run LTS
+
+variable {Symbol State : Type*}
+
+/-- `loop na` mimics `na`, but can nondeterministically decide to "loop back" by identifing
+an accepting state of `na` with a starting state of `na`.  This identification is achieved
+via a new dummy state `()`, which is the sole starting state and the sole accepting state
+of `loop na`. -/
+def loop (na : FinAcc State Symbol) : NA (Unit ⊕ State) Symbol where
```
chenson2018: Maybe this should be under the `FinAcc` namespace? The parameter being named `na` is slightly confusing too.

Cslib/Computability/Automata/NA/Loop.lean
```diff
@@ -0,0 +1,163 @@
...
+lemma loop_run_left_right_left {xs : ωSequence Symbol} {ss : ωSequence (Unit ⊕ State)}
+    (h : (loop na).Run xs ss) (n : ℕ) (h1 : 0 < n ∧ (ss n).isLeft)
+    (h2 : ∀ k, 0 < k → k < n → (ss k).isRight) : xs.take n ∈ language na := by
+  by_cases n = 1
+  · grind [loop_run_left_left]
+  · obtain ⟨s, t, h_mtr, _⟩ := loop_run_left_right h (n - 1) (by grind) (by grind)
+    obtain ⟨t', h_tr, _⟩ : ∃ t', na.Tr t (xs (n - 1)) t' ∧ t' ∈ na.accept := by
+      grind [loop, h.trans (n - 1)]
+    use s, by grind, t', by grind
+    grind [LTS.MTr.stepR na.toLTS h_mtr h_tr]
```
chenson2018: Same thing here:

```suggestion
    refine ⟨s, ?_, t', ?_⟩ <;> grind [LTS.MTr.stepR na.toLTS h_mtr h_tr]
```

Cslib/Computability/Automata/NA/Loop.lean
```diff
@@ -0,0 +1,163 @@
...
+theorem loop_run_one_iter {xs : ωSequence Symbol} {ss : ωSequence (Unit ⊕ State)}
+    (h : (loop na).Run xs ss) (h1 : ∃ k, 0 < k ∧ (ss k).isLeft) :
+    ∃ n, xs.take n ∈ language na - 1 ∧ (loop na).Run (xs.drop n) (ss.drop n) := by
+  let n := Nat.find h1
+  have : 0 < n ∧ (ss n).isLeft := Nat.find_spec h1
+  have : ∀ k, 0 < k → k < n → (ss k).isRight := by grind [Nat.find_min h1]
+  refine ⟨n, ⟨?_, ?_⟩, ?_⟩
+  · grind [loop_run_left_right_left]
+  · rw [mem_one, eq_nil_iff_length_eq_zero]
+    grind [Nat.find_spec h1]
```
chenson2018: It feels odd that `grind` doesn't get this one without manually rewriting. This works:

```suggestion
  · have neq : (ωSequence.take n xs).length ≠ 0 := by grind
    exact neq.imp (congrArg length)
```
but still is a bit manual. Not a big deal.

Cslib/Computability/Languages/OmegaRegularLanguage.lean
```diff
@@ -177,6 +178,17 @@ theorem IsRegular.hmul {l : Language Symbol} {p : ωLanguage Symbol}
...
+/-- The ω-power of a regular language is an ω-regular language. -/
+@[simp]
+theorem IsRegular.omegaPow [Inhabited Symbol] {l : Language Symbol}
+    (h : l.IsRegular) : (l^ω).IsRegular := by
+  obtain ⟨State, h_fin, na, rfl⟩ := Language.IsRegular.iff_nfa.mp h
+  let State' := Unit ⊕ State
+  let na' := NA.loop na
+  let acc' : Set State' := {inl ()}
+  use State', inferInstance, ⟨na', acc'⟩
+  rw [NA.Buchi.loop_language_eq]
```
chenson2018: I think this is clearer without the `let` and preferring `exact` to `rw`:

```suggestion
  use Unit ⊕ State, inferInstance, ⟨NA.loop na, {inl ()}⟩
  exact NA.Buchi.loop_language_eq
```

chenson2018 (changes requested):

Cslib/Computability/Languages/OmegaRegularLanguage.lean
chenson2018: Actually you could also make this a single `exact` like:
```lean
obtain ⟨State, h_fin, na, rfl⟩ := Language.IsRegular.iff_nfa.mp h
exact ⟨Unit ⊕ State, inferInstance, ⟨NA.loop na, {inl ()}⟩, NA.Buchi.loop_language_eq⟩
```

Cslib/Computability/Automata/NA/Loop.lean
ctchou: Fixed.

Cslib/Computability/Automata/NA/Loop.lean
ctchou: Fixed.

Cslib/Computability/Automata/NA/Loop.lean
ctchou: Fixed.

Cslib/Computability/Languages/OmegaRegularLanguage.lean
ctchou: I'll use the one with `use`.  The one-liner is a bit too cryptic for my taste.

ctchou: I'll also change the proof of `IsRegular,hmul` to follow the same pattern and do some minor reorg in this file.

Cslib/Computability/Automata/NA/Loop.lean
ctchou: I didn't put this construction in the `NA.FinAcc` namespace because its return value is in `NA`, not `NA.FinAcc`.  Later in the proof of `.IsRegular.omegaPow`, we have an expression `⟨NA.loop na, {inl ()}⟩`, which is in `NA.Buchi`.  So it seems to me that putting the construction in the `NA` namespace is more neutral.

As to the name `na`, I don't really have a better idea.  If I use `nfa`, it may give the wrong impression that the automaton is finite-state, which is not the case here.  I could have defined `NA.loop` to be of type:
```
NA.loop : NA State Symbol -> Set State -> NA State Symbol
```
But that seems even uglier.

chenson2018: It seems to me the type of the parameters, not the return type, often determines the namespace. So since you have
```
loop : FinAcc State Symbol -> NA (Unit ⊕ State) Symbol 
```
I thought the `FinAcc` namespace would be nice for dot notation based on some of the proofs in this PR.

(Not a strong preference, just seemed potentially convenient)

ctchou: OK, you convinced me.  Being able to use the dot-notation is a big plus. But I kept most theorems in NA/Loop.lean in the `NA` namespace, because they are about infinite executions, not finite ones.  The price to pay is that I have to use `FinAcc.loop` to supply `grind` with the definition of the loop construction.

chenson2018: LGTM. Next release we will have the option of custom "grind sets" that collect groups of grind rules, not necessarily bound to namespaces. I don't know the details yet, but maybe this will help a bit.

chenson2018 (approved):
My comments have been addressed, thanks!

fmontesi (approved):

PR #239, opened 2025-12-27
feat: prove that regular languages are closed under concatenation

ctchou:
In this patch we prove that regular languages are closed under concatenation, by leveraging the results already proved about the infinite executions of the automata concatenation construction in `NA/Concat.lean`.  In order to do so, we introduce the notion of a "total LTS", which means that the LTS has a next state for any given starting state and label, and a "LTS.totalize" construction that converts any LTS into a total LTS by adding a sink state and transitions from all states to the sink state.  Then the notion of "total" and the "totalize" construction are generalized to NA.  By "totalizing" NA's, we can extend any finite execution of a NA into an infinite execution and thus re-use the results already proved for infinite executions.

The same technique can be applied to `NA/Loop.lean` to prove that regular languages are closed under the Kleene star; see PR https://github.com/leanprover/cslib/pull/241.

### DIFF
--- a/Cslib/Computability/Automata/NA/Concat.lean
+++ b/Cslib/Computability/Automata/NA/Concat.lean
@@ -141,7 +141,7 @@ def finConcat (na1 : FinAcc State1 Symbol) (na2 : FinAcc State2 Symbol)
 variable {na1 : FinAcc State1 Symbol} {na2 : FinAcc State2 Symbol}
 
 /-- `finConcat na1 na2` is total. -/
-theorem finConcat_total : (finConcat na1 na2).Total where
+instance : (finConcat na1 na2).Total where
   total s x := match s with
     | inl _ => ⟨inl (inr ()), by grind [finConcat, concat, NA.totalize, LTS.totalize]⟩
     | inr _ => ⟨inr (inr ()), by grind [finConcat, concat, NA.totalize, LTS.totalize]⟩
@@ -153,7 +153,7 @@ theorem finConcat_language_eq [Inhabited Symbol] :
   ext xl
   constructor
   · rintro ⟨s, _, t, h_acc, h_mtr⟩
-    obtain ⟨xs, ss, h_ωtr, rfl, rfl⟩ := LTS.Total.mTr_ωTr (ht := finConcat_total) h_mtr
+    obtain ⟨xs, ss, h_ωtr, rfl, rfl⟩ := LTS.Total.mTr_ωTr h_mtr
     have hc : (finConcat na1 na2).Run (xl ++ω xs) ss := by grind [Run]
     have hr : (ss xl.length).isRight := by grind
     obtain ⟨n, _⟩ := concat_run_proj hc hr
--- a/Cslib/Computability/Automata/NA/Total.lean
+++ b/Cslib/Computability/Automata/NA/Total.lean
@@ -41,7 +41,7 @@ theorem totalize_run_mtr {xs : ωSequence Symbol} {ss : ωSequence (State ⊕ Un
 theorem totalize_mtr_run [Inhabited Symbol] {xl : List Symbol} {s t : State}
     (hs : s ∈ na.start) (hm : na.MTr s xl t) :
     ∃ xs ss, na.totalize.Run (xl ++ω xs) ss ∧ ss 0 = inl s ∧ ss xl.length = inl t := by
-  grind [totalize, Run, (LTS.totalize.total na.toLTS).mTr_ωTr, =_ LTS.totalize.mtr_left_iff]
+  grind [totalize, Run, LTS.Total.mTr_ωTr <| LTS.totalize.mtr_left_iff.mpr hm]
 
 namespace FinAcc
 
--- a/Cslib/Foundations/Semantics/LTS/Basic.lean
+++ b/Cslib/Foundations/Semantics/LTS/Basic.lean
@@ -325,7 +325,7 @@ def LTS.totalize (lts : LTS State Label) : LTS (State ⊕ Unit) Label where
     | inr (), inl _ => False
 
 /-- The LTS constructed by `LTS.totalize` is indeed total. -/
-theorem LTS.totalize.total (lts : LTS State Label) : lts.totalize.Total where
+instance (lts : LTS State Label) : lts.totalize.Total where
   total _ _ := by simp [LTS.totalize]
 
 /-- In `LTS.totalize`, there is no finite execution from the sink state to any non-sink state. -/


Cslib/Computability/Automata/NA/Concat.lean
```diff
@@ -74,17 +74,19 @@ lemma concat_run_right {xs : ωSequence Symbol} {ss : ωSequence (State1 ⊕ Sta
...
+  let n := Nat.find hr'
+  have hl (k) (h_k : k < n) := not_isRight.mp <| Nat.find_min hr' h_k
+  refine ⟨n, by grind, ?_, ?_⟩
   · by_cases h_n : n = 0
     · grind [concat_start_right]
     · grind [concat_run_left_right]
-  · have hr : (ss n).isRight := Nat.find_spec hr
-    grind [concat_run_right hc n hl hr]
+  · have hr'' : (ss n).isRight := Nat.find_spec hr'
+    grind [concat_run_right hc n hl hr'']
```
chenson2018: I would prefer exact to passing the same expression to `grind`

```suggestion
    exact concat_run_right hc n hl hr''
```

Cslib/Computability/Automata/NA/Concat.lean
```diff
@@ -130,4 +132,54 @@ theorem concat_language_eq {acc2 : Set State2} :
...
+
+variable {na1 : FinAcc State1 Symbol} {na2 : FinAcc State2 Symbol}
+
+/-- `finConcat na1 na2` is total. -/
+def finConcat_total : (finConcat na1 na2).Total where
+  next s x := match s with
+    | inl s1 => inl (inr ())
+    | inr s2 => inr (inr ())
+  total s x := by
+    cases s <;> simp [finConcat, concat, NA.totalize, LTS.totalize]
```
chenson2018: `grind` will try cases here:

```suggestion
  total s x := by grind [finConcat, concat, NA.totalize, LTS.totalize]
```

Cslib/Computability/Automata/NA/Concat.lean
```diff
@@ -130,4 +132,54 @@ theorem concat_language_eq {acc2 : Set State2} :
...
+
+/-- `finConcat na1 na2` accepts the concatenation of the languages of `na1` and `na2`. -/
+theorem finConcat_language_eq [Inhabited Symbol] :
```
chenson2018, line 150: I spent some time trying to tidy this up a bit:
```lean
theorem finConcat_language_eq [Inhabited Symbol] :
    language (FinAcc.mk (finConcat na1 na2) (inr '' (inl '' na2.accept))) =
    language na1 * language na2 := by
  ext xl
  constructor
  · rintro ⟨s, _, t, h_acc, h_mtr⟩
    obtain ⟨xs, ss, h_ωtr, rfl, rfl⟩ := LTS.Total.mTr_ωTr finConcat_total h_mtr
    have hc : (finConcat na1 na2).Run (xl ++ω xs) ss := by grind [Run]
    have hr : (ss xl.length).isRight := by grind
    obtain ⟨n, _⟩ := concat_run_proj hc hr
    refine ⟨xl.take n, ?_, xl.drop n, ?_, ?_⟩
    · grind [totalize_language_eq, take_append_of_le_length]
    · have : ss xl.length = (ss.drop n) (xl.length - n) := by grind
      grind [drop_append_of_le_length, take_append_of_le_length, totalize_run_mtr]
    · exact xl.take_append_drop n
  · rintro ⟨xl1, h_xl1, xl2, h_xl2, rfl⟩
    rw [← totalize_language_eq] at h_xl1
    obtain ⟨_, h_s2, _, _, h_mtr2⟩ := h_xl2
    obtain ⟨_, _, h_run2, _, _⟩ := totalize_mtr_run h_s2 h_mtr2
    obtain ⟨ss, ⟨_, h_ωtr⟩, _⟩ := concat_run_exists h_xl1 h_run2
    grind [
      finConcat, List.length_append, take_append_of_le_length,
      extract_eq_drop_take, =_ append_append_ωSequence, get_drop xl2.length xl1.length ss,
      LTS.ωTr_mTr h_ωtr (zero_le (xl1.length + xl2.length))
    ]
```
I am increasingly thinking there are some design issues around existential quantification, defeq abuse, and `grind`, but let's postpone that discussion.

Cslib/Computability/Automata/NA/Total.lean
```diff
@@ -0,0 +1,61 @@
...
+
+/-- Any finite execution of the original NA can be extended to an infinite execution of
+`NA.totalize`, provided that the alphabet is inbabited. -/
+theorem totalize_mtr_run [Inhabited Symbol] {xl : List Symbol} {s t : State}
+    (hs : s ∈ na.start) (hm : na.MTr s xl t) :
+    ∃ xs ss, na.totalize.Run (xl ++ω xs) ss ∧ ss 0 = inl s ∧ ss xl.length = inl t := by
+  obtain ⟨xs, ss, _⟩ := LTS.Total.mTr_ωTr
+    (LTS.totalize.total na.toLTS) (LTS.totalize.mtr_left_iff.mpr hm)
+  use xs, ss
+  grind [totalize, Run.mk]
```
chenson2018: No need to destructure here:

```suggestion
  grind [totalize, Run, (LTS.totalize.total na.toLTS).mTr_ωTr, =_ LTS.totalize.mtr_left_iff]
```

(some of the namespacing seems awkward here, I didn't look too closely)

Cslib/Computability/Languages/RegularLanguage.lean
```diff
@@ -121,4 +131,16 @@ theorem IsRegular.iSup {I : Type*} [Finite I] {s : Set I} {l : I → Language Sy
...
+/-- The concatenation of two regular languages is regular. -/
+@[simp]
+theorem IsRegular.mul [Inhabited Symbol] {l1 l2 : Language Symbol}
+    (h1 : l1.IsRegular) (h2 : l2.IsRegular) : (l1 * l2).IsRegular := by
+  rw [IsRegular.iff_nfa] at h1 h2 ⊢
+  obtain ⟨State1, h_fin1, nfa1, rfl⟩ := h1
+  obtain ⟨State2, h_fin1, nfa2, rfl⟩ := h2
+  use (State1 ⊕ Unit) ⊕ (State2 ⊕ Unit), inferInstance,
+    ⟨finConcat nfa1 nfa2, inr '' (inl '' nfa2.accept)⟩
+  rw [finConcat_language_eq]
```
chenson2018: More clear to use exact:

```suggestion
  exact finConcat_language_eq
```

Cslib/Computability/Automata/NA/Total.lean
```diff
@@ -0,0 +1,61 @@
...
+    (h : na.totalize.Run xs ss) (hl : (ss n).isLeft) :
+    ∃ s t, na.MTr s (xs.take n) t ∧ s ∈ na.start ∧ ss 0 = inl s ∧ ss n = inl t := by
+  obtain ⟨s, _, _⟩ := h.start
+  obtain ⟨t, _⟩ := isLeft_iff.mp hl
+  use s, t
+  refine ⟨?_, by grind⟩
+  rw [← LTS.totalize.mtr_left_iff, ← extract_eq_take]
+  have := LTS.ωTr_mTr h.trans (show 0 ≤ n by grind)
+  simp_all only
+  assumption
```
chenson2018: Now that I see the example, I do find this odd. Let's write this more explicitly and leave a TODO so the bot will open an issue:

```suggestion
  obtain ⟨s, _, eq₁⟩ := h.start
  obtain ⟨t, eq₂⟩ := isLeft_iff.mp hl
  use s, t
  refine ⟨?_, by grind⟩
  -- TODO: `grind` does not use congruence relations with `na.totalize.MTr`
  rw [← LTS.totalize.mtr_left_iff, ← extract_eq_take, eq₁, ← eq₂]
  exact LTS.ωTr_mTr h.trans (by grind)
```

chenson2018 (changes requested):
I'm going to leave review of LTS to Fabrizio, but I've left a few comments.

Cslib/Computability/Automata/NA/Concat.lean
ctchou: Fixed.

Cslib/Computability/Automata/NA/Concat.lean
ctchou: Fixed.

Cslib/Computability/Automata/NA/Total.lean
ctchou: Fixed.

Cslib/Computability/Languages/RegularLanguage.lean
ctchou: Fixed.

Cslib/Computability/Automata/NA/Total.lean
ctchou: Fixed.

Cslib/Computability/Automata/NA/Concat.lean
ctchou: Fixed.  Conceptually this proof is not hard, but I admit that it was pretty painful when doing it.  The problem with `grind` is that it is not obvious at the beginning what you should give to `grind` until you have gone through a lot of trials and errors.  That's why I have a `simp` in the middle to simplify an antecedent. I know that I can roll all `simp` lemmas into the final `grind`.  But until I get there, it is far more pleasant to simplify the proof state into something that is actually readable.

chenson2018: Yes, there is definitely a mismatch between the proof that makes sense for writing the first time and a very compact version. I sometimes find myself wishing that we could carry around both, because sometimes when a `grind` proof breaks, it is nice to have the original and I find myself digging through the git/PR history. (I might bring this up on Zulip later)

Cslib/Foundations/Semantics/LTS/Basic.lean
```diff
@@ -250,16 +250,119 @@ theorem LTS.ωTr_mTr {n m : ℕ} {hnm : n ≤ m} (h : lts.ωTr ss μs) :
...
+A LTS is total iff every state has a transition for every label.
+-/
+
+open Sum ωSequence
+
+variable {State Label : Type*} {lts : LTS State Label}
+
+/-- `LTS.Total` provides a witness that the LTS is total. -/
+structure LTS.Total (lts : LTS State Label) where
+  /-- `next` rovides a next state for any given starting state and label. -/
```
fmontesi: provides (lacks a p)

fmontesi:
There's a class `LTS.LeftTotal` in LTS/Basic.lean that I wrote for this kind of stuff, so I'd like to discuss the design choice of adding `LTS.Total`. One important difference is that I've used an existential:

```
class LTS.LeftTotal (lts : LTS State Label) where
  left_total : ∀ s μ, ∃ s', lts.Tr s μ s'
```

So I guess this would make some of the constructions noncomputable -- by contrast, `LTS.Total` is a witness from which we can build things! In particular, can't we noncomputably get the `next` function just from knowing that an LTS is `LeftTotal`?

Is there an example from mathlib we can refer to for this discussion maybe?

It might make sense to have both `LeftTotal` and `Total`. However, I'd probably change the name of the latter, since in mathlib people seem to stick to left total for this kind of things. I don't have a good proposal yet.

Cslib/Foundations/Semantics/LTS/Basic.lean

chenson2018:
To clarify, you are wondering if these are valuable to both have around? You can write:
```lean
instance (lt : LTS.Total lts) :  LTS.LeftTotal lts where
  left_total s μ := ⟨lt.next s μ, lt.total s μ⟩

noncomputable def LTS.leftTotal_total [inst : LTS.LeftTotal lts] : LTS.Total lts where
  next s μ := Classical.choose (inst.left_total s μ)
  total s μ :=  Classical.choose_spec (inst.left_total s μ)
```
So if you don't care about the computability aspect, it seems they can be consolidated.

ctchou:
I completely missed "LTS.LeftTotal".  Sorry!   (By the way, shouldn't it be "RightTotal"?). Anyway, I chose the name "total" by following the standard notion of a relation or function being "total" or "partial": https://en.wikipedia.org/wiki/Total_relation.

Actually I started with using an existential quantifier as in `LTS.LeftTotal`.  But then the need to define `LTS.makeωTr` and prove `LTS.Total.ωTr_exists` made me realize that an explicit witness is much easier to work with.  Furthermore, when we do an actual proof of existence (such as `finConcat_total` in `NA/Concat.lean`), we will typically present an explicit witnesses anyway.  This is analogous to assuming a type to be `Inhabited` vs assuming it to be `Nonempty`.  The former is easier to use, because it provides an explicit witness, and the typical proof of a type being nonempty is to present a witness.

fmontesi:
@ctchou Heh, funny thing: I had originally called it total too, but in mathlib they follow the 'left total' terminology. (Which makes sense, because you can also be concerned about right totality in general.)

@chenson2018 I expect them both to be useful, exactly because of what @ctchou wrote. I'm just wondering what we should call them, because having `LeftTotal` for the 'declarative' version and `Total` for the 'constructive' version is really confusing naming. Hence my wondering of whether there's a similar case in mathlib, to see if we can get some inspiration.

Just to give an example without caring too much about elegant names, the current `Total` could be called `ProofOfLeftTotal` (blergh!) or `LeftTotalWitness`. I hope we can come up with something better...? :-)

An observation: `Total` is basically equivalent to having an FLTS `flts` and showing that `flts.toLTS.Tr` is a sub-relation of `lts.Tr`. For any FLTS, `.toLTS.Tr` is always `LeftTotal`, so left-totality follows. I'm not sure if this observation is already useful, just wanted to point it out. Might lead to some theorems.

ctchou:
I find mathlib's use of "left" and "right" regarding relations is really unfortunate, because in the "constructive" versions of those notions the witness is for the opposite side.  Here's my proposal:

* We rename `LTS.LeftTotal` to `LTS.Total` and turn it into a `Prop`.  "Total" is a completely standard terminology in this context.
* We provide `LTS.chooseFLTS` using `Classical.choose` and prove what is now `LTS.Total.total`.  The rest of my code will be changed to use `LTS.chooseFLTS`.
* We will not have `LTS.RightTotal`, but we can define `LTS.reverse` to reverse an LTS.  This can come in handy in, for example, formalizing Brzozowski's algorithm: https://en.wikipedia.org/wiki/DFA_minimization#Brzozowski's_algorithm

How do you think about the above?

fmontesi:
I like it, modulo the fact that I'd keep totality a class instead of turning it into a Prop, since that's our current practice for similar properties. Or do you have a place where having it as a class is particularly inconvenient?

Full reasoning on terminology, etc., below -- some nuances might be important now or later.

In the context of relations, they made the most reasonable/immediate choice: it's pretty standard terminology for relations, otherwise there's no established way to name all interesting classes. When one can commit to being interested *only* in left-totality, saying 'total' is perfectly fine because that's the default meaning, but that's not good in general.
See p. 3 of https://homepages.hass.rpi.edu/heuveb/Teaching/Logic/CompLogic/Web/Handouts/SetsRelationsFunctions.pdf and https://en.wikipedia.org/wiki/Total_relation#:~:text=For%20relations%20R%20where%20x,that%20relation%20to%20something%20else.%22

However, in this discussion, we can consider also the glossary of transition systems. In this context, people are typically only interested in:
- Left totality. What we're discussing here. We could just call it 'total' (the standard default) or 'complete' (inspired by what people typically say in automata theory for 'complete' transition function). I'm ok with either.
- Right uniqueness. In LTS land, this is definitely called 'deterministic LTS'. In fact, I've already called it as such in `LTS.Deterministic`.

So we can either:
1. If you're ok with keeping `LTS.Deterministic` to talk also about automata -- so you'd get that a 'Deterministic' NA means that it's right-unique, which is sometimes what people mean but not always because some works use it to mean that there's a corresponding complete DA iirc -- then I think it's ok to rename `LeftTotal` to `Total` (or `Complete`, if you prefer it? do they say the same in graphs?).
2. Otherwise, if we risk having confusion, I'd keep `LeftTotal` and rename `Deterministic` to `RightUnique`.

Assuming (1) is ok (my preference), I'd suggest that we rename `LeftTotal` to `Total`, but keeping it as a class, and then:

```lean
-- In LTS/Basic.lean
noncomputable def LTS.chooseFLTS [inst : LTS.Total lts] : FLTS State Label [...]

theorem LTS.chooseFLTS_tr [LTS.Total lts] (s : State) (μ : Label) : lts.Tr s μ (lts.chooseFLTS s μ) := [...]

-- In FLTS/FLTSToLTS.lean.
instance (flts : FLTS State Label) : LTS.Total (flts.toLTS) where [...]
```

ctchou:
Let's stick with the names `Total` and `Deterministic`.  I'll update my code and have you review it.

ctchou:
I've updated the code.  Now `LTS.LeftTotal` is called `LTS.Total` and all results depending on `LTS.Total` have been updated accordingly.

Cslib/Foundations/Semantics/LTS/Basic.lean
```diff
@@ -250,16 +251,122 @@ theorem LTS.ωTr_mTr {n m : ℕ} {hnm : n ≤ m} (h : lts.ωTr ss μs) :
...
+
+/-- `LTS.totalize` constructs a total LTS from any given LTS by adding a sink state. -/
+def LTS.totalize (lts : LTS State Label) : LTS (State ⊕ Unit) Label where
+  Tr s' μ t' := match s', t' with
+    | inl s, inl t => lts.Tr s μ t
+    | _, inr () => True
+    | inr (), inl _ => False
+
+/-- The LTS constructed by `LTS.totalize` is indeed total. -/
+theorem LTS.totalize.total (lts : LTS State Label) : lts.totalize.Total where
```
fmontesi: Should this be an `instance`?

```
instance (lts : LTS State Label) : lts.totalize.Total [...]
```

(Or maybe LTS.Total lts.totalize if Lean doesn't accept the rhs.)

I haven't tested this, just curious. If it can be an instance, it probably should I guess? (That way the typeclass resolutor can figure it out for calls to totalize.)

fmontesi:
That looks good to me. Only one technical doubt left, see my comment on theorem vs instance.

Cslib/Foundations/Semantics/LTS/Basic.lean

ctchou:
Now "instance" instead "theorem" is used for LTS.Total.

fmontesi (approved):

PR #241, opened 2025-12-31
feat: prove that regular languages are closed under Kleene star

ctchou:
This patch proves that regular languages are closed under Kleene star.  The proof technique is similar to the one used in PR https://github.com/leanprover/cslib/pull/239 and re-uses much of the infrastructure introduce in that PR.

### DIFF
--- a/Cslib/Computability/Automata/NA/Loop.lean
+++ b/Cslib/Computability/Automata/NA/Loop.lean
@@ -4,8 +4,12 @@ Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Ching-Tsun Chou
 -/
 
-import Cslib.Computability.Automata.NA.Total
-import Cslib.Foundations.Data.OmegaSequence.Temporal
+module
+
+public import Cslib.Computability.Automata.NA.Total
+public import Cslib.Foundations.Data.OmegaSequence.Temporal
+
+@[expose] public section
 
 /-! # Loop construction on nondeterministic automata. -/
 
--- a/Cslib/Computability/Languages/RegularLanguage.lean
+++ b/Cslib/Computability/Languages/RegularLanguage.lean
@@ -4,15 +4,19 @@ Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Ching-Tsun Chou
 -/
 
-import Cslib.Computability.Automata.DA.Prod
-import Cslib.Computability.Automata.DA.ToNA
-import Cslib.Computability.Automata.NA.Concat
-import Cslib.Computability.Automata.NA.Loop
-import Cslib.Computability.Automata.NA.ToDA
-import Mathlib.Computability.DFA
-import Mathlib.Data.Finite.Sum
-import Mathlib.Data.Set.Card
-import Mathlib.Tactic.Common
+module
+
+public import Cslib.Computability.Automata.DA.Prod
+public import Cslib.Computability.Automata.DA.ToNA
+public import Cslib.Computability.Automata.NA.Concat
+public import Cslib.Computability.Automata.NA.Loop
+public import Cslib.Computability.Automata.NA.ToDA
+public import Mathlib.Computability.DFA
+public import Mathlib.Data.Finite.Sum
+public import Mathlib.Data.Set.Card
+public import Mathlib.Tactic.Common
+
+@[expose] public section
 
 /-!
 # Regular languages
--- a/.github/workflows/lean_action_ci.yml
+++ b/.github/workflows/lean_action_ci.yml
@@ -17,11 +17,14 @@ jobs:
       - uses: leanprover/lean-action@v1
         with:
           build-args: "--wfail --iofail"
-          mk_all-check: "true"
-      - name: "lake exe shake"
+      - name: "lake exe mk_all --check --module"
         run: |
           set -e
-          lake exe shake Cslib
+          lake exe mk_all --check --module
+      #- name: "lake exe shake"
+      #  run: |
+      #    set -e
+      #    lake exe shake Cslib
       - uses: leanprover-community/lint-style-action@main
         with:
           mode: check
--- a/Cslib.lean
+++ b/Cslib.lean
@@ -1,78 +1,80 @@
-import Cslib.Algorithms.Lean.MergeSort.MergeSort
-import Cslib.Algorithms.Lean.TimeM
-import Cslib.Computability.Automata.Acceptors.Acceptor
-import Cslib.Computability.Automata.Acceptors.OmegaAcceptor
-import Cslib.Computability.Automata.DA.Basic
-import Cslib.Computability.Automata.DA.Buchi
-import Cslib.Computability.Automata.DA.Prod
-import Cslib.Computability.Automata.DA.ToNA
-import Cslib.Computability.Automata.EpsilonNA.Basic
-import Cslib.Computability.Automata.EpsilonNA.ToNA
-import Cslib.Computability.Automata.NA.Basic
-import Cslib.Computability.Automata.NA.BuchiEquiv
-import Cslib.Computability.Automata.NA.BuchiInter
-import Cslib.Computability.Automata.NA.Concat
-import Cslib.Computability.Automata.NA.Hist
-import Cslib.Computability.Automata.NA.Loop
-import Cslib.Computability.Automata.NA.Prod
-import Cslib.Computability.Automata.NA.Sum
-import Cslib.Computability.Automata.NA.ToDA
-import Cslib.Computability.Automata.NA.Total
-import Cslib.Computability.Languages.ExampleEventuallyZero
-import Cslib.Computability.Languages.Language
-import Cslib.Computability.Languages.OmegaLanguage
-import Cslib.Computability.Languages.OmegaRegularLanguage
-import Cslib.Computability.Languages.RegularLanguage
-import Cslib.Foundations.Control.Monad.Free
-import Cslib.Foundations.Control.Monad.Free.Effects
-import Cslib.Foundations.Control.Monad.Free.Fold
-import Cslib.Foundations.Data.FinFun
-import Cslib.Foundations.Data.HasFresh
-import Cslib.Foundations.Data.Nat.Segment
-import Cslib.Foundations.Data.OmegaSequence.Defs
-import Cslib.Foundations.Data.OmegaSequence.Flatten
-import Cslib.Foundations.Data.OmegaSequence.InfOcc
-import Cslib.Foundations.Data.OmegaSequence.Init
-import Cslib.Foundations.Data.OmegaSequence.Temporal
-import Cslib.Foundations.Data.Relation
-import Cslib.Foundations.Lint.Basic
-import Cslib.Foundations.Semantics.FLTS.Basic
-import Cslib.Foundations.Semantics.FLTS.FLTSToLTS
-import Cslib.Foundations.Semantics.FLTS.LTSToFLTS
-import Cslib.Foundations.Semantics.FLTS.Prod
-import Cslib.Foundations.Semantics.LTS.Basic
-import Cslib.Foundations.Semantics.LTS.Bisimulation
-import Cslib.Foundations.Semantics.LTS.Simulation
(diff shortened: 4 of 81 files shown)


ctchou:
Rebase on https://github.com/leanprover/cslib/pull/239.

ctchou:
Rebase on https://github.com/leanprover/cslib/pull/239 again.

ctchou:
Rebase on https://github.com/leanprover/cslib/pull/239.

Cslib/Computability/Automata/NA/Loop.lean
```diff
@@ -156,8 +166,64 @@ theorem loop_language_eq [Inhabited Symbol] :
...
+namespace FinAcc
+
+open scoped Computability
+
+/-- `finLoop na` is the loop construction applied to the "totalized" version of `na`. -/
+def finLoop (na : FinAcc State Symbol) : NA (Unit ⊕ (State ⊕ Unit)) Symbol :=
+  FinAcc.loop ⟨na.totalize, inl '' na.accept⟩
+
+/-- `finLoop na` is total, assuming that `na` has at least one start state. -/
+instance [h : Fact na.start.Nonempty] : na.finLoop.Total where
```
chenson2018: This seems problematic, I'm going to take a closer look.

ctchou: It seems to me that my use of `Fact` is pretty close to that in `Mod.instField` in mathlib:
https://github.com/leanprover-community/mathlib4/blob/001c934c401de5997fba93beef118a544c7e1ed6/Mathlib/Algebra/Field/ZMod.lean#L29-L36

chenson2018: But here you're using fact with `Set.Nonempty`, which is already related to the class `Nonempty`. It seems wrong to use `Fact`.

ctchou: `Set.Nonempty` is a `Prop`, not a class:
https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Set/Defs.html#Set.Nonempty

kmill: Using `[Nonempty na.start]` is more standard — this uses the coercion from sets to types.

By the way, `Prop` and `class` are orthogonal concepts. The salient distinction is that `Set.Nonempty` is a `def` rather than a `class`. (Indeed, `Nonempty` is a `Prop` and a `class`.)

chenson2018: @ctchou Please see the commit I pushed. The only change is using `[Nonempty na.start]` as Kyle has described.

ctchou: Thanks for the patch!  I'd appreciate a review of the rest of the PR as well.

Cslib/Computability/Automata/NA/Loop.lean
```diff
@@ -77,16 +77,17 @@ lemma loop_run_from_left {xs : ωSequence Symbol} {ss : ωSequence (Unit ⊕ Sta
...
 of a nonempty finite run of `na` followed by a run of `na.loop`. -/
-theorem loop_run_one_iter {xs : ωSequence Symbol} {ss : ωSequence (Unit ⊕ State)}
-    (h : na.loop.Run xs ss) (h1 : ∃ k, 0 < k ∧ (ss k).isLeft) :
-    ∃ n, xs.take n ∈ language na - 1 ∧ na.loop.Run (xs.drop n) (ss.drop n) := by
-  let n := Nat.find h1
-  have : 0 < n ∧ (ss n).isLeft := Nat.find_spec h1
-  have : ∀ k, 0 < k → k < n → (ss k).isRight := by grind [Nat.find_min h1]
-  refine ⟨n, ⟨?_, ?_⟩, ?_⟩
+theorem loop_run_one_iter {xs : ωSequence Symbol} {ss : ωSequence (Unit ⊕ State)} {k : ℕ}
+    (h : na.loop.Run xs ss) (h1 : 0 < k ∧ (ss k).isLeft) :
```
chenson2018: Now that it is no longer an existential, it seems the conjunction `h1` should be split into two hypotheses.

Cslib/Computability/Automata/NA/Loop.lean
```diff
@@ -102,6 +103,15 @@ theorem loop_fin_run_exists {xl : List Symbol} (h : xl ∈ language na) :
   · use [inl ()] ++ (sl.extract 1 xl.length).map inr ++ [inl ()]
     grind [FinAcc.loop]
 
+/-- For any finite word in `language na`, there is a corresponding multistep transition
+of `na.loop`. -/
+theorem loop_fin_run_mtr {xl : List Symbol} (h : xl ∈ language na) :
+    na.loop.MTr (inl ()) xl (inl ()) := by
+  obtain ⟨sl, _, _, _, h_run⟩ := loop_fin_run_exists h
+  suffices ∀ k, ∀ _ : k ≤ xl.length, na.loop.MTr (inl ()) (xl.take k) sl[k] by grind
```
chenson2018: It is a bit less confusing and more standard I think to write this as:
```suggestion
  suffices ∀ k, (_ : k ≤ xl.length) → na.loop.MTr (inl ()) (xl.take k) sl[k] by grind
```
I would do the same in `loop_fin_run_exists` for consistency.

Cslib/Computability/Automata/NA/Loop.lean
chenson2018, line 99: Now that I see it used more, the second existential over an equality seems a bit odd to me. Is it possible to rewrite this, or could you explain to me why it's written this way?

Cslib/Computability/Automata/NA/Loop.lean
```diff
@@ -156,8 +166,64 @@ theorem loop_language_eq [Inhabited Symbol] :
...
+      · exact xl.take_append_drop n
+  · rintro (h | h)
+    · obtain ⟨xl1, ⟨h_xl1, _⟩, xl2, h_xl2, rfl⟩ := h
+      rw [← totalize_language_eq] at h_xl1
+      have := loop_fin_run_mtr h_xl1
+      obtain ⟨s1, _, s2, _, _⟩ := h_xl2
+      obtain ⟨rfl⟩ : s1 = inl () := by grind [finLoop, loop]
+      obtain ⟨rfl⟩ : s2 = inl () := by grind [finLoop, loop]
+      refine ⟨inl (), by assumption, inl (), by assumption, ?_⟩
+      apply LTS.MTr.comp <;> assumption
```
chenson2018: You can match the style you use below by being slightly more explicit:
```suggestion
      refine ⟨inl (), ?_, inl (), ?_, LTS.MTr.comp _ this ?_⟩ <;> assumption
```
though it is probably more clear and performant to just write everything explicitly since all the holes here are just single hypotheses.

Cslib/Computability/Automata/NA/Loop.lean
```diff
@@ -156,8 +166,64 @@ theorem loop_language_eq [Inhabited Symbol] :
...
+    by_cases h_xl : xl = []
+    · grind [mem_add, mem_one]
+    · have : Nonempty na.start := by
+        obtain ⟨_, s0, _, _⟩ := nonempty_iff_ne_empty.mpr h
+        use s0
+      obtain ⟨xs, ss, h_ωtr, rfl, rfl⟩ := LTS.Total.mTr_ωTr h_mtr
+      have h_run : na.finLoop.Run (xl ++ω xs) ss := by grind [Run]
+      have h1 : 0 < xl.length ∧ (ss xl.length).isLeft := by
+        simp only [mem_singleton_iff] at h_acc
+        grind  [List.eq_nil_iff_length_eq_zero]
```
chenson2018:
```suggestion
        grind
```

Cslib/Computability/Languages/RegularLanguage.lean
```diff
@@ -143,4 +144,16 @@ theorem IsRegular.mul [Inhabited Symbol] {l1 l2 : Language Symbol}
...
+/-- The Kleene star of a regular language is regular. -/
+@[simp]
+theorem IsRegular.kstar [Inhabited Symbol] {l : Language Symbol}
+    (h : l.IsRegular) : (l∗).IsRegular := by
+  by_cases h_l : l = 0
+  · simp [h_l]
+  · rw [IsRegular.iff_nfa] at h ⊢
+    obtain ⟨State, h_fin, nfa, rfl⟩ := h
+    use Unit ⊕ (State ⊕ Unit), inferInstance, ⟨finLoop nfa, {inl ()}⟩
+    grind [loop_language_eq]
```
chenson2018: Since the last proof can be a short `exact`, I'd just write:
```suggestion
    use Unit ⊕ (State ⊕ Unit), inferInstance, ⟨finLoop nfa, {inl ()}⟩, loop_language_eq h_l
```
for clarity and performance.

Cslib/Computability/Automata/NA/Loop.lean
```diff
@@ -156,8 +166,64 @@ theorem loop_language_eq [Inhabited Symbol] :
...
+      obtain ⟨n, h_n, _, _, h_ωtr'⟩ := loop_run_one_iter h_run h1
+      left; refine ⟨xl.take n, ?_, xl.drop n, ?_, ?_⟩
+      · grind [totalize_language_eq, take_append_of_le_length]
+      · refine ⟨ss n, by grind, ss xl.length, by grind, ?_⟩
+        have := LTS.ωTr_mTr h_ωtr' (show 0 ≤ xl.length - n by grind)
+        have : n + (xl.length - n) = xl.length := by grind
+        have : ((xl ++ω xs).drop n).extract 0 (xl.length - n) = xl.drop n := by
+          grind [extract_eq_take, drop_append_of_le_length, take_append_of_le_length,
+            List.take_drop, List.take_length, List.length_drop]
```
chenson2018: You can remove some theorems:
```suggestion
          grind [extract_eq_take, drop_append_of_le_length, take_append_of_le_length]
```

(Note to self: weird output from `grind?` here, maybe worth minimizing)

chenson2018 (changes requested):

Cslib/Computability/Automata/NA/Loop.lean
ctchou: Fixed.

Cslib/Computability/Automata/NA/Loop.lean
ctchou: Fixed in both places.

Cslib/Computability/Automata/NA/Loop.lean
ctchou: I tried to rewrite `∃ _ : sl.length = xl.length + 1,` as `(_ : sl.length = xl.length + 1) ∧`, but that didn't work.

I think the reason is that we need to create a new context in which there is an anonymous proof of `sl.length = xl.length + 1`, so that Lean knows to use it when proving that list indexing operations in the context are within bounds.  An existential quantifier creates a new context, but a conjunction doesn't.  On the other hand, `(_ : p) → ...` is literally `∀ _ : p, ...` in Lean.

Cslib/Computability/Automata/NA/Loop.lean
ctchou: I'll use the line you suggested above.  But I will continue to use `assumption` because the hypotheses in question are anonymous.

Cslib/Computability/Automata/NA/Loop.lean
ctchou: Fixed.

Cslib/Computability/Languages/RegularLanguage.lean
ctchou: Fixed.

Cslib/Computability/Automata/NA/Loop.lean
ctchou: Fixed.

ctchou:
Merge in `upstream/main` which includes the move to the new module system.

ctchou:
@fmontesi, could you review this PR?  Chris has approved it.  Thanks!

fmontesi (approved):

PR #245, opened 2026-01-05
chore: weekly lints for 2026-01-05

chenson2018:
As identified by [this workflow](https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/Weekly.20linting.20log/near/566277830).


chenson2018:
(I have added this week's run as well)

PR #247, opened 2026-01-07
feat: adding free reader monad API and minor edits to free writer and free cont

tannerduve:
This PR adds an API for the `Reader` monad represented as a free monad, to the list of existing free effects. Also made some minor changes to the `Writer` and `Cont` APIs in this same file

### DIFF
--- a/Cslib/Foundations/Control/Monad/Free/Effects.lean
+++ b/Cslib/Foundations/Control/Monad/Free/Effects.lean
@@ -261,41 +261,6 @@ instance [Monoid ω] : MonadWriter ω (FreeWriter ω) where
   listen := listen
   pass := pass
 
-/--
-Rewrite the log produced by `m` using `f`.
-
-This evaluates `m` via `run` and then re-emits a single `tell` containing the rewritten log.
--/
-def censor [Monoid ω] (f : ω → ω) (m : FreeWriter ω α) : FreeWriter ω α :=
-  let (a, w) := run m
-  liftBind (.tell (f w)) (fun _ => .pure a)
-
-/--
-Map each `tell w` to `tell (φ w)`, where `φ` is a monoid homomorphism.
-
-This is a syntactic transformation on the computation tree: it preserves the original sequencing
-of `tell`s, rewriting each one in place (rather than aggregating the log before rewriting).
--/
-def censorHom [Monoid ω] (φ : ω →* ω) : FreeWriter ω α → FreeWriter ω α
-  | .pure a => .pure a
-  | .liftBind (.tell w) k =>
-      .liftBind (.tell (φ w)) (fun _ => censorHom φ (k .unit))
-
-@[simp] theorem run_censor [Monoid ω] (f : ω → ω) (m : FreeWriter ω α) :
-    run (censor (α := α) f m) = (let (a, w) := run m; (a, f w)) := by
-  simp [censor, run, run_liftBind_tell, mul_one]
-
-@[simp] theorem run_censorHom [Monoid ω] (φ : ω →* ω) (m : FreeWriter ω α) :
-    run (censorHom (α := α) φ m) = (let (a, w) := run m; (a, φ w)) := by
-  induction m with
-  | pure a => simp [censorHom, run]
-  | liftBind op k ih =>
-    cases op; simp [censorHom, run, ih, map_mul]
-
-theorem run_censorHom_eq_run_censor [Monoid ω] (φ : ω →* ω) (m : FreeWriter ω α) :
-    run (censorHom (α := α) φ m) = run (censor (α := α) (fun w => φ w) m) := by
-  simp [run_censor, run_censorHom]
-
 end FreeWriter
 
 /-! ### Continuation Monad via `FreeM` -/
@@ -439,7 +404,7 @@ lemma run_read (k : σ → FreeReader σ α) (s₀ : σ) :
     run (liftBind .read k) s₀ = run (k s₀) s₀ := rfl
 
 instance instMonadWithReaderOf : MonadWithReaderOf σ (FreeReader σ) where
-  withReader := fun {α} f m =>
+  withReader {α} f m :=
     let rec go : FreeReader σ α → FreeReader σ α
       | .pure a => .pure a
       | .liftBind .read cont =>
@@ -453,14 +418,6 @@ instance instMonadWithReaderOf : MonadWithReaderOf σ (FreeReader σ) where
   | liftBind op cont ih =>
     cases op
     simpa [withTheReader, instMonadWithReaderOf, run] using (ih (f s) s)
-
-/-- Read the environment and return `g` applied to it. -/
-def asks (g : σ → α) : FreeReader σ α :=
-  g <$> (read : FreeReader σ σ)
-
-@[simp] lemma asks_run (g : σ → α) (s : σ) :
-    run (asks (σ := σ) (α := α) g) s = g s := by
-  simp [asks, run]
 
 end FreeReader
 


chenson2018:
@eric-wieser @quangvdao I recall you having thoughts on the initial PR, anything for this one?

quangvdao:
none, looks good to me. glad this is now in cslib!

Cslib/Foundations/Control/Monad/Free/Effects.lean
```diff
@@ -265,6 +269,23 @@ instance [Monoid ω] : MonadWriter ω (FreeWriter ω) where
...
 
+/-- Rewrite the log produced by `m` using `f`. -/
+def censor [Monoid ω] (f : ω → ω) (m : FreeWriter ω α) : FreeWriter ω α :=
+  pass (m >>= fun a => pure (a, f))
+
+/-- Map each `tell w` to `tell (φ w)`, where `φ` is a monoid homomorphism. -/
+def censorHom [Monoid ω] (φ : ω →* ω) : FreeWriter ω α → FreeWriter ω α
+  | .pure a => .pure a
+  | .liftBind (.tell w) k =>
+      .liftBind (.tell (φ w)) (fun _ => censorHom φ (k .unit))
```
eric-wieser: Are these two equal? Why have both?

Cslib/Foundations/Control/Monad/Free/Effects.lean
```diff
@@ -353,6 +374,93 @@ lemma run_callCC (f : MonadCont.Label α (FreeCont r) β → FreeCont r α) (k :
...
+
+/-- Reader monad via the `FreeM` monad -/
+abbrev FreeReader (σ) := FreeM (ReaderF σ)
+
+namespace FreeReader
+
+variable {σ : Type u} {α : Type u}
+
+instance : Monad (FreeReader σ) := inferInstance
+instance : LawfulMonad (FreeReader σ) := inferInstance
```
eric-wieser: I believe adding instances for `abbrev`s just slows things down, because these instances also fire when you write `FreeM (ReaderF σ)`
```suggestion
```

Cslib/Foundations/Control/Monad/Free/Effects.lean
```diff
@@ -353,6 +374,93 @@ lemma run_callCC (f : MonadCont.Label α (FreeCont r) β → FreeCont r α) (k :
...
+  match comp with
+  | .pure a => a
+  | .liftBind ReaderF.read a => run (a s₀) s₀
+
+/--
+The canonical interpreter `toReaderM` derived from `liftM` agrees with the hand-written
+recursive interpreter `run` for `FreeReader` -/
+@[simp]
+theorem run_toReaderM {α : Type u} (comp : FreeReader σ α) (s : σ) :
+  (toReaderM comp).run s = run comp s := by
```
eric-wieser: Indent is wrong here.

Cslib/Foundations/Control/Monad/Free/Effects.lean
```diff
@@ -353,6 +374,93 @@ lemma run_callCC (f : MonadCont.Label α (FreeCont r) β → FreeCont r α) (k :
...
+@[simp] theorem run_withReader (f : σ → σ) (m : FreeReader σ α) (s : σ) :
+    run (withReaderFree f m) s = run m (f s) := by
+    induction m generalizing s with
+    | pure a => rfl
+    | liftBind op cont ih =>
+      cases op; apply ih
+
+/-- Read the environment and return `g` applied to it. -/
+def asks (g : σ → α) : FreeReader σ α :=
+  (read : FreeReader σ σ) >>= fun s => pure (g s)
```
eric-wieser: Isnt't this just `g <$> read`? Why does this need API?

Cslib/Foundations/Control/Monad/Free/Effects.lean
```diff
@@ -353,6 +374,93 @@ lemma run_callCC (f : MonadCont.Label α (FreeCont r) β → FreeCont r α) (k :
...
+@[simp]
+lemma run_read (k : σ → FreeReader σ α) (s₀ : σ) :
+    run (liftBind .read k) s₀ = run (k s₀) s₀ := rfl
+
+/-- Transform the environment seen by all `read`s in a computation -/
+def withReaderFree (f : σ → σ) : FreeReader σ α → FreeReader σ α
+  | .pure a => .pure a
+  | .liftBind .read cont =>
+      let step : σ → FreeReader σ α := fun s => withReaderFree f (cont (f s))
+      .liftBind .read step
```
eric-wieser: I think it's better to inline this into the `WithReaderOf` instance below, and then write your theorems about the global `withReaderOf`.

Cslib/Foundations/Control/Monad/Free/Effects.lean
```diff
@@ -353,6 +381,89 @@ lemma run_callCC (f : MonadCont.Label α (FreeCont r) β → FreeCont r α) (k :
...
+@[simp]
+lemma run_pure (a : α) (s₀ : σ) :
+    run (.pure a : FreeReader σ α) s₀ = a := rfl
+
+@[simp]
+lemma run_read (k : σ → FreeReader σ α) (s₀ : σ) :
+    run (liftBind .read k) s₀ = run (k s₀) s₀ := rfl
+
+instance instMonadWithReaderOf : MonadWithReaderOf σ (FreeReader σ) where
+  withReader := fun {α} f m =>
```
eric-wieser:
```suggestion
  withReader {α} f m :=
```

Cslib/Foundations/Control/Monad/Free/Effects.lean
tannerduve: Not definitionally equal, `censor` rewrites the final accumulated log and `censorHom` rewrites each tell node in the syntax tree

They do agree on `run` (`run_censorHom_eq_run_censor` at line 295). I kept both since `censor` is standard in the `MonadWriter` API (eg. in [Haskell](https://hackage-content.haskell.org/package/mtl-2.3.2/docs/Control-Monad-Writer-Class.html)), and `censorHom` is a “map every tell” operation that may be useful.

Cslib/Foundations/Control/Monad/Free/Effects.lean
```diff
@@ -353,6 +381,89 @@ lemma run_callCC (f : MonadCont.Label α (FreeCont r) β → FreeCont r α) (k :
...
+    simpa [withTheReader, instMonadWithReaderOf, run] using (ih (f s) s)
+
+/-- Read the environment and return `g` applied to it. -/
+def asks (g : σ → α) : FreeReader σ α :=
+  g <$> (read : FreeReader σ σ)
+
+@[simp] lemma asks_run (g : σ → α) (s : σ) :
+    run (asks (σ := σ) (α := α) g) s = g s := by
+  simp [asks, run]
+
```
eric-wieser: I'd suggest just deleting this. If the `ReaderT` monad in Lean itself doesn't need this and no one has asked for it, I'm not sure why it would be useful here.

Cslib/Foundations/Control/Monad/Free/Effects.lean
```diff
@@ -265,6 +261,41 @@ instance [Monoid ω] : MonadWriter ω (FreeWriter ω) where
...
+  let (a, w) := run m
+  liftBind (.tell (f w)) (fun _ => .pure a)
+
+/--
+Map each `tell w` to `tell (φ w)`, where `φ` is a monoid homomorphism.
+
+This is a syntactic transformation on the computation tree: it preserves the original sequencing
+of `tell`s, rewriting each one in place (rather than aggregating the log before rewriting).
+-/
+def censorHom [Monoid ω] (φ : ω →* ω) : FreeWriter ω α → FreeWriter ω α
```
eric-wieser: Why does `φ` need to be a hom? Does the definition use this fact?

Cslib/Foundations/Control/Monad/Free/Effects.lean
```diff
@@ -265,6 +261,41 @@ instance [Monoid ω] : MonadWriter ω (FreeWriter ω) where
   listen := listen
   pass := pass
 
+/--
+Rewrite the log produced by `m` using `f`.
+
+This evaluates `m` via `run` and then re-emits a single `tell` containing the rewritten log.
+-/
+def censor [Monoid ω] (f : ω → ω) (m : FreeWriter ω α) : FreeWriter ω α :=
```
eric-wieser: Don't you want this to work for
```suggestion
def censor [Monoid ω] (f : ω → ω') (m : FreeWriter ω α) : FreeWriter ω α :=
```

Cslib/Foundations/Control/Monad/Free/Effects.lean
```diff
@@ -265,6 +261,41 @@ instance [Monoid ω] : MonadWriter ω (FreeWriter ω) where
...
+
+@[simp] theorem run_censorHom [Monoid ω] (φ : ω →* ω) (m : FreeWriter ω α) :
+    run (censorHom (α := α) φ m) = (let (a, w) := run m; (a, φ w)) := by
+  induction m with
+  | pure a => simp [censorHom, run]
+  | liftBind op k ih =>
+    cases op; simp [censorHom, run, ih, map_mul]
+
+theorem run_censorHom_eq_run_censor [Monoid ω] (φ : ω →* ω) (m : FreeWriter ω α) :
+    run (censorHom (α := α) φ m) = run (censor (α := α) (fun w => φ w) m) := by
```
eric-wieser: If running the two give equal results, do we really need both?

Cslib/Foundations/Control/Monad/Free/Effects.lean
eric-wieser: In mathlib this is called `WriterT.adapt`

Cslib/Foundations/Control/Monad/Free/Effects.lean
eric-wieser: If `censor` is standard in the `MonadWriter` API, then you should probably write it for `[MonadWriter ..]` not `FreeWriter`, that way you get it here for free.

tannerduve:
@eric-wieser i added `censor` because it is in the writer monad API in haskell, but it is probably not needed here since its not in the lean API

Since this PR is meant to be about adding the reader monad to the free effects library, I think I'll just remove the censor stuff for now

fmontesi (approved):

PR #249, opened 2026-01-07
feat: prove that an omega-language is regular iff it is a finite union of omega-languages of a special form

ctchou:
This PR proves that an ω-language is regular iff it is the finite union of ω-languages of the form `L * M^ω`, where all `L`s and `M`s are regular languages.  In addition to being of independent interest, thiss result will also be used in proving that ω-regular languages are closed under complementation, where the complement of an ω-regular language will be given as a finite union of this form.  As part of this PR, we prove the theorem `LTS.ωTr.flatten`, which says that an infinite sequence of finite executions of a LTS connecting an infinite sequence of intermediate states can be concatenated into an infinite execution of the LTS.  We also use this theorem to simplify the proof of `loop_run_exists` in Loop.lean.


ctchou:
Rebase on https://github.com/leanprover/cslib/pull/241.

ctchou:
Rebase on https://github.com/leanprover/cslib/pull/241.

ctchou:
Rebase on #241 .

ctchou:
Rebase on #241, which includes the move to the new module system.

ctchou:
Rebase on upstream/main.

Cslib/Computability/Automata/NA/Pair.lean
```diff
@@ -0,0 +1,81 @@
...
+theorem LTS.mem_pairLang (lts : LTS State Symbol) (s t : State) (xl : List Symbol) :
+    xl ∈ (lts.pairLang s t) ↔ lts.MTr s xl t := Iff.rfl
+
+/-- `LTS.pairLang s t` is a regular language if there are only finitely many states. -/
+@[simp]
+theorem LTS.pairLang_regular [Finite State] (lts : LTS State Symbol) (s t : State) :
+    (lts.pairLang s t).IsRegular := by
+  rw [IsRegular.iff_nfa]
+  use State, inferInstance, (NA.FinAcc.mk ⟨lts, {s}⟩ {t})
+  ext xl; simp [LTS.mem_pairLang]
```
chenson2018: Either of these works:

```suggestion
  ext
  simp
```
or

```suggestion
  ext xl
  grind [NA.FinAcc.instAcceptor]
```
(is this a missing `grind` annotation maybe?)

Cslib/Computability/Automata/NA/Pair.lean
```diff
@@ -0,0 +1,81 @@
...
+theorem language_eq_fin_iSup_hmul_omegaPow
+    [Inhabited Symbol] [Finite State] (na : NA.Buchi State Symbol) :
+    language na = ⨆ s ∈ na.start, ⨆ t ∈ na.accept, (na.pairLang s t) * (na.pairLang t t)^ω := by
+  ext xs
+  simp only [NA.Buchi.instωAcceptor, ωAcceptor.mem_language,
+    ωLanguage.mem_iSup, ωLanguage.mem_hmul, LTS.mem_pairLang]
+  constructor
+  · rintro ⟨ss, h_run, h_inf⟩
+    obtain ⟨t, _, h_t⟩ := frequently_in_finite_type.mp h_inf
+    refine ⟨ss 0, by grind [h_run.start], t, by assumption, ?_⟩
```
chenson2018: Two things here:
- when it is very easy to replace `assumption` with the term I prefer this, for both readability and because I think sometimes the check for definitional equality can be expensive
- Using `h_run.start` is a bit odd, because `grind` should handle projecting out of structures

```suggestion
    obtain ⟨t, tmem, h_t⟩ := frequently_in_finite_type.mp h_inf
    use ss 0, by grind [NA.Run], t, tmem
```

Cslib/Computability/Automata/NA/Pair.lean
```diff
@@ -0,0 +1,81 @@
...
+    obtain ⟨zls, rfl, h_zls⟩ := mem_omegaPow.mp h_zs
+    let ts := ωSequence.const t
+    have h_mtr (n : ℕ) : na.MTr (ts n) (zls n) (ts (n + 1)) := by
+      grind [Language.mem_sub_one, LTS.mem_pairLang]
+    have h_pos (n : ℕ) : (zls n).length > 0 := by
+      grind [Language.mem_sub_one, List.eq_nil_iff_length_eq_zero]
+    obtain ⟨zss, h_zss, _⟩ := LTS.ωTr.flatten h_mtr h_pos
+    have (n : ℕ) : zss (zls.cumLen n) = t := by grind
+    obtain ⟨xss, _, _, _, _⟩ := LTS.ωTr.append h_yl h_zss (by grind [cumLen_zero (ls := zls)])
+    refine ⟨xss, by grind [NA.Run.mk], ?_⟩
```
chenson2018:
```suggestion
    use xss, by grind [NA.Run]
```

Cslib/Computability/Automata/NA/Pair.lean
```diff
@@ -0,0 +1,81 @@
...
+  ext xs
+  simp only [NA.Buchi.instωAcceptor, ωAcceptor.mem_language,
+    ωLanguage.mem_iSup, ωLanguage.mem_hmul, LTS.mem_pairLang]
+  constructor
+  · rintro ⟨ss, h_run, h_inf⟩
+    obtain ⟨t, _, h_t⟩ := frequently_in_finite_type.mp h_inf
+    refine ⟨ss 0, by grind [h_run.start], t, by assumption, ?_⟩
+    obtain ⟨f, h_mono, h_f⟩ := frequently_iff_strictMono.mp h_t
+    refine ⟨xs.take (f 0), ?_, xs.drop (f 0), ?_, by grind⟩
+    · grind [extract_eq_drop_take, LTS.ωTr_mTr h_run.trans (Nat.zero_le (f 0))]
```
chenson2018: I think this is a bit more clear if you split it up and pass off arithmetic to `grind`:

```suggestion
    · have : na.MTr (ss 0) (xs.extract 0 (f 0)) (ss (f 0)) := by grind [LTS.ωTr_mTr, NA.Run]
      grind [extract_eq_drop_take]
```

Cslib/Computability/Automata/NA/Pair.lean
```diff
@@ -0,0 +1,81 @@
...
+    · simp only [omegaPow_seq_prop, LTS.mem_pairLang]
+      refine ⟨(f · - f 0), by grind [Nat.base_zero_strictMono], by simp, ?_⟩
+      intro n
+      grind [extract_drop, h_mono.monotone (Nat.zero_le n), h_mono.monotone (Nat.zero_le (n + 1)),
+        LTS.ωTr_mTr h_run.trans <| h_mono.monotone (show n ≤ n + 1 by grind)]
```
chenson2018: Here's what I came up with for readability:

```suggestion
      use (f · - f 0)
      split_ands
      · grind [Nat.base_zero_strictMono]
      · grind
      · intro n
        have mono_f (k : ℕ) : f 0 ≤ f (n + k) := h_mono.monotone (by grind)
        grind [extract_drop, mono_f 0, LTS.ωTr_mTr h_run.trans <| h_mono.monotone (?_ : n ≤ n + 1)]
```

In general, I think we are doing something wrong with `grind` if we're ever writing something as low level as `Nat.zero_le`.

Cslib/Computability/Automata/NA/Pair.lean
```diff
@@ -0,0 +1,81 @@
...
+  use State, inferInstance, (NA.FinAcc.mk ⟨lts, {s}⟩ {t})
+  ext xl; simp [LTS.mem_pairLang]
+
+namespace NA.Buchi
+
+open Set Filter ωSequence ωLanguage ωAcceptor
+
+/-- The ω-language accepted by a finite-state Büchi automaton is the finite union of ω-languages
+of the form `L * M^ω`, where all `L`s and `M`s are regular languages. -/
+theorem language_eq_fin_iSup_hmul_omegaPow
```
chenson2018, line 46: If you try to open the LTS scope here, `grind` times out on some proofs, which is a bit at odds with us right above adding an annotation for  `LTS.mem_pairLang`.

Soon there will be an equivalent for `simp` sets for `grind`, where you can specify custom sets of `grind` rules as opposed to having everything attached to scopes. I think this could help separate out what I think are problems with the LTS annotations being too costly. It would be helpful for me if you can think a bit about what `grind` sets you might like for automata once core releases this feature, e.g. one set for all of automata or if more refinement is helpful.

Cslib/Computability/Automata/NA/Pair.lean
```diff
@@ -0,0 +1,81 @@
...
+def LTS.pairLang (lts : LTS State Symbol) (s t : State) : Language Symbol :=
+  { xl | lts.MTr s xl t }
+
+@[simp, scoped grind =]
+theorem LTS.mem_pairLang (lts : LTS State Symbol) (s t : State) (xl : List Symbol) :
+    xl ∈ (lts.pairLang s t) ↔ lts.MTr s xl t := Iff.rfl
+
+/-- `LTS.pairLang s t` is a regular language if there are only finitely many states. -/
+@[simp]
+theorem LTS.pairLang_regular [Finite State] (lts : LTS State Symbol) (s t : State) :
```
chenson2018: It seems these should be in the `LTS` module.

Cslib/Foundations/Semantics/LTS/Basic.lean
```diff
@@ -264,17 +264,54 @@ theorem LTS.ωTr.cons (htr : lts.Tr s μ t) (hωtr : lts.ωTr ss μs) (hm : ss 0
...
 /-- Prepends an infinite execution with a finite execution. -/
-theorem LTS.ωTr.append (hmtr : lts.MTr s μl t) (hωtr : lts.ωTr ss μs)
-    (hm : ss 0 = t) : ∃ ss', lts.ωTr ss' (μl ++ω μs) ∧ ss' 0 = s ∧ ss' μl.length = t := by
+theorem LTS.ωTr.append
+    (hmtr : lts.MTr s μl t) (hωtr : lts.ωTr ss μs) (hm : ss 0 = t) :
+    ∃ ss', lts.ωTr ss' (μl ++ω μs) ∧ ss' 0 = s ∧ ss' μl.length = t ∧ ss'.drop μl.length = ss := by
   obtain ⟨sl, _, _, _, _⟩ := LTS.MTr.exists_states hmtr
-  refine ⟨sl ++ω ss.drop 1, ?_, by grind [get_append_left], by grind [get_append_left]⟩
+  refine ⟨sl.take μl.length ++ω ss, ?_,
+    by grind [get_append_left], by grind [get_append_left], by grind [drop_append_of_ge_length]⟩
```
chenson2018: My rule of thumb would be if it doesn't fit on a single line and you have a nicely ordered conjunction as your goal, do

```suggestion
  use sl.take μl.length ++ω ss
  split_ands
```
and then a bullet for each proof.

Cslib/Foundations/Semantics/LTS/Basic.lean
```diff
@@ -264,17 +264,54 @@ theorem LTS.ωTr.cons (htr : lts.Tr s μ t) (hωtr : lts.ωTr ss μs) (hm : ss 0
...
+  · have h1 : segs.cumLen (segment segs.cumLen n + 1) = n + 1 := by
+      grind [segment_upper_bound h_mono h_zero n]
+    have h2 : segment segs.cumLen (n + 1) = segment segs.cumLen n + 1 := by
+      simp [← h1, segment_idem h_mono]
+    have : n + 1 - segs.cumLen (segment segs.cumLen n) = (μls (segment segs.cumLen n)).length := by
+      grind
+    have h3 : ts (segment segs.cumLen n + 1) =
+        (sls (segment segs.cumLen n))[n + 1 - segs.cumLen (segment segs.cumLen n)]! := by
+      grind
+    simp [h1, h2, h_seg0, h3]
```
chenson2018, line 373: Is this file very slow to edit for you as well? This is all the proof golfing I could stand to do:

```suggestion
    simp [h1, h2, h_seg0]
```

Cslib/Foundations/Semantics/LTS/Basic.lean
```diff
@@ -264,17 +264,54 @@ theorem LTS.ωTr.cons (htr : lts.Tr s μ t) (hωtr : lts.ωTr ss μs) (hm : ss 0
   induction i <;> grind
 
 /-- Prepends an infinite execution with a finite execution. -/
-theorem LTS.ωTr.append (hmtr : lts.MTr s μl t) (hωtr : lts.ωTr ss μs)
-    (hm : ss 0 = t) : ∃ ss', lts.ωTr ss' (μl ++ω μs) ∧ ss' 0 = s ∧ ss' μl.length = t := by
+theorem LTS.ωTr.append
+    (hmtr : lts.MTr s μl t) (hωtr : lts.ωTr ss μs) (hm : ss 0 = t) :
+    ∃ ss', lts.ωTr ss' (μl ++ω μs) ∧ ss' 0 = s ∧ ss' μl.length = t ∧ ss'.drop μl.length = ss := by
```
chenson2018, line 326: Should this new condition have been here from the initial addition of this theorem?

chenson2018 (changes requested):

Cslib/Computability/Automata/NA/Pair.lean
ctchou: Fixed with the 1st option.

Cslib/Computability/Automata/NA/Pair.lean
ctchou: Fixed.

Cslib/Computability/Automata/NA/Pair.lean
ctchou: Fixed.

Cslib/Computability/Automata/NA/Pair.lean
ctchou: Fixed.

Cslib/Computability/Automata/NA/Pair.lean
ctchou: Fixed.

Cslib/Computability/Automata/NA/Pair.lean
ctchou: I'm not even sure what "simp sets" are.  Is there an example I can look at?

Cslib/Computability/Automata/NA/Pair.lean
ctchou: Currently `LTS` does not depend on `Language`.  I'm not sure we want to add that dependency, since `LTS` can potentially be used by many other things than automata theories.  The notion of `pairLang` is really specific to `NA`.  The only reason I put it in the `LTS` namespace is that I don't want to drag in the `NA.start` states, which are not used here.  Also, I will shortly define `pairAccLang` that will depend on a set of accepting states.  I would like to keep both definitions in the same file under `Automata.NA`.

Cslib/Computability/Automata/NA/Pair.lean
chenson2018: Ah sorry. There is documentation [here](https://lean-lang.org/doc/reference/latest/The-Simplifier/Simp-sets/) and you can find some examples in Mathlib.

It is just a way of organizing `simp` rules past them all being in one default set. So when this feature is released we can specify which `grind` rules should run for automata. The benefit should be nicer syntax because you'll have less `open scoped` and performance by not having to bring in every annotation that `LTS` uses.

Cslib/Foundations/Semantics/LTS/Basic.lean

Cslib/Computability/Automata/NA/Pair.lean
chenson2018: I see, that makes sense given the current module structure. We might revisit if we split up the `LTS` module later down the road.

Cslib/Foundations/Semantics/LTS/Basic.lean
ctchou: I would like to keep `h3` even if it is not strictly needed.  This is a tricky proof and I'd like to keep all the intermediate results so that I will know in the future what the key facts are.

And, yes, it took me quite a while to get `grind` under control.  This proof should be essentially the same as the one in `Loop.lean` which is now a corollary of this theorem.  But for reasons I don't understand, `grind` behaved much more badly here than in the old proof in `Loop.lean`.

More generally, I really wish I knew how to define and reason about `OmegaSequence.flatten` without using `Nat.segment`, which is a pain to work with.  But, alas, I don't know how.

Cslib/Foundations/Semantics/LTS/Basic.lean
ctchou: Not sure what exactly you mean.  The new conjunct `ss'.drop μl.length = ss` asserts that `ss'` has `ss` as an infinite suffix.  Since `ss` is among the antecedents of this theorem, this conjunct "anchors" the `ss'` asserted to exist by the theorem by fixing its suffix.

Cslib/Computability/Automata/NA/Loop.lean
```diff
@@ -120,30 +120,12 @@ the state `()` marks the boundaries between the finite words in `xls`. -/
...
-    by_cases h_n : n + 1 < segs.cumLen (segment segs.cumLen n + 1)
-    · grind [segment_range_val h_mono (by grind) h_n]
-    · have h1 : segs.cumLen (segment segs.cumLen n + 1) = n + 1 := by
-        grind [segment_upper_bound h_mono h_zero n]
-      have h2 : segment segs.cumLen (n + 1) = segment segs.cumLen n + 1 := by
-        simp [← h1, segment_idem h_mono]
-      simp [h1, h2, h_seg0]
-      grind
-  · simp [h_len, flatten_def, segment_idem h_mono, h_seg0]
+  let ts := ωSequence.const (inl () : Unit ⊕ State)
```
chenson2018, line 123: I meant to say earlier, this is a very nice golf!

Cslib/Foundations/Semantics/LTS/Basic.lean
chenson2018: What I meant is that this theorem was here before this pull request without this conjunct, and a proof downstream (`LTS.Total.mTr_ωTr`) seems to now ignore this addition. The question here is if it is important that they are all true at the same time, or if they can be split up to some extent into separate theorems.

ctchou:
Rebase on upstream/main to incorporate #258.

Cslib/Foundations/Semantics/LTS/Basic.lean
ctchou: I understand your question now.  This change was made because when `LTS.ωTr.append` was used in proving `language_eq_fin_iSup_hmul_omegaPow` in Pair.lean, I found that the original condition `ss' μl.length = t` was not strong enough.  Note that the new condition `ss'.drop μl.length = ss` implies the old condition `ss' μl.length = t`.

chenson2018: I see, that clears it up. Thanks!

Cslib/Computability/Automata/NA/Loop.lean
ctchou: Thanks!

ctchou:
Merge in the current upstream/main.

PR #252, opened 2026-01-08
feat: correctness of eta-expansion for CLL

fmontesi:
Proves that eta-expansion (Proposition.expand) for CLL is correct, i.e., all axiom applications in the returned proof treat only atomic propositions.


chenson2018 (approved):
I fixed a lint issue, everything else looks good.

PR #258, opened 2026-01-11
feat: add notion of finite execution for LTS

fmontesi:
This PR adds a notion of finite execution for LTS analogous to the infinite counterpart.

### DIFF
--- a/Cslib/Foundations/Semantics/LTS/Basic.lean
+++ b/Cslib/Foundations/Semantics/LTS/Basic.lean
@@ -104,7 +104,6 @@ theorem LTS.MTr.single {s1 : State} {μ : Label} {s2 : State} :
   · apply LTS.MTr.refl
 
 /-- Any multistep transition can be extended by adding a transition. -/
-@[scoped grind <=]
 theorem LTS.MTr.stepR {s1 : State} {μs : List Label} {s2 : State} {μ : Label} {s3 : State} :
   lts.MTr s1 μs s2 → lts.Tr s2 μ s3 → lts.MTr s1 (μs ++ [μ]) s3 := by
   intro h1 h2
@@ -162,12 +161,10 @@ theorem LTS.IsExecution.refl (lts : LTS State Label) (s : State) : lts.IsExecuti
   grind
 
 /-- Equivalent of `MTr.stepL` for executions. -/
-@[scoped grind →]
 theorem LTS.IsExecution.stepL {lts : LTS State Label} (htr : lts.Tr s1 μ s2)
     (hexec : lts.IsExecution s2 μs s3 ss) : lts.IsExecution s1 (μ :: μs) s3 (s1 :: ss) := by grind
 
 /-- Deconstruction of executions with `List.cons`. -/
-@[scoped grind →]
 theorem LTS.isExecution_cons_invert (h : lts.IsExecution s1 (μ :: μs) s2 (s1 :: ss)) :
     lts.IsExecution (ss[0]'(by grind)) μs s2 ss := by
   obtain ⟨_, _, _, h4⟩ := h
@@ -538,7 +535,6 @@ theorem LTS.mem_setImage {lts : LTS State Label} :
   simp only [setImage, Set.mem_iUnion, exists_prop]
   grind
 
-@[scoped grind →]
 theorem LTS.tr_setImage {lts : LTS State Label} (hs : s ∈ S) (htr : lts.Tr s μ s') :
   s' ∈ lts.setImage S μ := by grind
 
@@ -669,7 +665,6 @@ theorem LTS.saturate_tr_sTr [HasTau Label] {lts : LTS State Label} :
   lts.saturate.Tr = lts.STr := by rfl
 
 /-- Any transition is also a saturated transition. -/
-@[scoped grind →]
 theorem LTS.STr.single [HasTau Label] (lts : LTS State Label) :
     lts.Tr s μ s' → lts.STr s μ s' := by
   intro h
--- a/Cslib/Computability/Automata/NA/Loop.lean
+++ b/Cslib/Computability/Automata/NA/Loop.lean
@@ -103,7 +103,7 @@ theorem loop_fin_run_exists {xl : List Symbol} (h : xl ∈ language na) :
   · use [inl ()]
     grind
   · use [inl ()] ++ (sl.extract 1 xl.length).map inr ++ [inl ()]
-    grind [FinAcc.loop]
+    grind [FinAcc.loop, → LTS.tr_setImage]
 
 /-- For any finite word in `language na`, there is a corresponding multistep transition
 of `na.loop`. -/


Cslib/Foundations/Semantics/LTS/Basic.lean
```diff
@@ -140,20 +140,71 @@ theorem LTS.MTr.nil_eq (h : lts.MTr s1 [] s2) : s1 = s2 := by
...
+
+/-- Every execution has a start state. -/
+@[scoped grind →]
+theorem LTS.isExecution_nonEmpty_states (h : lts.IsExecution ss μs) : ss ≠ [] := by grind
+
+/-- Every state has an execution of zero steps terminating in itself. -/
+@[scoped grind ⇒]
+theorem LTS.IsExecution.refl (lts : LTS State Label) (s : State) :
+    lts.IsExecution [s] [] := by
+  exists (by grind)
```
chenson2018: I don't have an alternative to suggest, but I really dislike this pattern that using a propositional existential for indexing causes. I think it'd be worth minimizing and asking in the general Zulip channel if there's a better alternative to this.

chenson2018:
There are lots of `grind` lints, but they are very close to the threshold, so I'm not overly concerned.

Cslib/Foundations/Semantics/LTS/Basic.lean
```diff
@@ -140,20 +140,71 @@ theorem LTS.MTr.nil_eq (h : lts.MTr s1 [] s2) : s1 = s2 := by
...
-/-- For every multistep transition, there exists a sequence of intermediate states
-which satisfies the single-step transition at every step. -/
-theorem LTS.MTr.exists_states {lts : LTS State Label} {s1 s2 : State} {μs : List Label}
-    (h : lts.MTr s1 μs s2) : ∃ ss : List State, ∃ _ : ss.length = μs.length + 1,
-    ss[0] = s1 ∧ ss[μs.length] = s2 ∧ ∀ k, ∀ _ : k < μs.length, lts.Tr ss[k] μs[k] ss[k + 1] := by
+/-- A finite execution, or sequence of transitions. -/
+@[scoped grind =]
+def LTS.IsExecution (lts : LTS State Label)
+    (ss : List State) (μs : List Label) : Prop :=
+  ∃ hlen : ss.length = μs.length + 1, ∀ k, {valid : k < μs.length} → lts.Tr ss[k] μs[k] ss[k + 1]
```
ctchou: My main comment about this definition is that the beginning state `s1` and ending state `s2` of `ss` should be among the arguments of `LTS.IsExecution` and the constraints `ss[0] = s1` and `ss[μs.length] = s2` should be included in this definition.  The consequence of not doing so is that they have to be repeated again and again in the theorems I pointed out below.  It is particularly ugly that the condition `ss.length = μs.length + 1` has to be repeated several times below, only because it is needed to discharge the list indexing proof obligations.

Do we really need to give names `hlen` and `valid` to the two subexpressions?  I've checked that replacing both by `_` still works and I can't think of a use case where we actually need those names.

Also, is there any advantage in using `{...}` instead of `(...)` for `_ : k < μs.length`?

Cslib/Foundations/Semantics/LTS/Basic.lean
ctchou: A single `grind` proves this theorem.

Cslib/Foundations/Semantics/LTS/Basic.lean
```diff
@@ -140,20 +140,71 @@ theorem LTS.MTr.nil_eq (h : lts.MTr s1 [] s2) : s1 = s2 := by
...
+  exists (by grind)
+  intro k valid
+  specialize h2 k <;> grind
+
+open scoped LTS.IsExecution in
+/-- A multistep transition implies the existence of an execution. -/
+@[scoped grind →]
+theorem LTS.mTr_isExecution {lts : LTS State Label} {s1 : State} {μs : List Label}
+    {s2 : State} (h : lts.MTr s1 μs s2) : ∃ ss : List State, ∃ _ : ss.length = μs.length + 1,
+    ss[0] = s1 ∧ ss[ss.length - 1] = s2 ∧ lts.IsExecution ss μs := by
```
ctchou: You should have included `s1` and `s2` among the arguments of `LTS.IsExecution`.  Then you don't need to keep repeating the `ss.length = μs.length + 1` condition here.

Cslib/Foundations/Semantics/LTS/Basic.lean
```diff
@@ -140,20 +140,71 @@ theorem LTS.MTr.nil_eq (h : lts.MTr s1 [] s2) : s1 = s2 := by
...
-    use [t1] ++ ss'
+  case stepL t1 μ t2 μs t3 htr hmtr ih =>
+    obtain ⟨ss', hlen, ih⟩ := ih
+    use t1 :: ss'
     grind
 
+/-- Converts an execution into a multistep transition. -/
+@[scoped grind →]
+theorem LTS.isExecution_mTr (hexec : lts.IsExecution ss μs) :
+    lts.MTr (ss[0]'(by grind)) μs (ss[ss.length - 1]'(by grind)) := by
```
ctchou: Again, if you have included the beginning and ending states of `ss` among the arguments of `LTS.IsExecution`, then you wouldn't have to invoke `grind` in the statement of the theorem.

Cslib/Foundations/Semantics/LTS/Basic.lean
```diff
@@ -140,20 +140,71 @@ theorem LTS.MTr.nil_eq (h : lts.MTr s1 [] s2) : s1 = s2 := by
...
+  induction ss generalizing μs
+  case nil => grind
+  case cons s1 ss ih =>
+    cases μs <;> grind
+
+/-- Correspondence of multistep transitions and executions. -/
+@[scoped grind =]
+theorem LTS.mTr_isExecution_iff : lts.MTr s1 μs s2 ↔
+    ∃ ss : List State, ∃ _ : ss.length = μs.length + 1,
+    ss[0] = s1 ∧ ss[ss.length - 1] = s2 ∧ lts.IsExecution ss μs := by
```
ctchou: Ditto.  If you have included `s1` and `s2` among the arguments of `LTS.IsExecution`, then you don't need to talk about `length` at all.  The statement of the theorem will simply be something like:
```
lts.MTr s1 μs s2 ↔ ∃ ss, lts.IsExecution s1 μs s2 ss 
```

ctchou (changes requested):
Please include the beginning and ending states of `ss` among the arguments of `LTS.IsExecution`.  That will make the statements of several theorems shorter and prettier.  My personal preference of the argument order is:
```
lts.IsExecution s1 μs s2 ss
```
which makes `LTS.IsExecution` just an extension of `LTS.MTr` by one additional argument.

Also, this PR will interact with my PRs #241 and #249.  So we need to figure out how they are sequenced.

Cslib/Foundations/Semantics/LTS/Basic.lean
fmontesi: Correct, I had a dumber set of lemmas in a previous version.

Cslib/Foundations/Semantics/LTS/Basic.lean
fmontesi: hlen and valid are remnants of a previous version, removed.

Re the start and final states: I'm gonna try a bit to see if I can remove the disadvantage you mention without introducing start and final states in the signature. If possible, I'd like to keep the signature as close as possible to the omega-variant. The only reason for having them is formalisation convenience (which I agree shouldn't be sacrificed *too much*, but let's see..).

fmontesi: The reason for `{...}` is to try and tell Lean that that part should be figured out automatically from the context. We can always revert it if we don't find it very useful.

Ah, now I remember, that one had a valid reason for being called `valid` (pun intended): it's so that we can instantiate it directly when necessary by using the `(valid := ...)` syntax.

ctchou: I view `LTS.IsExecution` as a natural extension of `LTS.MTr` that provides a witness to the execution path.  There is no advantage in trying to shoehorn it to match `LTS.IsOmegaExecution`.  It just creates more contorted code.

fmontesi: Ah, but that (valid := ...) might not be too useful since it's not a formal param of the def... removed for now.

fmontesi: Ah, seeing it as an extension of `MTr` kinda makes that signature more palatable to me for some reason. Ok, then, I'll add the params.

Btw, for memory, I've tried using head? and getLast? instead of indexes and it still feels like shoehorning.

fmontesi:
I think it's pretty usable now and I've integrated the major comments, so we can discuss it again as a proper PR.
CI is failing because of grind lints, I think.. @chenson2018 any ideas?

ctchou:
I merged in the latest `upstream/main` and fixed the `CslibTests.GrindLint` errors in the following branch:

https://github.com/ctchou/cslib/tree/lts-execution-prop-ctchou

The changes I made are the last two commits.

A few words about how I fixed the `GrintLint` errors: I simply removed the `grind` annotations of the offending theorems.  In all cases except one, `lake build` continued to work.  In that one exception (the theorem `LTS.tr_setImage`), a `grind` in `Loop.lean` failed.  I fixed it by passing the removed `grind` annotation to the `grind` call explicitly.

ctchou (approved):
LGTM.

fmontesi:
Think I got it now. @chenson2018 can you tick the review box? :)

chenson2018 (approved):

PR #265, opened 2026-01-14
feat: define right congruence and prove its basic properties

ctchou:
This PR has been absorbed into #278.

> This PR defines the notion of a "right congruence", which is an equivalence relation between finite words that is preserved by concatenation on the right, and proves its basic properties:
> - There is a natural deterministic automaton corresponding to each right congruence whose states are the equivalence classes of the congruence and whose transitions correspond to concatenation on the right of the input symbols.
> - If a right congruence is of finite index, then each of its equivalence classes is a regular language.
> 
> We will eventually prove the closure of omega-regular languages under complementation by constructing a right congruence of finite index with certain special properties.
> 
> Right congruences are also intimately related to the Myhill-Nerode theorem, but currently we do not plan to pursue that direction.

### DIFF
--- a/scripts/bench/build/README.md
+++ b/scripts/bench/build/README.md
@@ -22,3 +22,6 @@ The following metrics are collected individually for each module:
 
 - `build/module/<name>//lines`
 - `build/module/<name>//instructions`
+
+If the file `build_upload_lakeprof_report` is present in the repo root,
+the lakeprof report will be uploaded once the benchmark run concludes.
--- a/scripts/bench/build/lakeprof_report_template.html
+++ b/scripts/bench/build/lakeprof_report_template.html
@@ -1,0 +1,74 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="UTF-8" />
+    <title>Lakeprof Report</title>
+  </head>
+  <body>
+    <h1>Lakeprof Report</h1>
+    <button type="button" id="btn_fetch">View build trace in Perfetto</button>
+    <pre><code>__LAKEPROF_REPORT__</code></pre>
+
+    <script type="text/javascript">
+      // https://perfetto.dev/docs/visualization/deep-linking-to-perfetto-ui
+      // https://gist.github.com/chromy/170c11ce30d9084957d7f3aa065e89f8
+
+      const BASE_URL = __BASE_URL__;
+      const ORIGIN = "https://ui.perfetto.dev";
+
+      const btnFetch = document.getElementById("btn_fetch");
+
+      async function fetchAndOpen(traceUrl) {
+        const resp = await fetch(traceUrl);
+        // Error checking is left as an exercise to the reader.
+        const blob = await resp.blob();
+        const arrayBuffer = await blob.arrayBuffer();
+        openTrace(arrayBuffer, traceUrl);
+      }
+
+      function openTrace(arrayBuffer, traceUrl) {
+        const win = window.open(ORIGIN);
+        if (!win) {
+          btnFetch.style.background = "#f3ca63";
+          btnFetch.onclick = () => openTrace(arrayBuffer);
+          btnFetch.innerText =
+            "Popups blocked, click here to open the trace file";
+          return;
+        }
+
+        const timer = setInterval(() => win.postMessage("PING", ORIGIN), 50);
+
+        const onMessageHandler = (evt) => {
+          if (evt.data !== "PONG") return;
+
+          // We got a PONG, the UI is ready.
+          window.clearInterval(timer);
+          window.removeEventListener("message", onMessageHandler);
+
+          const reopenUrl = new URL(location.href);
+          reopenUrl.hash = `#reopen=${traceUrl}`;
+          win.postMessage(
+            {
+              perfetto: {
+                buffer: arrayBuffer,
+                title: "Lake Build Trace",
+                url: reopenUrl.toString(),
+              },
+            },
+            ORIGIN
+          );
+        };
+
+        window.addEventListener("message", onMessageHandler);
+      }
+
+      // This is triggered when following the link from the Perfetto UI's sidebar.
+      if (location.hash.startsWith("#reopen=")) {
+        const traceUrl = location.hash.substr(8);
+        fetchAndOpen(traceUrl);
+      }
+
+      btnFetch.onclick = () => fetchAndOpen(`${BASE_URL}/lakeprof.trace_event`);
+    </script>
+  </body>
+</html>
--- a/scripts/bench/build/lakeprof_report_upload.py
+++ b/scripts/bench/build/lakeprof_report_upload.py
@@ -1,0 +1,44 @@
+#!/usr/bin/env python3
+
+import json
+import subprocess
+import sys
+from pathlib import Path
+
+
+def run(*args: str) -> None:
+    subprocess.run(args, check=True)
+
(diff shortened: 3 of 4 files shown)


ctchou:
Rebase on upstream/main.

Cslib/Computability/Automata/DA/Congr.lean
```diff
@@ -0,0 +1,68 @@
...
+
+/-! # Deterministic automaton corresponding to a right congruence. -/
+
+namespace Cslib
+
+open scoped FLTS RightCongruence
+
+variable {Symbol : Type*}
+
+/-- Every right congruence gives rise to a DA whose states are the euqivalence classes of
```
chenson2018: Typo:
```suggestion
/-- Every right congruence gives rise to a DA whose states are the equivalence classes of
```

Cslib/Computability/Languages/Congruences/RightCongruence.lean
```diff
@@ -0,0 +1,50 @@
...
+-/
+
+namespace Cslib
+
+/-- A right congruence is an equivalence relation on finite sequences (represented by lists)
+that is preserved by concatenation on the right.  The equivalence relation is represented
+by a setoid to to enable ready access to the quotient construction. -/
+class RightCongruence (α : Type*) extends eq : Setoid (List α) where
+  /-- If `u` an `v` are congruent, then `u ++ w` and `v ++ w` are also congruent for any `w`. -/
+  right_congr : ∀ u v, eq u v → ∀ w, eq (u ++ w) (v ++ w)
```
chenson2018: Better spelling:

```suggestion
  right_congr {u v} (huv : u ≈ v) (w : List α) : u ++ w ≈ v ++ w
```

Cslib/Computability/Languages/Congruences/RightCongruence.lean
```diff
@@ -0,0 +1,50 @@
+/-
+Copyright (c) 2026 Ching-Tsun Chou. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Ching-Tsun Chou
+-/
+
+module
+
+public import Cslib.Computability.Languages.Language
```
chenson2018: `shake` is disabled at the moment, but eventually this should be

```suggestion
public import Cslib.Init
public import Mathlib.Computability.Language
```

chenson2018 (changes requested):
This all looks reasonable to me, but would you mind posting in the "Is There Code for X?" channel to see if anyone has thoughts about `RightCongruence`/`RightCongruence.QuotType`/`RightCongruence.eqvCls`?

Cslib/Computability/Automata/DA/Congr.lean
ctchou: Fixed.

Cslib/Computability/Languages/Congruences/RightCongruence.lean
ctchou: Fixed.  I'm surprised that this works.  Isn't the `≈` notation defined only in the instance following this definition?

Cslib/Computability/Languages/Congruences/RightCongruence.lean
ctchou: Fixed.

Cslib/Computability/Languages/Congruences/RightCongruence.lean
ctchou: Ah, the change you suggested does not work after all: it breaks `DA/Congr.lean`.  In particular, it breaks the proof inside the definition of `RightCongruence.toDA`.  It looks like Lean gets confused about what `≈` refers to.  I'll keep my old code.

chenson2018: It works, you just have to update the proof.

ctchou: Git it.  You made `u` and `v` implicit.

chenson2018: Please look at the commit I just pushed, it is the exact same proof but omits the implicit arguments.

ctchou: Got it.  Thanks!

ctchou:
Per your suggestion, I have posted a query on Zulip: https://leanprover.zulipchat.com/#narrow/channel/217875-Is-there-code-for-X.3F/topic/Right.20congruence/with/568624176

ctchou:
I'm not sure the [`CovariantClass`](https://leanprover-community.github.io/mathlib4_docs/find/?pattern=CovariantClass#doc) that Eric Wieser pointed out in the Zulip thread is appropriate here.  The comments at the beginning of `Mathlib.Algebra.Order.Monoid.Unbundled.Defs` indicate that the underlying binary relation is intended to be an order, but in our application that relation is an equivalence relation.

chenson2018:
> I'm not sure the [`CovariantClass`](https://leanprover-community.github.io/mathlib4_docs/find/?pattern=CovariantClass#doc) that Eric Wieser pointed out in the Zulip thread is appropriate here. The comments at the beginning of `Mathlib.Algebra.Order.Monoid.Unbundled.Defs` indicate that the underlying binary relation is intended to be an order, but in our application that relation is an equivalence relation.

I followed up on the Zulip thread.

Cslib/Computability/Languages/Congruences/RightCongruence.lean
ctchou: @chenson2018 In view of the discussion on Zulip:
https://leanprover.zulipchat.com/#narrow/channel/217875-Is-there-code-for-X.3F/topic/Right.20congruence/with/569421397
does the change you suggested still make sense?  Wouldn't the `≈ ` refer to the specific equivalence  relation:
https://leanprover-community.github.io/mathlib4_docs/Init/Data/List/Perm.html#List.isSetoid
rather than a general equivalence relation which is what I want here?

chenson2018: No, here it is still fine, just a mistake I made in the Zulip thread. You can use `pp.all` and check that the notation is properly using the provided instance in `right_congr`, and this still does correspond to `CovariantClass`.

I am working on a commit that does this, and your proofs stay almost exactly the same in #278. I will push it to that PR so we can discuss with the full context of what it looks like to actually use it.

ctchou: Is there any advantage in using the `≈` notation here at all?  It just creates possible confusion with the existing `List.Perm` and doesn't really improve readability.  In fact, so far as I can tell, the only occasion to use the `≈` notation is to state the `right_congr` condition.

chenson2018: If a notation exists, my opinion is it is almost always preferable to use for reasons including making theorems easier to search for and avoiding problems around definitional equality.

The confusion here is really inherent to there already being an existing `Setoid` instance for lists, but I don't think it will cause actually ambiguity (in the sense of a typeclass diamond) because the instances are parameterized by an automaton.

ctchou: OK, I'll wait for your commit.

ctchou:
This PR has been absorbed into #278.

PR #269, opened 2026-01-17
feat: Single Tape Turing Machines

BoltonBailey:
I have now marked this ready for review but it's quite a large PR, reviewers please tell me if I should reorganize things or submit in a different way.

This PR adds a model of Single Tape Turing machines, which inputs and outputs `List α` for a finite alphabet type `α`, with the following features:

* In `Cslib/Foundations/Data`
  * A file for `StackTape`, a stack-like data structure for optional alphabet symbols that extends infinitely (in one direction) with blank (none) symbols. Represented as a finite list that tructates the infinite stream of none elements
  * A file for `BiTape`, a bidirectionally infinite Tape, made from two `StackTape`
* In `Cslib/Computability/Machines/SingleTapeTuring/Basic.lean`
	* `structures` representing computability within a time bound, and polynomial time computability.
	* Machine that computes the identity function, c.f. Mathlib's [Turing.idComputer](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Computability/TMComputable.html#Turing.idComputer)
	* Machine that computes the composition of the function computed by two individual machines. (Mathlib currently doesn't have this)
	  * Proof this machine preserves polynomial time computation 

This PR was authored with the support of Claude Code and [Project Numina's Lean MCP fork](https://github.com/project-numina/lean-lsp-mcp)

---

## My Design Notes

While Mathlib has a definition of polynomial time computability based on a more complicated model of TMs, I think it makes sense to have one based on a simple single-tape model here.

Longer term, I would like to try to create a unified framework for talking about TMs with variable amounts of tapes / stacks, so that we can eventually prove theorems about the time and space overhead incurred by switching from one model to another (e.g. [2 tape TMs simulate k-tape TMs with log overhead](https://www.cs.toronto.edu/tss/files/papers/HennieStearns66.pdf), 1 tape TMs simulate 2-tape TMs with quadratic overhead, 2-stack machines simulate 1 tape TMs ).

I ultimately did not reuse much if any of the Mathlib infrastructure around `Tape`s, because of the injectivity issue I described in the docstring.

### DIFF
--- a/Cslib/Computability/Machines/SingleTapeTuring/Basic.lean
+++ b/Cslib/Computability/Machines/SingleTapeTuring/Basic.lean
@@ -93,7 +93,7 @@ structure SingleTapeTM Symbol [Inhabited Symbol] [Fintype Symbol] where
   /-- type of state labels -/
   (State : Type)
   /-- finiteness of the state type -/
-  [StateFintype : Fintype State]
+  [stateFintype : Fintype State]
   /-- Initial state -/
   (q₀ : State)
   /-- Transition function, mapping a state and a head symbol to a `Stmt` to invoke,
@@ -116,7 +116,7 @@ variable [Inhabited Symbol] [Fintype Symbol] (tm : SingleTapeTM Symbol)
 
 instance : Inhabited tm.State := ⟨tm.q₀⟩
 
-instance : Fintype tm.State := tm.StateFintype
+instance : Fintype tm.State := tm.stateFintype
 
 instance inhabitedStmt : Inhabited (Stmt Symbol) := inferInstance
 


Cslib/Computability/Machines/SingleTapeTuring/Basic.lean
```diff
@@ -0,0 +1,487 @@
...
+/-- Get the movement direction from a `Stmt`. -/
+def Stmt.movement : Stmt α → Option Dir
+  | (_, movement) => movement
+
+end SingleTapeTM
+
+/-- A SingleTapeTM over the alphabet of Option α (none is blank BiTape symbol). -/
+structure SingleTapeTM α where
+  /-- Inhabited instance for the alphabet -/
+  [Inhabitedα : Inhabited α]
```
chenson2018: In Mathlib they write this reversed like

```suggestion
  [αInhabited : Inhabited α]
```
which I think is slightly easier to read.

Cslib/Foundations/Data/StackTape.lean
```diff
@@ -0,0 +1,141 @@
...
+
+namespace Turing
+
+/--
+An infinite tape representation using a list of `Option` values,
+where the list is eventually `none`.
+
+Represented as a `List (Option α)` that does not end with `none`.
+-/
+structure StackTape (α : Type) where
```
chenson2018: This is not a fully formed thought, but is this similar to `Stream'.Seq`? Just wondering if we can reuse some of that infrastructure.

Cslib/Foundations/Data/BiTape.lean
```diff
@@ -0,0 +1,138 @@
...
+
+namespace Turing
+
+/--
+A structure for bidirectionally-infinite Turing machine tapes
+that eventually take on blank `none` values
+-/
+structure BiTape (α : Type) where
+  /-- The symbol currently under the tape head -/
+  (head : Option α)
```
chenson2018: Style question: why the parentheses around these fields?

Cslib/Foundations/Data/BiTape.lean
```diff
@@ -0,0 +1,138 @@
...
+
+/--
+Given a `List` of `α`, construct a `BiTape` by mapping the list to `some` elements
+and laying them out to the right side,
+with the head under the first element of the list if it exists.
+-/
+def mk₁ {α} (l : List α) : BiTape α :=
+  match l with
+  | [] => nil
+  | h :: t => ⟨some h, StackTape.nil, StackTape.map_some t⟩
```
chenson2018: I'd go ahead and use the `∅`/`{}` notation here.

Cslib/Computability/Machines/SingleTapeTuring/Basic.lean
```diff
@@ -0,0 +1,487 @@
...
+This section defines the configurations of a Turing machine,
+the step function that lets the machine transition from one configuration to the next,
+and the intended initial and final configurations.
+-/
+
+variable (tm : SingleTapeTM α)
+
+instance : Inhabited tm.Λ := ⟨tm.q₀⟩
+
+instance : Fintype tm.Λ := tm.FintypeΛ
```
chenson2018: Do you want instances for the other bundled typeclasses as well?

Cslib/Computability/Machines/SingleTapeTuring/Basic.lean
```diff
@@ -0,0 +1,487 @@
...
+    (h : tm.OutputsWithinTime l l' t) :
+    l'.length ≤ max 1 l.length + t := by
+  simp only [OutputsWithinTime] at h
+  obtain ⟨steps, hsteps_le, hevals⟩ := h
+  replace hevals := hevals.apply_le_apply_add
+  specialize hevals (Cfg.space_used tm)
+  simp only [Cfg.space_used_initCfg, Cfg.space_used_haltCfg] at hevals
+  suffices l'.length ≤ max 1 l.length + steps by lia
+  specialize hevals fun a b hstep ↦ Cfg.space_used_step a b (Option.mem_def.mp hstep)
+  omega
```
chenson2018: A little cleaning wouldn't hurt:
```suggestion
  obtain ⟨steps, hsteps_le, hevals⟩ := h
  replace hevals := hevals.apply_le_apply_add
  specialize hevals (Cfg.space_used tm)
  specialize hevals fun a b hstep ↦ Cfg.space_used_step a b (Option.mem_def.mp hstep)
  grind [Cfg.space_used_initCfg, Cfg.space_used_haltCfg]
```

If you wanted to fully golf with `grind`, I ended up with:
```suggestion
  have (a b) (hstep : tm.TransitionRelation a b) : tm.step a = some b := Option.mem_def.mp hstep
  have := hevals.apply_le_apply_add (Cfg.space_used tm)
  grind [→ Cfg.space_used_step, Cfg.space_used_initCfg, Cfg.space_used_haltCfg]
```

Cslib/Computability/Machines/SingleTapeTuring/Basic.lean
```diff
@@ -0,0 +1,487 @@
...
+This is important for guaranteeing that composition of polynomial time Turing machines
+remains polynomial time, as the input to the second machine
+is bounded by the output length of the first machine.
+-/
+lemma output_length_le_input_length_add_time (tm : SingleTapeTM α) (l l' : List α) (t : ℕ)
+    (h : tm.OutputsWithinTime l l' t) :
+    l'.length ≤ max 1 l.length + t := by
+  simp only [OutputsWithinTime] at h
+  obtain ⟨steps, hsteps_le, hevals⟩ := h
+  replace hevals := hevals.apply_le_apply_add
```
chenson2018: `apply_le_apply_add` seems to have somewhat awkward ordering of paramaters?

Cslib/Computability/Machines/SingleTapeTuring/Basic.lean
```diff
@@ -0,0 +1,487 @@
...
+
+section Computers
+
+variable [Inhabited α] [Fintype α]
+
+/-- A Turing machine computing the identity. -/
+def idComputer : SingleTapeTM α where
+  Λ := PUnit
+  q₀ := PUnit.unit
+  M := fun _ b => ⟨(b, none), none⟩
```
chenson2018: I find
```suggestion
  M _ b := ⟨(b, none), none⟩
```
slightly more readable.

Cslib/Computability/Machines/SingleTapeTuring/Basic.lean
```diff
@@ -0,0 +1,487 @@
...
+  cases cfg2 with
+  | mk state BiTape =>
+    cases state with
+    | none =>
+      simp only [step, toCompCfg_right, Option.map_none, compComputer]
+    | some q =>
+      simp only [step, toCompCfg_right, compComputer, Option.map_some]
+      generalize hM : tm2.M q BiTape.head = result
+      obtain ⟨⟨wr, dir⟩, nextState⟩ := result
+      cases nextState <;> grind [toCompCfg_right]
```
chenson2018: `grind` will do cases on options by default

```suggestion
      grind [toCompCfg_right]
```

chenson2018: More generally, when I see a `simp only` followed by terminal `grind`, it often indicates to me that some annotations are missing. Somewhat unusually though there seems to be some difference between `simp` and `grind` as

```suggestion
      generalize hM : tm2.M q BiTape.head = result
      obtain ⟨⟨wr, dir⟩, nextState⟩ := result
      simp only [compComputer]
      grind [toCompCfg_right, step, toCompCfg_right, compComputer]
```
breaks if you remove `compComputer` from either.

Cslib/Computability/Machines/SingleTapeTuring/Basic.lean
```diff
@@ -0,0 +1,487 @@
...
+
+/-- A Turing machine + a time function +
+a proof it outputs `f` in at most `time(input.length)` steps. -/
+structure TimeComputable (f : List α → List α) where
+  /-- the underlying bundled SingleTapeTM -/
+  tm : SingleTapeTM α
+  /-- a bound on runtime -/
+  time_bound : ℕ → ℕ
+  /-- proof this machine outputs `f` in at most `time_bound(input.length)` steps -/
+  outputsFunInTime : ∀ a, tm.OutputsWithinTime a (f a) (time_bound a.length)
```
chenson2018: I believe Mathlib style usually would dictate

```suggestion
  outputsFunInTime (a) : tm.OutputsWithinTime a (f a) (time_bound a.length)
```

Cslib/Computability/Machines/SingleTapeTuring/Basic.lean
```diff
@@ -0,0 +1,487 @@
...
+* The time taken by the first machine on the input size
+* added to the time taken by the second machine on the output size of the first machine
+  (which is itself bounded by the time taken by the first machine)
+
+Note that we require the time function of the second machine to be monotone;
+this is to ensure that if the first machine returns an output
+which is shorter than the maximum possible length of output for that input size,
+then the time bound for the second machine still holds for that shorter input to the second machine.
+-/
+def TimeComputable.comp {f g : List α → List α} (hf : TimeComputable f) (hg : TimeComputable g)
```
chenson2018: I think we could use some notation, but this can come in follow-up PRs.

Cslib/Computability/Machines/SingleTapeTuring/Basic.lean
```diff
@@ -0,0 +1,487 @@
...
+
+variable {α : Type}
+
+namespace SingleTapeTM
+
+/--
+A Turing machine "statement" is just a `Option`al command to move left or right,
+and write a symbol on the `BiTape`.
+-/
+def Stmt (α : Type) := Option α × Option Dir
```
chenson2018: This is a bit of an outlier as the only type `def` versus using a `structure`, any particular reason why?

chenson2018 (changes requested):
This looks very well done and documented, thank you. The comments I've left are mostly stylistic, with the primary exception being the design of `StackTape`.

Please note that not all suggestions regarding `grind` are mandatory, but I felt it appropriate to mention since you were using it a bit already. Some of these equally apply to `left`/`right` versions of theorems, I just commented on the ones I saw first.

Normally I might raise questions about duplication from Mathlib, but in this case the representation is somewhat the point of the formalization, and you've closely followed the style of `TM1`, `TM2`, etc.

Cslib/Foundations/Data/StackTape.lean
BoltonBailey: ~~Unfortunately, I don't see a clean way to combine these:~~

StackTape is not a subset of [Stream'.Seq](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Seq/Defs.html#Stream'.Seq) because it allows one to write a none in the middle of the list (to disallow this would be complicated, as the TM would have to know adjacent cells to be sure this was allowed.

OTOH it's not a superset either because StackTape *must* have finitely many `somes`, and Stream'.Seq doesn't.

Another question is, could StackTape simply be made a subtype of [Stream'](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Stream/Defs.html#Stream') (Option α) (like Seq but with a different predicate). This would be possible, but I worry that I will at some point want to print out the tape as a finite list, or the length, and if I am going to do this, I need to know the length explicitly, not just that such a length exists. 

Perhaps it could be some kind of length, stream pair / structure?

chenson2018: Hmm, I'll trust your judgement on this. If on further thought one of the alternatives stands out to you as an improvement in terms of the combination of code reuse and ergonomics, feel free to make that change. Otherwise, I am fine with this as is.

Cslib/Computability/Machines/SingleTapeTuring/Basic.lean
```diff
@@ -0,0 +1,487 @@
...
+  /-- Finiteness of the alphabet -/
+  [Fintypeα : Fintype α]
+  /-- type of state labels -/
+  (Λ : Type)
+  /-- finiteness of the state type -/
+  [FintypeΛ : Fintype Λ]
+  /-- Initial state -/
+  (q₀ : Λ)
+  /-- Transition function, mapping a state and a head symbol
+  to a Stmt to invoke, and optionally a new state (none for halt) -/
```
crei: This could maybe clarify that in case of `halt`, the statement is still executed before the machine halts.

Maybe:
`and optionally the new state to transition to afterwards (none for halt)`

Cslib/Computability/Machines/SingleTapeTuring/Basic.lean
```diff
@@ -0,0 +1,487 @@
...
+    -- If in the halting state, there is no next configuration
+    none
+  | ⟨some q', t⟩ =>
+    -- If in state q', perform look up in the transition function
+    match tm.M q' t.head with
+    -- and enter a new configuration with state q'' (or none for halting)
+    -- and tape updated according to the Stmt
+    | ⟨(wr, dir), q''⟩ => some ⟨q'', (t.write wr).optionMove dir⟩
+
+/-- The initial configuration corresponding to a list in the input alphabet. -/
```
crei: There is currently no distinction between input, output and work alphabet, so maybe just say "alphabet"? (I also think distinguishing between the alphabets is a complication that should not be introduced)

Cslib/Computability/Machines/SingleTapeTuring/Basic.lean
```diff
@@ -0,0 +1,487 @@
...
+
+namespace Turing
+
+variable {α : Type}
+
+namespace SingleTapeTM
+
+/--
+A Turing machine "statement" is just a `Option`al command to move left or right,
+and write a symbol on the `BiTape`.
```
crei: Maybe clarify that "writing `none`" means "writing the blank symbol" instead of "not writing" (as opposed to the move action where none means "do not move")?

Cslib/Foundations/Data/BiTape.lean
BoltonBailey: Not sure, perhaps I was just pattern matching from structure FinTM2 in Mathlib where I guess it's done to differentiate the typeclass args from otherwise?

Cslib/Computability/Machines/SingleTapeTuring/Basic.lean
BoltonBailey: The second one didn't compile for me

BoltonBailey: Ok I've tweaked it a bit and it works

Cslib/Computability/Machines/SingleTapeTuring/Basic.lean
BoltonBailey: ah ok, yes, I'll reorder and make the m implicit so that the first explicit argument is the RelatesInSteps.

Cslib/Computability/Machines/SingleTapeTuring/Basic.lean
BoltonBailey: Perhaps the issue is that step for comp computer should be defined separately? It feels to me like that should be part of the definition, though, so I'm not sure what's best, I'll just reorder the proof

Cslib/Computability/Machines/SingleTapeTuring/Basic.lean
BoltonBailey: Sounds good

Cslib/Computability/Machines/SingleTapeTuring/Basic.lean
BoltonBailey: Hmm perhaps I just was coding quickly and saw there were only two fields and thought to use a pair, I'll change it.

Cslib/Computability/Machines/SingleTapeTuring/Basic.lean
BoltonBailey: > There is currently no distinction between input, output and work alphabet, so maybe just say "alphabet"? 

Well, that's not quite true, the input alphabet is \alpha but the work alphabet (i.e. the type of entries of the tape cells, as they are read and written by the machine) is Option \alpha. That is why I clarify here.

The reason for having this distinction is that I feel it is good to make sure that I can define a initCfg map which is as natural as possible while still guaranteeing that the map is injective, I explain this a little in the design section of the BiTape module docstring, but I'll also put a note here.

Cslib/Foundations/Data/StackTape.lean
BoltonBailey: After thinking more and playing around, I think the fact that I want to use `length` makes it easier to keep as is, I'll leave a note in the docstring about different possible future design choices.

Cslib/Computability/Machines/SingleTapeTuring/Basic.lean
chenson2018: Looks like I was wrong, as these new instances trigger the `impossibleInstance` linter.

BoltonBailey:
I am experiencing some kind of CI error with `mk_all`, is it right that I should be changing CSLibTests.lean?

chenson2018:
> I am experiencing some kind of CI error with `mk_all`, is it right that I should be changing CSLibTests.lean?

Are you passing the `--module` flag?

BoltonBailey:
No, the error message just told me to run `lake exe mk_all`

The file 'Cslib.lean' is out of date: run `lake exe mk_all` to update it

chenson2018:
> No, the error message just told me to run `lake exe mk_all`
> 
> The file 'Cslib.lean' is out of date: run `lake exe mk_all` to update it

The docs upstream are likely out of date given how recent this change is.

chenson2018:
It looks like you've addressed my suggestions, thank you! Do you mind identifying some of the theorems you are repeatedly passing to `grind` and adding annotations to them?

BoltonBailey:
Ok I have gone and added grind annotations to the point where I think I haven't used any one lemma more than once, so I guess that's the letter of what you've said. I haven't looked heavily into the different variety of grind annotations though, so I'm a bit nervous that I've left ones out, but I also don't want to add too many annotations (would that cause grind to explode or something?). LMK if you want me to do more research into how to annotate so I can give the best annotations here.

BoltonBailey:
Like for example `length_tail_le {α} (l : StackTape α) : l.tail.length ≤ l.length`.

It would seem nice if this would be triggered whenever `l.tail.length` appears, but not when `l.length` appears, is there a way of annotating this?

Edit, I guess this is what grind_pattern is for!

chenson2018:
> Edit, I guess this is what grind_pattern is for!

Yeah in this case I think what you're asking for is `grind_pattern length_tail_le => l.tail`, which I believe is equivalent to `@[grind! .]`.

chenson2018:
Can you also please use `scoped` with the annotations you have added? (remember that if they are inconveniently nested deep in a namespace, you can add them afterward as an attribute to avoid this)

Edit: I am not sure when, but I remember hearing there will be something like `simp` sets to make this nicer in the future. We aggressively scope `grind` annotations for performance reasons.

chenson2018 (approved):
Thanks again for the contribution! I think all my feedback has been addressed. I will wait to merge in case @fmontesi has any comments.

Cslib/Computability/Machines/SingleTapeTuring/Basic.lean
```diff
@@ -0,0 +1,490 @@
...
+
+section PolyTimeComputable
+
+open Polynomial
+
+variable [Inhabited α] [Fintype α]
+
+/-- A Turing machine + a polynomial time function +
+a proof it outputs `f` in at most `time(input.length)` steps. -/
+structure PolyTimeComputable (f : List α → List α) extends TimeComputable f where
```
crei: Just saw the discussion on zulib about computable polynomials.

This is probably for a future change, but maybe this could also instead use something like

∃ c₁ c₂ c₃, ∀ n, time_bound n ≤ c1 * n^c2 + c3

(as it is done in most textbooks).

And then only use mathlib's polynomial library to show relations between expressions of that form. It could also help to use Mathlib.Analysis.Asymptotics, although that is quite a heavy machinery. It probably makes sense to have a CsLib-tailored definition of "BigO" in general.

And then we would be able to say here

```
PolyTimeComputable (f : List α → List α) extends TimeComputable f where
  k : ℕ
  bounds : time_bound ∈ O(·^k)
```

BoltonBailey: Yes, I agree something like this would be nice, especially as "polynomial time" is just one time-based complexity class, and it would be nice to express others.

In the case here I think `Polynomial.comp` is doing the work of proving that this class of functions is closed under composition, and since I don't want to reprove a result like this now, I'll leave it in.

Cslib/Computability/Machines/SingleTapeTuring/Basic.lean
```diff
@@ -0,0 +1,490 @@
...
+A single-tape Turing machine
+over the alphabet of `Option α` (where `none` is the blank `BiTape` symbol).
+-/
+structure SingleTapeTM α where
+  /-- Inhabited instance for the alphabet -/
+  [αInhabited : Inhabited α]
+  /-- Finiteness of the alphabet -/
+  [αFintype : Fintype α]
+  /-- type of state labels -/
+  (Λ : Type)
```
fmontesi: Why not just `State` as in automata, LTS, etc.?

fmontesi: Why is the state type bundled, as opposed to how we do things in automata, reductions, and lts?

Cslib/Computability/Machines/SingleTapeTuring/Basic.lean
```diff
@@ -0,0 +1,490 @@
...
+  movement : Option Dir
+deriving Inhabited
+
+end SingleTapeTM
+
+/--
+A single-tape Turing machine
+over the alphabet of `Option α` (where `none` is the blank `BiTape` symbol).
+-/
+structure SingleTapeTM α where
```
fmontesi: You could also use `Symbol` or `Label` for alpha here, as we do in automata/lts.

Cslib/Computability/Machines/SingleTapeTuring/Basic.lean
```diff
@@ -0,0 +1,490 @@
...
+  [αFintype : Fintype α]
+  /-- type of state labels -/
+  (Λ : Type)
+  /-- finiteness of the state type -/
+  [ΛFintype : Fintype Λ]
+  /-- Initial state -/
+  (q₀ : Λ)
+  /-- Transition function, mapping a state and a head symbol to a `Stmt` to invoke,
+  and optionally the new state to transition to afterwards (`none` for halt) -/
+  (M : Λ → Option α → SingleTapeTM.Stmt α × Option Λ)
```
fmontesi: We typically write `tr` for transition functions (in the automata module).

fmontesi: Could you explain the design decisions a bit here?

If I look at the pen-and-paper def, there's usually a notion of accept states (as in automata), and this looks more like (sorry for mixing unicode and backslash notation..):

`tr : {s : State // s \notin accept} → Option Symbol → Option (Stmt \times Symbol)`

fmontesi (changes requested):
Sorry it took me a while to circle back to this.

I don't have first-hand experience in the formalisation of Turing machines, so I might well be off, but I think it'd still be good to have answers to my questions for future reference and other people who will use this.

Cslib/Computability/Machines/SingleTapeTuring/Basic.lean
chenson2018: I've been meaning to ask about this actually. This is one area where we allow deviations from the Mathlib style guide, using names instead of Greek letter for type variables. This is fine, but is it a requirement or just something we allow? There are some places that I find the names pretty verbose and occasionally I do find it confusing. 

My preference would be to allow this to be author discretion, especially when it's something like this that closely interacts with a Mathlib formalization. I've already done this for instance in `Cslib.Foundations.Data.Relation`, where I think we agree following the convention for relations is preferable. I think it is similarly nice here that this lines up with Mathlib's `TM0`, `TM1`, and `TM2`.

chenson2018:
@fmontesi I think the answer to most of your questions here is consistency with Mathlib, which has several notions of Turing Machine that share for instance a naming convention.

(It doesn't hurt to reexamine design questions, just pointing out the context here.)

Cslib/Computability/Machines/SingleTapeTuring/Basic.lean
crei: I'm not sure about @BoltonBailey 's motivation, but I have worked on top of both this definition and a definition that uses accepting states instead and I prefer the one where the transition function goes to `Option`. The annoying thing with accepting states is that when you prove the semantics of a TM you not only have to reason that the configuration is reachable but also that it is the first accepting configuration that is reached. For the `Option` one, you only need to prove reachability. Also sequential composition is easier because you don't have to combine the accepting state of the first TM with the initial state of the second TM.

Complexity textbook definitions also have the notion of accepting and rejecting states (in addition to "regular" states), but it can be easily replaced by a write-only output tape, so I don't think this is an issue here.

Cslib/Computability/Machines/SingleTapeTuring/Basic.lean
fmontesi: I'm still brainstorming myself about this, but this case is at least weird to see.
The key difference here is that a relation is clearly intended as a concept for 'whatever type', so there's no intuition to give about that type and something not descriptive as `α` works well. But in a TM, we have an opportunity to give intuition as to what these types are meant to represent within the machinery.

Cslib/Computability/Machines/SingleTapeTuring/Basic.lean
fmontesi: Good point, thanks! Indeed a key difference between TM and automata is that a TM cannot move anymore after reaching an accept state, whereas an automaton can.

Cslib/Computability/Machines/SingleTapeTuring/Basic.lean
chenson2018: I think my example of relations has caused some confusion; let me try to be more clear. These *particular* symbols are specified (see the doc on [Naming Conventions](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Computability/PostTuringMachine.html#Naming-conventions)) for this usage in the several models of Turing machines that Mathlib has. I'll also note that these align somewhat with what is used in the literature.

The tradeoff here is between a longer descriptive name versus this connection to the literature and Mathlib notation. My personal position is that that latter is preferable for readability and consistency, but more generally I'm okay with this being at author's discretion (within reason). For instance, if it were up to me automata and LTS would also use Greek letters for type variables, but I felt like this is something that didn't need to be strictly dictated either way.

chenson2018: (I also found it helpful when reviewing that these aligned, as it made comparing them much easier)

Cslib/Computability/Machines/SingleTapeTuring/Basic.lean
BoltonBailey: This was a holdover from the Mathlib version I based this on, now fixed!

BoltonBailey: I think that with Turing machines, the state set is usually considered not important. I think I would like to later be able to write theorems of the form "If a is an oracle machine with A functionality and b is a Turing machine with B functionality, then a calling b has C functionality", more concisely than "if s_a is a type of states and a is an oracle machine over state type s_a with A functionality and s_b is another state type and b is a Turing machine over state type s_b with B functionality, then the Turing machine a calling b has C functionality".

Perhaps the key difference is that TMs are more compositional, and its more expected that you will have more functions that take in TMs and return other TMs?

Cslib/Computability/Machines/SingleTapeTuring/Basic.lean
BoltonBailey: I like \alpha because it evokes "alphabet" which is the term I'm most familiar with, but I have also heard "Symbol" so I'll switch to that.

Cslib/Computability/Machines/SingleTapeTuring/Basic.lean
BoltonBailey: After trying it out, even in this PR, I think it is awkward, because the type signature of compComputer would have to go from 

`def compComputer (tm1 tm2 : SingleTapeTM Symbol) : SingleTapeTM Symbol`

To something like 

`def compComputer  (tm1 : SingleTapeTM Symbol State1) (tm2 : SingleTapeTM Symbol State2) : SingleTapeTM Symbol (State1 ⊕ State2)`

And I feel lucky the state of the combined machine isn't a more complicated expression. So I think I am going to leave this, but let me know if it's very important.

Cslib/Computability/Machines/SingleTapeTuring/Basic.lean
chenson2018: I've unresolved this comment just because I'd like to continue the more general discussion. (This can move to Zulip if we'd like)

Cslib/Computability/Machines/SingleTapeTuring/Basic.lean
BoltonBailey: I have elaborated on the module docstring to give an account of the reasoning behind these choices.

fmontesi:
I'm pretty happy with this now.

Before merging, @chenson2018, can you double check if the casing of things like `[SymbolInhabited : Inhabited Symbol]` is correct? Or is it supposed to be `symbolInhabited`? (That's what I'd do.)

chenson2018:
> I'm pretty happy with this now.
> 
> Before merging, @chenson2018, can you double check if the casing of things like `[SymbolInhabited : Inhabited Symbol]` is correct? Or is it supposed to be `symbolInhabited`? (That's what I'd do.)

@BoltonBailey While discussing the naming convention, the topic of why the instances for `Symbol` are bundled came up. It perhaps makes sense for `State` since they are all bundled, but probably not to have a mixture. Is this an easy fix?

BoltonBailey:
> While discussing the naming convention, the topic of why the instances for `Symbol` are bundled came up. It perhaps makes sense for `State` since they are all bundled, but probably not to have a mixture. Is this an easy fix?

Sorry I'm not quite sure what you're asking here. The reason State is bundled is because it's convenient to not have to specify state type before defining a new TM, and so the Fintype State is also bundled because it comes after it. 

I guess the question is why the Fintype and Inhabited for Symbol aren't unbundled? I think this probably comes from the mathlib version, because in the mathlib version it's a multi-tape/stack machine with a different symbol alphabet for each tape, the multi tape case is a case where we might like to make a new TM with one tape for each tape in a collection of input machines, and I don't want to have to know in advance and pass in the symbol types for each of those when I am constructing my machine (similarly to how for  `State` I don't want to have to know how to make the state type of composition machines before I define composition machines).

In this case we have a single tape and single alphabet, so it's not necessary to bundle, and it seems to work fine to unbundle. I'm not totally sure how this will affect generalizing this to multiple stacks and tapes. Mostly, I think I am less interested in generalizing machines to having different symbols in each tape, because I think the theory there is less interesting, but maybe it's something that will become a practical necessity?

I will unbundle the symbol typeclasses here, and hopefully any future versions with multiple symbols can just have unbundled input tape symbol with bundled work tape symbols. If that's not a good solution due to being an unintuitive mixture of bundles and unbundles, we can cross that bridge later.

chenson2018:
Yes, this is exactly what I meant, just unbundling the `Symbol` instances. If we need to revist anything on generalization, it is no problem. I fixed one naming and will set to automerge.

PR #270, opened 2026-01-17
feat: add time step counting for relations

BoltonBailey:
This PR adds `Reduction.relates{In,Within}Steps`, functions that communicate whether a relation forms a directed path of length (at most) `n` between two elements.

Specifically, it includes:

* Definitions of these functions
* Theorems about reflexivity and transitivity
* Theorems that track the growth of ancillary functions with respect to the relation

These are a prerequisite to defining time bounds, perhaps for multiple models of computation, but in particular for Turing machines, see PR #269.

### DIFF
--- a/Cslib.lean
+++ b/Cslib.lean
@@ -36,6 +36,7 @@ public import Cslib.Foundations.Data.OmegaSequence.Flatten
 public import Cslib.Foundations.Data.OmegaSequence.InfOcc
 public import Cslib.Foundations.Data.OmegaSequence.Init
 public import Cslib.Foundations.Data.OmegaSequence.Temporal
+public import Cslib.Foundations.Data.RelatesInSteps
 public import Cslib.Foundations.Data.Relation
 public import Cslib.Foundations.Lint.Basic
 public import Cslib.Foundations.Semantics.FLTS.Basic


chenson2018:
This looks great to have, thanks for working on this! I am wondering though if we should at least initially put this in `Cslib.Foundations.Data.Relation` instead  (or make an appropriate subdirectory) and work with just unbundled relations. My reasoning is:
- there is some overlap there, for instance the definition [Relation.Reducible](https://api.cslib.io/docs/Cslib/Foundations/Data/Relation.html#Relation.Reducible) and [Relation.Normal](https://api.cslib.io/docs/Cslib/Foundations/Data/Relation.html#Relation.Normal) could be reused
- I am a bit unsure about fully committing to using `ReductionSystem` as an API because of the duplication of proofs about relations that would be required

What do you think?

BoltonBailey:
> What do you think?

Sounds fine to me! I have moved this content to the `Relation` file and made some name changes that might make more sense in this context.

Cslib/Foundations/Data/Relation.lean
```diff
@@ -339,4 +339,176 @@ theorem reflTransGen_compRel : ReflTransGen (CompRel r) = EqvGen r := by
...
+section Steps
+
+/--
+A relation `r` relates two elements of `α` in `n` steps
+if there is a chain of `n` pairs `(t_i, t_{i+1})` such that `r t_i t_{i+1}` for each `i`,
+starting from the first element and ending at the second.
+-/
+inductive relatesInSteps (r : α → α → Prop) : α → α → ℕ → Prop
+  | refl (t : α) : relatesInSteps r t t 0
+  | cons (t t' t'' : α) (n : ℕ) (h₁ : r t t') (h₂ : relatesInSteps r t' t'' n) :
```
chenson2018: I think it would be good for these constructors to line up with the definition of the reflexive transitive closure. In thinking about this I wrote:
```lean
theorem relatesInSteps.reflTransGen (h : relatesInSteps r a b n) : ReflTransGen r a b := by
  induction h with
  | refl => rfl
  | cons _ _ _ _ h _ ih => exact .head h ih

theorem ReflTransGen.relatesInSteps (h : ReflTransGen r a b) : ∃ n, relatesInSteps r a b n := by
  induction h using ReflTransGen.head_induction_on with
  | refl => exact ⟨0, .refl b⟩
  | head _ _ ih =>
    obtain ⟨n, _⟩ := ih
    exact ⟨n + 1, by grind [relatesInSteps]⟩
```
and this highlighted that `relatesInSteps` builds in the opposite direction as `ReflTransGen`. At minimum I'd call it `head` to match `ReflTransGen.head_induction_on`, or do you think it's worth rewriting to the `tail` version?

Cslib/Foundations/Data/Relation.lean
```diff
@@ -339,4 +339,176 @@ theorem reflTransGen_compRel : ReflTransGen (CompRel r) = EqvGen r := by
...
     | trans _ _ _ _ _ ih₁ ih₂ => exact ih₁.trans ih₂
 
+section Steps
+
+/--
+A relation `r` relates two elements of `α` in `n` steps
+if there is a chain of `n` pairs `(t_i, t_{i+1})` such that `r t_i t_{i+1}` for each `i`,
+starting from the first element and ending at the second.
+-/
+inductive relatesInSteps (r : α → α → Prop) : α → α → ℕ → Prop
```
chenson2018: I think that naming conventions would have this be called `RelatesInSteps`, based on:

> Props and Types (or Sort) (inductive types, structures, classes) are in UpperCamelCase.

(We are not perfect about following Mathlib naming, but I try my best for the simple rules at least...)

Cslib/Foundations/Data/Relation.lean
```diff
@@ -339,4 +339,176 @@ theorem reflTransGen_compRel : ReflTransGen (CompRel r) = EqvGen r := by
...
+  | refl (t : α) : relatesInSteps r t t 0
+  | cons (t t' t'' : α) (n : ℕ) (h₁ : r t t') (h₂ : relatesInSteps r t' t'' n) :
+      relatesInSteps r t t'' (n + 1)
+
+lemma relatesInSteps.single {a b : α} (h : r a b) :
+    relatesInSteps r a b 1 := by
+  exact relatesInSteps.cons a b b 0 h (relatesInSteps.refl b)
```
chenson2018: The signature fits on one line, and I'd remove the `by exact`
```suggestion
lemma relatesInSteps.single {a b : α} (h : r a b) : relatesInSteps r a b 1 :=
  relatesInSteps.cons a b b 0 h (relatesInSteps.refl b)
```

Cslib/Foundations/Data/Relation.lean
```diff
@@ -339,4 +339,176 @@ theorem reflTransGen_compRel : ReflTransGen (CompRel r) = EqvGen r := by
...
+-/
+lemma relatesInSteps.apply_le_apply_add {a b : α} (h : α → ℕ)
+    (h_step : ∀ a b, r a b → h b ≤ h a + 1)
+    (m : ℕ) (hevals : relatesInSteps r a b m) :
+    h b ≤ h a + m := by
+  induction hevals with
+  | refl _ => simp
+  | cons t t' t'' k h_red _ ih =>
+    have h_step' := h_step t t' h_red
+    omega
```
chenson2018: I have a very slight preference for `lia` for future maintenance purposes.

chenson2018 (changes requested):
Thanks for moving this over to this module. I left some additional style/naming comments.

Cslib/Foundations/Data/Relation.lean
BoltonBailey: I somehow hadn't realized that `ReflTransGen` goes the other way (I think I based this on [SimpleGraph.Walk](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Combinatorics/SimpleGraph/Walks/Basic.html#SimpleGraph.Walk)). It's sort of surprising to me how `ReflTransGen` is defined, since I think this way is more like how `List` recursion works.

But I agree it would be good to correspond to this one in particular, so I will switch the recursion, and then just make a `head` theorem artificially / replicate the other type of induction.

Cslib/Foundations/Data/Relation.lean
```diff
@@ -339,4 +339,207 @@ theorem reflTransGen_compRel : ReflTransGen (CompRel r) = EqvGen r := by
...
+  | refl (a : α) : RelatesInSteps r a a 0
+  | tail (t t' t'' : α) (n : ℕ) (h₁ : RelatesInSteps r t t' n) (h₂ : r t' t'') :
+      RelatesInSteps r t t'' (n + 1)
+
+theorem RelatesInSteps.reflTransGen (h : RelatesInSteps r a b n) : ReflTransGen r a b := by
+  induction h with
+  | refl => rfl
+  | tail _ _ _ _ h ih => exact .tail ih h
+
+theorem ReflTransGen.RelatesInSteps (h : ReflTransGen r a b) : ∃ n, RelatesInSteps r a b n := by
```
chenson2018: This becomes `ReflTransGen.relatesInSteps`, per your last commit?

BoltonBailey:
With the additional lemmas this is getting longer, and there is more `ReflTransGen` content from Mathlib.Logic.Relation that could be replicated. Would you like me to move this to a new file?

chenson2018:
> With the additional lemmas this is getting longer, and there is more `ReflTransGen` content from Mathlib.Logic.Relation that could be replicated. Would you like me to move this to a new file?

Sure, feel free. Another file in this same directory seems appropriate for now?

chenson2018:
I think you need to run `lake exe mk_all --module`

PR #273, opened 2026-01-21
feat: Bisimilarity can be characterized through symmetric simulation relations

benkeks:
In formalized proofs, the characterization of bisimilarity through symmetric simulations often is more convenient than the one through bisimulation relations. (The reason is that paper proofs about bisimulation often say “the converse case is analogous,” which is not possible in proof assistants. The background usually is that the witness simulation is symmetric or can easily be adapted to be symmetric, which can usually be handled conveniently in a proof assistant.)

Being about alternative characterizations of bisimilarity notions, the PR also proposes to remove the `SWBisimilarity` relation and only see `SWBisimulation` as an alternative characterization mechanism for `WeakBisimilarity`. The thought behind this is that there usually are many ways of characterizing the same notion of equivalence. Especially the weak spectrum has rich aliasing of notions, e.g. with a range of different definitions all leading to the same branching bisimilarity relation. It seems impractical to grant each equivalent definition its own relation symbol.

Also, some proofs are shortened a bit using `grind`.


Cslib/Foundations/Semantics/LTS/Bisimulation.lean
```diff
@@ -756,6 +752,24 @@ theorem Bisimulation.simulation_iff {lts : LTS State Label} {r : State → State
   have _ (s1 s2) : r s1 s2 → flip r s2 s1 := id
   grind [Simulation, flip]
 
+open LTS in
+/-- Bisimilarity can also be characterized through symmetric simulations. -/
+@[scoped grind =]
+theorem Bisimilarity.symm_simulation {lts : LTS State Label} :
+    p ~[lts] q = ∃ r, r p q ∧ IsSymm State r ∧ Simulation lts r := by
```
fmontesi: IsSymm has recently been deprecated, so I guess we should use Symm?

fmontesi: For consistency with weak bisimilarity, shouldn't we just

```
Bisimilarity lts =
    fun s1 s2 => ∃ r : State → State → Prop, r p q ∧ IsSymm State r ∧ Simulation lts r
```

fmontesi:
Very nice, thanks! I just have a couple of minor comments.

Cslib/Foundations/Semantics/LTS/Bisimulation.lean
benkeks: You're right, `Std.Symm` is much nicer anyway :+1: (Being unsure about the namespace import policy, I went with qualified name so far.)

benkeks: I've made the change (and also changed `p` `q` to `s1` `s2` .)

PR #276, opened 2026-01-22
feat: Adding some `CompRel` lemmas to Relation for symmetric closure support

abhamra:
This PR adds some preliminary support for [`CompRel`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Order/Comparable.html#CompRel) in `Relation.lean`, as it doesn't currently have a concrete notion of symmetric closure. In particular, we add:
-  Helper lemmas for symmetry relationships between `ReflGen`, `ReflTransGen`, and `CompRel`
- A theorem showing `EqvGen r = ReflTransGen (CompRel r)`, bridging the gap between `Gen` structures.

This should hopefully help with future development surrounding symmetric closures in `Relation.lean`.

### DIFF
--- a/Cslib/Foundations/Data/Relation.lean
+++ b/Cslib/Foundations/Data/Relation.lean
@@ -45,6 +45,8 @@ theorem TransGen.to_eqvGen (h : TransGen r a b) : EqvGen r a b := by
 theorem ReflTransGen.to_eqvGen (h : ReflTransGen r a b) : EqvGen r a b := by
   induction h <;> grind
 
+-- TODO: topNamespace environment linter fails for CompRel.to_eqvGen
+@[nolint topNamespace]
 theorem _root_.CompRel.to_eqvGen (h : CompRel r a b) : EqvGen r a b := by
   induction h <;> grind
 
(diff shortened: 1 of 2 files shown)


chenson2018 (approved):
With the minor changes I made, this looks good. So you can understand what I changed and why:

- I changed the last theorem to `ReflTransGen (CompRel r) = EqvGen r`, because for `simp`/`grind` purposes we want simplify into a single inductive (and these go left to right)
- For similar reasons, I removed `ReflTransGen.compRel_swap`. Where you used this, `reflTransGen_swap.mp` also works.
- I ran `lake exe lint-style --fix` to remove some trailing whitespace
- I used the name `_root_.CompRel.to_eqvGen`, because otherwise this would confusingly be namespaced as `Relation.CompRel.to_eqvGen`
- There was a weird linter bug (not your fault) that I added a `nolint` for

abhamra:
> With the minor changes I made, this looks good. So you can understand what I changed and why:
> 
>     * I changed the last theorem to `ReflTransGen (CompRel r) = EqvGen r`, because for `simp`/`grind` purposes we want simplify into a single inductive (and these go left to right)
> 
>     * For similar reasons, I removed `ReflTransGen.compRel_swap`. Where you used this, `reflTransGen_swap.mp` also works.
> 
>     * I ran `lake exe lint-style --fix` to remove some trailing whitespace
> 
>     * I used the name `_root_.CompRel.to_eqvGen`, because otherwise this would confusingly be namespaced as `Relation.CompRel.to_eqvGen`
> 
>     * There was a weird linter bug (not your fault) that I added a `nolint` for

This was very helpful; I definitely wouldn't have considered the first point but they all make sense.

PR #278, opened 2026-01-22
feat: define BuchiCongruence and prove that it is a right congruence of finite index

ctchou:
This PR defines the notion of a "right congruence", which is an equivalence relation between finite words that is preserved by concatenation on the right, and proves its basic properties:

- There is a natural deterministic automaton corresponding to each right congruence whose states are the equivalence classes of the congruence and whose transitions correspond to concatenation on the right of the input symbols.
- If a right congruence is of finite index, then each of its equivalence classes is a regular language.

Furthermore, this PR also defines a special type of right congruences introduced by J.R. Buchi and proves that it is of finite index if the underlying Buchi automaton is finite-state.  The purpose of Buchi congruence is to prove the closure of ω-regular languages under complementation.  But more work will be needed before we reach that goal.

The old PR #265 has been absorbed into this PR.


ctchou:
Rebase on the current upstream/main.

chenson2018:
Please take a look at the first commit that I just pushed. (This took me quite a while to think over, so please consider the differences carefully)

I used the following definition:
```lean
class RightCongruence (α : Type*) extends
  eq : Setoid (List α) , CovariantClass _ _ (fun x y => y ++ x) eq where
```
so there is no confusion about notation referring to the existing setoid instance for lists, and all of your proofs are the same with the exception of some arguments changing order.

I also made some changes related to where you were passing `RightCongruence` as an explicit paramater versus a typeclass paramater. A nice side effect is that using `RightCongruence.eqvCls` is a bit more succinct. Personally, I would prefer to do away with `RightCongruence.QuotType` and just write `Quotient` explicitly, but I am fine leaving it in this form where it is an abbrev used to infer the setoid when possible.

Lastly I removed the `HasEquiv` notation, which I don't think was doing what you intended, as this was on the setoid itself. I don't see anywhere this was used, but please clarify if you think this is still needed.

ctchou:
@chenson2018  Your changes look good to me.  I made one further change: I gave a name `right_cov` to the `CovariantClass` field of RightCongruence.  One can imagine defining a `LeftCongruence` with a `left_cov` field and a (two-sided) `Congruence` with both `left_cov` and `right_cov` fields in the future.

Also, I will close #265 and merge it into this PR, because the changes you made impact both PR's.

chenson2018:
This looks like a good change. Sorry about pushing changes here, but it was just too hard to see how different definitions worked out otherwise. I will review the rest of this PR once Fabrizio merges #249.

ctchou:
Merge in the current upstream/main and fix a conflict with the new IsRegular in Mathlib.Algebra.Regular.Defs.

fmontesi:
@chenson2018 fyi, I've merged #249  :)

ctchou:
Merge in the current upstream/main, which now contains #249.

Cslib/Computability/Languages/Congruences/BuchiCongruence.lean
```diff
@@ -0,0 +1,85 @@
...
+    · obtain ⟨r, h1, _⟩ := LTS.pairLang_split h
+      grind [<= LTS.pairLang_append, h_eq.left.mp h1]
+    · obtain ⟨r, h1, _⟩ := LTS.pairLang_split h
+      grind [<= LTS.pairLang_append, h_eq.left.mpr h1]
+    · obtain ⟨r, ⟨h1, _⟩ | ⟨h1, _⟩⟩ := LTS.pairViaLang_split h
+      · grind [<= LTS.pairViaLang_append_pairLang, h_eq.right.mp h1]
+      · grind [<= LTS.pairLang_append_pairViaLang, h_eq.left.mp h1]
+    · obtain ⟨r, ⟨h1, _⟩ | ⟨h1, _⟩⟩ := LTS.pairViaLang_split h
+      · grind [<= LTS.pairViaLang_append_pairLang, h_eq.right.mpr h1]
+      · grind [<= LTS.pairLang_append_pairViaLang, h_eq.left.mpr h1]
```
chenson2018: There is no need to do all this splitting manually:
```suggestion
    grind [Covariant, → LTS.pairLang_split, <= LTS.pairLang_append, → LTS.pairViaLang_split,
      <= LTS.pairViaLang_append_pairLang, <= LTS.pairLang_append_pairViaLang]
```

I would add a grind attribute to `Covariant` where `RightCongruence` is defined to help. I know opening the `LTS` scope for `grind` is problematic, so probably fine that the rest are manually specified.

Cslib/Foundations/Semantics/LTS/Basic.lean
```diff
@@ -215,6 +215,26 @@ theorem LTS.mTr_isExecution_iff : lts.MTr s1 μs s2 ↔
...
+theorem LTS.IsExecution.split
+    {lts : LTS State Label} {s t : State} {μs : List Label} {ss : List State}
+    (he : lts.IsExecution s μs t ss) (n : ℕ) (hn : n ≤ μs.length) :
+    lts.IsExecution s (μs.take n) (ss[n]'(by grind)) (ss.take (n + 1)) ∧
+    lts.IsExecution (ss[n]'(by grind)) (μs.drop n) t (ss.drop n) := by
+  split_ands
+  · grind
+  · use by grind, by grind, ?_, by grind
+    simp [List.getElem_drop, show n + (ss.length - n - 1) = ss.length - 1 by grind]
+    grind
```
chenson2018: I feel this part of the library has some issues with `Prop`-valued `def`s (like `LTS.mTr_isExecution`). When we do the refactoring pass we've discussed, I'd like to do something like make them structures or at least irreducible.

In the meantime, you can at least unnest the use of `simp` and `grind` a bit:

```suggestion
  have : n + (ss.length - n - 1) = ss.length - 1 := by grind
  simp [IsExecution]
  grind
```

chenson2018 (changes requested):
(sorry, I accidentally hit the approve button instead of "request changes")

Cslib/Computability/Languages/Congruences/BuchiCongruence.lean
ctchou: Wow!  I didn't realize `grind` can do this.  Fixed.

Cslib/Foundations/Semantics/LTS/Basic.lean

Cslib/Computability/Languages/Congruences/RightCongruence.lean
```diff
@@ -0,0 +1,41 @@
...
+NOTE: Left congruences and two-sided congruences can be similarly defined.
+But they are left to future work because they are not needed for now.
+-/
+
+namespace Cslib
+
+/-- A right congruence is an equivalence relation on finite sequences (represented by lists)
+that is preserved by concatenation on the right.  The equivalence relation is represented
+by a setoid to to enable ready access to the quotient construction. -/
+class RightCongruence (α : Type*) extends eq : Setoid (List α) where
```
chenson2018, line 28: I don't understand why your last commit changed this, can you explain? It seems better to me as it was with it extending both classes.

ctchou: I like this better because the two-sided congruence can be defined as extending both the right congruence and the left congruence.  I tried to do that with the old definition but was not able to.

ctchou:
@chenson2018 Any more comments?  I explained why I changed the definition of `RightCongruence` slightly.  But if you really don't like it, I can revert it back.

chenson2018:
> @chenson2018 Any more comments? I explained why I changed the definition of `RightCongruence` slightly. But if you really don't like it, I can revert it back.

@ctchou  Where is that explanation? I don't see a reply to that comment.

Edit: After discussing on DM, I find this reasonable in preparation for also having left congruence. There's a question of `CovariantClass` versus just `Covariant`, but this is easily changed later.

chenson2018 (approved):
I feel somewhat out of my depth with the quotient definitions, but all of my comments have been addressed.

PR #275, opened 2026-01-21
feat: query complexity model for algorithms

Shreyas4991:
Generalizing @tannerduve 's free monad take on the query complexity idea. This model of algorithmic complexity provides a lightweight approach to complexity verification of algorithms, similar to #165 

However it offers several improvements over #165  :
1. No explicit ticks needed, so the chances of making mistakes like forgetting to count specific calls is removed.
2. The operations which are counted are made explicit as an inductive type and a model of this type.
3. The model of a query type accepts custom cost structures so long as you can define addition and a 0 operation on them (basically additive monoids suffice).
4. With a notion of reductions between models, this approach is more modular for the algorithm specifier, allowing high level specs of operations which can be translated into simpler query models later. 

~~**Update to the below drawback** : The model requires users to choose a cost for each pure operation upfront by a typeclass instance for a given type of `Cost`. This is a departure from `TimeM`s model. But with custom Cost types with a specific field for `pure`, these costs can be separated from that of calls to queries.~~

One drawback : It is still possible to sneak in free operations inside `pure`. However this is unavoidable in any lightweight monadic approach. A deeply embedded DSL is the only foolproof way to avoid this. Nevertheless this approach removes annotation and review burden and ensures that any actual call to a query will be counted. Thus it is easy to notice when a monadic operation is not being called. 

Zulip Discussion Thread: https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/Query.20model.20for.20Algorithms.20complexity.20.3A.20updates/near/569606456

PR #280, opened 2026-01-23
feat: Adding proofs of correctness and time complexity of insertion sort

bhargavkulk:
I have added the file `InsertionSort.lean` in this PR. I am aware that `List.insertionSort` exists, but it is not instrumented with `tick`, which does not make it possible to write the complexity proof. I have chosen the worst case time bound as `n^2` but the same proof discharges for `n(n+1)/2`. I know I have not added the required documentation in this PR, but I first want to get an OK for the actual proofs implemented.

I think insertion sort, while a very simple algorithm, merits its inclusion because more complicated sorts implemented in the standard libraries of programming languages (like "TimSort" in Python) use insertion sort to quickly sort smaller sub-arrays as insertion sort is faster than divide-and-conquer based comparison sorts like merge and quick sort for very small arrays. I believe that the proofs in this file would aid future proofs about these hybrid in-actual-use sorts.

### DIFF
--- a/Cslib/Algorithms/Lean/InsertionSort/InsertionSort.lean
+++ b/Cslib/Algorithms/Lean/InsertionSort/InsertionSort.lean
@@ -17,7 +17,7 @@ public import Mathlib.Data.Nat.Log
 # Insertion Sort on a list
 
 In this file we introduce `insert` and `insertionSort` algorithms that returns a time monad
-over the list `TimeM (List α)`. The time complexity of `insertionSort` is the number of comparisons.
+over the list `TimeM ℕ (List α)`. The time complexity of `insertionSort` is the number of comparisons.
 
 --
 ## Main results
@@ -34,11 +34,11 @@ namespace Cslib.Algorithms.Lean.TimeM
 variable {α : Type} [LinearOrder α]
 
 /-- Inserts an element into a list, counting comparisons as time cost.
-Returns a `TimeM (List α)` where the time represents the number of comparisons performed. -/
-def insert : α → List α → TimeM (List α)
+Returns a `TimeM ℕ (List α)` where the time represents the number of comparisons performed. -/
+def insert : α → List α → TimeM ℕ (List α)
   | x, [] => return [x]
   | x, y :: ys => do
-    let c ← ✓ (x ≤ y : Bool)
+    ✓ let c := (x ≤ y : Bool)
     if c then
       return (x :: y :: ys)
     else
@@ -46,8 +46,8 @@ def insert : α → List α → TimeM (List α)
       return (y :: rest)
 
 /-- Sorts a list using the insertion sort algorithm, counting comparisons as time cost.
-Returns a `TimeM (List α)` where the time represents the total number of comparisons. -/
-def insertionSort (xs : List α) : TimeM (List α) := do
+Returns a `TimeM ℕ (List α)` where the time represents the total number of comparisons. -/
+def insertionSort (xs : List α) : TimeM ℕ (List α) := do
   match xs with
   | [] => return []
   | x :: xs' => do
@@ -61,11 +61,8 @@ open List
 @[grind →]
 theorem mem_either_insert (xs : List α) (a b : α) (hz : a ∈ ⟪insert b xs⟫) : a = b ∨ a ∈ xs := by
   fun_induction insert with
-  | case1 _ => simp only [Pure.pure, pure, mem_cons, not_mem_nil, or_false] at hz; exact Or.inl hz
-  | case2 x y ys ih =>
-    simp_all only [Bind.bind, Pure.pure]
-    simp at hz
-    grind
+  | case1 _ => grind
+  | case2 x y ys ih => grind
 
 /-- A list is sorted if it satisfies the `Pairwise (· ≤ ·)` predicate. -/
 abbrev IsSorted (l : List α) : Prop := List.Pairwise (· ≤ ·) l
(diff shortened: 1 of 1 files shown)


Shreyas4991:
I think this PR should be put on hold until we reach a decision on #275, specifically [this question](https://github.com/leanprover/cslib/pull/275#discussion_r2723854311)

chenson2018:
Hi Bhargav, thanks for interest in contributing! Just to add context to the above comment, I think this sort of PR will be welcome, but there is some infrastructure flux at the moment.

In the meantime could you add the standard copyright header and address the CI failure? Thanks!

bhargavkulk:
I have added the copyright header and fixed the CI error.

fmontesi:
@sorrachai Can you review this code?

bhargavkulk:
@sorrachai I have added doc comments and edited the code to meet the style guide. Let me know id anything else is missing!

Cslib/Algorithms/Lean/InsertionSort/InsertionSort.lean
```diff
@@ -0,0 +1,143 @@
...
+
+set_option autoImplicit false
+
+namespace Cslib.Algorithms.Lean.TimeM
+
+variable {α : Type} [LinearOrder α]
+
+/-- Inserts an element into a list, counting comparisons as time cost.
+Returns a `TimeM (List α)` where the time represents the number of comparisons performed. -/
+def insert : α → List α → TimeM (List α)
```
chenson2018: Please merge `main`. This signature should now use `TimeM ℕ (List α)`.

Cslib/Algorithms/Lean/InsertionSort/InsertionSort.lean
```diff
@@ -0,0 +1,143 @@
...
+open List
+
+@[grind →]
+theorem mem_either_insert (xs : List α) (a b : α) (hz : a ∈ ⟪insert b xs⟫) : a = b ∨ a ∈ xs := by
+  fun_induction insert with
+  | case1 _ => simp only [Pure.pure, pure, mem_cons, not_mem_nil, or_false] at hz; exact Or.inl hz
+  | case2 x y ys ih =>
+    simp_all only [Bind.bind, Pure.pure]
+    simp at hz
+    grind
```
chenson2018: You should not need to simplify `pure` in this way. This can all be just

```suggestion
  fun_induction insert <;> grind
```

Cslib/Algorithms/Lean/InsertionSort/InsertionSort.lean
```diff
@@ -0,0 +1,143 @@
...
+
+/-- A list is sorted if it satisfies the `Pairwise (· ≤ ·)` predicate. -/
+abbrev IsSorted (l : List α) : Prop := List.Pairwise (· ≤ ·) l
```
chenson2018, line 68: I think we should use `List.SortedLE` for this.

At minimum you can't have this duplicated because it will cause a failure in importing this alongside the merge sort module. This declaration `Cslib.Algorithms.Lean.TimeM.IsSorted` should probably never have been namespaced in this way.

Cslib/Algorithms/Lean/InsertionSort/InsertionSort.lean
```diff
@@ -0,0 +1,143 @@
...
+
+theorem sorted_insert {x : α} {xs : List α} (hxs : IsSorted xs) : IsSorted ⟪insert x xs⟫ := by
+  fun_induction insert x xs with
+  | case1 _ => simp
+  | case2 x y ys ih =>
+    simp only [Bind.bind, Pure.pure]
```
chenson2018: Once you merge `main`, you'll find that all of these `simp` are not needed.

Cslib/Algorithms/Lean/InsertionSort/InsertionSort.lean
```diff
@@ -0,0 +1,143 @@
...
+  | case2 x xs ih =>
+    simp [Bind.bind]
+    grind [insert_length]
+
+/-- Time complexity of `insertionSort`. -/
+theorem insertionSort_time (xs : List α) : (insertionSort xs).time ≤ xs.length * xs.length:= by
+  fun_induction insertionSort with
+  | case1 => simp
+  | case2 x xs ih =>
+    simp [Bind.bind]
```
chenson2018: This is the one exception, this should be
```suggestion
    simp only [time_bind, List.length_cons]
```

Shreyas4991:
I suggest merging main. The simp set for TimeM has changed. In fact depending on when main was merged last time, there might be considerable breakage.

Cslib/Algorithms/Lean/InsertionSort/InsertionSort.lean
Shreyas4991: I think SortedLE is deprecated in favour of Pairwise, so this usage is correct. The abbreviation is not necessary however.

Shreyas4991:
Fwiw this PR duplicates List.orderedInsert. See my Prog version of insertionSort in #275.

Cslib/Algorithms/Lean/InsertionSort/InsertionSort.lean
bhargavkulk: `MergeSort.lean` also defines the `IsSorted` abbrev. So, I am not sure if I should inline it.

bhargavkulk:
@chenson2018 I have made the requested changes. See my comment on the `IsSorted` abbrev, I'm not sure if i should inline it; I'll defer to your decision.

Cslib/Algorithms/Lean/InsertionSort/InsertionSort.lean
chenson2018: @Shreyas4991 I'm not sure that is correct. The deprecation on `List.Sorted` does refer to `List.Pairwise`, but the comment specifically says to consider using `SortedLE`, which is not deprecated and has `sortedLE_iff_pairwise` as a `grind =` lemma. I think it is correct to use `SortedLE` here. (And in the merge sort file, but let's make a separate PR)

Shreyas4991:
Just as a heads up, #275  contains insertionSort already. Also the PR does not seem to pass CI  at the moment.

bhargavkulk:
closing this in favor of #372.

PR #281, opened 2026-01-23
feat: prove the Ramsey theorem for infinite graphs

ctchou:
This PR proves the Ramsey theorem for infinite graphs, which will be needed in Buchi's proof that omega-regular languages are closed under complementation.

Ideally, this result should be in mathlib, but it is not and I don't know when it will be:
https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/Infinite.20Ramsey.20theory/with/568507228
So I now prove what I will need.  Once mathlib has the desired result, this proof can be deprecated.

### DIFF
--- a/Cslib/Foundations/Combinatorics/InfiniteGraphRamsey.lean
+++ b/Cslib/Foundations/Combinatorics/InfiniteGraphRamsey.lean
@@ -70,7 +70,7 @@ private lemma goodSelection_exists (ivs : InfVSet Vertex) :
   let f u := color {v, u}
   obtain ⟨c, vs, h_inf, h_vs, h_col⟩ := infinite_pigeonhole_principle f <|
     Set.Infinite.diff ivs.inf (finite_singleton v)
-  simp only [subset_diff, disjoint_singleton_right] at h_vs
+  simp only [subset_diff] at h_vs
   let ivs' := InfVSet.mk vs h_inf
   use {vs := ivs', v := v, c := c}
   grind [GoodSelection]


Cslib/Foundations/Combinatorics/InfiniteGraphRamsey.lean
```diff
@@ -0,0 +1,149 @@
...
+  obtain ⟨y, h_inf'⟩ := Finite.exists_infinite_fiber (s.restrict f)
+  have h_inf_iff := Equiv.infinite_iff <|
+    Equiv.subtypeSubtypeEquivSubtypeInter (· ∈ s) (fun x ↦ f x = y)
+  simp only [coe_eq_subtype, mem_preimage, restrict_apply, mem_singleton_iff, h_inf_iff] at h_inf'
+  have h_inf'' := (infinite_coe_iff (s := { x | x ∈ s ∧ f x = y })).mp h_inf'
+  use y, {x | x ∈ s ∧ f x = y}
+  grind
+
+/-- An `InfVSet` consists of a set of vertices and a proof that the set is infinite. -/
+structure InfVSet (Vertex : Type*) where
```
chenson2018: What parts of this file do you plan to use elsewhere? I don't like these structures, but if all you need for instance is the final `infinite_graph_ramsey` theorem, you could make everything else private and minimize how much is exposed elsewhere until there is the Mathlib replacement you mentioned.

chenson2018: (besides this, everything else seems fine)

ctchou: I've made everything `private` except for the first and the last theorems.

chenson2018:
(that merge made `mk_all` fail because the alphabetic order of imports changed, I have fixed it)

ctchou:
@chenson2018 Thanks for the fix!

PR #284, opened 2026-01-23
feat: syntactic contexts and congruence relations

fmontesi:
This PR introduces notions of (single hole) syntactic contexts (`HasContext`) and congruence relations (`Congruence`), along with the standard notation `c[t]` for 'fill the hole in the context `c` with term `t`'.

The development of CCS is adapted accordingly and should hold for many other calculi. The idea actually comes from an ongoing development on the generalisation of sequent calculi and logics, but these require more concepts. Also, in the future, it might be interesting to study approaches similar to `HasContext` in semantics that use different kinds of contexts.

The PR also scopes the `grind` annotations in CCS.

### DIFF
--- a/Cslib/Foundations/Syntax/Congruence.lean
+++ b/Cslib/Foundations/Syntax/Congruence.lean
@@ -7,14 +7,14 @@ Authors: Fabrizio Montesi
 module
 
 public import Cslib.Foundations.Syntax.Context
+public import Mathlib.Algebra.Order.Monoid.Unbundled.Defs
 
 @[expose] public section
 
 namespace Cslib
 
 /-- An equivalence relation preserved by all contexts. -/
-class Congruence (α : Sort*) [HasContext α] (r : α → α → Prop) extends IsEquiv α r where
-  /-- Equivalence is preserved by contexts. -/
-  is_congruence (a b : α) (c : HasContext.Context α) : r a b → r c[a] c[b]
+class Congruence (α : Type*) [HasContext α] (r : α → α → Prop) extends
+  IsEquiv α r, covariant : CovariantClass (HasContext.Context α) α (·[·]) r
 
 end Cslib
--- a/Cslib/Languages/CCS/BehaviouralTheory.lean
+++ b/Cslib/Languages/CCS/BehaviouralTheory.lean
@@ -434,7 +434,7 @@ theorem bisimilarity_is_congruence
 /-- Bisimilarity is a congruence in CCS. -/
 instance bisimilarityCongruence :
     Congruence (Process Name Constant) (Bisimilarity (lts (defs := defs))) where
-  is_congruence := bisimilarity_is_congruence
+  covariant := ⟨by grind [Covariant, bisimilarity_is_congruence]⟩
 
 end CCS
 


Cslib/Foundations/Syntax/Congruence.lean
```diff
@@ -0,0 +1,20 @@
...
+public import Cslib.Foundations.Syntax.Context
+
+@[expose] public section
+
+namespace Cslib
+
+/-- An equivalence relation preserved by all contexts. -/
+class Congruence (α : Sort*) [HasContext α] (r : α → α → Prop) extends IsEquiv α r where
+  /-- Equivalence is preserved by contexts. -/
+  is_congruence (a b : α) (c : HasContext.Context α) : r a b → r c[a] c[b]
```
ctchou: You may want to take a look at how right congruence over finite words is defined in #278, where @chenson2018 had the idea of using `CovariantClass` to state the congruence condition.  Also, if you ever want to do the quotient construction, you may wan to consider using a `Setoid` instead of `IsEquiv`.

chenson2018: I'm not exactly sure when you're supposed to use `Setoid` versus `IsEquiv`, but I'll note that `Setoid` seems more common (502 versus 20 declarations according to Loogle) and `Quotient.instIsEquivEquiv` exists to convert from a `Setoid` if needed.

fmontesi: I was unaware of CovariantClass, I'll see if I can build on it, thanks!

I was aware of Setoid. I'm not using it on purpose, because it bundles the relation. So it seems to say 'this is the canonical equivalence for this type' to me, whereas the unbundled IsEquiv says 'this particular relation is an equivalence' (which is what I want to say). Do you agree with my interpretation of the difference?

chenson2018: Yes, Eric pointed out `CovariantClass` and it seems to fit well for automata.

This distinction makes sense. The terminology "carries data" applies here I believe. If you are not needing to construct quotients as mentioned above, this seems good.

Cslib/Foundations/Syntax/Context.lean
```diff
@@ -0,0 +1,24 @@
...
+public import Cslib.Init
+
+@[expose] public section
+
+namespace Cslib
+
+/-- Class for types (`Term`) that have a notion of (single-hole) contexts (`Context`). -/
+class HasContext (Term : Sort*) where
+  /-- The type of contexts. -/
+  Context : Sort*
```
chenson2018, line 18: Can you explain why this sort is bundled?

Cslib/Foundations/Syntax/Context.lean
```diff
@@ -0,0 +1,24 @@
...
+  /-- Replaces the hole in the context with a term. -/
+  fill (c : Context) (t : Term) : Term
+
+@[inherit_doc] notation:max c "[" t "]" => HasContext.fill c t
```
chenson2018, line 22: I'm wary of adding any notation using `[` and `]`, as these can be problematic in clashing with `GetElem` for arrays and lists.

Cslib/Foundations/Syntax/Congruence.lean
ctchou: `Setoid` does not strike me as implying any notion of being a "canonical" equivalence relation.  It's just a relation bundled with a proof that the relation is an equivalence relation.  What really surprised me is that there are two apparently equivalent ways to state that a relation is an equivalence relations:
https://leanprover-community.github.io/mathlib4_docs/Mathlib/Order/Defs/Unbundled.html#IsEquiv
https://leanprover-community.github.io/mathlib4_docs/Init/Core.html#Equivalence
There are some discussions on Zulip about this and other duplications:
https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/Relation.20properties.20duplication

chenson2018: The problem is similar to when I was confused by the Setoid instance for lists. When you have bundled a particular relation into a typeclass, as opposed to a proposition about the relation unbundled, you are in some sense making it canonical since we try to avoid typeclass diamonds.

ctchou: Well, in my case I knew I needed to do the quotient construction from the beginning.  So `Setoid` is the natural way to go.

chenson2018: Yes, but the quotient part is irrelevant to the point we are making. (Though if you did want quotients here, I would agree with you it's natural to use Setoid)

`Setoid` bundles in a particular relation, and if that is made an instance, anywhere else that has access to this instance is in some sense forced to use it, or at minimum avoid a diamond with it. The difference here is between a typeclass `Setoid` that carries this relation as data, versus `isEquiv` which is a proposition.

Cslib/Foundations/Syntax/Context.lean
fmontesi: For every syntax/grammar, the notion of single hole context (as defined in CCS) is typically canonical. This enables writing [HasContext Term] instead of [IsContext Term Context], which the refactoring of CCS immediately showed cumbersome. This is consistent with HasEquiv, which bundles the relation.

One can then have non-canonical notions of context, like sometimes in semantics there are many evaluation contexts, but those are not designed to reconstruct the grammar from atoms as the 'normal' contexts (see CCS.Context.complete for what I mean). We can revisit this when we look at those, maybe when it comes to giving access to the syntax. Maybe that should be allowed for the unbundled IsContext, and then HasContext is a special case.

Cslib/Foundations/Syntax/Context.lean
fmontesi: Didn't seem to be a problem in any situation I tried, lean could manage. If it becomes an issue we can revert this.

This is a *very* standard notation, so we should have a good reason not to have it.

chenson2018: We can wait and see, I just felt it best to point out now. It may not be an issue depending on where all this ends up being used.

Cslib/Foundations/Syntax/Context.lean
chenson2018: Okay, I trust your judgment on this; it sound like you've thought this out.

Cslib/Foundations/Syntax/Congruence.lean
```diff
@@ -0,0 +1,20 @@
...
+module
+
+public import Cslib.Foundations.Syntax.Context
+
+@[expose] public section
+
+namespace Cslib
+
+/-- An equivalence relation preserved by all contexts. -/
+class Congruence (α : Sort*) [HasContext α] (r : α → α → Prop) extends IsEquiv α r where
```
chenson2018: Does `α` need to be a `Sort` here? This builds if you make it a `Type`, so I wondered if you did this just because you were extending `IsEquiv`.

If this can be in `Type`, the following is equivalent:
```lean
class Congruence (α : Type*) [HasContext α] (r : α → α → Prop) extends
  IsEquiv α r,
  is_congruence : CovariantClass (HasContext.Context α) α (·[·]) r
```
up to ordering of arguments, e.g.
```lean
instance bisimilarityCongruence :
    Congruence (Process Name Constant) (Bisimilarity (lts (defs := defs))) where
  is_congruence := ⟨by grind [Covariant, bisimilarity_is_congruence]⟩
```
works in `CCS.BehaviouralTheory`.

If we end up using this sort of thing in multiple places (i.e. here and @ctchou's #278), I'd prefer that we use the same naming like `covariant : CovariantClass ...` so that you get the constructor `covariant.elim : Covariant ...`.

ctchou: In #278, I used the name `right_cov`, because one can also define a congruence that is both a left congruence and a right congruence (in which case there will be two covariant fields).  Perhaps `context_cov` is a good name here.

fmontesi:
I was doing it to be consistent with HasEquiv in Lean core, because I don't
need it to be a type, strictly speaking.

But I think using Type is fine for our purposes, so integrating with
mathlib should take priority here. I'll change it to what you suggest, I
think it's a good idea!

fmontesi:
All done. Looks pretty. :-)

PR #286, opened 2026-01-24
feat: define the `Transducer` class as recognizing a function from input strings to weights

LiamSchilling:
This definition easily specializes to:
- weighted acceptors, by attaching an annihilator `none` to the weights
- functional string transducers, by having output strings as the weights
- partial functional string transducers, by having output strings with an annihilator `none` as the weights

The design is similar to the related `Acceptor.lean`

### DIFF
--- a/Cslib/Computability/Transductions/RationalTransduction.lean
+++ b/Cslib/Computability/Transductions/RationalTransduction.lean
@@ -12,6 +12,14 @@ public import Cslib.Computability.Transductions.Transduction
 @[expose] public section
 
 /-! # Rational Transductions
+
+A transduction `IsRational` if it is recognized by a deterministic printer
+with finite-state look-left and look-right (a `Bimachine`).
+Equivalently, a rational transduction is recognized by
+a nondeterministic finite-state left-to-right reader-printer.
+A transduction `IsSubsequential` if it is recognized by
+a deterministic finite-state left-to-right reader-printer (a `DetTransducer`).
+All subsequential transductions are rational transductions.
 -/
 
 namespace Cslib.Transduction
@@ -31,7 +39,7 @@ def IsSubsequential (f : Transduction Symbol Weight) : Prop :=
   ∃ σ : Type, ∃ _ : Fintype σ, ∃ dt : DetTransducer σ Symbol Weight,
     transduceLeft dt = f
 
-/-- A rational function is recognized by a bimachine. -/
+/-- A rational transduction is recognized by a bimachine. -/
 def IsRational (f : Transduction Symbol Weight) : Prop :=
   ∃ σl σr : Type, ∃ _ : Fintype σl, ∃ _ : Fintype σr, ∃ bm : Bimachine σl σr Symbol Weight,
     transduceLeft bm = f


Cslib/Computability/Automata/Transducers/Transducer.lean
```diff
@@ -0,0 +1,93 @@
...
+
+The definition specializes to:
+- weighted acceptors,
+  by attaching an annihilator `none` to the weights
+- functional string transducers,
+  by having output strings as the weights
+- partial functional string transducers,
+  by having output strings with an annihilator `none` as the weights -/
+class Transducer
+    (T : Type u) (Symbol : outParam (Type v)) (Weight : outParam (Type w)) [Semigroup Weight] where
```
ctchou: What is the purpose of the `T` and `t : T` arguments here and below?  They never seem to be used.

Cslib/Computability/Automata/Transducers/Transducer.lean
```diff
@@ -0,0 +1,93 @@
...
+namespace Transducer
+
+/- We choose the direction of simplification that pulls multiplication into the accumulator.
+The motivation is cases like `transduceFrom t xs 1 1 * acc = transduceFrom t xs 1 acc`,
+which are not handled by the other choice of direction. -/
+attribute [simp, scoped grind =] Transducer.mul_transduceFrom Transducer.transduceFrom_mul
+
+variable {Symbol : Type v} {Weight : Type w}
+
+section
```
ctchou: Please give a name to this section.

LiamSchilling:
A good reference for weighed automata is Droste et al. 2009. As in the chapter Mohri 2009, weighted automata have shortest-distance interpretations when weights are from an idempotent semiring. In this commit, just a semigroup is sufficient for stating the definition.

Cslib/Computability/Automata/Transducers/Transducer.lean
LiamSchilling: `T` is a type of transducers implementing the type class `Transducer`. The function recognized by some transducer `t : T` can then be accessed via `transduce t`.

e.g. If we have `def WeightedDFA : Type := ...`, then the function recognized by a `wdfa : WeightedDFA` will be implemented as `transduceFrom wdfa` in `instance : Transducer WeightedDFA Symbol Weight where ...`.

In this sense, `T` and `t : T` are analogous to `A` and `a : A` in `Acceptor.lean`.

Cslib/Computability/Automata/Transducers/Transducer.lean
LiamSchilling: Thanks for the feedback 👍 This is now named `Monoid` as the section that takes weights to be from a `Monoid`.

Cslib/Computability/Automata/Transducers/Transducer.lean
ctchou: Thanks for the explanation.  Perhaps something like you wrote above should be part of the doc-string.

LiamSchilling:
Thanks for all the feedback @ctchou. Here's a second iteration that works with an implementation of typical deterministic FSTs (`DetTransducer`) and bimachines (`Bimachine`) in a downstream branch. It's a lot of code altogether, so I'm thinking it's best to split into 3 PRs:
- this one, defining `Transducer`
- one defining `DetTransducer` with a `Transducer` implementation
- one defining `Bimachine` with a `Transducer` implementation

Cslib/Computability/Automata/DT/Basic.lean
```diff
@@ -0,0 +1,272 @@
...
+
+## References
+
+* [Stoyan Mihov, Klaus U. Schulz,
+  *Finite-State Techniques: Automata, Transducers and Bimachines*][Mihov2019]
+-/
+
+namespace Cslib.Automata
+
+namespace DA
```
ctchou: I think you can merge these two into `namespace Cslib.Automata.DA`.

Cslib/Computability/Transductions/Algebra.lean
```diff
@@ -0,0 +1,87 @@
...
+@[expose] public section
+
+/-! ## Left GCD Monoids
+
+A left GCD monoid is a generalization of `GCDMonoid` that relaxes commutativity.
+
+A GCD will be necessary for defining the tail function of a transduction,
+but commutativity is too strict to accommodate cases we care about
+such as string-to-string transduction.
+-/
```
ctchou: This file is pure algebra.  I think it should be moved to `Cslib/Foundations/Algebra/LeftGCDMonoid.lean`.

Cslib/Computability/Transductions/Transduction.lean
```diff
@@ -0,0 +1,74 @@
...
+which are typically from a `Monoid` or a `Semiring`.
+A *string-to-string* transduction has output strings as its weights.
+-/
+
+namespace Cslib
+
+open Semigroup LeftSetGCDMonoid
+
+/- A transduction is a function with string inputs. -/
+def Transduction (α β) :=
```
ctchou: \alpha and \beta are undecleared.

ctchou (changes requested):

Cslib/Foundations/Algebra/Div.lean
```diff
@@ -0,0 +1,26 @@
...
+Authors: William (Liam) Schilling
+-/
+
+module
+
+public import Cslib.Init
+public import Mathlib.Algebra.Divisibility.Basic
+
+@[expose] public section
+
```
ctchou, line 13: There should be a `/-! ... -/` comment about what this file is for.

Cslib/Foundations/Algebra/Div.lean
```diff
@@ -0,0 +1,26 @@
...
+namespace Semigroup
```
ctchou, line 14: If the purpose of this file is to prove additional results about semigroup that are not in mathlib, this file should be named `Semigroup.lean`.

ctchou (changes requested):
Sorry, I did not mean to approve.  I was requesting changes in the last batch of comments.

PR #211, opened 2025-12-11
feat: add several definitions and theorems about reduction systems

IvanRenison:
This PR adds some additional API to `ReductionSystem`. While it duplicates from the relation namespace, this seems like an acceptable tradeoff for unifying proofs that use the corresponding notation, and ultimately a user can still simplify to the underlying relation if they desire. (description added by Chris Henson)

PR #299, opened 2026-01-28
feat(Computability): add Unlimited Register Machines

jessealama:
## Summary

This PR introduces Unlimited Register Machines (URMs) and extends `ReductionSystem` with general termination concepts.

### URM definitions

**Core definitions** (`Defs.lean`):
- `Instr`: The four instruction types (Z, S, T, J)
- `Program`: Lists of instructions
- `Regs`: Register contents as `ℕ → ℕ`
- `State`: Machine configuration (program counter + registers)

**Basic lemmas** (`Basic.lean`):
- Register read/write lemmas, state lemmas
- Jump instruction predicates and bounded jump targets

**Execution semantics** (`Execution.lean`):
- `Step`/`Steps`: Single-step and multi-step execution
- `stepRs`: ReductionSystem integration
- `Halts`, `Diverges`, `HaltsWithResult`: Defined via `ReductionSystem.Terminates`/`Diverges`
- `eval`: Evaluation as a partial function
- `ProgramEquiv`: Program equivalence with `Setoid` instance
- Notation: `p ↓ inputs`, `p ↑ inputs`, `p ↓ inputs ≫ result`, `s ⭢ᵉ s'`, `s ↠ᵉ s'`

**Computability** (`Computable.lean`):
- `Computes`, `Computable`: URM-computability definitions

**Straight-line programs** (`StraightLine.lean`):
- `IsStraightLine`, `straight_line_halts`

**Standard form** (`StandardForm.lean`):
- `IsStandardForm`, `toStandardForm`, equivalence and idempotence theorems

### ReductionSystem extensions (`Basic.lean`)

- `Deterministic` class: each term has at most one successor
- `Terminates(Halted, t)`: term can reach a halted state
- `Diverges(Halted, t)`: term cannot reach any halted state
- `Deterministic.unique_halted`: uniqueness of reachable halted states

### LTS connection (`LTS/Basic.lean`)

- `Trapped`: state cannot reach terminated state (corresponds to `Diverges`)
- `Functional`: at most one successor regardless of label (corresponds to `Deterministic`)
- `toReductionSystem`: convert LTS to ReductionSystem by forgetting labels
- Equivalence theorems: `MayTerminate ↔ Terminates`, `Trapped ↔ Diverges`

### Design decisions

- 0-indexed registers (Lean/Mathlib convention)
- Functional register representation (`Regs := ℕ → ℕ`)
- Relational semantics following Mathlib's Turing machine style
- URM `Halts`/`Diverges` defined via general `ReductionSystem` concepts

### DIFF
--- a/Cslib/Computability/URM/StandardForm.lean
+++ b/Cslib/Computability/URM/StandardForm.lean
@@ -82,11 +82,13 @@ theorem straight_line_IsStandardForm {p : Program} (hsl : p.IsStraightLine) :
   intro instr hinstr
   exact Instr.jumpsBoundedBy_of_nonJump (hsl instr hinstr) p.length
 
-/-! ## Functional Equivalence
+/-! ## Bisimulation
 
-`p` and `p.toStandardForm` are functionally equivalent: they halt on the same
-inputs with the same results. The key insight is that jumps with target `q > p.length`
-land in halted states in both programs. -/
+`p` and `p.toStandardForm` are bisimilar: each step in one program corresponds to a
+step in the other that either reaches the same state or reaches a halted state with the
+same registers. This bisimulation implies functional equivalence (`eval_toStandardForm`).
+The key insight is that jumps with target `q > p.length` land in halted states in both
+programs. -/
 
 /-- Forward step correspondence: if p steps from s to s', then either:
     (1) p.toStandardForm steps from s to s' (same step), or


Cslib/Computability/Urm/Basic.lean
```diff
@@ -0,0 +1,49 @@
...
+public import Cslib.Computability.Urm.Instr
+public import Cslib.Computability.Urm.State
+public import Cslib.Computability.Urm.Program
+public import Cslib.Computability.Urm.Config
+public import Cslib.Computability.Urm.Execution
+public import Cslib.Computability.Urm.StraightLine
+public import Cslib.Computability.Urm.StandardForm
+public import Cslib.Computability.Urm.Computable
+
+/-! # URM Basic Definitions - Hub File
```
BoltonBailey: Generally, we don't create files without definitions like this, if another file needs to import results from multiple files in a folder it will just import them separately.

`Basic.lean` is usually a file name used for files that create basic definitions for a type (or `Defs.lean`, if it doesn't have lemmas). Perhaps it would make sense to move some of the definitions you have at the start of other files to a `Defs.lean` and make all of the files in the folder depend on that?

Cslib/Computability/Urm/Execution.lean
```diff
@@ -0,0 +1,263 @@
...
+-/
+@[grind]
+inductive Step : Config → Config → Prop where
+  | zero {c : Config} {n : ℕ}
+      (h : p[c.pc]? = some (Instr.Z n)) :
+      Step c ⟨c.pc + 1, c.state.write n 0⟩
+  | succ {c : Config} {n : ℕ}
+      (h : p[c.pc]? = some (Instr.S n)) :
+      Step c ⟨c.pc + 1, c.state.write n (c.state.read n + 1)⟩
+  | trans {c : Config} {m n : ℕ}
```
BoltonBailey: `trans` in Mathlib/Lean generally usually refers to "transitivity", but I take it this refers to "transfer" here. Perhaps it would be best to not abbreviate to avoid this collision.

Cslib/Computability/Urm/State.lean
```diff
@@ -0,0 +1,71 @@
...
+
+@[expose] public section
+
+namespace Cslib.Urm
+
+/-- Register state: maps register indices to natural number contents.
+
+Uses the functional representation `ℕ → ℕ` for efficiency with rewrites,
+following the advice from the `grind` tactic documentation. -/
+abbrev State := ℕ → ℕ
```
BoltonBailey: Representing states as functions from Nat to Nat is convenient here, but I worry that it would lead to problems down the road. The other choice I have in mind is to let the registers of a UTM be indexed by an arbitrary (or perhaps `[Fintype]`) type, and then have State be defined `State (m : UTM) := u.RegisterIndex -> Nat` as machine-dependent type. You would have to include this extra variable. But the benefit would be that if you want to define a composition of the functionalities of different UTMs (like I do for TMs in #269), you can do so by taking a direct product of the individual register indices, which would hopefully make functions on that register type easier to define. It would also have the benefit, if the register type were finite, of having the register state be printable data.

Perhaps this change is not necessary since the textbook defines it this way, but it's something to think about.

Cslib/Computability/Urm/StraightLine.lean
```diff
@@ -0,0 +1,144 @@
...
+
+namespace Cslib.Urm
+
+/-! ## Straight-Line Programs -/
+
+/-- A program is "straight-line" if it contains no jump instructions. -/
+def Program.is_straight_line (p : Program) : Bool :=
+  p.all (fun i => !i.isJump)
+
+/-- Prop version: a program is straight-line (no jumps). -/
```
BoltonBailey: `Bool`s can be coerced to `Prop`s in this way, so I don't know if this function is necessary.

Cslib/Computability/Urm/Instr.lean
```diff
@@ -0,0 +1,218 @@
...
+
+/-- J instruction is a jump (Prop version). -/
+theorem J_IsJump (m n q : ℕ) : (J m n q).IsJump := rfl
+
+/-- shift_jumps is identity for non-jumping instructions. -/
+theorem shift_jumps_of_not_isJump {instr : Instr}
+    (h : instr.isJump = false) (offset : ℕ) : instr.shift_jumps offset = instr := by
+  cases instr with
+  | Z _ | S _ | T _ _ => rfl
+  | J _ _ _ => simp [isJump] at h
```
chenson2018: `grind` works here

Cslib/Computability/Urm/Instr.lean
```diff
@@ -0,0 +1,218 @@
...
+def JumpsBoundedBy (len : ℕ) (instr : Instr) : Prop := instr.jumps_bounded_by len = true
+
+instance (len : ℕ) (instr : Instr) : Decidable (instr.JumpsBoundedBy len) :=
+  decidable_of_iff (instr.jumps_bounded_by len = true) Iff.rfl
+
+/-- Non-jumping instructions have bounded jumps for any length. -/
+theorem JumpsBoundedBy_of_not_IsJump {instr : Instr} (h : ¬instr.IsJump)
+    (len : ℕ) : instr.JumpsBoundedBy len := by
+  unfold IsJump at h; simp only [Bool.not_eq_true] at h
+  cases instr <;> grind [isJump, jumps_bounded_by, JumpsBoundedBy]
```
chenson2018: This `unfold` is not needed. The minimal change would be to place `IsJump` in the `simp`, but I would personally rewrite to

```suggestion
  grind [IsJump, isJump, jumps_bounded_by, JumpsBoundedBy]
```

Cslib/Computability/Urm/Instr.lean
```diff
@@ -0,0 +1,218 @@
...
+
+/-! ## Jump Instructions -/
+
+/-- An instruction is a jump instruction if it's J. -/
+def isJump : Instr → Bool
+  | J _ _ _ => true
+  | _ => false
+
+/-- Prop version: an instruction is a jump. -/
+def IsJump (instr : Instr) : Prop := instr.isJump = true
```
chenson2018: I think that all of this duplication between `Bool` and `Prop` is unnecessary.

Cslib/Computability/Urm/Instr.lean
```diff
@@ -0,0 +1,218 @@
...
+  cases instr <;> grind [isJump, jumps_bounded_by, JumpsBoundedBy]
+
+/-- jumps_bounded_by is monotonic: if bounded for len1, then bounded for any len2 ≥ len1. -/
+theorem JumpsBoundedBy.mono {instr : Instr} {len1 len2 : ℕ}
+    (h : instr.JumpsBoundedBy len1) (hle : len1 ≤ len2) :
+    instr.JumpsBoundedBy len2 := by
+  unfold JumpsBoundedBy at h ⊢
+  cases instr with
+  | Z _ | S _ | T _ _ => simp [jumps_bounded_by]
+  | J _ _ q => simp only [jumps_bounded_by, decide_eq_true_eq] at h ⊢; omega
```
chenson2018: You're not obliged to use `grind` everywhere but it works well here

```suggestion
  grind [JumpsBoundedBy, Instr.shift_jumps, jumps_bounded_by]
```

Cslib/Computability/Urm/Program.lean
```diff
@@ -0,0 +1,61 @@
...
+/-- The maximum register index referenced by any instruction in the program. -/
+@[scoped grind =]
+def max_register (p : Program) : ℕ :=
+  p.foldl (fun acc instr => max acc instr.max_register) 0
+
+/-- Shift all jump targets in a program by `offset`.
+Used when concatenating programs: the second program's jumps must be adjusted
+by the length of the first program. -/
+@[scoped grind =]
+def shift_jumps (offset : ℕ) (p : Program) : Program :=
```
chenson2018: Not too important, but I feel like the `Program` argument would usually go first here and below in `shift_registers`.

Cslib/Computability/URM/Execution.lean
```diff
@@ -0,0 +1,263 @@
...
+
+Each constructor corresponds to one of the four instruction types:
+- `zero`: Execute `Z n` (set register n to 0)
+- `succ`: Execute `S n` (increment register n)
+- `trans`: Execute `T m n` (copy register m to register n)
+- `jump_eq`: Execute `J m n q` when registers m and n are equal (jump to q)
+- `jump_ne`: Execute `J m n q` when registers m and n differ (proceed to next)
+-/
+@[grind]
+inductive Step : Config → Config → Prop where
```
chenson2018: If you use
```suggestion
@[grind, reduction_sys rs "ᵉ"]
inductive Step : Config → Config → Prop where
```
it will set up notation `⭢ᵉ` and `↠ᵉ` for the reflexive transitive closure. You can use whatever symbol you'd like, and it parameterizes the notation accordingly. See the lambda calculus modules for examples.

Cslib/Computability/Urm/Execution.lean
chenson2018: I agree, and think since this is something you'll take the reflexive transitive closure of makes it additionally confusing.

Edit: GitHub paced this comment oddly, I was saying I agreed about `trans` not being a great name.

Cslib/Computability/URM/Execution.lean
```diff
@@ -0,0 +1,263 @@
...
+/-- Multi-step execution: the reflexive-transitive closure of `Step`. -/
+abbrev Steps : Config → Config → Prop := Relation.ReflTransGen (Step p)
+
+namespace Step
+
+variable {p : Program}
+
+/-- The step relation is deterministic: each configuration has at most one successor. -/
+theorem deterministic {c c' c'' : Config} (h1 : Step p c c') (h2 : Step p c c'') : c' = c'' := by
+  cases h1 <;> cases h2 <;> grind
```
chenson2018:
```suggestion
  grind
```

Cslib/Computability/URM/Execution.lean
```diff
@@ -0,0 +1,263 @@
...
+
+/-- A halted configuration has no successor in the step relation. -/
+theorem no_step_of_halted {c c' : Config} (hhalted : c.is_halted p) : ¬Step p c c' := by
+  intro hstep
+  cases hstep <;> grind [Config.is_halted]
```
chenson2018:
```suggestion
  grind [Config.is_halted]
```

Cslib/Computability/URM/Execution.lean
```diff
@@ -0,0 +1,263 @@
...
+      (h : p[c.pc]? = some (Instr.J m n q))
+      (heq : c.state.read m = c.state.read n) :
+      Step c ⟨q, c.state⟩
+  | jump_ne {c : Config} {m n q : ℕ}
+      (h : p[c.pc]? = some (Instr.J m n q))
+      (hne : c.state.read m ≠ c.state.read n) :
+      Step c ⟨c.pc + 1, c.state⟩
+
+/-- Multi-step execution: the reflexive-transitive closure of `Step`. -/
+abbrev Steps : Config → Config → Prop := Relation.ReflTransGen (Step p)
```
chenson2018: This also gets defined with the above attribute.

Cslib/Computability/URM/Execution.lean
```diff
@@ -0,0 +1,263 @@
...
+  | jump_eq _ _ | jump_ne _ _ => rfl
+
+end Step
+
+namespace Steps
+
+variable {p : Program}
+
+/-- Steps compose transitively. -/
+theorem trans {c₁ c₂ c₃ : Config} (h1 : Steps p c₁ c₂) (h2 : Steps p c₂ c₃) : Steps p c₁ c₃ :=
```
chenson2018: If you use the notation I have described above, you should get this API for free from the bundled `ReductionSystem` it generates. (This is new from #211, let me know if you hit any issues)

Cslib/Computability/URM/Execution.lean
```diff
@@ -0,0 +1,263 @@
...
+theorem preserves_register {c c' : Config} {r : ℕ}
+    (hsteps : Steps p c c')
+    (hr : ∀ instr, instr ∈ p → instr.writes_to ≠ some r) :
+    c'.state.read r = c.state.read r := by
+  induction hsteps using Relation.ReflTransGen.head_induction_on with
+  | refl => rfl
+  | head hstep _ ih =>
+    rw [ih]
+    exact Step.preserves_register hstep fun instr hinstr =>
+      hr instr (List.mem_of_getElem? hinstr)
```
chenson2018: I'm not marking all the `grind` simplifications possible (they are not a requirement), but as an example:
```suggestion
  induction hsteps <;> grind [Step.preserves_register]
```

Cslib/Computability/URM/Execution.lean
```diff
@@ -0,0 +1,263 @@
...
+    cases h2 using Relation.ReflTransGen.head_induction_on with
+    | refl =>
+      -- c₂ = init is halted, but init can step - contradiction
+      exact absurd hstep_c (Step.no_step_of_halted hh2)
+    | head hstep_c' hrest' =>
+      -- init → c' → ... → c₂
+      -- By determinism, c = c'
+      have heq : _ = _ := Step.deterministic hstep_c hstep_c'
+      subst heq
+      exact ih hrest'
```
chenson2018: Another example:
```lean
theorem eq_of_halts {init c₁ c₂ : Config}
    (h1 : Steps p init c₁) (hh1 : c₁.is_halted p)
    (h2 : Steps p init c₂) (hh2 : c₂.is_halted p) : c₁ = c₂ := by
  induction h1 using Relation.ReflTransGen.head_induction_on with
  | refl => grind [cases Step, Step.no_step_of_halted]
  | head =>
    cases h2 using Relation.ReflTransGen.head_induction_on with
    | refl => grind [Step.no_step_of_halted]
    | head => grind [Step.deterministic]
```
Maybe some of the recurring theorems here could be forward rules. I'd also omit the comments, it seems clear enough already.

Cslib/Computability/URM/Execution.lean
```diff
@@ -0,0 +1,263 @@
...
+with `result` in R[0]. -/
+scoped notation:50 p " ↓ " inputs:51 " ≫ " result:51 => HaltsWithResult p inputs result
+
+namespace HaltsWithResult
+
+variable {p : Program}
+
+/-- If a program halts with a result, it halts. -/
+theorem toHalts {inputs : List ℕ} {result : ℕ}
+    (h : p ↓ inputs ≫ result) : p ↓ inputs :=
```
chenson2018, line 224: This fits on one line.

Cslib/Computability/Urm/Execution.lean
```diff
@@ -0,0 +1,263 @@
...
+
+end HaltsWithResult
+
+namespace Halts
+
+variable {p : Program}
+
+/-- If a program halts, it halts with the output of the final configuration. -/
+theorem toHaltsWithResult {inputs : List ℕ} (h : p ↓ inputs) :
+    p ↓ inputs ≫ (Classical.choose h).state.output :=
```
chenson2018: Is this the best way to do this? Having choice in the signature seems awkward.

Cslib/Computability/URM/Execution.lean
```diff
@@ -0,0 +1,263 @@
...
+  ∀ inputs : List ℕ, eval p inputs = eval q inputs
+
+/-- Program equivalence is an equivalence relation. -/
+theorem ProgramEquiv.equivalence : Equivalence ProgramEquiv where
+  refl := fun _ _ => rfl
+  symm := fun h inputs => (h inputs).symm
+  trans := fun h₁ h₂ inputs => (h₁ inputs).trans (h₂ inputs)
+
+/-- Setoid instance for programs, enabling the ≈ notation. -/
+instance : Setoid Program where
```
chenson2018: There are a couple of ways `Setoid` and `Equivalence` interact, see https://loogle.lean-lang.org/?q=Equivalence%2C+Setoid

Cslib/Computability/URM/Execution.lean
```diff
@@ -0,0 +1,263 @@
...
+  r := ProgramEquiv
+  iseqv := ProgramEquiv.equivalence
+
+/-- Equivalence is reflexive. -/
+theorem ProgramEquiv.refl (p : Program) : p ≈ p := Setoid.refl p
```
chenson2018: Not sure you need to define this API if you already have a `Setoid` instance.

Cslib/Computability/Urm/StraightLine.lean
```diff
@@ -0,0 +1,144 @@
...
+/-- A non-jumping instruction produces a step that increments PC by 1. -/
+theorem Step.of_nonJumping {p : Program} {c : Config} (hlt : c.pc < p.length)
+    (hnonjump : (p[c.pc]'hlt).isJump = false) :
+    ∃ c', Step p c c' ∧ c'.pc = c.pc + 1 := by
+  have hinstr : p[c.pc]? = some p[c.pc] := List.getElem?_eq_getElem hlt
+  cases hp : (p[c.pc]'hlt) with
+  | Z n => exact ⟨_, Step.zero (hp ▸ hinstr), rfl⟩
+  | S n => exact ⟨_, Step.succ (hp ▸ hinstr), rfl⟩
+  | T m n => exact ⟨_, Step.trans (hp ▸ hinstr), rfl⟩
+  | J _ _ _ => simp [hp, Instr.isJump] at hnonjump
```
chenson2018: I think this is much clearer if you make the data explicit and let `grind` figure out the indexing

```suggestion
  cases hp : p[c.pc]'hlt with
  | Z n =>
    use { pc := c.pc + 1, state := c.state.write n 0}
    grind
  | S n =>
    use { pc := c.pc + 1, state := c.state.write n (c.state.read n + 1) }
    grind
  | T m n =>
    use { pc := c.pc + 1, state := c.state.write n (c.state.read m) }
    grind
  | J _ _ _ => grind [Instr.isJump]
```

Cslib/Computability/URM/StraightLine.lean
```diff
@@ -0,0 +1,144 @@
...
+  | _ remaining ih =>
+    by_cases hhalted : c.pc ≥ p.length
+    · exact ⟨c, Relation.ReflTransGen.refl, by omega⟩
+    · push_neg at hhalted
+      unfold Program.IsStraightLine Program.is_straight_line at hsl
+      simp only [List.all_eq_true, Bool.not_eq_true'] at hsl
+      have hnonjump := hsl p[c.pc] (List.getElem_mem hhalted)
+      obtain ⟨c', hstep', hpc'⟩ := Step.of_nonJumping hhalted hnonjump
+      obtain ⟨c'', hsteps'', hpc''⟩ := ih (p.length - c'.pc) (by omega) c' (by omega) rfl
+      exact ⟨c'', Relation.ReflTransGen.head hstep' hsteps'', hpc''⟩
```
chenson2018:
```suggestion
    · grind
    · grind [Relation.ReflTransGen.head, Step.of_nonJumping, Program.IsStraightLine, Program.is_straight_line]
```

Cslib/Computability/URM/StraightLine.lean
```diff
@@ -0,0 +1,144 @@
...
+  | zero => exact ⟨⟨0, s⟩, Relation.ReflTransGen.refl, rfl⟩
+  | succ n ih =>
+    obtain ⟨c_n, hsteps_n, hpc_n⟩ := ih (Nat.le_of_succ_le htarget)
+    have hn_lt : n < p.length := Nat.lt_of_succ_le htarget
+    have hpc_lt : c_n.pc < p.length := hpc_n ▸ hn_lt
+    unfold Program.IsStraightLine Program.is_straight_line at hsl
+    simp only [List.all_eq_true, Bool.not_eq_true'] at hsl
+    have hnonjump := hsl p[c_n.pc] (List.getElem_mem hpc_lt)
+    obtain ⟨c', hstep', hpc'⟩ := Step.of_nonJumping hpc_lt hnonjump
+    exact ⟨c', Relation.ReflTransGen.tail hsteps_n hstep', hpc_n ▸ hpc'⟩
```
chenson2018:
```suggestion
  | succ => grind [Step.of_nonJumping, Program.IsStraightLine, Program.is_straight_line]
```

Cslib/Computability/Urm/StandardForm.lean
```diff
@@ -0,0 +1,301 @@
...
+    (c'.is_halted p ∧ ∃ c₂, Step p.to_standard_form c c₂ ∧
+      c₂.is_halted p.to_standard_form ∧ c'.state = c₂.state) := by
+  cases hstep with
+  | zero hinstr =>
+    left; exact Step.zero (by simp [Program.getElem?_to_standard_form, hinstr])
+  | succ hinstr =>
+    left; exact Step.succ (by simp [Program.getElem?_to_standard_form, hinstr])
+  | trans hinstr =>
+    left; exact Step.trans (by simp [Program.getElem?_to_standard_form, hinstr])
+  | jump_ne hinstr hne =>
```
chenson2018: Here and below, please use something like

```suggestion
  | @jump_ne m n q hinstr hne =>
```
to avoid the use of `rename_i`.

Cslib/Computability/Urm/StandardForm.lean
```diff
@@ -0,0 +1,301 @@
...
+        simp [Program.getElem?_to_standard_form, hinstr, Instr.cap_jump, Nat.min_eq_left hbounded]
+      exact Step.jump_eq hcap heq
+    · right
+      have hgt : q > p.length := Nat.not_le.mp hbounded
+      have hcap : p.to_standard_form[c.pc]? = some (Instr.J m n p.length) := by
+        simp [Program.getElem?_to_standard_form, hinstr, Instr.cap_jump,
+              Nat.min_eq_right (Nat.le_of_lt hgt)]
+      exact ⟨by simp [Config.is_halted]; omega,
+             ⟨p.length, c.state⟩, Step.jump_eq hcap heq,
+             by simp [Config.is_halted, Program.to_standard_form_length], rfl⟩
```
chenson2018: Again, the recurring rules can be annotated, but here is a version with them inlined to `grind`
  
```suggestion
  | @jump_eq m n q hinstr heq =>
    have (x : ℕ) (h : min q p.length = x) : p.to_standard_form[c.pc]? = some (Instr.J m n x) := by
      grind [Program.getElem?_to_standard_form, Instr.cap_jump]
    by_cases q ≤ p.length
    · grind [Step.jump_eq]
    · right
      split_ands
      · grind [Config.is_halted]
      · use ⟨p.length, c.state⟩
        grind [Config.is_halted, Program.to_standard_form_length]
```

Cslib/Computability/Urm/StandardForm.lean
```diff
@@ -0,0 +1,301 @@
...
+    simp only [Config.is_halted, Program.to_standard_form_length]; exact hhalted
+  | head hstep hrest ih =>
+    rcases Step.to_standard_form hstep with
+      hsame | ⟨hhalted_mid, c_mid, hstep_mid, hhalted_mid', hstate_eq⟩
+    · obtain ⟨c₂, hsteps₂, hhalted₂, hstate_eq⟩ := ih
+      exact ⟨c₂, Steps.trans (Steps.single hsame) hsteps₂, hhalted₂, hstate_eq⟩
+    · rename_i c_next
+      have hrest_trivial : c_next = c' := Steps.eq_of_halts (Steps.refl _) hhalted_mid hrest hhalted
+      subst hrest_trivial
+      exact ⟨c_mid, Steps.single hstep_mid, hhalted_mid', hstate_eq⟩
```
chenson2018: I didn't try to fully golf this one, but note that you can pass expressions to `grind` like

```suggestion
    · grind [(Steps.refl _).eq_of_halts hhalted_mid hrest hhalted]
```

Cslib/Computability/Urm/StandardForm.lean
```diff
@@ -0,0 +1,301 @@
...
+    (c'.is_halted p.to_standard_form ∧ ∃ c₂, Step p c c₂ ∧
+      c₂.is_halted p ∧ c'.state = c₂.state) := by
+  cases hstep with
+  | zero hinstr =>
+    left
+    rw [Program.getElem?_to_standard_form] at hinstr
+    simp only [Option.map_eq_some_iff] at hinstr
+    obtain ⟨instr, hinstr', hcap⟩ := hinstr
+    cases instr with
+    | Z n' => simp only [Instr.cap_jump, Instr.Z.injEq] at hcap; subst hcap; exact Step.zero hinstr'
```
chenson2018: Personal style nitpick: I think putting so many tactics onto one line with semicolons hurts readability

Cslib/Computability/URM/StandardForm.lean
```diff
@@ -0,0 +1,301 @@
...
+    Halts p.to_standard_form inputs := by
+  obtain ⟨c, hsteps, hhalted⟩ := h
+  obtain ⟨c₂, hsteps₂, hhalted₂, _⟩ := Steps.to_standard_form_halts hsteps hhalted
+  exact ⟨c₂, hsteps₂, hhalted₂⟩
+
+/-- Reverse step correspondence: if p.to_standard_form steps from c to c', then either:
+    (1) p steps from c to c' (same step), or
+    (2) c' is halted in p.to_standard_form, and p steps to a config that is also halted
+        with the same state (this only happens for jumps with unbounded targets). -/
+theorem Step.from_to_standard_form {p : Program} {c c' : Config}
```
chenson2018: I would guess this can be `grind`ed significantly.

chenson2018 (changes requested):
This looks pretty good to me, and thank you for documenting things well. Here are some overarching thoughts:

- I think the duplication of API between `Bool` and `Prop` that happens in several places is unnecessary 
- I made a lot of comments about how you can use `grind`. These are _not_ all mandatory (I have a personal inclination to use it a lot) but I think worth it overall. Definitions/theorems I passed directly to `grind` often should have annotations for these.
- I wasn't completely comfortable with the use of choice in some type signatures, but I need to think more carefully about this

Cslib/Computability/Urm/Instr.lean
```diff
@@ -0,0 +1,218 @@
...
+@[expose] public section
+
+namespace Cslib.Urm
+
+/-- URM instructions.
+- `Z n`: Set register n to zero
+- `S n`: Increment register n by one
+- `T m n`: Transfer (copy) the contents of register m to register n
+- `J m n q`: If registers m and n have equal contents, jump to instruction q;
+             otherwise proceed to the next instruction
```
ctchou: If the jump instruction's semantics is changed to a relative jump (namely, the jump target is specified via an offset relative to the jump instruction's own position, rather than an absolute position), would `shift_jumps` and all related stuff still be needed?  Of course this would require either `q` be an integer or the jump instruction be split into a jump_forward and a jump_backward instructions.

Cslib/Computability/Urm/Basic.lean
jessealama: Ah, thanks! My thinking was backwards about this. I basically just made a barrel file. I'll think about setting up a `Defs.lean`. Stay tuned!

jessealama: I've rearranged some of the definitions and put them in a new `Defs.lean`. `Basic.lean` now contains basic material that gets used throughout the formalization.

Cslib/Computability/Urm/Execution.lean
jessealama: Good point. I think `transfer` is a better name. (`copy` might be even better, but I'm trying to stay true to the terminology of URMs. One could use this formalization as an opportunity to _update_ that language, but I think `transfer` is better than `trans` here.)

Cslib/Computability/Urm/StraightLine.lean
jessealama: Yeah, this wasn't great. I wasn't really comfortable with this but it made sense early on in the formalization. I've opted for a `Prop`-based approach to predicates now (also addresses one of @chenson2018 's concerns).

Cslib/Computability/Urm/Instr.lean
jessealama: Interesting idea! I think relative offsets would be fine (and shouldn't affect the class of computable functions). This change would indeed obviate the need for `shift_jumps` and similar bureaucratic details, but some details would just get shifted to other places. On balance, though, the net gain of a relative approach might be worth it. I'm not sure. I'd be happy to explore this idea.

Cslib/Computability/URM/StandardForm.lean
jessealama: I took a look at your examples and indeed found that we can compress the proof by about 50% with more liberal `grind`ing. Done!

Cslib/Computability/Urm/Instr.lean
jessealama: Agree. I've cleaned that up; sorry for the noise.

Cslib/Computability/Urm/Instr.lean
jessealama: Thanks! I've applied that. I'm still warming up to `grind`. Examples like this help.

Cslib/Computability/Urm/Program.lean
jessealama: Nice idea. Done

Cslib/Computability/Urm/StandardForm.lean
jessealama: I've taken care of this case and a few others.

Cslib/Computability/Urm/Instr.lean
jessealama: Thanks -- I've taken care of that.

Cslib/Computability/Urm/Instr.lean
jessealama: done

Cslib/Computability/Urm/State.lean
fmontesi: If this takes too long to figure out, I think it's fine to proceed with the current version. In general, we might wanna have a general 'RegisterMachine' concept and then apply it here, but it might be good to accumulate some experience first.

Cslib/Computability/Urm/Basic.lean
```diff
@@ -0,0 +1,183 @@
...
+- `State.write_read_self`, `State.write_read_of_ne`: state read/write lemmas
+- `Config.is_halted_iff`, `Config.ext`: configuration lemmas
+- `JumpsBoundedBy.mono`: bounded jumps are monotonic in the bound
+- `JumpsBoundedBy.shift_jumps`: shifting preserves bounded jumps
+- `Program.mem_max_register`: instruction max_register bounded by program max_register
+-/
+
+@[expose] public section
+
+namespace Cslib.Urm
```
fmontesi: I guess that according to Mathlib's conventions this should be all capitalised `Cslib.URM`?

Cslib/Computability/URM/Basic.lean
```diff
@@ -0,0 +1,183 @@
...
+Non-jump instructions trivially satisfy this. -/
+def JumpsBoundedBy (len : ℕ) : Instr → Prop
+  | J _ _ q => q ≤ len
+  | _ => True
+
+instance (len : ℕ) (instr : Instr) : Decidable (instr.JumpsBoundedBy len) := by
+  cases instr <;> simp only [JumpsBoundedBy] <;> infer_instance
+
+/-- Non-jumping instructions have bounded jumps for any length. -/
+theorem JumpsBoundedBy_of_nonJump {instr : Instr} (h : ¬instr.IsJump)
```
fmontesi: Should be `jumpsBoundedBy_of_nonJump`, see https://leanprover-community.github.io/contribute/naming.html#capitalization

Same for all the rest.

Cslib/Computability/URM/Basic.lean
```diff
@@ -0,0 +1,183 @@
...
+    (len : ℕ) : instr.JumpsBoundedBy len := by
+  cases instr with
+  | J _ _ _ => exact absurd trivial h
+  | _ => trivial
+
+/-- JumpsBoundedBy is monotonic: if bounded for len1, then bounded for any len2 ≥ len1. -/
+theorem JumpsBoundedBy.mono {instr : Instr} {len1 len2 : ℕ}
```
fmontesi, line 118: Sidenote: differently from above, this is ok because it's under JumpsBoundedBy.

Cslib/Computability/URM/Basic.lean
```diff
@@ -0,0 +1,183 @@
...
+    (h : instr.JumpsBoundedBy len) :
+    (instr.shift_jumps offset).JumpsBoundedBy (offset + len) := by
+  cases instr with
+  | J _ _ q => simp only [Instr.shift_jumps, JumpsBoundedBy] at h ⊢; omega
+  | _ => trivial
+
+/-! ## Jump Target Capping -/
+
+/-- Cap a jump target to be at most `len`. Non-jump instructions are unchanged. -/
+def cap_jump (len : ℕ) : Instr → Instr
```
fmontesi: Should be `capJump`.

Cslib/Computability/URM/Basic.lean
```diff
@@ -0,0 +1,183 @@
...
+  | Z n => Z n
+  | S n => S n
+  | T m n => T m n
+  | J m n q => J m n (min q len)
+
+@[simp]
+theorem cap_jump_Z (len n : ℕ) : (Z n).cap_jump len = Z n := rfl
```
fmontesi: Should be `capJump_Z`, etc.

jessealama:
> * I wasn't completely comfortable with the use of choice in some type signatures, but I need to think more carefully about this

I've refactored the theorems that use choice to use the `Part` API, in particular `Part.get`. That pushes choice out of the URM API. (I think for this formalization we can't totally avoid choice.)

Cslib/Computability/URM/Defs.lean
```diff
@@ -0,0 +1,202 @@
...
+  | S : ℕ → Instr
+  | T : ℕ → ℕ → Instr
+  | J : ℕ → ℕ → ℕ → Instr
+deriving DecidableEq, Repr
+
+namespace Instr
+
+/-- The registers read by an instruction. -/
+@[scoped grind =]
+def reads_from : Instr → Finset ℕ
```
fmontesi: Should be `readsFrom`

fmontesi: (same comment for below)

Cslib/Computability/URM/Defs.lean
```diff
@@ -0,0 +1,202 @@
...
+def read (σ : State) (n : ℕ) : ℕ := σ n
+
+/-- Write value v to register n. -/
+@[scoped grind =]
+def write (σ : State) (n : ℕ) (v : ℕ) : State := Function.update σ n v
+
+/-- Initialize state with input values in registers 0, 1, ..., k-1.
+Registers beyond the inputs are initialized to 0. -/
+@[scoped grind =]
+def of_inputs (inputs : List ℕ) : State := fun n => inputs.getD n 0
```
fmontesi: `ofInputs`, etc.

Cslib/Computability/URM/Defs.lean
```diff
@@ -0,0 +1,202 @@
...
+@[scoped grind =]
+def shift_registers (p : Program) (offset : ℕ) : Program :=
+  p.map (Instr.shift_registers offset)
+
+end Program
+
+/-! ## Machine Configuration -/
+
+/-- Machine configuration: program counter (0-indexed) and register state. -/
+structure Config where
```
fmontesi: A configuration in the textbook is just the sequence of values for the registers. Why are you including the `pc` here as well?
Maybe `Config` is `State` and `State` is `Config`? :-)

Cslib/Computability/URM/Execution.lean
fmontesi: @chenson2018, in LTS we have the class `Deterministic` to cover exactly this. Would it make sense to introduce a similar class for `ReductionSystem` and use it here?

Cslib/Computability/URM/Execution.lean
```diff
@@ -0,0 +1,263 @@
...
+namespace Steps
+
+variable {p : Program}
+
+/-- Steps compose transitively. -/
+theorem trans {c₁ c₂ c₃ : Config} (h1 : Steps p c₁ c₂) (h2 : Steps p c₂ c₃) : Steps p c₁ c₃ :=
+  Relation.ReflTransGen.trans h1 h2
+
+/-- One step implies multi-step. -/
+theorem single {c c' : Config} (h : Step p c c') : Steps p c c' :=
```
fmontesi: As Chris mentioned, some of these things can come for free from `ReductionSystem`.

Cslib/Computability/URM/Execution.lean
```diff
@@ -0,0 +1,263 @@
...
+      -- By determinism, c = c'
+      have heq : _ = _ := Step.deterministic hstep_c hstep_c'
+      subst heq
+      exact ih hrest'
+
+end Steps
+
+/-- A program halts on given inputs if there exists a halted configuration reachable from
+the initial configuration. -/
+def Halts (inputs : List ℕ) : Prop :=
```
fmontesi, line 195: @chenson2018: if we don't/cannot generalise this to `ReductionSystem` now, I'd like a concurrent issue to be opened so that we remember.

In general, concepts like divergence and termination should probably be unified under general definitions for `ReductionSystem` and `LTS`.

Cslib/Computability/Urm/Instr.lean
jessealama: I thought about this some more, and I suggest exploring this in a follow-on PR. A syntactic conversion between programs is straightforward and an equivalence proof should also be pretty easy. In my opinion, this kind of exercise belongs in CSLib. What do you think?

Cslib/Computability/URM/Execution.lean
jessealama: I really like that idea. I'll investigate using `Deterministic`. I tried building a bit on top of LTSs but I couldn't think of a good match with URM semantics, but it looks like I need to investigate LTSs some more.

Cslib/Computability/Urm/Basic.lean
jessealama: Ah, OK, thanks. Done!

Cslib/Computability/URM/Basic.lean
jessealama: done, thanks

Cslib/Computability/URM/Basic.lean
jessealama: Sounds good -- I've camelCase'd a bunch of identifiers.

Cslib/Computability/URM/Defs.lean
jessealama: Done, thanks!

Cslib/Computability/URM/Basic.lean
jessealama: done!

Cslib/Computability/URM/Defs.lean
jessealama: You're right, the naming was backwards. Fixed — swapped `State` ↔ `Config` (now `Regs` for register contents, `State` for full machine state).

Cslib/Computability/URM/Execution.lean
chenson2018: Yes, this makes sense. I can add this once `Step` is using `ReductionSystem`.

Cslib/Computability/Urm/Instr.lean
ctchou: It's up to you, for you will have to do the proofs.  PC-relative jump is in fact an old idea.  Every modern instruction set architecture has it, precisely for the purpose of supporting position-independent code.

ctchou: A related idea is to add a base register and/or literal to each instruction, so that each address is relative to that base, rather than an absolute address.  Here you can probably get away with using only Nat for address.

chenson2018: There's some spectrum of tradeoff here from staying minimal and abstract to mirroring a real instruction set. I'm fine if this stays simple at first, especially if that will make your future plans for connecting to SKI simpler.

Cslib/Computability/URM/Execution.lean
chenson2018: I think `ReuctionSystem` is pretty straightforward in replacing `Step`/`Steps`. I'd say give it a try and if we hit any significant difficulties we open an issue as Fabrizio has suggested.

Cslib/Computability/Urm/Instr.lean
ctchou: Come to think of it: you can have a special register (say, registter 0) which contains a "base PC".  The real target of a jump instruction is the target + the value of the base PC.  This will also give you position-independent code at the expense of changing the base PC value when switching between different blocks of code.

ctchou: I agree with Chris.  Adding more "realistic" instruction features is easy.  The hard part is to show that the fancier instructions can be implemented by a more primitive formalism like SKI.

jessealama: My first instinct here is to investigate the relative jump idea. That ought to simplify the proofs of closure of URM-computable functions under composition, primitive recursion, and minimization. And it should help with the SKI equivalence proof, too.

Cslib/Computability/Urm/State.lean
jessealama: I love the idea of a general register machine concept! What I might suggest is implementing another register machine model, say RASP machines (but that's just one possibility!).

jessealama: One challenge with the use of an arbitrary index type for the register indices, or `[Fintype]`, would be to ensure the coherence of programs. Using natural numbers, one has the luxury of knowing that any program is coherent in the sense that any register index appearing in it refers to a register that actually exists. Programs and machines (or states) are sort of independent. But perhaps I'm overthinking this; making the definition of machine states and programs share an index type could work (and might have some nice theoretical benefits such as enabling a notion of product of machines).

Cslib/Computability/URM/Execution.lean
jessealama: I'll give that a try; I'd really like to align with existing infrastructure. Stay tuned!

Cslib/Computability/URM/Execution.lean
jessealama: I've taken a look and applied some of `ReductionSystem`. For now, the `Step.deterministic` theorem remains as-is. A `ReductionSystem.Deterministic` class would be a nice addition; I'm happy to open a follow-up issue or PR for that if useful.

Cslib/Computability/URM/Execution.lean
jessealama: Should I open an issue to track generalizing `Diverges`/`Halts` concepts to `ReductionSystem` (and maybe to LTS, too)?

chenson2018: Sure, if you'd like to leave this for now feel free to open an issue after the PR is merged.

jessealama: I'll sketch out how halting/divergence could work and add it to this PR.

jessealama: @fabriziomontesi Here's a sketch of how we can unify `Diverges` and `Halts` under `ReductionSystem`. For `ReductionSystem`, we could add a `Deterministic` class (each term has at most one successor) as well as a `Terminates` and `Diverges` parameterized by a halting predicate.

One approach to unify LTSs with reduction systems would be to forget labels. How does "trapped" sound as the LTS analogue of divergence (terminated state can't be reached)? As for `Deterministic` in the conext of LTSs, how about "functional" (at most one succcessor of a state, regardlless of label)?

chenson2018: Could we please postpone this to another PR? I think it is much cleaner organizationally to separate this out from an already large PR and I have thoughts about how this best fits in with what we already have about terminating relations.

I would much rather you just use `ReductionSystem` for `Step`/`Steps` via the attribute, which should be straightforward and not require any new machinery. (And remember that you can always just simplify to the relation it bundles)

Cslib/Computability/Urm/Instr.lean
jessealama: I'd be interested in exploring the base PC idea with relative offsets, but perhaps in a separate PR. I think that approach would give us a bit more experience with a generalized register machine concept. One could also have a relative-offset, without base PC, semantics, too.

Cslib/Foundations/Semantics/ReductionSystem/Basic.lean
```diff
@@ -198,6 +198,73 @@ theorem ReductionSystem.isNormalizing_of_isTerminating (h : rs.Terminating) :
...
+/-! ## Determinism, termination, and divergence -/
+
+/-- A reduction system is deterministic if each term has at most one successor. -/
+class ReductionSystem.Deterministic (rs : ReductionSystem Term) : Prop where
+  /-- Each term has at most one successor under the reduction relation. -/
+  deterministic : ∀ t t₁ t₂, rs.Red t t₁ → rs.Red t t₂ → t₁ = t₂
+
+/-- A term terminates (reaches a halted state) given a halting predicate.
+Parameterized by `Halted : Term → Prop` to allow different notions of halting. -/
+def ReductionSystem.Terminates (Halted : Term → Prop) (t : Term) : Prop :=
```
chenson2018: I think this needs more discussion, as this doesn't align with what we already have defined in [Relation.Terminating](https://api.cslib.io/docs/Cslib/Foundations/Data/Relation.html#Relation.Terminating) or the design of all the the API of `ReductionSystem` being defeq to the definitions in `Cslib.Foundations.Data.Relation`

As I mention in my other comment, I would prefer that this PR not make changes to `ReductionSystem` and `LTS` at this time, for now using `ReductionSystem` as it currently is and following up after this is merged.

chenson2018: To make it clear why I say this doesn't align with `Relation.Terminating`: the idea is that the encoding of halting should happen in the relation itself and this is why we are able to reuse the infrastructure from Lean itself to define termination using [WellFounded](https://leanprover-community.github.io/mathlib4_docs/Init/WF.html#WellFounded). Requiring an outside proposition on terms to determine if we are at a "halted state" is contrary to this design, which already captures termination. There are some options to unify these different encodings which I think we can consider after this PR.

Cslib/Computability/URM/Execution.lean
```diff
@@ -0,0 +1,281 @@
...
+      Step s ⟨q, s.regs⟩
+  | jump_ne {s : State} {m n q : ℕ}
+      (h : p[s.pc]? = some (Instr.J m n q))
+      (hne : s.regs.read m ≠ s.regs.read n) :
+      Step s ⟨s.pc + 1, s.regs⟩
+
+/-- ReductionSystem for URM execution, parameterized by program.
+
+This integrates URM execution with CSLib's `ReductionSystem` infrastructure. -/
+def stepRs : ReductionSystem State := ⟨Step p⟩
```
chenson2018: As I described in [this comment](https://github.com/leanprover/cslib/pull/299#discussion_r2738648716), you should always be using the `reduction_sys` attribute to define a `ReductionSystem`, and use the notation it generates as the canonical way to write `Steps`. 

We usually follow this with an API lemma that the notation is equivalent to the inductive, e.g. [fullBetaRs_Red_eq](https://github.com/leanprover/cslib/blob/b2e8d7c14918761760f01feb3cf21e788e0440c0/Cslib/Languages/LambdaCalculus/LocallyNameless/Untyped/FullBeta.lean#L53-L55).

jessealama: I tried using `@[reduction_sys stepRs "ᵉ "]` on `Step`. The issue is that `Step` has a value parameter `(p : Program)` rather than a type parameter. The attribute correctly generates `stepRs : Program → ReductionSystem State`, but the notation expands to `(ReductionSystem.Red stepRs) t t'` — it doesn't apply `stepRs` to the program.

We could work around this by defining `Config (p : Program) := State` and making `Step : Config p → Config p → Prop`, but that feels like bending the domain model to fit the metaprogramming. For now I've added stepRs manually with a TODO comment. Happy to dig into extending `reduction_sys` if that's useful. It could also be that I'm being thick and not seeing something obvious here.

Cslib/Foundations/Semantics/ReductionSystem/Basic.lean
jessealama: Sounds good -- let's talk about this separately. I think I misunderstood; sorry about that.

Cslib/Computability/URM/Execution.lean
chenson2018: Oh I see. This is my fault for not noticing this earlier, this is just out of scope for what this is designed for. I have just removed the `ReductionSystem` piece completely.

jessealama: You were right — we found a cleaner solution. The issue was that `Step` was parameterized by programs `(Step p : State → State → Prop)`, which didn't fit the `reduction_sys` attribute well. I fixed this by bundling the program and state into `Configuration`. Then we can define `Step : Configuration → Configuration → Prop`, and `@[reduction_sys stepRs "ᵉ "]` works as intended. The notation generates correctly, and we can remove the manual wrappers (`Steps.trans`, `Steps.single`, etc.) since `ReductionSystem` provides them.

Thanks for pushing on this — the result is cleaner.

jessealama:
Ah, I see you're working on it. I was able to find a solution that I think address the issue with ReductionSystem. Let me know when things look ready. I'm happy to get a first working solution and continue the discussion going in follow-on PRs. Thanks for your help!

jessealama:
> Jesse, please see the commits I pushed. They just remove the unused `ReductionSystem` and address a few of my earlier comments. Please let me know if you have any concerns. (And I tried to be careful, but also double-check that I didn't mangle any of your docs while merging)
> 
> With these items addressed I think this looks good, thank you again for the contribution!

Taking a look now -- thank you! It looks like you bumped into the issue with ReductionSystem. I did manage to work out a solution by packing the program argument in to a new type. That works, but it's not great imo. I see you're working more directly with `ReflTransGen`. That strikes me as fine for now. I'll take a look at your work.

chenson2018:
Sorry, I think we worked past each other a bit here! I have just seen your comment amount bundling and making the `ReductionSystem` work. If you feel this is still better, maybe we do this in a follow up PR so the diff is easier to read.

jessealama:
> Jesse, please see the commits I pushed. They just remove the unused `ReductionSystem` and address a few of my earlier comments. Please let me know if you have any concerns. (And I tried to be careful, but also double-check that I didn't mangle any of your docs while merging)
> 
> With these items addressed I think this looks good, thank you again for the contribution!

Looks good -- I found a couple of outdated bits of documentation and removed them. Otherwise, I think it's OK!

jessealama:
ping @fmontesi I think Chris and I took care of your concerns -- care to take a quick second look?

Cslib/Computability/URM/Execution.lean
```diff
@@ -0,0 +1,289 @@
...
+      · exact hnormal ⟨_, Step.jump_eq (hp ▸ hinstr) heq⟩
+      · exact hnormal ⟨_, Step.jump_ne (hp ▸ hinstr) heq⟩
+
+/-- The step relation is confluent.
+
+For a deterministic relation, any two execution paths from the same state must follow
+the same sequence of steps, so if both reach some state, they reach the same state.
+We prove this directly rather than via Diamond since Diamond requires the relation
+to always have successors. -/
+theorem Step_confluent (p : Program) : Relation.Confluent (Step p) := by
```
fmontesi: Should be `step_confluent`.

Cslib/Computability/URM/Execution.lean
```diff
@@ -0,0 +1,289 @@
...
+  | zero {s : State} {n : ℕ}
+      (h : p[s.pc]? = some (Instr.Z n)) :
+      Step s ⟨s.pc + 1, s.regs.write n 0⟩
+  | succ {s : State} {n : ℕ}
+      (h : p[s.pc]? = some (Instr.S n)) :
+      Step s ⟨s.pc + 1, s.regs.write n (s.regs.read n + 1)⟩
+  | transfer {s : State} {m n : ℕ}
+      (h : p[s.pc]? = some (Instr.T m n)) :
+      Step s ⟨s.pc + 1, s.regs.write n (s.regs.read m)⟩
+  | jump_eq {s : State} {m n q : ℕ}
```
fmontesi, line 70: I notice now we're a bit inconsistent in how we name inductive cases like this in cslib, where we use snake_case here and in lambda-calculi, and camelCase in CCS. Does anybody have a reference for this? I can't figure it out from what's written in the mathlib conventions. CC @chenson2018

Cslib/Computability/URM/StandardForm.lean
```diff
@@ -0,0 +1,249 @@
...
+
+/-- Straight-line programs are in standard form. -/
+theorem straight_line_IsStandardForm {p : Program} (hsl : p.IsStraightLine) :
+    p.IsStandardForm := by
+  intro instr hinstr
+  exact Instr.jumpsBoundedBy_of_nonJump (hsl instr hinstr) p.length
+
+/-! ## Behavioral Equivalence
+
+`p` and `p.toStandardForm` are behaviorally equivalent: they halt on the same
```
fmontesi: Just for the docstring: what you write here after the colon is functional equivalence (or something like that), not behavioural equivalence, right? (The way you prove it though seems to build a sort of bisimulation.)

Cslib/Computability/URM/Execution.lean
chenson2018: The way it works is that the naming rules for constructors follow the same rules based on their type (rule 7). 

So constructors like this that are propositional should be snake case, while pretty much only constructors of structures/classes in `Type` should have camel case.

Cslib/Computability/URM/Execution.lean
jessealama: done

Cslib/Computability/URM/StandardForm.lean
jessealama: Right, it's functional equivalence as stated though the proof is essentially a bisimulation. I've updated these docstrings appropriately.

PR #302, opened 2026-01-29
feat: add linear logic tests/benchmark

arademaker:
For all logics that are likely to be implemented, it would be interesting to have a test suite that not only demonstrates how the definitions of syntax and semantics work, but also serves a didactic purpose through concrete instantiations of the logic.

In this PR, I suggest some initial tests for the linear logic already implemented in CSLib. In addition to tests for syntax and trivial equivalences, I present a proof of a linear logic theorem that I picked at random from https://arxiv.org/abs/1904.06850. From this PR onward, we can think about more systematic methods for constructing proof benchmarks.

```
@article{Olarte_2019,
   title={The ILLTP Library for Intuitionistic Linear Logic},
   volume={292},
   ISSN={2075-2180},
   url={http://dx.doi.org/10.4204/EPTCS.292.7},
   DOI={10.4204/eptcs.292.7},
   journal={Electronic Proceedings in Theoretical Computer Science},
   publisher={Open Publishing Association},
   author={Olarte, Carlos and de Paiva, Valeria and Pimentel, Elaine and Reis, Giselle},
   year={2019},
   month=apr,
   pages={118–132}}
```

### DIFF
--- a/CslibTests/CLL.lean
+++ b/CslibTests/CLL.lean
@@ -6,10 +6,9 @@ Authors: Alexandre Rademaker
 
 import Cslib.Logics.LinearLogic.CLL.Basic
 
-/-! # Tests for Classical Linear Logic
+namespace CslibTests
 
-This file contains tests for the CLL implementation in
-`Cslib.Logics.LinearLogic.CLL.Basic`.
+/-! # Tests for Classical Linear Logic
 
 I use `Proposition Nat` as the concrete instantiation for atoms.
 -/
@@ -17,8 +16,6 @@ I use `Proposition Nat` as the concrete instantiation for atoms.
 open Cslib.CLL
 
 /-! ## Proposition construction tests -/
-
-section PropositionTests
 
 -- Define some atomic propositions for testing
 abbrev P := Proposition Nat
@@ -46,11 +43,7 @@ example : P := (a ⊗ b) ⅋ c
 example : P := !(a ⊗ b)
 example : P := a ⊕ (b & c)
 
-end PropositionTests
-
 /-! ## Duality tests -/
-
-section DualityTests
 
 -- dual_involution: a⫠⫠ = a
 example : (a⫠⫠ : P) = a := Proposition.dual_involution a
@@ -75,11 +68,8 @@ example : (1 : P) ≠ (1 : P)⫠ := Proposition.dual_neq 1
 -- dual_inj: duality is injective
 example : (a⫠ = b⫠) ↔ (a = b) := Proposition.dual_inj a b
 
-end DualityTests
 
 /-! ## Basic proof tests -/
-
-section ProofTests
 
 -- Axiom: ⊢ a, a⫠
 example : ⇓({a, a⫠} : Sequent Nat) := Proof.ax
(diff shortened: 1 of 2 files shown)


CslibTests/CLL.lean
```diff
@@ -0,0 +1,243 @@
...
+
+Breaking down the formula:
+
+              A ⊸ B = A⫠ ⅋ B (linear implication)
+           !(A ⊸ B) = !(A⫠ ⅋ B)
+        (!(A ⊸ B))⫠ = ʔ((A⫠ ⅋ B)⫠) = ʔ(A ⊗ B⫠)
+       !(A ⊸ B) ⊸ B = (!(A ⊸ B))⫠ ⅋ B = ʔ(A ⊗ B⫠) ⅋ B
+ !A ⊸ B = (!A)⫠ ⅋ B = ʔA⫠ ⅋ B
+       B ⊸ (!A ⊸ B) = B⫠ ⅋ (ʔA⫠ ⅋ B) -/
+--/
```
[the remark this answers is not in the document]
arademaker: done

CslibTests/CLL.lean
```diff
@@ -0,0 +1,243 @@
...
+
+/-! # Tests for Classical Linear Logic
+
+This file contains tests for the CLL implementation in
+`Cslib.Logics.LinearLogic.CLL.Basic`.
+
+I use `Proposition Nat` as the concrete instantiation for atoms.
+-/
+
+open Cslib.CLL
```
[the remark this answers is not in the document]
arademaker: Fixed

CslibTests/CLL.lean
```diff
@@ -0,0 +1,218 @@
...
+
+         A ⊸ B = A⫠ ⅋ B (linear implication)
+      !(A ⊸ B) = !(A⫠ ⅋ B)
+   (!(A ⊸ B))⫠ = ʔ((A⫠ ⅋ B)⫠) = ʔ(A ⊗ B⫠)
+  !(A ⊸ B) ⊸ B = (!(A ⊸ B))⫠ ⅋ B = ʔ(A ⊗ B⫠) ⅋ B
+        !A ⊸ B = (!A)⫠ ⅋ B = ʔA⫠ ⅋ B
+  B ⊸ (!A ⊸ B) = B⫠ ⅋ (ʔA⫠ ⅋ B)
+-/
+example : ⇓({b⫠, (!(a ⊸ b) ⊸ b) ⊗ (b ⊸ (!a ⊸ b))} : Sequent Nat) := by
+  apply Proof.rwConclusion (Multiset.pair_comm ..)
```
chenson2018, line 203: We try to follow the convention of not constructing data (non-propositions) with tactics, though it is very awkward for linear logic.

chenson2018 (commented):
I think it is a good general idea to have these sorts of examples as a sanity check that we've correctly defined various logics and languages.

(Leaving approval to Fabrizio since he wrote this module)

CslibTests/CLL.lean
arademaker: Hi @chenson2018, I thought about that. The first version of the proof was a term-based proof. But the bottom-up approach to the tactics - at least in the case of LL - makes the proof much clearer. But I am open to respecting any decision.

chenson2018: I completely agree that it is natural to write this way. I'm fine with leaving it this way for a test file. If it breaks for whatever reason, this is also interesting information and easily fixed.

fmontesi (approved):
LGTM, thanks!

PR #303, opened 2026-01-29
chore: simplify the connection between `HasFresh` and `Infinite`

chenson2018:
This PR adds an instance directly showing that `HasFresh` implies `Infinite`. This replaces `HasFresh.not_of_finite` which triggered a linter and is now equivalent to just `not_finite`. `HasFresh.of_infinite` is also changed to an instance with a slightly more direct definition.

### DIFF
--- a/Cslib/Foundations/Data/HasFresh.lean
+++ b/Cslib/Foundations/Data/HasFresh.lean
@@ -132,15 +132,17 @@ end
 
 export HasFresh (fresh fresh_notMem fresh_exists)
 
-set_option linter.unusedFintypeInType false in
-lemma HasFresh.not_of_finite (α : Type u) [Fintype α] : IsEmpty (HasFresh α) :=
-  ⟨fun f ↦ (f.fresh_notMem .univ).elim (Finset.mem_univ _)⟩
+/-- `HasFresh α` implies a computably infinite type. -/
+instance HasFresh.to_infinite (α : Type u) [HasFresh α] : Infinite α := by
+  apply Infinite.of_not_fintype
+  rintro ⟨elems, _⟩
+  grind [fresh_notMem elems]
 
 /-- All infinite types have an associated (at least noncomputable) fresh function.
-This, in conjunction with `HasFresh.not_of_finite`, characterizes `HasFresh`. -/
-noncomputable def HasFresh.of_infinite (α : Type u) [Infinite α] : HasFresh α where
-  fresh s := s.finite_toSet.infinite_compl.nonempty.choose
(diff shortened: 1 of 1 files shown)


Cslib/Foundations/Data/HasFresh.lean
```diff
@@ -132,15 +132,17 @@ end
...
+  grind [fresh_notMem elems]
 
 /-- All infinite types have an associated (at least noncomputable) fresh function.
-This, in conjunction with `HasFresh.not_of_finite`, characterizes `HasFresh`. -/
-noncomputable def HasFresh.of_infinite (α : Type u) [Infinite α] : HasFresh α where
-  fresh s := s.finite_toSet.infinite_compl.nonempty.choose
-  fresh_notMem s := s.finite_toSet.infinite_compl.nonempty.choose_spec
+This, in conjunction with `HasFresh.to_infinite`, characterizes `HasFresh`. -/
+noncomputable instance HasFresh.of_infinite (α : Type u) [Infinite α] : HasFresh α where
+  fresh s := Infinite.exists_notMem_finset s |>.choose
```
ctchou, line 144: I really don't know what the preferred style is: should there be a space between `|>` and `.`?

ctchou (approved):
LGTM.

Cslib/Foundations/Data/HasFresh.lean
chenson2018: These are semantically different:
- `|>.choose` is projecting from the pipe ([Lean.Parser.Term.pipeProj](https://leanprover-community.github.io/mathlib4_docs/Lean/Parser/Term.html#Lean.Parser.Term.pipeProj)) and resolves to `Exists.choose`
- `|> .choose` is piping to [Lean.Parser.Term.dotIdent](https://leanprover-community.github.io/mathlib4_docs/Lean/Parser/Term.html#Lean.Parser.Term.dotIdent), which cannot infer the meaning of `.choose` in this case because `Classical.choose` also exists

ctchou: Good to know that.  Thanks!

ctchou: In this case, I wonder if `(Infinite.exists_notMem_finset s).choose` is actually clearer.

chenson2018: I am fairly ambivalent, but I think this is a common enough core syntax to not be considered obfuscatory. I wrote it without thinking, not as a golfing attempt.

PR #304, opened 2026-01-30
feat(Foundations/Semantics/ReductionSystem): extend `ReductionSystem`

IvanRenison:
- Add more definitions and theorems about reduction systems
- Use namespace `ReductionSystem`
- Add sections

### DIFF
--- a/Cslib/Foundations/Semantics/ReductionSystem/Basic.lean
+++ b/Cslib/Foundations/Semantics/ReductionSystem/Basic.lean
@@ -257,22 +257,21 @@ end Termination
 section Convergence
 
 /-- A reduction system is convergent when it is both confluent and terminating. -/
-structure Convergent : Prop where
-  isConfluent : rs.Confluent
-  isTerminating : rs.Terminating
+def Convergent : Prop := Relation.Convergent rs.Red
 
-theorem Convergent_iff :
-    rs.Convergent ↔ rs.Confluent ∧ rs.Terminating :=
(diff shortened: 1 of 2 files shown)


chenson2018:
Please add a description to this PR.

Cslib/Foundations/Semantics/ReductionSystem/Basic.lean
```diff
@@ -198,6 +256,58 @@ theorem ReductionSystem.isNormalizing_of_isTerminating (h : rs.Terminating) :
   · unfold Normal
     use a
 
+theorem Terminating.isConfluent_iff_all_unique_Normal (ht : rs.Terminating) :
```
chenson2018, line 251: Any additions to this module should be definitionally equal to something in `Cslib.Foundations.Data.Relation`. These definitions and theorems may be welcome, but I would ask for anything new like this that you add them there first for unbundled relations.

chenson2018 (changes requested):

IvanRenison:
Should I also add the `Convergent` things to `Relation`?

chenson2018:
> Should I also add the `Convergent` things to `Relation`?

Yes. For the moment I would add it as an `abbrev` that is the conjunction of the two conditions. (We might revisit this later, but let's start with this for consistency)

chenson2018 (approved):
This looks good, thank you! 

You may have seen on Zulip I am considering removing `ReductionSystem`, but this works helps make it more usable in the meantime and everything added to `Cslib.Foundations.Data.Relation` will remain and is great to have.

PR #319, opened 2026-02-05
feat: Hennessy-Milner logic

fmontesi:
This PR adds Hennessy-Milner Logic (HML) in its general form (parameterised over any LTS). We have proven the correctness of denotational semantics and its key result (the Hennessy-Milner theorem, stating that theory equivalence coincides with bisimilarity).

### DIFF
--- a/CslibTests.lean
+++ b/CslibTests.lean
@@ -5,6 +5,7 @@ public import CslibTests.CCS
 public import CslibTests.DFA
 public import CslibTests.FreeMonad
 public import CslibTests.GrindLint
+public import CslibTests.HML
 public import CslibTests.HasFresh
 public import CslibTests.ImportWithMathlib
 public import CslibTests.LTS


Cslib/Logics/HML/Basic.lean
```diff
@@ -0,0 +1,309 @@
...
+theorem satisfies_finiteOr {lts : LTS State Label} {s : State}
+    {as : List (Proposition Label)} :
+    Satisfies lts s (Proposition.finiteOr as) ↔ ∃ a ∈ as, Satisfies lts s a := by
+  induction as <;> grind
+
+/-- Two states are theory-equivalent iff they are denotationally equivalent. -/
+theorem theoryEq_denotation_eq {lts : LTS State Label} :
+    TheoryEq lts s1 s2 ↔
+    (∀ a : Proposition Label, s1 ∈ a.denotation lts ↔ s2 ∈ a.denotation lts) := by
+  apply Iff.intro <;> intro h
```
chenson2018: Having `theory` as an `abbrev` is why this proof is so long. If you add some API to it you can do something like:
```lean
@[grind →]
theorem satisfies_theory (h : Satisfies lts s1 a) : a ∈ theory lts s1 := by
  grind

theorem theoryEq_denotation_eq {lts : LTS State Label} :
    TheoryEq lts s1 s2 ↔
    (∀ a : Proposition Label, s1 ∈ a.denotation lts ↔ s2 ∈ a.denotation lts) := by
  grind [_=_ satisfies_mem_denotation]
```

Cslib/Logics/HML/Basic.lean
```diff
@@ -0,0 +1,309 @@
...
+  · intro a
+    apply Iff.intro <;> intro hmem
+    · have : a ∈ theory lts s1 := by grind
+      grind
+    · have : a ∈ theory lts s1 := by grind
+      grind
+  · grind
+
+/-- If two states are not theory equivalent, there exists a distinguishing proposition. -/
+lemma not_theoryEq_satisfies (h : ¬(TheoryEq lts) s1 s2) :
```
chenson2018:
```suggestion
lemma not_theoryEq_satisfies (h : ¬ TheoryEq lts s1 s2) :
```

Cslib/Logics/HML/Basic.lean
```diff
@@ -0,0 +1,309 @@
...
+    ∃ a, (Satisfies lts s1 a ∧ ¬Satisfies lts s2 a) := by
+  grind [neg_satisfies]
+
+/-- If two states are theory equivalent and the former satisfies a proposition, the latter does as
+well. -/
+theorem theoryEq_satisfies {lts : LTS State Label} (h : TheoryEq lts s1 s2)
+    (hs : Satisfies lts s1 a) : Satisfies lts s2 a := by
+  unfold TheoryEq theory at h
+  rw [Set.ext_iff] at h
+  exact (h a).mp hs
```
chenson2018, line 175: Can be just `grind` if you'd like.

Cslib/Logics/HML/Basic.lean
```diff
@@ -0,0 +1,309 @@
...
+  case diamond μ a ih =>
+    cases hs
+    rename_i s1' htr hs
+    have := hrb.follow_fst hr htr
+    grind
+  case box μ a ih =>
+    apply Satisfies.box
+    cases hs
+    rename_i hs
+    grind
```
chenson2018: Please try to avoid using `rename_i`. It is also nice to combine identical cases for readability:
 
```suggestion
  induction a generalizing s1 s2 with
  | diamond => cases hs with | diamond htr _ => grind [hrb.follow_fst hr htr]
  | box => grind [cases Satisfies]
  | _ => grind
```

Cslib/Logics/HML/Basic.lean
```diff
@@ -0,0 +1,309 @@
...
+    TheoryEq lts s1 s2 := by
+  unfold TheoryEq theory
+  ext a
+  apply Iff.intro <;> intro h
+  case mp =>
+    grind
+  case mpr =>
+    have hbs : s2 ~[lts] s1 := by grind [Bisimilarity.symm]
+    obtain ⟨r', hr', hr'b⟩ := hbs
+    grind
```
chenson2018: `grind` will unfold the `abbrev`s automatically:

```suggestion
  have : s2 ~[lts] s1 := by grind [Bisimilarity.symm]
  grind
```

Cslib/Logics/HML/Basic.lean
```diff
@@ -0,0 +1,309 @@
...
+well. -/
+theorem theoryEq_satisfies {lts : LTS State Label} (h : TheoryEq lts s1 s2)
+    (hs : Satisfies lts s1 a) : Satisfies lts s2 a := by
+  unfold TheoryEq theory at h
+  rw [Set.ext_iff] at h
+  exact (h a).mp hs
+
+/-- Theory equivalence is a bisimulation. -/
+@[scoped grind ⇒]
+theorem theoryEq_isBisimulation (lts : LTS State Label)
```
chenson2018, line 202: Could we try to break this up a bit? There's no hard limit, but this is pretty extreme. I'd rather someone with more domain expertise take a stab at it first before I approach from a purely proof golfing perspective.

Cslib/Logics/HML/Basic.lean
```diff
@@ -0,0 +1,309 @@
...
+
+/-- Characterisation theorem for the denotational semantics. -/
+@[scoped grind =]
+theorem satisfies_mem_denotation {lts : LTS State Label} :
+    Satisfies lts s a ↔ s ∈ a.denotation lts := by
+  induction a generalizing s <;> grind
+
+/-- A state satisfies a proposition iff it does not satisfy the negation of the proposition. -/
+theorem neg_satisfies {lts : LTS State Label} :
+    Satisfies lts s a ↔ ¬Satisfies lts s a.neg := by
```
chenson2018: For these `neg_*` theorems with `↔` signatures, my intuition is to have `neg` and `¬` on the lhs for `simp`/`grind` purposes.

Cslib/Logics/HML/Basic.lean
fmontesi: I tried and it failed? Something's amiss maybe.. :-)

chenson2018: This works fine for me when I copy paste... I can commit it to make things simpler.

Cslib/Logics/HML/Basic.lean
chenson2018: I tried this and helps with some of the `#grind_lint` issues, so I think this is the right thing to do.

chenson2018: (to be clear, I would flip both `neg_satisfies` and `neg_denotation`)

Cslib/Logics/HML/Basic.lean
fmontesi: We might be able to do so, but I'd like to do in a separate PR, there's a bit of conceptual work that might await us to do it elegantly.

Cslib/Logics/HML/Basic.lean
fmontesi: Took it a step further, even. Thanks!

fmontesi:
Addressed almost everything. I couldn't figure out how not to make grind super slow or not working at all while removing the abbrevs... but I do agree it'd be nice if they were proper defs. :-\

Cslib/Logics/HML/Basic.lean
chenson2018: Beyond being elegant, there is some very basic cleaning to be done here. I think it is important that we insist upon this as it comes along, as temporary proofs have a tendency to become longstanding technical debt.

I've pushed a commit with how I would clean this up, please let me know if you have any concerns.

fmontesi: Cleaning up is definitely good, thank you very much.

Cslib/Logics/HML/Basic.lean
fmontesi: This is much slower on my machine for some reason that I haven't figured out yet. Do you experience the same?

fmontesi:
Applied a couple of changes, although I don't have really intuitive names for the auxiliary defs/theorems yet.

The only thing remaining at this point is the use of `abbrev`.

Cslib/Logics/HML/Basic.lean
chenson2018: I didn't notice, but there's no harm in leaving it explicit just in case.

chenson2018:
I am fine leaving the `abbrev` for now if you'd like. 

Looks like you need to run `lake exe mk_all --module` once more, but I'll go ahead and approve.

Cslib/Logics/HML/Basic.lean
chenson2018: You're right, using the profiler I see an ~3.5x slowdown for elaboration of the proof. This is a proof where `grind` relies on the `abbrev` transparency, which seems to have a side effect of some expensive checks of definitional equality.

So best to leave explicit for now and a motivation for making these use `def` in the future. (Unless you want me to try fixing this now.)

Cslib/Logics/HML/Basic.lean
```diff
@@ -0,0 +1,266 @@
...
+
+Hennessy-Milner Logic (HML) is a logic for reasoning about the behaviour of nondeterministic and
+concurrent systems.
+
+## Implementation notes
+There are two main versions of HML. The original [Hennessy1985], which includes a negation
+connective, and a variation without negation, for example as in [Aceto1999].
+We follow the latter, which is used in many recent papers. Negation is recovered as usual, by having
+a `false` atomic proposition and a function that, given any proposition, returns its negated form
+(see `Proposition.neg`).
```
benkeks, line 23: As somebody who did a lot of stuff with HML in the last years, I'm surprised by the recency claim. The variant with negation + diamond modality seems to appear more regularly in 2020s papers.

The background is that it's the more convenient basis to work from: One nice functionally complete set of three operators for propositional logic $\\{ \land, \neg, \top \\}$ and *one* modality $\langle \cdot \rangle$. Also, many interesting sublanguages of HML are much cleaner to describe if there are negations instead of necessities (as opposed to reconstructing modal alternations), and only one canonical empty element in syntax trees.

Having 4+2 instead of 3+1 constructors makes every induction more cumbersome---and especially the modal cases boiler-platy in formalizations. So it seems sensible to default to the smaller inductive language.

Maybe there's a way to reconsider this design decision?

fmontesi:
Thanks for this comment! It's 'more recent' but not the 'most recent' then.
:-)

We can certainly reconsider. One problem I can remember with negation was
that having a negative premise in Satisfy made Lean unhappy. Have you tried
coding that perhaps?

benkeks:
> having a negative premise in Satisfy made Lean unhappy. Have you tried coding that perhaps?

Ah, yes, I see what you mean regarding `Satisfy`. It can be solved by pushing the negations around. I have an adapted version over at:

https://github.com/leanprover/cslib/compare/main...benkeks:cslib:small-hml

I've still kept the prior operators as abbrevs. If there's interest, I can try to make it into a full PR with proper documentation.

(As I've made the mistake of playing around with other stuff in the branch, here's the diff with stable commit ids: https://github.com/leanprover/cslib/compare/f0d87b6e107dddfe4959dcd2f44ba985e3f1c0d2...benkeks:cslib:4d48abcd54eef29170cff8d5af30218d887d83ea)

fmontesi:
Huh, interesting fix indeed. :-)

Those rules are not very intuitive though, and even introduce double-negation as a primitive rule. So there are fewer cases in the language of propositions, but semantics gets tricky in its own way because of the formalisation. Mmh..

benkeks:
Basically, the cases of `Satisfies` are the same as in the version with `or` and `box`, just that those become `negAnd` and `negDiamond`.

Instead of pushing around the `neg`, one might also express it as a mutual recursion highlighting the two-player nature of the definition:

```lean
mutual
  /-- Satisfaction relation. `Satisfies lts s a` means that, in the LTS `lts`,
    the state `s` satisfies the proposition `a`. -/
  @[scoped grind]
  inductive Satisfies (lts : LTS State Label) : State → Proposition Label → Prop where
    | true {s : State} : Satisfies lts s .true
    | and {s : State} {a b : Proposition Label} :
      Satisfies lts s a → Satisfies lts s b →
      Satisfies lts s (.and a b)
    | diamond {s s' : State} {μ : Label} {a : Proposition Label}
      (htr : lts.Tr s μ s') (hs : Satisfies lts s' a) : Satisfies lts s (.diamond μ a)
    | neg {s : State} {a : Proposition Label} :
      Unsatisfies lts s a →
      Satisfies lts s (.neg a)

  /-- Unsatisfaction relation. `Unsatisfies lts s a` means that, in the LTS `lts`,
    the state `s` does not satisfy the proposition `a`. -/
  @[scoped grind]
  inductive Unsatisfies (lts : LTS State Label) : State → Proposition Label → Prop where
    | negAnd₁ {s : State} {a b : Proposition Label} :
      Unsatisfies lts s a →
      Unsatisfies lts s (.and a b)
    | negAnd₂ {s : State} {a b : Proposition Label} :
      Unsatisfies lts s b →
      Unsatisfies lts s (.and a b)
    | negDiamond {s : State} {μ : Label} {a : Proposition Label} :
      (∀ s', lts.Tr s μ s' → Unsatisfies lts s' a) →
      Unsatisfies lts s (.diamond μ a)
    | negNeg {s : State} {a : Proposition Label} :
      Satisfies lts s a →
      Unsatisfies lts s (.neg a)
end
```

This has the upside of having an easier termination argument.

All options have the property to be easy in the sense that one can `repeat constructor` everything away that's non-modal:

```lean
example : Satisfies lts s (.neg .false) := by
  repeat constructor

example : Satisfies lts s (.and (.neg .false) (.box a .true)) := by
  repeat constructor
  intro s htr
  repeat constructor
```

fmontesi:
I like the mutual approach, but Unsatisfies still feels like hardcoding what we mean by negation in all cases. So I'm still a bit on the fence regarding which system is better. @arademaker @mperessotti any thoughts?

Are we perhaps facing a situation where we should offer *both* systems and prove they're equivalent? @chenson2018 this might be akin to the Lambda calculus situation.

arademaker:
Thank you, @fmontesi, for leading the PR reviews so far, and @benkeks for your interest in our work and the time you put to present your point of view. I recognize your extensive experience with HML (your https://www.isa-afp.org/entries/Weak_Spectroscopy.html is in my reading list!)

I think your perspective on game semantics makes sense for your area of work. And the implementation of Satisfies/Unsatisfies reflects this view.

On the other hand, aligned with @Fabrizio, we have the impression that we may also have applications—which I cannot anticipate—in which the current presentation may be more appropriate (for different reasons: pedagogical, clarity, simplicity, etc.)

Furthermore, I think this type of situation will become increasingly common in CSLib as more logics are added; we will certainly have different presentations, some completely compatible, others where differences in syntax may have greater implications for semantics and for the formulation or proof of desirable theorems.

That said, I prefer that we consider how to integrate both approaches into CSLib and, if possible, demonstrate their equivalence. It would be an interesting parallel with data structures and algorithms in general; multiple approaches sometimes prove their value, especially when we can achieve translations or mappings between them.

benkeks:
I'm also really wondering how to handle such situations cleanly for a big library in systems like Lean. One would want to profit, for instance, from the fact that, syntactically, `HML = Propositional Logic - vars + modality` and `µ-calculus = HML + fixpoints`, without too much duplicate boilerplate. Does it make sense to build/use something like [MathLib.ModelTheory](https://leanprover-community.github.io/mathlib4_docs/Mathlib/ModelTheory/Basic.html) to be shared between several logics or rather not?

Similar story with adding “weak” modalities for weak/branching bisimilarity etc. without reimplementing core HML. ( @arademaker  I'm happy to hear that you take interest in behavioral *spectroscopies*. :D Find my [strange Hennessy–Milner thm for our quite special brand of HML](https://www.isa-afp.org/thys/AFP/Weak_Spectroscopy/Branching_Bisimilarity.html#Branching_Bisimilarity.lts_tau.sr_branching_bisim_is_hmlsrbb|thm) over there. ;))

Is there a nice way of having one's cake and eating it?

fmontesi:
I've been working on exactly this kind of problem for a while now, and the natural way to address reuse about this kind of stuff in Lean seems to be defining appropriate typeclasses for different logical concepts (including connective, equivalences, etc.). I hope to have something to start a discussion in a few weeks. Extending inductives is unfortunately hard to do without boilerplate... :-\

PR #323, opened 2026-02-06
chore: golf proofs about merge

eric-wieser:
The lemmas this golfs are now unused, and could also be deleted.

### DIFF
--- a/Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
+++ b/Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
@@ -66,10 +66,7 @@ open List
 /-- Our merge computes the one already in mathlib. -/
 @[simp, grind =]
 theorem ret_merge (xs ys : List α) : ⟪merge xs ys⟫ = xs.merge ys := by
-  fun_induction merge
-  · simp
-  · simp
-  · grind [cons_merge_cons]
+  fun_induction merge with grind [nil_merge, merge_right, cons_merge_cons]
 
 /-- A list is sorted if it satisfies the `Pairwise (· ≤ ·)` predicate. -/
 abbrev IsSorted (l : List α) : Prop := List.Pairwise (· ≤ ·) l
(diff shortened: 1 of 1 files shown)


Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
```diff
@@ -63,6 +63,14 @@ section Correctness
 
 open List
 
+/-- Our merge computes the one already in mathlib. -/
+@[simp, grind =]
+theorem ret_merge (xs ys : List α) : ⟪merge xs ys⟫ = xs.merge ys := by
+  fun_induction merge
+  · simp
+  · simp
+  · grind [cons_merge_cons]
```
chenson2018: I usually try to stay consistent about using `simp` or `grind` when they are both just working with theorems that make sense as `@[simp, grind =]` attributes, e.g.

```suggestion
  · grind [nil_merge]
  · grind [merge_right]
  · grind [cons_merge_cons]
```
(If these were already annotated upstream I'd just write `fun_induction merge <;> grind`, but I don't think it's helpful when you are specifying theorems for different branches manually.)

Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
```diff
@@ -177,16 +178,13 @@ theorem timeMergeSortRec_le (n : ℕ) : timeMergeSortRec n ≤ T n := by
 
 @[simp] theorem merge_ret_length_eq_sum (xs ys : List α) :
     ⟪merge xs ys⟫.length = xs.length + ys.length := by
-  fun_induction merge with
-  | case3 =>
-    grind
-  | _ => simp
+  simp
```
chenson2018, line 173: If you address the `simpNF` linter by removing `ret_merge`, this is the only affected proof that then becomes

```suggestion
  grind [List.length_merge]
```

Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
chenson2018: Seems like this should follow suit and be `grind [List.merge_perm_append]`

chenson2018: If golfing, this could just be `grind [Nat.clog_of_one_lt]`.

Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
eric-wieser: I've combined onto one line.

Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
eric-wieser: I think `ret_merge` is the better simp lemma, I've removed this one.

PR #325, opened 2026-02-06
feat: prove that the Buchi congruence has the saturation property

ctchou:
This PR proves that the Buchi congruence introduced in #278 has the saturation property defined in `Cslib.Foundations.Data.Set.Saturation`.  More precisely, the family of omega-languages of the form `U * V^ω`, where `U` and `V` are equivalence classes of the Buchi congruence, saturates the omega-language accepted by the underlying Buchi automaton.  This proof is the hardest step in proving the closure of ω-regular languages under complementation and explains why the Buchi congruence is defined the way it is.  Some miscellaneous results about LTS and infinite sequences that are needed by the proof are also added.

### DIFF
--- a/Cslib.lean
+++ b/Cslib.lean
@@ -89,6 +89,7 @@ public import Cslib.Languages.LambdaCalculus.LocallyNameless.Fsub.Typing
 public import Cslib.Languages.LambdaCalculus.LocallyNameless.Fsub.WellFormed
 public import Cslib.Languages.LambdaCalculus.LocallyNameless.Stlc.Basic
 public import Cslib.Languages.LambdaCalculus.LocallyNameless.Stlc.Safety
+public import Cslib.Languages.LambdaCalculus.LocallyNameless.Stlc.StrongNorm
 public import Cslib.Languages.LambdaCalculus.LocallyNameless.Untyped.Basic
 public import Cslib.Languages.LambdaCalculus.LocallyNameless.Untyped.FullBeta
 public import Cslib.Languages.LambdaCalculus.LocallyNameless.Untyped.FullBetaConfluence
@@ -98,6 +99,7 @@ public import Cslib.Languages.LambdaCalculus.LocallyNameless.Untyped.LcAt
 public import Cslib.Languages.LambdaCalculus.LocallyNameless.Untyped.MultiApp
 public import Cslib.Languages.LambdaCalculus.LocallyNameless.Untyped.MultiSubst
 public import Cslib.Languages.LambdaCalculus.LocallyNameless.Untyped.Properties
+public import Cslib.Languages.LambdaCalculus.LocallyNameless.Untyped.StrongNorm
 public import Cslib.Languages.LambdaCalculus.Named.Untyped.Basic
 public import Cslib.Logics.HML.Basic
 public import Cslib.Logics.HML.LogicalEquivalence
--- a/Cslib/Languages/LambdaCalculus/LocallyNameless/Stlc/StrongNorm.lean
+++ b/Cslib/Languages/LambdaCalculus/LocallyNameless/Stlc/StrongNorm.lean
@@ -1,0 +1,124 @@
+/-
+Copyright (c) 2025 David Wegmann. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: David Wegmann
+-/
+
+module
+
+public import Cslib.Foundations.Data.HasFresh
+public import Cslib.Foundations.Data.Relation
+public import Cslib.Languages.LambdaCalculus.LocallyNameless.Stlc.Basic
+public import Cslib.Languages.LambdaCalculus.LocallyNameless.Untyped.FullBeta
+public import Cslib.Languages.LambdaCalculus.LocallyNameless.Untyped.StrongNorm
+public import Cslib.Languages.LambdaCalculus.LocallyNameless.Untyped.LcAt
+public import Cslib.Languages.LambdaCalculus.LocallyNameless.Untyped.MultiSubst
+
+@[expose] public section
+
+set_option linter.unusedDecidableInType false
+
+namespace Cslib
+
+universe u v
+
+namespace LambdaCalculus.LocallyNameless.Stlc
+
+open Untyped Typing LambdaCalculus.LocallyNameless.Untyped.Term
+
+variable {Var : Type u} {Base : Type v} [DecidableEq Var] [HasFresh Var]
+
+open LambdaCalculus.LocallyNameless.Stlc
+open scoped Term
+
+/-- A set of terms is called saturated if it
+
+    1. only contains locally closed terms,
+    2. only contains strongly normalizing terms,
+    3. contains all neutral locally closed terms, and
+    4. is closed under top-level β-reduction of the form (λ M) N P₁ … Pₙ → M ^ N P₁ … Pₙ.
+-/
+@[scoped grind]
+structure Saturated (S : Set (Term Var)) : Prop where
+  lc : ∀ M ∈ S, LC M
+  sn : ∀ M ∈ S, SN M
+  neutal_lc : ∀ M, Neutral M → LC M → M ∈ S
+  multiApp : ∀ M N P, LC N → SN N → multiApp (M ^ N) P ∈ S → multiApp (M.abs.app N) P ∈ S
+
+/-- The semantic map maps each type to a corresponding saturated set of terms.
+    For the strong normalization proof to work, we must ensure that
+    Γ ⊢ t ∶ τ implies that t is in the set of terms corresponding to τ.
+
+    Strong normalization later follows from the fact that terms in saturated
+    sets are strongly normalizing.
+-/
+@[simp, scoped grind =]
+def semanticMap : Ty Base → Set (Term Var)
+  | .base _ => { t | SN t ∧ LC t }
+  | .arrow τ₁ τ₂ => { t | ∀ s, s ∈ semanticMap τ₁ → app t s ∈ semanticMap τ₂ }
+
+/-- The sets constructed by semanticMap are saturated -/
+lemma semanticMap_saturated (τ : Ty Base) : @Saturated Var (semanticMap τ) := by
+  induction τ with
+  | base => grind [sn_abs_app_multiApp, sn_neutral, open_abs_lc]
+  | arrow τ₁ τ₂ ih₁ ih₂ =>
+    constructor
+    · grind [ih₁.neutal_lc (fvar <| fresh {}) (.fvar <| fresh {}) (.fvar <| fresh {}), cases LC]
+    · grind [sn_app_left (Var := Var) (N := fvar <| fresh {})]
+    · grind
+    · intro M N P _ _ _ s _
+      grind [ih₂.multiApp M N (s :: P)]
+
+/-- The `entails_context` predicate ensures that each variable in the context
+    is mapped to a term in the corresponding semantic map. -/
+abbrev entails_context (E : Term.Env Var) (Γ : Context Var (Ty Base)) :=
+  ∀ {x τ}, ⟨x, τ⟩ ∈ Γ → (multiSubst E (fvar x)) ∈ semanticMap τ
+
+/-- The empty context is entailed by any environment. -/
+lemma entails_context_empty {Γ : Context Var (Ty Base)} : entails_context [] Γ := by
+  have := semanticMap_saturated (Var := Var) (Base := Base)
+  grind
+
+open scoped Context in
+omit [HasFresh Var] in
+/-- The `entails_context` predicate is preserved when extending the context
+    with a new variable, provided the new variable is fresh and its
+    substitution is in the corresponding semantic map. -/
+lemma entails_context_cons (E : Term.Env Var) (Γ : Context Var (Ty Base))
+    (x : Var) (τ : Ty Base) (sub : Term Var)
+    (h_fresh : x ∉ E.dom ∪ E.fv ∪ Γ.dom)
+    (h_mem : sub ∈ semanticMap τ) :
+    entails_context E Γ → entails_context (⟨ x, sub ⟩ :: E) (⟨ x, τ ⟩ :: Γ) := by
+  grind [multiSubst_fvar_fresh, subst_fresh, multiSubst_preserves_not_fvar]
+
+/-- The `entails` predicate states that a term `t` is
+    semantically valid with respect to a context `Γ` and a type `τ` -/
+abbrev entails (Γ : Context Var (Ty Base)) (t : Term Var) (τ : Ty Base) :=
+    ∀ E, env_LC E → (entails_context E Γ) → (multiSubst E t) ∈ semanticMap τ
+
+/-- The `soundness` lemma states that if a term `t` has type `τ` in context `Γ`,
+    then `t` is semantically valid with respect to `Γ` and `τ` -/
+lemma soundness {Γ : Context Var (Ty Base)} (derivation_t : Γ ⊢ t ∶ τ) : entails Γ t τ := by
+  induction derivation_t with
+  | var Γ xσ_mem_Γ => grind
+  | @abs σ Γ t τ L HL IH =>
+    intro E _ _ s
+    have sat_semMap_σ := semanticMap_saturated (Var := Var) σ
+    have sat_semMap_τ := semanticMap_saturated (Var := Var) τ
+    have := sat_semMap_τ.multiApp (multiSubst E t) s []
+    let := multiSubst E t
+    have ⟨x, _⟩ := fresh_exists <| E.dom ∪ free_union [fv, Context.dom, Env.fv] Var
+    have := IH (x := x) (E := ⟨x,s⟩ :: E)
+    grind [multiSubst_abs, entails_context_cons, multiSubst_open_var]
+  | app => grind [multiSubst_app]
+
+/-- Using soundness and the fact that the empty context
+    is entailed by any environment, we can conclude that
+    a well-typed term is strongly normalizing. -/
+theorem strong_norm {t : Term Var} {τ : Ty Base} (der : Γ ⊢ t ∶ τ) : SN t := by
+  apply (semanticMap_saturated τ).sn
+  apply (soundness der [] (by grind) entails_context_empty)
+
+end LambdaCalculus.LocallyNameless.Stlc
+
+end Cslib
--- a/Cslib/Languages/LambdaCalculus/LocallyNameless/Untyped/StrongNorm.lean
+++ b/Cslib/Languages/LambdaCalculus/LocallyNameless/Untyped/StrongNorm.lean
@@ -1,0 +1,153 @@
+/-
+Copyright (c) 2025 David Wegmann. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: David Wegmann
+-/
+
+module
+
+public import Cslib.Languages.LambdaCalculus.LocallyNameless.Untyped.FullBeta
+public import Cslib.Languages.LambdaCalculus.LocallyNameless.Untyped.MultiApp
+public import Cslib.Languages.LambdaCalculus.LocallyNameless.Untyped.LcAt
+
+@[expose] public section
+
+set_option linter.unusedDecidableInType false
+
+namespace Cslib
+
+universe u
+
+namespace LambdaCalculus.LocallyNameless.Untyped.Term
+
+variable {Var : Type u} {t t' : Term Var}
+
+open FullBeta
+
+attribute [grind =] Finset.union_singleton
+
+/-- A term is strongly normalizing if every reduction sequence terminates at some point.
+    This is ensured by the following type as inductive data must always be finite. -/
+inductive SN {α} : Term α → Prop
+| sn t : (∀ t', t ⭢βᶠ t' → SN t') → SN t
+
+attribute [scoped grind .] SN.sn
+
+/-- A single β-reduction step preserves strong normalization. -/
+lemma sn_step (t_st_t' : t ⭢βᶠ t') (sn_t : SN t) : SN t' := by
+  grind [cases SN]
+
+/-- Multiple β-reduction steps also preserve strong normalization. -/
+lemma sn_steps (t_st_t' : t ↠βᶠ t') (sn_t : SN t) : SN t' := by
+  induction t_st_t' with grind [sn_step]
+
+/-- Free variables are strongly normalizing. -/
+lemma sn_fvar {x : Var} : SN (fvar x) := by
+  grind [cases FullBeta]
+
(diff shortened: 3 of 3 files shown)


Cslib/Computability/Languages/Congruences/BuchiCongruence.lean
```diff
@@ -72,4 +74,128 @@ lemma buchiCongrParam_surjective : Surjective na.BuchiCongrParam := by
...
 
+/-- If `xl` and `yl` belong to the same equivalence class of `na.BuchiCongruence`
+and `xl` can move `na` from state `s` to state `t`, then so can `yl` and, furthermore,
+if `xl` makes `na` go through an accepting state of `na`, then so can `yl`. -/
+lemma buchiCongruence_transfer
+    {a : Quotient na.BuchiCongruence.eq} {xl yl : List Symbol} {s t : State}
+    (hc : xl ∈ na.BuchiCongruence.eqvCls a) (hc' : yl ∈ na.BuchiCongruence.eqvCls a)
+    (hp : xl ∈ na.pairLang s t) : ∃ sl, na.IsExecution s yl t sl ∧
+      ( xl ∈ na.pairViaLang na.accept s t → ∃ r ∈ na.accept, r ∈ sl ) := by
+  have h_eq : na.BuchiCongruence.eq xl yl := by
```
chenson2018, line 86: That you are having to manually specify this setoid instead of using notation is troubling.

Cslib/Computability/Languages/Congruences/BuchiCongruence.lean
```diff
@@ -72,4 +74,128 @@ lemma buchiCongrParam_surjective : Surjective na.BuchiCongrParam := by
...
+  have := h_eq s t
+  by_cases h_xl : xl ∈ na.pairViaLang na.accept s t
+  · have h_yl : yl ∈ na.pairViaLang na.accept s t := by grind
+    obtain ⟨r, _, yl1, yl2, h_yl1, h_yl2, rfl⟩ := LTS.mem_pairViaLang.mp h_yl
+    obtain ⟨sl1, h_yl1⟩ := LTS.mTr_isExecution h_yl1
+    obtain ⟨sl2, h_yl2⟩ := LTS.mTr_isExecution h_yl2
+    have := LTS.IsExecution.comp h_yl1 h_yl2
+    grind [LTS.IsExecution]
+  · have h_yl : yl ∈ na.pairLang s t := by grind
+    grind [LTS.mTr_isExecution h_yl]
```
chenson2018: The destructing is redundant here:

```suggestion
  have h_yl : yl ∈ na.pairLang s t := by grind
  have := LTS.mTr_isExecution h_yl
  grind [LTS.mem_pairViaLang, LTS.IsExecution, → LTS.IsExecution.comp, → LTS.mTr_isExecution]
```

Cslib/Computability/Languages/Congruences/BuchiCongruence.lean
```diff
@@ -72,4 +74,128 @@ lemma buchiCongrParam_surjective : Surjective na.BuchiCongrParam := by
...
+private lemma frequently_via_accept [Inhabited Symbol]
+    {xl : List Symbol} {xls : ωSequence (List Symbol)} {ss : ωSequence State}
+    (h_acc : ∃ᶠ (k : ℕ) in atTop, ss k ∈ na.accept) (h_exec : na.ωTr ss (xl ++ω xls.flatten))
+    (h_xls_p : ∀ (k : ℕ), (xls k).length > 0)
+    (f : ℕ → ℕ) (h_f : f = fun k => xl.length + xls.cumLen k)
+    (ts : ωSequence State) (h_ts : ts = ωSequence.mk (fun k ↦ ss (f k))) :
+    ∃ᶠ (k : ℕ) in atTop, xls k ∈ na.pairViaLang na.accept (ts k) (ts (k + 1)) := by
+  have hm : StrictMono f := by
+    intro m n h_mn
+    grind [cumLen_strictMono h_xls_p h_mn]
```
chenson2018:
```suggestion
  have hm : StrictMono f := by grind [StrictMono, cumLen_strictMono]
```

Cslib/Computability/Languages/Congruences/BuchiCongruence.lean
```diff
@@ -72,4 +74,128 @@ lemma buchiCongrParam_surjective : Surjective na.BuchiCongrParam := by
...
+  apply Frequently.mono <| frequently_in_strictMono hm h_acc
+  rintro n ⟨k, _, _⟩
+  apply LTS.mem_pairViaLang.mpr
+  use ss (f n + k), by grind
+  use (xls n).take k, (xls n).drop k
```
chenson2018:
```suggestion
  use ss (f n + k), by grind, (xls n).take k, (xls n).drop k
```

Cslib/Computability/Languages/Congruences/BuchiCongruence.lean
```diff
@@ -72,4 +74,128 @@ lemma buchiCongrParam_surjective : Surjective na.BuchiCongrParam := by
...
+    suffices (xl ++ω xls.flatten).extract (f n) (f n + k) = (xls n).take k by grind
+    have : xls.flatten.extract (xls.cumLen n) (xls.cumLen n + k) = (xls n).take k := by
+      grind [extract_flatten h_xls_p n]
+    grind only [extract_apppend_right_right]
+  · have := LTS.isExecution_mTr na.toLTS <|
+      LTS.ωTr_isExecution h_exec (show f n + k ≤ f (n + 1) by grind)
+    suffices (xl ++ω xls.flatten).extract (f n + k) (f (n + 1)) = (xls n).drop k by grind
+    have : xls.flatten.extract (xls.cumLen n + k) (xls.cumLen (n + 1)) = (xls n).drop k := by
+      grind [extract_flatten h_xls_p n]
+    grind only [= cumLen_succ, extract_apppend_right_right]
```
chenson2018: I understand performance was a consideration here, but it seems with this extracted we can remove the `grind only` and clean this up a bit. Here's what I came up with:

```suggestion
  have := extract_flatten h_xls_p n
  have exec {m n} (h : m ≤ n) := LTS.isExecution_mTr na.toLTS <| LTS.ωTr_isExecution h_exec h
  split_ands
  · have h : f n ≤ f n + k := by lia
    specialize exec h
    grind [extract_apppend_right_right]
  · have h : f n + k ≤ f (n + 1) := by lia
    specialize exec h
    grind [extract_apppend_right_right]
```

Cslib/Computability/Languages/Congruences/BuchiCongruence.lean
```diff
@@ -72,4 +74,128 @@ lemma buchiCongrParam_surjective : Surjective na.BuchiCongrParam := by
...
+  obtain ⟨yl, yls, h_yl_c, h_yls_c, rfl⟩ := mem_buchiFamily.mp h_ys
+  obtain ⟨ss, ⟨h_init, h_exec⟩, h_acc⟩ := h_lang
+  have h_xls_p (k : ℕ) : (xls k).length > 0 := by
+    grind [Language.mem_sub_one, List.ne_nil_iff_length_pos]
+  have h_yls_p (k : ℕ) : (yls k).length > 0 := by
+    grind [Language.mem_sub_one, List.ne_nil_iff_length_pos]
+  let f (k : ℕ) := xl.length + xls.cumLen k
+  let ts := ωSequence.mk (fun k ↦ ss (f k))
+  have h_xl_e : xl ∈ na.pairLang (ss 0) (ts 0) := by
+    have := LTS.ωTr_mTr h_exec (show 0 ≤ xl.length by grind)
```
chenson2018: This is going to take some more time for me to work through this, but could we start by unfolding all these nested uses of grind? The pattern of `show ... by grind` hurts readability IMO, I would use a `have` on the previous line.

Cslib/Computability/Languages/Congruences/BuchiCongruence.lean
```diff
@@ -72,4 +74,128 @@ lemma buchiCongrParam_surjective : Surjective na.BuchiCongrParam := by
...
+  have h_yls (k : ℕ) : ∃ sl, na.IsExecution (ts k) (yls k) (ts (k + 1)) sl ∧
+      ( (xls k) ∈ na.pairViaLang na.accept (ts k) (ts (k + 1)) → ∃ s ∈ na.accept, s ∈ sl ) := by
+    exact buchiCongruence_transfer ((h_xls_c k).left) ((h_yls_c k).left) (h_xls_e k)
+  choose sls h_yls_e h_yls_a using h_yls
+  obtain ⟨ss1, h_ss1_run, h_ss1_seg⟩ := LTS.IsExecution.flatten h_yls_e h_yls_p
+  obtain ⟨ss2, _, _, _, _⟩ := LTS.ωTr.append h_yl_e h_ss1_run (by
+    have : ss1 0 = (sls 0)[0]'(by grind) := by
+      grind [get_extract (xs := ss1) (show 0 < yls.cumLen 1 - yls.cumLen 0 by grind)]
+    have : ts 0 = (sls 0)[0]'(by grind) := by grind [LTS.IsExecution]
+    grind)
```
chenson2018: Similarly for these proofs with indexing, I would separate them out. Currently this also has repeated proofs, so this will also be a (small) help to performance. Like this for example:

```suggestion
    have h : 0 < yls.cumLen 1 - yls.cumLen 0 := by grind
    have : 0 < (sls 0).length := by grind
    have : ss1 0 = (sls 0)[0] := by grind [get_extract (xs := ss1) h]
    have : ts 0 = (sls 0)[0] := by grind [LTS.IsExecution]
    grind)
```

More broadly, I would also move this above to a separate `have : ss1 0 = ts 0 := by ...`.

Cslib/Foundations/Semantics/LTS/Basic.lean
```diff
@@ -215,6 +215,36 @@ theorem LTS.mTr_isExecution_iff : lts.MTr s1 μs s2 ↔
...
+    {lts : LTS State Label} {s r t : State} {μs1 μs2 : List Label} {ss1 ss2 : List State}
+    (h1 : lts.IsExecution s μs1 r ss1) (h2 : lts.IsExecution r μs2 t ss2)
+    (k : ℕ) (h_k : k < ss2.length) :
+    (ss1 ++ ss2.tail)[μs1.length + k]'(by grind) = ss2[k] := by
+  by_cases h : k = 0
+  · simp (disch := grind) only [h, add_zero, List.getElem_append_left]
+    grind
+  · simp (disch := grind) only [List.getElem_append_right, List.getElem_tail]
+    have : μs1.length + k - ss1.length + 1 = k := by grind
+    grind only
```
chenson2018: This doesn't seem to impact performance

```suggestion
    grind
```

Cslib/Foundations/Semantics/LTS/Basic.lean
```diff
@@ -215,6 +215,36 @@ theorem LTS.mTr_isExecution_iff : lts.MTr s1 μs s2 ↔
...
+  split_ands
+  · grind
+  · have := LTS.IsExecution.comp_seg2 h1 h2 μs2.length (by grind)
+    grind only [IsExecution, = List.length_append]
+  · intro k h_k
+    by_cases k < μs1.length
+    · grind only [IsExecution, = List.getElem_append]
+    · have := LTS.IsExecution.comp_seg2 h1 h2 (k - μs1.length) (by grind)
+      have := LTS.IsExecution.comp_seg2 h1 h2 (k - μs1.length + 1) (by grind)
+      grind
```
chenson2018: You can defer some `grind` calls to the end. I'm actually not sure, but maybe this is better for performance (and also less cluttered regardless)

```suggestion
    · have := LTS.IsExecution.comp_seg2 h1 h2 (k - μs1.length)
      have := LTS.IsExecution.comp_seg2 h1 h2 (k - μs1.length + 1)
      grind
```

Cslib/Foundations/Semantics/LTS/Basic.lean
```diff
@@ -215,6 +215,36 @@ theorem LTS.mTr_isExecution_iff : lts.MTr s1 μs s2 ↔
...
+/-- The composition of two executions is an execution. -/
+theorem LTS.IsExecution.comp
+    {lts : LTS State Label} {s r t : State} {μs1 μs2 : List Label} {ss1 ss2 : List State}
+    (h1 : lts.IsExecution s μs1 r ss1) (h2 : lts.IsExecution r μs2 t ss2) :
+    lts.IsExecution s (μs1 ++ μs2) t (ss1 ++ ss2.tail) := by
+  have h0 : (ss1 ++ ss2.tail).length = (μs1 ++ μs2).length + 1 := by grind
+  use h0
+  split_ands
+  · grind
+  · have := LTS.IsExecution.comp_seg2 h1 h2 μs2.length (by grind)
```
chenson2018: Same thing here

```suggestion
  · have := LTS.IsExecution.comp_seg2 h1 h2 μs2.length
```

Cslib/Computability/Languages/Congruences/BuchiCongruence.lean
ctchou: Unfortunately `≈` doesn't work here, because it resolves to:
https://leanprover-community.github.io/mathlib4_docs/Init/Data/List/Perm.html#List.isSetoid
not `na.BuchiCongruence.eq`.

chenson2018: Can you open an issue for this? I'd like to revisit this later.

Cslib/Computability/Languages/Congruences/BuchiCongruence.lean
ctchou: Wow, how did you find out that the case splitting is unnecessary here?  When doing the proof in my mind, I always case-split.

Fixed.

Cslib/Computability/Languages/Congruences/BuchiCongruence.lean
ctchou: Fixed.

Cslib/Computability/Languages/Congruences/BuchiCongruence.lean
ctchou: Fixed.

Cslib/Computability/Languages/Congruences/BuchiCongruence.lean
ctchou: New issues: https://github.com/leanprover/cslib/issues/352

BTW, I had a similar issue when working on `Segment.lean`.  I was using `φ` for a generic `ℕ → ℕ` function in the `Nat` namespace and got all sorts weird behaviors.  Finally I realized that `φ` was already used by the Euler totient function in that namespace.  So I switched to using `f` instead.

Cslib/Computability/Languages/Congruences/BuchiCongruence.lean
ctchou: Fixed.

Cslib/Computability/Languages/Congruences/BuchiCongruence.lean
ctchou: Fixed.

Cslib/Computability/Languages/Congruences/BuchiCongruence.lean
ctchou: Fixed.

Cslib/Foundations/Semantics/LTS/Basic.lean

Cslib/Foundations/Semantics/LTS/Basic.lean

Cslib/Foundations/Semantics/LTS/Basic.lean

Cslib/Foundations/Semantics/LTS/Basic.lean
```diff
@@ -215,6 +215,36 @@ theorem LTS.mTr_isExecution_iff : lts.MTr s1 μs s2 ↔
     ∃ ss : List State, lts.IsExecution s1 μs s2 ss := by
   grind
 
+lemma LTS.IsExecution.comp_seg2
```
fmontesi: Might be a problem with my editor, but I can't find a `comp_seg`, so why is this called `comp_seg2`?

Cslib/Foundations/Semantics/LTS/Basic.lean
```diff
@@ -317,20 +347,17 @@ def LTS.ωTr (lts : LTS State Label) (ss : ωSequence State) (μs : ωSequence L
 
 variable {lts : LTS State Label}
 
-open scoped ωSequence in
+open ωSequence
+
 /-- Any finite execution extracted from an infinite execution is valid. -/
+theorem LTS.ωTr_isExecution (h : lts.ωTr ss μs) {n m : ℕ} (hnm : n ≤ m) :
```
fmontesi, line 350: I'd call this `LTS.ωTr_extract_isExecution`

Cslib/Computability/Languages/Congruences/BuchiCongruence.lean
```diff
@@ -72,4 +74,118 @@ lemma buchiCongrParam_surjective : Surjective na.BuchiCongrParam := by
...
+  use ss (f n + k), by grind, (xls n).take k, (xls n).drop k
+  have := extract_flatten h_xls_p n
+  have exec {m n} (h : m ≤ n) := LTS.isExecution_mTr na.toLTS <| LTS.ωTr_isExecution h_exec h
+  split_ands
+  · have h : f n ≤ f n + k := by lia
+    specialize exec h
+    grind [extract_apppend_right_right]
+  · have h : f n + k ≤ f (n + 1) := by lia
+    specialize exec h
+    grind [extract_apppend_right_right]
```
fmontesi: This makes me notice that apppend has an extra p.

Cslib/Foundations/Semantics/LTS/Basic.lean
chenson2018: And in general, I don't think we should have names with numbers like this.

chenson2018:
My delay in approving this is because of `buchiFamily_saturation`. I have looked at this off and on and could make some incremental `grind` improvements, but I don't have the subject expertise here. Is there anything that can be split off in separate lemmas? The existential from `mem_buchiFamily` seems to make this awkward, but is there any way this could be reworked more cleanly?

ctchou:
We can work on this after I get back from my trip on March 12. I don't have my laptop with me now.

Cslib/Foundations/Semantics/LTS/Basic.lean
ctchou: I changed the theorem's name to `LTS.IsExecution.comp_helper`, because it is just a helper lemma used in proving `LTS.IsExecution.comp`.

Cslib/Foundations/Semantics/LTS/Basic.lean
ctchou: Fixed.

Cslib/Computability/Languages/Congruences/BuchiCongruence.lean
ctchou: Fixed.

ctchou:
@chenson2018  Could you elaborate on your last comment?  I'm not sure I understand it.  `buchiFamly` is an indexed family of omega-regular languages, where each index is a pair of equivalence classes of `na.BuchiCongruence`.  The theorem `men_buchiFamily` merely relates a member of such a language with the two equivalence classes of its index.  I don't think the proof of `buchiFamily_saturation` can be simplified much.  If you still have the survey article by W. Thomas I sent you, this proof is essentially a formalization of the proof of part (a) of lemma 2.2 on page 140.

chenson2018:
@ctchou Let me take another look and at least add the cleanups to `grind` usage I'd done locally. My question was essentially if you can rephrase the existentials to be "constructive" so it'd be easier to extract some smaller lemmas.

ctchou:
I don't think that's possible.  In essence, here we are starting with something like `x ∈ U * V^ω`.  Both the concatenation operation `*` and the omega power operation `^ω` are essentially nondeterministic operations.  For example, `x ∈ U * V^ω` merely says that `x` is the concatenation of some finite sequence from `U` followed by some infinite sequence from `V^ω`.  We have no other information and no way to avoid the existential quantifiers.

chenson2018:
@ctchou Please see my last commit and let me know what you think. My aim here was to as much as possible organize things into as few calls to `grind` as possible and to keep the number of available hypothesis minimized at each area of the proof. 

(There is one particularly problematic call to grind I will point out separately.)

Cslib/Computability/Languages/Congruences/BuchiCongruence.lean
```diff
@@ -72,4 +75,110 @@ lemma buchiCongrParam_surjective : Surjective na.BuchiCongrParam := by
...
+    have h_xl_e : xl ∈ na.pairLang (ss 0) (ts 0) := by
+      grind [LTS.ωTr_mTr h_exec (?_ : 0 ≤ xl.length), extract_append_zero_right, LTS.mem_pairLang]
+    have h_yl_e : yl ∈ na.pairLang (ss 0) (ts 0) := by
+      grind [buchiCongruence_transfer h_xl_c h_yl_c h_xl_e, LTS.mem_pairLang, LTS.isExecution_mTr]
+    have h_ss1_ts : ss1 0 = ts 0 := by
+      have h : 0 < yls.cumLen 1 - yls.cumLen 0 := by grind
+      have : 0 < (sls 0).length := by grind
+      have : ss1 0 = (sls 0)[0] := by grind [get_extract (xs := ss1) h]
+      -- TODO:: `grind` with `LTS.IsExecution` is problematic
+      have : ts 0 = (sls 0)[0] := by grind [LTS.IsExecution]
```
chenson2018: This `grind [LTS.IsExecution]` fails pretty hard if you try to use `grind?` (lists no theorems and even invalid anchors). Would it be possible to rewrite just this one manually? I just don't want to be stuck with this on nightly testing and have no idea how to proceed.

ctchou: @chenson2018  I found that `grind only [LTS.IsExecution]` works as well.  Is that good enough?  I can do a more manual proof if that's not good enough.

chenson2018: Oh, I see. I guess this is fine then, though I might have flipped the equality to just write
```lean
have : (sls 0)[0] = ts 0 := h_yls_e 0 |>.choose_spec |>.1
```
(This would look a lot nicer if we refactor `LTS.IsExecution` into a structure)

ctchou: Great!  I'll use your new proof.

ctchou: Your other changes look fine, by the way.

chenson2018 (approved):
Thanks for the patience with a longer review!

ctchou:
@chenson2018 It's a pleasure working with you.

PR #340, opened 2026-02-17
chore: create GitHub releases from version tags

kim-em:
## Summary

This PR adds a GitHub Actions workflow that automatically creates GitHub releases when version tags are pushed, mirroring the existing functionality in batteries.

See also https://github.com/leanprover-community/mathlib4/pull/35425

🤖 Prepared with Claude Code

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,21 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+"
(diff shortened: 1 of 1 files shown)


chenson2018:
I requeued so that I could change the title to pass the commit convention check. We can add `ci` as an allowed prefix to match Mathlib, it just isn't there currently because I took the workflow from the core repo.

fmontesi:
Thanks, Chris. I think allowing ci as a prefix is a good idea, makes it easy to spot these PRs.

PR #341, opened 2026-02-17
chore: use authenticated gh api in nightly bump workflow

kim-em:
This PR replaces unauthenticated `curl` calls to the GitHub API with `gh api` in the nightly bump workflow. The `gh api` command automatically uses the app token already generated in the workflow, avoiding transient failures from the unauthenticated rate limit (60 req/hour).

See [this mathlib4-nightly-testing failure](https://github.com/leanprover-community/mathlib4-nightly-testing/actions/runs/22084056453/job/63814863096) for an example of the problem: the API returned a rate-limit error object instead of the expected array, and `jq` crashed trying to index it.

Corresponding PRs:
- https://github.com/leanprover-community/mathlib4/pull/35427
- https://github.com/leanprover-community/batteries/pull/1678

🤖 Prepared with Claude Code

### DIFF
--- a/.github/workflows/bump_toolchain_nightly-testing.yml
+++ b/.github/workflows/bump_toolchain_nightly-testing.yml
@@ -28,12 +28,21 @@ jobs:
 
     - name: Get latest release tag from leanprover/lean4-nightly
       id: get-latest-release
+      env:
+        GH_TOKEN: ${{ steps.app-token.outputs.token }}
       run: |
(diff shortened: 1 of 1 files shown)


fmontesi (approved):

chenson2018:
@kim-em This seems to be failing, see https://github.com/leanprover/cslib/actions/runs/22094383045/job/63847471569

PR #342, opened 2026-02-17
chore(CombinatoryLogic): use Relation infrastructure

### LINKED ISSUE #333: SKI.ParallelReduction should use standard relation notation
Notation for parallel reduction -/

https://github.com/leanprover/cslib/blob/939bd13108ef482e217efda27fd01d9f57fe8a1b/Cslib/Languages/CombinatoryLogic/Confluence.lean#L62-L63

### LINKED ISSUE #334: inline `SKI.CommonReduct` as `Relation.MJoin Red`
Express that two terms have a reduce to a common term. -/

https://github.com/leanprover/cslib/blob/939bd13108ef482e217efda27fd01d9f57fe8a1b/Cslib/Languages/CombinatoryLogic/Defs.lean#L115-L116

### LINKED ISSUE #335: `SKI.redexFree_iff` and related theorems should use `Relation.Normal`
A term is redex free iff it has no one-step reductions. -/

https://github.com/leanprover/cslib/blob/939bd13108ef482e217efda27fd01d9f57fe8a1b/Cslib/Languages/CombinatoryLogic/Evaluation.lean#L147-L148

thomaskwaring:
... plus some small golfing & style.

closes #333 
closes #334 
closes #335

### DIFF
--- a/Cslib/Languages/CombinatoryLogic/Confluence.lean
+++ b/Cslib/Languages/CombinatoryLogic/Confluence.lean
@@ -118,13 +118,8 @@ lemma K_irreducible (a : SKI) (h : K ⭢ₚ a) : a = K := by
 
 lemma Ka_irreducible (a c : SKI) (h : (K ⬝ a) ⭢ₚ c) : ∃ a', a ⭢ₚ a' ∧ c = K ⬝ a' := by
   cases h
-  case refl =>
-    use a
-    exact ⟨ParallelReduction.refl a, rfl⟩
-  case par b a' h h' =>
-    use a'
-    rw [K_irreducible b h]
-    exact ⟨h', rfl⟩
+  case refl => use a, .refl a
+  case par b a' h h' => rw [K_irreducible b h]; use a'
 
 lemma S_irreducible (a : SKI) (h : S ⭢ₚ a) : a = S := by
   cases h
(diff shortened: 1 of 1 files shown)


Cslib/Languages/CombinatoryLogic/Basic.lean
```diff
@@ -107,37 +107,29 @@ theorem Polynomial.elimVar_correct {n : Nat} (Γ : SKI.Polynomial (n + 1)) {ys :
...
-    · simp_rw [SKI.Polynomial.eval]
-      replace hi := Nat.eq_of_lt_succ_of_not_lt i.isLt hi
-      simp_rw [Fin.getElem_fin, hi]
-      have app_len : (ys ++ [z]).length = n+1 := by simpa
+    · replace hi := Nat.eq_of_lt_succ_of_not_lt i.isLt hi
+      have app_len : (ys ++ [z]).length = n + 1 := by simpa
       have : (ys ++ [z])[n]'(by rw [app_len]; exact Nat.lt_add_one n) = z := by
         rw [List.getElem_append_right] <;> simp [hys]
-      rw [this]
+      simp_rw [SKI.Polynomial.eval, Fin.getElem_fin, hi, this]
```
chenson2018, line 117: I think this might be a bit slower, but note that this entire branch can be:
```lean
  | n, .var i =>
    have : n < (ys ++ [z]).length := by grind
    have : (ys ++ [z])[n] = z := by grind
    grind [elimVar, eval, Nat.mod_eq_of_lt]
```
(This seems like a nice use case for the interactive `grind` mode, but I'm not sure how to tell it to split the `if`.)

Cslib/Languages/CombinatoryLogic/Confluence.lean
```diff
@@ -112,15 +108,15 @@ Irreducibility for the (partially applied) primitive combinators.
...
   cases h
   rfl
 
-lemma K_irreducible (a : SKI) (h : K ⇒ₚ a) : a = K := by
+lemma K_irreducible (a : SKI) (h : K ⭢ₚ a) : a = K := by
   cases h
   rfl
 
-lemma Ka_irreducible (a c : SKI) (h : K ⬝ a ⇒ₚ c) : ∃ a', a ⇒ₚ a' ∧ c = K ⬝ a' := by
+lemma Ka_irreducible (a c : SKI) (h : (K ⬝ a) ⭢ₚ c) : ∃ a', a ⭢ₚ a' ∧ c = K ⬝ a' := by
```
chenson2018, line 119: Since we're golfing... if you change the the ordering of tactics in these `*_irreducible` theorems a bit, `use` will infer the `rfl` parts, for example:
```lean
lemma Ka_irreducible (a c : SKI) (h : (K ⬝ a) ⭢ₚ c) : ∃ a', a ⭢ₚ a' ∧ c = K ⬝ a' := by
  cases h
  case refl => use a, .refl a
  case par b a' h _ => rw [K_irreducible b h]; use a'
```
More generally, adjacent `use` can be combined with the terms separated by commas.

Cslib/Languages/CombinatoryLogic/Basic.lean
chenson2018: Also style-wise, you can be a bit more terse by making use of the namespace you're in, like above where I write `.var` instead of `SKI.Polynomial.var`. Maybe you know this already and did so for readability, but I personally like the shorter style

thomaskwaring: it seems noticeably slower to me so i might leave as-is, thanks though!

Cslib/Languages/CombinatoryLogic/Confluence.lean
thomaskwaring: nice! thanks

PR #344, opened 2026-02-17
chore: bump toolchain to v4.29.0-rc1

kim-em:


chenson2018:
@kim-em I've set `backward.isDefEq.respectTransparency` in the other places it is needed at the moment.

When you get a chance can you look at my comment on https://github.com/leanprover/cslib/pull/341? Not sure if I missed these because of that workflow being broken.

PR #357, opened 2026-02-20
feat: generalize `TimeM` to arbitrary cost types providing `AddZero`

eric-wieser:
This also fixes some insufficiently-general `simp` lemmas about `seqLeft`, `seqRight`, and `seq`.

Discussed in [#CSLib > Lightweight algorithms in CSLib :Cslib#275 @ 💬](https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/Lightweight.20algorithms.20in.20CSLib.20.3ACslib.23275/near/574964064); whatever we end up doing with #275, this change is a natural generalization and can be used to experiment with cleanups in #275. If we ultimately converge on dropping `TimeM`, then we can drop this generalization along with the rest of `TimeM` at that time.

### DIFF
--- a/Cslib/Algorithms/Lean/TimeM.lean
+++ b/Cslib/Algorithms/Lean/TimeM.lean
@@ -42,8 +42,8 @@ See [Danielsson2008] for the discussion.
 namespace Cslib.Algorithms.Lean
 
 /-- A monad for tracking time complexity of computations.
-`TimeM C α` represents a computation that returns a value of type `α`
-and accumulates a time cost (represented as a type `C`, typically `ℕ`). -/
+`TimeM T α` represents a computation that returns a value of type `α`
+and accumulates a time cost (represented as a type `T`, typically `ℕ`). -/
 @[ext]
 structure TimeM (T : Type*) (α : Type*) where
   /-- The return value of the computation -/


Cslib/Algorithms/Lean/TimeM.lean
```diff
@@ -38,60 +42,89 @@ See [Danielsson2008] for the discussion.
...
+instance [AddMonoid T] : LawfulMonad (TimeM T) := .mk'
   (id_map := fun x => rfl)
   (pure_bind := fun _ _ => by ext <;> simp)
-  (bind_assoc := fun _ _ _ => by ext <;> simp [Nat.add_assoc])
+  (bind_assoc := fun _ _ _ => by ext <;> simp [add_assoc])
+  (seqLeft_eq := fun _ _ => by ext <;> simp)
+  (bind_pure_comp := fun _ _ => by ext <;> simp)
 
-/-- Creates a `TimeM` computation with a time cost.
-The time cost defaults to 1 if not provided. -/
```
eric-wieser, line 93: This default was provided by the notation anyway, so no need to keep it here.

Shreyas4991:
You want `AddCommMonoid`. I can't imagine a situation where the cost type does not have commutative addition. It just makes life simpler in proofs.

Cslib/Algorithms/Lean/TimeM.lean
```diff
@@ -38,60 +42,88 @@ See [Danielsson2008] for the discussion.
 namespace Cslib.Algorithms.Lean
 
 /-- A monad for tracking time complexity of computations.
-`TimeM α` represents a computation that returns a value of type `α`
-and accumulates a time cost (represented as a natural number). -/
+`TimeM C α` represents a computation that returns a value of type `α`
+and accumulates a time cost (represented as a type `C`, typically `ℕ`). -/
```
chenson2018: In this docstring you write `TimeM C α` where elsewhere it has `T` for the type variable, was this intentional?

eric-wieser: Nope, this was an error.
```suggestion
`TimeM T α` represents a computation that returns a value of type `α`
and accumulates a time cost (represented as a type `T`, typically `ℕ`). -/
```

eric-wieser:
> You want `AddCommMonoid`. I can't imagine a situation where the cost type does not have commutative addition. It just makes life simpler in proofs.

Downstream code is free to work in the generality of commutative costs. There's no benefit to making the lawful instance less general.

fmontesi:
LGTM, thanks!

PR #360, opened 2026-02-21
feat: additive writer monad

Shreyas4991:
This PR adds a writer monad whose log type is additive. This fills a gap in the mathlib API whose writer monad treats the log as multiplicative, which is a mistake as far as computer science is concerned.

## Reference
-  ["As logging-effect in computer science"](https://ncatlab.org/nlab/show/action+monad)

### DIFF
--- a/Cslib/Foundations/Control/Monad/AddWriter.lean
+++ b/Cslib/Foundations/Control/Monad/AddWriter.lean
@@ -6,6 +6,7 @@ Authors: Simon Hudon, Edward Ayers, Shreyas Srinivas
 module
 
 public import Mathlib
+public import Cslib.Init
 
 /-!
 # Additive Writer monads


eric-wieser:
> which is a mistake as far as computer science is concerned.

Citation needed. Computer science does not make the `Monoid`/`AddMonoid` distinction made by mathlib; and in fact haskell's monoid system makes you use `Writer (Sum X)` and `Writer (Product X)` to choose between `+` and `*` as the accumulation operator (instead of the default of `++`)

There is a zulip discussion [#mathlib4 > Writer should use an additive monoid @ 💬](https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/Writer.20should.20use.20an.20additive.20monoid/near/574990415). If CSLib reviewers have an opinion, I'd propose they leave it there rather than encouraging duplication into cslib.

Shreyas4991:
Extra reference: https://homepages.inf.ed.ac.uk/wadler/papers/marktoberdorf/baastad.pdf

PR #365, opened 2026-02-23
feat: insertionSort

ichxorya:
Implement the insertion sort algorithm with few changes to existing merge sort code (generalize some parts)

- Separate shared sorting definitions like min/max and sorted/strictly sorted.
- Implement proofs for insertion sort, notably: best/worst case time complexity, proof of correctness, largest bound of time complexity

### DIFF
--- a/Cslib/Algorithms/Lean/InsertionSort/InsertionSort.lean
+++ b/Cslib/Algorithms/Lean/InsertionSort/InsertionSort.lean
@@ -23,7 +23,10 @@ comparisons.
 
 - `insertionSort_correct`: `insertionSort` permutes the list into a sorted one.
 - `insertionSort_time`: The number of comparisons of `insertionSort` is at most `(n * (n - 1)) / 2`.
-
+- `insertionSort_time_worst_case`: In the worst case, the number of comparisons
+is exactly `(n * (n - 1)) / 2`.
+- `insertionSort_time_best_case`: In the best case, the number of comparisons
+is `n - 1` for a non-empty list, and `0` for an empty list.
 -/
 
 set_option autoImplicit false


eric-wieser:
Note that #275 also contains an insertionSort algorithm.

Cslib/Algorithms/Lean/Sorting.lean
```diff
@@ -0,0 +1,107 @@
...
+
+/-- A list is strictly sorted (ascending/increasing)
+if it satisfies the `Pairwise (· < ·)` predicate.
+-/
+abbrev IsStrictlySortedAscending (l : List α) : Prop := List.Pairwise (· < ·) l
+
+/-- A list is strictly sorted (descending/decreasing)
+if it satisfies the `Pairwise (· > ·)` predicate.
+-/
+abbrev IsStrictlySortedDescending (l : List α) : Prop := List.Pairwise (· > ·) l
```
eric-wieser, line 54: Mathlib already has `List.SortedLE` and `List.SortedGE`  and `List.SortedLT` and `List.SortedGT` for this purpose.

ichxorya: I would refactor this, thank you for your suggestions. Originally Sorrachai included AscSorted in Merge Sort, hence I think we could bring it to a separated module and extend it with more concepts

ichxorya:
> Note that #275 also contains an insertionSort algorithm.

My bad for not looking at the PR. Let me check for other PR as well, since there have been people worked on this before me.

Edit: I'll be waiting for 275's status. I hope that my proof of best/worst case can still possibly be useful in later updates.

PR #370, opened 2026-02-25
LongestCommonSublist: TimeM proof

kuotsanhsu:
Time complexity using `TimeM` is proven, but `TimeM` is messing up my lemmas `aux2_1` and `cons_aux0_cons` for the correctness proof `lcsDP` whereas my [original non-monadic implementation](https://gist.github.com/kuotsanhsu/e21ec4cfb78f3b43e898ed19d55d3625) can utilize definitional equality of `List.foldr`. Reviewers can start giving feedback while I work on the 2 lemmas with `sorry`.

### DIFF
--- a/Cslib/Algorithms/Lean/LongestCommonSublist.lean
+++ b/Cslib/Algorithms/Lean/LongestCommonSublist.lean
@@ -1,0 +1,354 @@
+module
+
+public import Cslib.Algorithms.Lean.TimeM
+public import Mathlib.Algebra.Group.Nat.Defs
+
+/-!
+# Longest Common Sublist (Subsequence)
+
+Two implementations, `lcsBrute` and `lcsDP`, of the longest common sublist (LCS) algorithm are
+provided. Their unit tests take the form of `#guard_msgs in #eval`. There could be multiple LCS for
+two given lists. We consistently yield an *opinionated* one among many.
+
+## Main results
+
+- `lcsBrute_correct`: The correctness proof of the slow-recursive implementation `lcsBrute`.
+- `lcsDP_eq_lcsBrute`: The functional equality between `lcsDP` and `lcsBrute`.
+- `lcsDP_correct`: The correctness proof of the dynamic programming implementation `lcsDP`.
+- `lcsDP_time`: The number of calls to `lcsDPM.dp` is proportional to the product of the length of
+  the two input lists.
+-/
+
+set_option autoImplicit false
+
+def test_string (lcs : List Char → List Char → List Char) (s t : String) : String :=
+  .ofList <| lcs s.toList t.toList
+
+namespace List
+
+variable {α} [DecidableEq α]
+
+/-- `cs` is a common sublist of the lists `xs` and `ys`. -/
+public structure CommonSublist (xs ys cs : List α) : Prop where
+  sublist₁ : cs <+ xs
+  sublist₂ : cs <+ ys
+
+/-- `lcs` is a longest common sublist of the lists `xs` and `ys`. -/
+public structure LCS (xs ys lcs : List α) : Prop extends CommonSublist xs ys lcs where
+  longest : ∀ cs, CommonSublist xs ys cs → cs.length ≤ lcs.length
+
+/-- Straightforward recursion but slow: not even the length of LCS is cached. -/
+def lcsBrute : List α → List α → List α
+  | [], _
+  | _, [] => []
+  | x :: xs, y :: ys =>
+    if x = y then
+      x :: lcsBrute xs ys
+    else
+      let xx := lcsBrute xs (y :: ys)
+      let yy := lcsBrute (x :: xs) ys
+      if xx.length ≥ yy.length then xx else yy
+
+section unittest
+
+/-- info: "" -/
+#guard_msgs(info) in
+#eval test_string lcsBrute "" ""
+
+/-- info: "" -/
+#guard_msgs(info) in
+#eval test_string lcsBrute "ABCD" ""
+
+/-- info: "" -/
+#guard_msgs(info) in
+#eval test_string lcsBrute "" "ACBAD"
+
+/-- info: "ACD" -/
+#guard_msgs(info) in
+#eval test_string lcsBrute "ABCD" "ACBAD"
+
+/-- info: "AC" -/
+#guard_msgs(info) in
+#eval test_string lcsBrute "GAC" "AGCAT"
+
+/-- info: "MJAU" -/
+#guard_msgs(info) in
+#eval test_string lcsBrute "XMJYAUZ" "MZJAWXU"
+
+end unittest
+
+theorem nil_lcsBrute : {ys : List α} → lcsBrute [] ys = []
+  | [] | _ :: _ => lcsBrute.eq_def ..
+
+theorem lcsBrute_nil : {xs : List α} → lcsBrute xs [] = []
+  | [] | _ :: _ => lcsBrute.eq_def ..
+
+theorem lcsBrute_correct : {xs ys : List α} → LCS xs ys (lcsBrute xs ys)
+  | [], ys =>
+    suffices LCS [] ys [] from nil_lcsBrute.symm.rec this
+    {
+      sublist₁ := show [] <+ [] from .slnil
+      sublist₂ := show [] <+ ys from ys.nil_sublist
+      longest
+      | cs, ⟨(h : cs <+ []), _⟩ =>
+        show cs.length ≤ [].length from
+        suffices cs = [] from this.rec Nat.le.refl
+        eq_nil_of_sublist_nil h
+    }
+  | xs, [] =>
+    suffices LCS xs [] [] from lcsBrute_nil.symm.rec this
+    {
+      sublist₁ := show [] <+ xs from xs.nil_sublist
+      sublist₂ := show [] <+ [] from .slnil
+      longest
+      | cs, ⟨_, (h : cs <+ [])⟩ =>
+        show cs.length ≤ [].length from
+        suffices cs = [] from this.rec Nat.le.refl
+        eq_nil_of_sublist_nil h
+    }
+  | x :: xs, y :: ys =>
+    if h₁ : x = y then
+      let xy := lcsBrute xs ys ; have hxy : LCS xs ys xy := lcsBrute_correct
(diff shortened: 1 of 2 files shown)


Cslib/Algorithms/Lean/LongestCommonSublist.lean
```diff
@@ -0,0 +1,354 @@
...
+    suffices LCS [] ys [] from nil_lcsBrute.symm.rec this
+    {
+      sublist₁ := show [] <+ [] from .slnil
+      sublist₂ := show [] <+ ys from ys.nil_sublist
+      longest
+      | cs, ⟨(h : cs <+ []), _⟩ =>
+        show cs.length ≤ [].length from
+        suffices cs = [] from this.rec Nat.le.refl
+        eq_nil_of_sublist_nil h
+    }
```
eric-wieser, line 97: Usually we would format this as
```suggestion
    { sublist₁ := show [] <+ [] from .slnil
      sublist₂ := show [] <+ ys from ys.nil_sublist
      longest
      | cs, ⟨(h : cs <+ []), _⟩ =>
        show cs.length ≤ [].length from
        suffices cs = [] from this.rec Nat.le.refl
        eq_nil_of_sublist_nil h }
```
(and throughout)

Cslib/Algorithms/Lean/LongestCommonSublist.lean
```diff
@@ -0,0 +1,354 @@
...
+
+def test_string (lcs : List Char → List Char → List Char) (s t : String) : String :=
+  .ofList <| lcs s.toList t.toList
+
+namespace List
+
+variable {α} [DecidableEq α]
+
+/-- `cs` is a common sublist of the lists `xs` and `ys`. -/
+public structure CommonSublist (xs ys cs : List α) : Prop where
```
eric-wieser, line 32:
```suggestion
/-- `IsCommonSublist xs ys xs` indicates that `cs` is a common sublist of the lists `xs` and `ys`. -/
public structure IsCommonSublist (xs ys cs : List α) : Prop where
```

Cslib/Algorithms/Lean/LongestCommonSublist.lean
```diff
@@ -0,0 +1,354 @@
...
+#guard_msgs(info) in
+#eval test_string lcsBrute "XMJYAUZ" "MZJAWXU"
+
+end unittest
+
+theorem nil_lcsBrute : {ys : List α} → lcsBrute [] ys = []
+  | [] | _ :: _ => lcsBrute.eq_def ..
+
+theorem lcsBrute_nil : {xs : List α} → lcsBrute xs [] = []
+  | [] | _ :: _ => lcsBrute.eq_def ..
```
eric-wieser, line 84: These should be `@[simp]`

Cslib/Algorithms/Lean/LongestCommonSublist.lean
```diff
@@ -0,0 +1,354 @@
...
+
+theorem lcsBrute_correct : {xs ys : List α} → LCS xs ys (lcsBrute xs ys)
+  | [], ys =>
```
eric-wieser, line 87: Instead of doing a manual pattern match here, I think it would be better to use `fun_induction`, which will aso handle the `x = y` case split for you.

Cslib/Algorithms/Lean/LongestCommonSublist.lean
```diff
@@ -0,0 +1,354 @@
...
+  aux2 (x : α) (dps : DPS α) : TimeM ℕ (DPS α × DP α × DP α) := dps.foldrM (aux3 x) default
+  aux3 (x : α) : α × DP α → DPS α × DP α × DP α → TimeM ℕ (DPS α × DP α × DP α)
+    | (y, xx), (dps, yy, xy) => do
+      ✓ let dp := dp x y xx yy xy
+      return ((y, dp) :: dps, dp, xx)
+  dp (x y : α) (xx yy xy : DP α) : DP α :=
+    if x = y then
+      ⟨x :: xy.lcs, xy.length + 1, xy.length_eq.rec rfl⟩
+    else
+      if xx.length ≥ yy.length then xx else yy
```
eric-wieser, line 201: I'd suggest
```suggestion
  aux3 (x : α) : α × DP α → DPS α × DP α × DP α → TimeM ℕ (DPS α × DP α × DP α)
    | (y, xx), (dps, yy, xy) => do
      let dp ← dp x y xx yy xy
      return ((y, dp) :: dps, dp, xx)
  dp (x y : α) (xx yy xy : DP α) : TimeM (DP α) := do
    ✓ if x = y then
      return ⟨x :: xy.lcs, xy.length + 1, xy.length_eq.rec rfl⟩
    else
      return if xx.length ≥ yy.length then xx else yy
```

Cslib/Algorithms/Lean/LongestCommonSublist.lean
eric-wieser: You might want to put the `cs` argument first so that you can write `cs.IsCommonSublist xs ys` too

Cslib/Algorithms/Lean/LongestCommonSublist.lean
```diff
@@ -0,0 +1,354 @@
+module
+
```
Shreyas4991, line 2: you might want to put `@[expose] public section` here. This exposes all your definitions by default.

Cslib/Algorithms/Lean/LongestCommonSublist.lean
Shreyas4991, line 32: Explicit `public` is not required.

Shreyas4991, line 32: Also why not just 
```
abbrev CommonSublist (xs ys cs : List α) : Prop := (cs <+ xs) ∧ (cs <+ ys)
```
?

Cslib/Algorithms/Lean/LongestCommonSublist.lean
```diff
@@ -0,0 +1,354 @@
...
+variable {α} [DecidableEq α]
+
+/-- `cs` is a common sublist of the lists `xs` and `ys`. -/
+public structure CommonSublist (xs ys cs : List α) : Prop where
+  sublist₁ : cs <+ xs
+  sublist₂ : cs <+ ys
+
+/-- `lcs` is a longest common sublist of the lists `xs` and `ys`. -/
+public structure LCS (xs ys lcs : List α) : Prop extends CommonSublist xs ys lcs where
```
Shreyas4991, line 37: Following my suggestion on `CommonSubList`, this could just be 

`abbrev LCS (xs ys lcs : List α) := CommonSubList xs ys lcs ∧ ∀ cs, CommonSublist xs ys cs → cs.length ≤ lcs.length`

Cslib/Algorithms/Lean/LongestCommonSublist.lean
Shreyas4991: Also, I am pretty sure `simp` could handle this case, maybe with some  sublist lemmas. Especially if you follow my `Prop` def suggestion.

Cslib/Algorithms/Lean/LongestCommonSublist.lean
```diff
@@ -0,0 +1,354 @@
...
+                  suffices x = y from absurd this h₁
+                  cons.inj (e.symm.trans h) |>.left
+              calc cs.length
+              _  ≤ xx.length := hxx.longest cs ⟨this, hy⟩
+              _  ≤ yy.length := Nat.le_of_not_ge h₂
+        }
+
+namespace lcsDP
+
+structure DP α where
```
Shreyas4991, line 169: For LCS you should really just take the list length as an input parameter. Even better, solve for a general `l` which denotes a prefix of the list (including the whole list if `l \geq n`) to an auxiliary function, and then pass `List.length` as the  argument to this parameter in the actual function

Cslib/Algorithms/Lean/LongestCommonSublist.lean
```diff
@@ -0,0 +1,354 @@
...
+  lcs : List α
+  length : Nat
+  length_eq : lcs.length = length
+
+scoped instance : Inhabited (DP α) where default := ⟨[], 0, rfl⟩
+example : @Eq (DP α) default ⟨[], 0, rfl⟩ := rfl
+abbrev DPS α := List (α × DP α)
```
Shreyas4991, line 176: why this second definition at all?

Cslib/Algorithms/Lean/LongestCommonSublist.lean
```diff
@@ -0,0 +1,354 @@
...
+    | (y, xx), (dps, yy, xy) => do
+      ✓ let dp := dp x y xx yy xy
+      return ((y, dp) :: dps, dp, xx)
+  dp (x y : α) (xx yy xy : DP α) : DP α :=
+    if x = y then
+      ⟨x :: xy.lcs, xy.length + 1, xy.length_eq.rec rfl⟩
+    else
+      if xx.length ≥ yy.length then xx else yy
+
+section unittest
```
Shreyas4991, line 203: This entire section goes in the `CslibTests` folder. Please don't put tests in the main library folder.

Cslib/Algorithms/Lean/LongestCommonSublist.lean
```diff
@@ -0,0 +1,354 @@
...
+#guard_msgs(info) in
+#eval test_string (⟪lcsDP · ·⟫) "XMJYAUZ" "MZJAWXU"
+
+end unittest
+
+namespace lcsDP
+
+variable {x y : α} {xs ys : List α}
+
+theorem cons_aux0 : ⟪aux0 (x :: xs) ys⟫ = ⟪aux1 x ⟪aux0 xs ys⟫⟫ :=
```
Shreyas4991, line 235: The four lemmas about `aux0` sound like things you will get free as equational lemmas from lean. You should probably just `fun_induction` wherever you use these lemmas. I could be wrong so I'll tag @eric-wieser to check. If these lemmas stay, they probably ought to be `@[simp]`

Shreyas4991:
A more CI-passing and representative title of the PR might be "feat: longest common subsequence in TimeM" or "feat: time complexity of longest common subsequence"

PR #384, opened 2026-03-02
feat: Multi-tape Turing machine

crei:
This introduces a multi-tape Turing machine built in a similar way to the single-tape Turing machine in #269 

I did not define how a function is computed by a multi-tape Turing machine. Instead I would like to focus on tuples of tapes being transformed by Turing machines. The idea is that this allows us to build complex Turing machines from basic building blocks by composition without the need to serialize and de-serialize all the time.

Especially when talking about logarithmic space it will become necessary to define dedicated read-only input and write-only output tapes, but this is not yet included.

I have not yet properly combined the common parts between single- and multi-tape Turing machines and am hoping for some ideas there.


Cslib/Computability/Machines/MultiTapeTuring/Basic.lean
```diff
@@ -0,0 +1,364 @@
...
+public structure MultiTapeTM k Symbol [Inhabited Symbol] [Fintype Symbol] where
+  /-- type of state labels -/
+  (State : Type)
+  /-- finiteness of the state type -/
+  [stateFintype : Fintype State]
+  /-- initial state -/
+  (q₀ : State)
+  /-- transition function, mapping a state and a tuple of head symbols to a `Stmt` to invoke
+  for each tape and optionally the new state to transition to afterwards (`none` for halt) -/
+  (tr : State → (Fin k → Option Symbol) → ((Fin k → (SingleTapeTM.Stmt Symbol)) × Option State))
```
chenson2018: Some of these parentheses are not needed
```suggestion
  State : Type
  /-- finiteness of the state type -/
  [stateFintype : Fintype State]
  /-- initial state -/
  q₀ : State
  /-- transition function, mapping a state and a tuple of head symbols to a `Stmt` to invoke
  for each tape and optionally the new state to transition to afterwards (`none` for halt) -/
  tr : State → (Fin k → Option Symbol) → (Fin k → SingleTapeTM.Stmt Symbol) × Option State
```

Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
```diff
@@ -0,0 +1,364 @@
+/-
+Copyright (c) 2026 Christian Reitwiessner. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Christian Reitwiessner
+-/
+
+module
+
```
chenson2018, line 8: At minimum I'd have a public section so you don't have to write `public` often. I'm not sure if it is intentional that definitions are not exposed.

Cslib/Computability/Machines/MultiTapeTuring/Basic.lean
```diff
@@ -0,0 +1,364 @@
...
+/-- Any number of positive steps run from a halting configuration lead to `none`. -/
+@[simp, scoped grind =]
+public lemma step_iter_none_eq_none (tapes : Fin k → BiTape Symbol) (n : ℕ) :
+    (Option.bind · tm.step)^[n + 1] (some ⟨none, tapes⟩) = none := by
+  rw [Function.iterate_succ_apply]
+  induction n with
+  | zero => simp [step]
+  | succ n ih =>
+    simp only [Function.iterate_succ_apply', ih]
+    simp [step]
```
chenson2018: Nitpick: the terminal `simp` splitting is awkward here. I might have instead written

```suggestion
  induction n with
  | zero => rfl
  | succ => grind [Function.iterate_succ_apply']
```

Cslib/Computability/Machines/MultiTapeTuring/Basic.lean
```diff
@@ -0,0 +1,364 @@
...
+    (tm : MultiTapeTM k Symbol)
+    (tapes tapes' : Fin k → BiTape Symbol)
+    (t : ℕ) : Prop :=
+  RelatesWithinSteps tm.TransitionRelation (tm.initCfgTapes tapes) (tm.haltCfgTapes tapes') t
+
+/-- The Turing machine `tm` transforms tapes `tapes` to `tapes'`. -/
+public def TransformsTapes
+    (tm : MultiTapeTM k Symbol)
+    (tapes tapes' : Fin k → BiTape Symbol) : Prop :=
+  ∃ t, tm.TransformsTapesInExactTime tapes tapes' t
```
chenson2018: Some of these theorem would fit one a single line if not for writing `public`

Cslib/Computability/Machines/MultiTapeTuring/Basic.lean
```diff
@@ -0,0 +1,364 @@
...
+    (cfg₁ cfg₂ : tm.Cfg)
+    (t : ℕ) :
+  RelatesInSteps tm.TransitionRelation cfg₁ cfg₂ t ↔
+    (Option.bind · tm.step)^[t] cfg₁ = .some cfg₂ := by
+  induction t generalizing cfg₁ cfg₂ with
+  | zero => simp
+  | succ t ih =>
+    rw [RelatesInSteps.succ_iff, Function.iterate_succ_apply']
+    constructor
+    · grind only [TransitionRelation, = Option.bind_some]
```
chenson2018: I don't think there's official style guidance yet in Mathlib to follow, but my impression is we generally shouldn't squeeze `grind` unless there's a significant performance problem.

Cslib/Computability/Machines/MultiTapeTuring/Basic.lean
```diff
@@ -0,0 +1,364 @@
...
+      grind [haltsAtStep, Nat.find_spec h_dom]
+    · exact h_get
+  · intro ⟨t, h_iter⟩
+    rw [TransformsTapesInExactTime, relatesInSteps_iff_step_iter_eq_some] at h_iter
+    rw [← configs] at h_iter
+    have h_halts_at_t : tm.haltsAtStep tapes t := by simp [haltsAtStep, h_iter]
+    let h_halts : ∃ t, tm.haltsAtStep tapes t := ⟨t, h_halts_at_t⟩
+    use h_halts
+    have h_eq : Nat.find h_halts = t := halting_step_unique (Nat.find_spec h_halts) h_halts_at_t
+    simp [h_eq, h_iter]
```
chenson2018: You're not obliged to use `grind` as extensively as I am personally inclined to, but consider

```suggestion
  · intro ⟨h_dom, h_get⟩
    use Nat.find h_dom
    grind [TransformsTapesInExactTime, configs, haltCfgTapes, haltsAtStep]
  · intro ⟨t, h_iter⟩
    rw [TransformsTapesInExactTime, relatesInSteps_iff_step_iter_eq_some, ← configs] at h_iter
    have h_halts_at_t : tm.haltsAtStep tapes t := by grind [haltsAtStep]
    have : ∃ t, tm.haltsAtStep tapes t := ⟨t, h_halts_at_t⟩
    grind [haltCfgTapes, halting_step_unique]
```

chenson2018 (changes requested):
A few style and golfing reviews:

crei:
@chenson2018 could you take another look, please? @SamuelSchlesinger has another PR lined up that introduces different kinds of tapes and thus allows to define the function computed by a multi-tape TM, which is quite useful, I think.

crei:
This has been stale for some months now. I changed the code such that all multi-tape Turing machines now have an explicit read-only input tape and an explicit write-only output tape, so that we can properly define space-bounded computations below linear space. I hope this increases the motivation to merge this PR.

The parameter `k` is now the number of work tapes. Due to that, a 0-tape Turing machine should now be equivalent to a finite transducer. I started a proof that regular languages coincide with the languages decided in "zero space", but I got a bit lost in the definitions and maybe an export in labeled transition systems could help here.

Cslib/Computability/Machines/TuringCommon.lean
```diff
@@ -0,0 +1,28 @@
...
+
+namespace Turing
+
+/--
+A Turing machine "statement" is just a `Option`al command to move left or right,
+and write a symbol (i.e. an `Option Symbol`, where `none` is the blank symbol) on the `BiTape`
+-/
+structure Stmt (Symbol : Type) where
+  /-- The symbol to write at the current head position -/
+  symbol : Option Symbol
```
crei: I'm tempted to make this `Option (Option Symbol)` to allow an explicit "do not write", which would make some proofs easier.

Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
```diff
@@ -0,0 +1,414 @@
...
+    time and space bound in the input length
+* `DecidesLanguageInTimeAndSpace`: a proof that a TM decides a language within a certain time
+    and space bound
+
+There are two ways to talk about the behaviour of a multi-tape Turing machine, and they are
+proven to be equivalent.
+
+* `MultiTapeTM.configs`: a sequence of configurations by execution step
+* `RelatesInSteps tm.TransitionRelation cfg cfg' t`: a proof that `tm` transforms the configuration
+    `cfg` into `cfg'` in exactly `t` steps
```
Shreyas4991, line 86: The reference textbooks that you mention need to be cited here in the module doctoring

Cslib/Computability/Machines/MultiTapeTuring/Basic.lean
```diff
@@ -0,0 +1,414 @@
...
+@[ext]
+structure Cfg : Type where
+  /-- the state of the TM (or none for the halting state) -/
+  state : Option tm.State
+  /-- the input -/
+  input : List Symbol
+  /-- the position of the input head, shifted by one -/
+  inputPos : Fin (input.length + 2)
+  /-- the work tape -/
+  workTapes : Fin k → BiTape Symbol
```
Shreyas4991: Are we still using `BiTapes`? I thought we decided in favor of your design to use functions from `\Z` to `\Gamma`

Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
```diff
@@ -0,0 +1,414 @@
...
+  /-- An optional symbol to output. -/
+  outS : Option Symbol
+  /-- The successor state or none to halt. -/
+  q' : Option State
+
+/--
+A multi-tape Turing machine with `k` work tapes over the alphabet of `Option Symbol` (where `none`
+is the blank `BiTape` symbol).
+-/
+structure MultiTapeTM k Symbol [Inhabited Symbol] [Fintype Symbol] where
```
eric-wieser:
```suggestion
structure MultiTapeTM (k Symbol : Type*) [Inhabited Symbol] [Fintype Symbol] where
```

Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
```diff
@@ -0,0 +1,414 @@
...
+  /-- type of state labels -/
+  State : Type
+  /-- finiteness of the state type -/
+  [stateFintype : Fintype State]
```
eric-wieser: Are you sure you want to bundle these?

Cslib/Computability/Machines/MultiTapeTuring/Basic.lean
```diff
@@ -0,0 +1,414 @@
...
+  /-- type of state labels -/
+  State : Type
+  /-- finiteness of the state type -/
+  [stateFintype : Fintype State]
+  /-- initial state -/
+  q₀ : State
+  /-- transition function, mapping a state, the current input symbol and a tuple of head symbols
+  to a movement for the input head, actions on the work tape, optionally a symbol to output and
+  the successor state -/
+  tr : State → (Option Symbol) → (Fin k → Option Symbol) → TransitionOut k Symbol State
```
eric-wieser:
```suggestion
  tr (s : State) (input : Option Symbol) (heads : Fin k → Option Symbol) : TransitionOut k Symbol State
```
etc, makes it much easier to interpret the type

Cslib/Computability/Machines/MultiTapeTuring/Basic.lean
```diff
@@ -0,0 +1,414 @@
...
+
+@[expose] public section
+
+open Cslib Relation
+
+namespace Turing
+
+open BiTape StackTape
+
+variable {Symbol : Type}
```
eric-wieser:
```suggestion
variable {Symbol : Type*}
```

Cslib/Foundations/Data/BiTape.lean
```diff
@@ -114,6 +114,15 @@ lemma moveLeft_moveRight (t : BiTape Symbol) : t.moveLeft.moveRight = t := by
 lemma moveRight_moveLeft (t : BiTape Symbol) : t.moveRight.moveLeft = t := by
   simp [moveLeft, moveRight]
 
+/-- Translate an optional direction into a head movement offset, where the positive
+direction is to the right. -/
+@[scoped grind =]
+def optionDirToInt (d : Option Dir) : ℤ :=
```
eric-wieser: This function appears to be defined twice.

Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
crei: The exact state set is never relevant to "the outside", while the number of tapes and the alphabet can be, although I'm not sure what the downsides might be.

Cslib/Computability/Machines/MultiTapeTuring/Basic.lean
```diff
@@ -0,0 +1,414 @@
...
+  | i => ∑ t' ∈ Finset.range t, (tm.configs cfg t').elim 0 (tm.headMovements · i)
+
+/--
+The number of work tape cells touched by the head of tape `i` in the computation starting from
+configuration `cfg` up to step `t`.
+-/
+def spaceUsedByTape (cfg : tm.Cfg) (t : ℕ) (i : Fin k) : ℕ :=
+  let positions := (Finset.range (t + 1)).image (fun t' => tm.headPositions cfg t' i)
+  have ne := Finset.image_nonempty.mpr ⟨0, by simp⟩
+  (positions.max' ne - positions.min' ne).toNat + 1
```
Shreyas4991: You might want to use Int.natAbs here since you might potentially write backwards on the work tape. `toNat` will turn all negative numbers to `0` which can become a cheat code to use space without accounting for it.

Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
```diff
@@ -0,0 +1,414 @@
...
+    | some q => optionDirToInt
+      (tm.transitionOutput q (tm.inputSymbol cfg) cfg.workTapes |>.stmts i |>.2)
+
+/-- The head positions of the work tapes as a function of the number of steps, relative to
+the starting position in `cfg`. -/
+def headPositions (cfg : tm.Cfg) (t : ℕ) : Fin k → ℤ
+  | i => ∑ t' ∈ Finset.range t, (tm.configs cfg t').elim 0 (tm.headMovements · i)
+
+/--
+The number of work tape cells touched by the head of tape `i` in the computation starting from
```
Shreyas4991, line 243: Isn't this the max - min tape cell written to be the tape head?

Shreyas4991 (commented):
Initial review. Focusses on the definitions for now

Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
crei: No, it's max - min tape cells touched. Note that writing is optional: https://github.com/leanprover/cslib/pull/384/changes#diff-574328057b08dc32a92432c942fe520d03c79c34872fceaef9166ed0337e82b0R168

Cslib/Computability/Machines/MultiTapeTuring/Basic.lean
crei: This is `max' - min'` of the same set, so it should always be non-negative.

Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
crei: This is now derived by the use of `TransitionOut` - should I add it nevertheless?

Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
```diff
@@ -0,0 +1,409 @@
...
+The transition function can optionally output one symbol, which models the write-only output tape.
+Because of these restrictions, we ignore the input and output tapes for space usage of the machine.
+The space usage is defined as the total number of cells the work tape heads visited during
+execution.
+
+Restricting the movement of the input head is not essential, but useful because it allows
+us to easily bound the number of possible configurations of a space-bounded machine. Most textbooks
+have this restriction.
+
+Instead of considering the cells _visited_ by the work tape heads, some textbooks
```
Shreyas4991, line 41: I think this docstring needs to be changed. Concretely, while dspace(log log n) = dspace(1) might be seen as an issue in some problems, I'd like to avoid entirely ruling out the Barak Arora definition entirely (since it is the basis of the standard textbook in complexity theory). Per our discussion in the meeting I recall that we agreed to allow both definitions of space complexity to co-exist. The documentation must be altered to reflect this. I can pass this comment on to complexity experts and get back with their opinions

Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
```diff
@@ -0,0 +1,409 @@
...
+open Cslib Relation
+
+namespace Turing
+
+variable {Symbol : Type*}
+
+variable {k : ℕ}
+
+/-- Possible moves of tape heads. -/
+inductive Move
```
Shreyas4991: I think this is called Dir in other PRs. Let's stay consistent with the nomenclature here.

Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
```diff
@@ -0,0 +1,409 @@
...
+  | left => -1
+  | stay => 0
+  | right => 1
+
+@[scoped grind .]
+lemma Move.toInt_bound (m : Move) : -1 ≤ m.toInt ∧ m.toInt ≤ 1 := by
+  rcases m with _ | d <;> decide
+
+/-- The output of the transition function. -/
+structure TransitionOut (k : ℕ) (Symbol State : Type*) where
```
Shreyas4991, line 105: I am of the opinion that we should use an inductive type here, and separate `move (tape : Fin k) (dir : Dir)`, `read (tape : Fin k)`, `write` etc.

Shreyas4991 (changes requested):
I am sticking to design comments for now.

Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
eric-wieser: Alternatively you could drop it entirely and use `SignType`

Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
eric-wieser: My suggestion was intended just as a style comment; if you are keeping this `strcuture` then yes you should add the `()`s.

Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
```diff
@@ -0,0 +1,409 @@
...
+  /-- The successor state or none to halt. -/
+  q' : Option State
+
+/--
+A multi-tape Turing machine with `k` work tapes over the alphabet of `Option Symbol` (where `none`
+is the blank `BiTape` symbol).
+-/
+structure MultiTapeTM k Symbol [Inhabited Symbol] [Fintype Symbol] where
+  /-- type of state labels -/
+  State : Type*
```
eric-wieser: If your state type is finite, you can use
```suggestion
  State : Type
```
without loss of generality.

If you don't do this, then `MultiTapeTM.{_,_,0}` and `MultiTapeTM.{_,_,1}` are different type families, distinguished not by the state, but the universe in which the state resides.

Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
```diff
@@ -0,0 +1,409 @@
...
+
+/-- This lemma translates between the relational notion and the iterated step notion. The latter
+can be more convenient especially for deterministic machines as we have here. -/
+@[scoped grind =]
+lemma relatesInSteps_iff_configs_eq
+    (tm : MultiTapeTM k Symbol)
+    (cfg₁ cfg₂ : tm.Cfg)
+    (t : ℕ) :
+  RelatesInSteps tm.TransitionRelation cfg₁ cfg₂ t ↔ tm.configs cfg₁ t = cfg₂ := by
+  unfold configs
```
eric-wieser, line 352: style:
```suggestion
    (t : ℕ) :
    RelatesInSteps tm.TransitionRelation cfg₁ cfg₂ t ↔ tm.configs cfg₁ t = cfg₂ := by
  unfold configs
```

references.bib
```diff
@@ -476,3 +476,29 @@ @mastersthesis{Calisto2022
...
+}
+
+
+@book{Papadimitriou94,
+  title={Computational Complexity},
+  author={Papadimitriou, Christos H.},
+  year={1994},
+  publisher={Addison-Wesley},
+  address={Reading, Massachusetts}
+}
```
eric-wieser:
```suggestion
}

```

Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
Shreyas4991: After prolonged discussion, the conclusion is that the Papadmitriou definition is both convenient and in some edge cases like simulating counter machines, yields the correct space complexity.

Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
Shreyas4991: That could also work.

Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
```diff
@@ -0,0 +1,409 @@
...
+  if h₁ : cfg.inputPos = 0 then none
+  else if h₂ : cfg.inputPos = cfg.input.length + 1 then none
+  else cfg.input[cfg.inputPos.val - 1]'(by grind)
+
+@[simp]
+lemma inputSymbolInner {cfg : tm.Cfg} (p : ℕ)
+    (h₁ : cfg.inputPos.val = 1 + p)
+    (h₂ : p < cfg.input.length) :
+    tm.inputSymbol cfg = some cfg.input[p] := by
+  simp [inputSymbol, h₁]
```
Shreyas4991: non-terminal simp.

Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
```diff
@@ -0,0 +1,409 @@
...
+The number of work tape cells touched by a computation starting from configuration
+`cfg` up to step `t`.
+-/
+def spaceUsed (cfg : tm.Cfg) (t : ℕ) : ℕ := ∑ i, tm.spaceUsedByTape cfg t i
+
+/-- A zero-tape Turing machine uses zero space. -/
+@[simp]
+lemma spaceUsed_zero_tapes_eq_zero (cfg : tm.Cfg) (t : ℕ) (h_zero : k = 0) :
+    tm.spaceUsed cfg t = 0 := by
+  unfold spaceUsed
```
Shreyas4991, line 259: I think `simp [spaceUsed, h_zero]` should solve this goal

Shreyas4991 (commented):
Small tactic nitpicks

Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
crei: But then the same is true for `Symbol`, right? (cf. https://github.com/leanprover/cslib/pull/384#discussion_r3453856543)

Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
crei: Unfortunately not, simp does not make use of h_zero.

crei:
I removed the bundling and generally removed the `Fintype` requirements.

Issue #308, opened 2026-01-31
grind sets


chenson2018:
https://github.com/leanprover/lean4/pull/11765 adds user-defined `grind` attributes, analogous to custom [`simp` sets](https://lean-lang.org/doc/reference/latest/The-Simplifier/Simp-sets/#simp-sets). I think this could have several advantages over the current practice of using `scoped grind`. Tying `grind` rules to namespaces can be problematic when they are deeply nested, causing us to make annotations with `attribute` away from their definitions or keep track of overly specific scopes. With grind sets, we could define high level sets of rules (LTS, Automata, etc.) without being concerned about these naming details.

xvade:
Hello, I'm interested in working on this issue. This will be my first contribution to CSLib, so I thought I would ask for some advice/make sure I'm in the right place. I've read [CODE_OF_CONDUCT.md](https://github.com/leanprover/cslib/blob/main/CODE_OF_CONDUCT.md) and [CONTRIBUTING.md](https://github.com/leanprover/cslib/blob/main/CONTRIBUTING.md). I'm planning on familiarizing myself with the general structure of CSLib before I start making changes, but I'm wondering if there are any files I should read in more depth for this issue.

chenson2018:
Hi @xvade, nice to see someone interested in implementing this. I gather from your GitHub you are relatively new to Lean and this is part of a course project? I'm not sure exactly how much experience you have, so I will err on the side of providing detailed context and advice.

A nice thing about this task is that it can be done in pieces without having to migrate everything at once. I specifically had the the example of the `grind` rules in the [automata directory](https://github.com/leanprover/cslib/tree/main/Cslib/Computability/Automata) as something we could try this with first. Along with the things I mention above, a problem here is that [Cslib.LTS](https://api.cslib.io/docs/Cslib/Foundations/Semantics/LTS/Basic.html#Cslib.LTS), a structure we extend for automata, has some rules that are required but seem to cause performance issues when opening the entire scope.

What I was considering was to make a first `grind` set that includes just what we need in this directory. My plan was roughly:

- add a `register_grind_attr automata` in [Cslib.Init](https://github.com/leanprover/cslib/blob/main/Cslib/Init.lean). (This is a file that gets imported everywhere in CSLib)
- in general we would replace `grind` annotations with `automata` annotations, which should remove almost all of the `open scoped`.
[...]

PR #385, opened 2026-03-02
refactor: introduce grind sets for Automata (see issue #308)

xvade:
Adds the `automata` grind attribute and replaces most instances of `scoped grind` in the Automata directory with it. Updates proofs to match.

### DIFF
--- a/lake-manifest.json
+++ b/lake-manifest.json
@@ -5,7 +5,7 @@
    "type": "git",
    "subDir": null,
    "scope": "leanprover-community",
-   "rev": "a237616fe5742014bd87366490691a7c1de9d1a7",
+   "rev": "891dd961017425184bf9499bdcf6c88a68076667",
    "name": "mathlib",
    "manifestFile": "lake-manifest.json",
    "inputRev": "master",


CslibTests/GrindLint.lean
```diff
@@ -40,13 +40,8 @@ open_scoped_all Cslib
 #grind_lint skip Cslib.ωSequence.drop_const
 #grind_lint skip Cslib.ωSequence.get_cons_append_zero
 #grind_lint skip Cslib.ωSequence.map_id
-#grind_lint skip Cslib.Automata.DA.buchi_eq_finAcc_omegaLim
 #grind_lint skip Cslib.LTS.MTr.stepL
 #grind_lint skip Cslib.LTS.STr.trans_τ
-#grind_lint skip Cslib.Automata.DA.FinAcc.toNAFinAcc_language_eq
```
chenson2018, line 46: Having to remove these is a pretty serious tradeoff I wasn't expecting. I have [asked on Zulip](https://leanprover.zulipchat.com/#narrow/channel/270676-lean4/topic/.23grind_lint.20for.20non-default.20annotations/with/576794486) about this, let's see what is said there before I review the rest of this closely.

lake-manifest.json
```diff
@@ -5,7 +5,7 @@
    "type": "git",
    "subDir": null,
    "scope": "leanprover-community",
-   "rev": "9b1001cbdbfe51cbc92398637317609eb7bbf139",
+   "rev": "a237616fe5742014bd87366490691a7c1de9d1a7",
```
chenson2018: Generally PRs should not change the manifest (I assume this was via `lake update`).

Cslib/Init.lean
```diff
@@ -10,6 +10,8 @@ public import Cslib.Foundations.Lint.Basic
 public import Mathlib.Init
 public import Mathlib.Tactic.Common
 
+register_grind_attr automata
```
fmontesi, line 13: This means that:
1. We get to write `grind [automata]`, which I like.
2. We annotate theorems with `@[automata]`, which I'm not sure I like. It doesn't communicate that we're making a grind annotation. Perhaps we should consider a prefix, like `grind_` (or `grind:`, if it's supported)?

chenson2018: I think for this tradeoff I'd be more annoyed by the verbosity of `grind [grind_automata, ...`. In the attributes too, because there would eventually be theorems with multiple of these like `@[grind_lts, grind_automata, ...]`.

fmontesi:
Mmmh true.. albeit I already find it annoying that we'll have to change the
name of the existing lts annotation. :⁠-⁠\

PR #391, opened 2026-03-02
feat: define the category of LTSs

### LINKED ISSUE #387: Define the category of LTSs
As discussed with @chenson2018 on Zulip.

ayberkt:
This PR resolves #387.


Cslib/Foundations/Semantics/LTS/Category.lean
```diff
@@ -0,0 +1,74 @@
...
+
+module
+
+public import Cslib.Foundations.Semantics.LTS.Basic
+public import Mathlib.CategoryTheory.Category.Basic
+open Cslib
+
+@[expose] public section
+
+variable {State Label State₃ Label₃ State₁ State₂ Label₁ Label₂ : Type}
```
fmontesi: I'd order it like this: State Label State₁ State₂ State₃ Label₁ Label₂ Label₃

Cslib/Foundations/Semantics/LTS/LTSCat/Basic.lean
```diff
@@ -0,0 +1,74 @@
...
+public import Mathlib.CategoryTheory.Category.Basic
+open Cslib
+
+@[expose] public section
+
+variable {State Label State₃ Label₃ State₁ State₂ Label₁ Label₂ : Type}
+variable (lts₁ : LTS State₁ Label₁)
+variable (lts₂ : LTS State₂ Label₂)
+variable (lts₃ : LTS State₃ Label₃)
+
```
fmontesi, line 18: From what I see below, you don't need these lts variables that much? Consider removing them maybe?

Cslib/Foundations/Semantics/LTS/Category.lean
fmontesi: `Type` is too restrictive here, you want each type to be in its own universe to state the definitions below more generally. I suggest uᵢ for the States and vᵢ for the labels.

Cslib/Foundations/Semantics/LTS/Category.lean
```diff
@@ -0,0 +1,74 @@
...
+
+/-!
+# Category of Labelled Transition Systems
+-/
+
+/-! ## Definition of LTS morphism -/
+
+/-- A morphism between two labelled transition systems, consisting of a function on states, a
+function on labels, and a proof that transitions are preserved. -/
+structure LTSMorphism (lts₁ : LTS State₁ Label₁) (lts₂ : LTS State₂ Label₂) : Type where
```
fmontesi: The resulting type should live in the max of the universes of the params.

fmontesi: I'd put this under the LTS namespace, so LTS.Morphism. Same for everything else actually, so I'd remove `open Cslib` at the beginning and put `namespace Cslib.LTS` somewhere around before Morphism.

Cslib/Foundations/Semantics/LTS/Category.lean
```diff
@@ -0,0 +1,74 @@
...
+  toFun : State₁ → State₂
```
fmontesi: I'm a bit confused by the names. I suggest: stateMap, labelMap, and labelMap_tr.

Cslib/Foundations/Semantics/LTS/Category.lean
```diff
@@ -0,0 +1,74 @@
...
+  labelMap : Label₁ → Label₂
```
fmontesi: Reading from the reference you provided on Zulip, this function is supposed to be partial. Is it a conscious decision to make it total here?

Cslib/Foundations/Semantics/LTS/LTSCat/Basic.lean
```diff
@@ -0,0 +1,74 @@
...
+/-!
+# Category of Labelled Transition Systems
```
fmontesi: Would be good to add a reference.

fmontesi (changes requested):

fmontesi:
Not urgent, but please fix the title of the PR at some point to start with the required `feat:` prefix, or CI won't let us merge this.

Cslib/Foundations/Semantics/LTS/Category.lean
ayberkt: Good point. I wasn’t sure about this. I’ve definitely seen it defined like this elsewhere, but it might be better to stick with the definitions from the Winskel and Nielsen reference as it is quite standard. I will update the definition.

I am also looking at Daniel Mroz’s [PhD thesis](https://research.manchester.ac.uk/en/studentTheses/a-categorical-setting-for-transition-systems/) and the category of LTSs there is over a fixed label set.

Cslib/Foundations/Semantics/LTS/Category.lean
```diff
@@ -0,0 +1,75 @@
...
+                            → (l : Label₁)
+                            → lts₁.Tr s l s'
+                            → lts₂.Tr (toFun s) (labelMap l) (toFun s')
+
+/-- The identity LTS morphism. -/
+def LTS.Morphism.id (lts : LTS State Label) : LTS.Morphism lts lts :=
+  { toFun                     := _root_.id
+  , labelMap                  := _root_.id
+  , fun_preserves_transitions := fun _ _ _ h => h
+  }
```
chenson2018: A couple of style points:
- you can use `where` with `def` as well for better readability
- please eliminate the extra whitespace here and everywhere else. I personally like it too aesthetically, but we prioritize the uniformity of the Mathlib style.
- For `fun_preserves_transitions` move parameters to the left for clarity

All together this gives:

```suggestion
def LTS.Morphism.id (lts : LTS State Label) : LTS.Morphism lts lts where
  toFun := _root_.id
  labelMap := _root_.id
  fun_preserves_transitions  _ _ _ h := h
```

Cslib/Foundations/Semantics/LTS/Category.lean
```diff
@@ -0,0 +1,75 @@
...
+
+/-!
+# Category of Labelled Transition Systems
+-/
+
+/-! ## Definition of LTS morphism -/
+
+/-- A morphism between two labelled transition systems, consisting of a function on states, a
+function on labels, and a proof that transitions are preserved. -/
+structure LTS.Morphism (lts₁ : LTS State₁ Label₁) (lts₂ : LTS State₂ Label₂) : Type where
```
chenson2018: More significantly this needs to follow suit with Mathlib's concrete categories such as [AddMonCat](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/Category/MonCat/Basic.html#AddMonCat), [AlgCat](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/Category/AlgCat/Basic.html#AlgCat), [GrpCat](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Algebra/Category/Grp/Basic.html#GrpCat), etc. As I describe below, I think you end up formalizing a definition different from what you are intending.

I would expect this structure to follow their naming and to similarly bundle a separately defined type of LTS homomorphism.

Cslib/Foundations/Semantics/LTS/Category.lean
```diff
@@ -0,0 +1,75 @@
...
+  toFun : State₁ → State₂
+  labelMap : Label₁ → Label₂
+  fun_preserves_transitions : (s s' : State₁)
+                            → (l : Label₁)
+                            → lts₁.Tr s l s'
+                            → lts₂.Tr (toFun s) (labelMap l) (toFun s')
```
chenson2018: Style-wise this should go mostly on one line

```suggestion
  fun_preserves_transitions (s s' : State₁) (l : Label₁) :
    lts₁.Tr s l s' → lts₂.Tr (toFun s) (labelMap l) (toFun s')
```

Cslib/Foundations/Semantics/LTS/Category.lean
```diff
@@ -0,0 +1,75 @@
...
+    let r := by intros _ _ _ h
+                apply q
+                apply p
+                exact h
+    ⟨g ∘ f, ν ∘ μ, r⟩
+
+/-! ## LTSs and LTS morphisms form a category -/
+
+/-- `LTS.Morphism` provides a category structure on the `LTS` type. -/
+instance {State Label : Type} : CategoryTheory.CategoryStruct (LTS State Label) where
```
chenson2018: You shouldn't define this separately from the `Category` instance.

Cslib/Foundations/Semantics/LTS/Category.lean
```diff
@@ -0,0 +1,75 @@
...
+  Hom                       := LTS.Morphism
+  id                        := LTS.Morphism.id
+  comp {lts₁} {lts₂} {lts₃} := LTS.Morphism.comp lts₁ lts₂ lts₃
+
+/-- Proof that the above structure actually forms a category. -/
+instance {State Label : Type} : CategoryTheory.Category (LTS State Label) where
```
chenson2018: You'll see in the examples I link above that they would define a fully bundled equivalent
```lean
structure LTSCat where
  State : Type u
  Label : Type v
  str : LTS State Label
```
Otherwise, we aren't talking about *the* category of LTSs where morphisms are between LTS with different states/labels, because a polymorphic instance like this is defining a category for *each* state/label type.

chenson2018 (changes requested):

Cslib/Foundations/Semantics/LTS/Category.lean
chenson2018: It is common in category theory formalization to have some encoding of partial functions though dependent types, so I wasn't particularly surprised by this definition. Please see my reviews below however, I believe these definitions have a more fundamental problem.

Cslib/Foundations/Semantics/LTS/Category.lean
chenson2018: It might be best also to go ahead and have the module structure match up as `Cslib/Foundations/Semantics/LTS/LTSCat/Basic.lean` since it looks like you have immediate plans to extend this further.

Cslib/Foundations/Semantics/LTS/Category.lean
```diff
@@ -0,0 +1,84 @@
...
+
+/-- The notion of labelled transition system -/
+structure BundledLTS where
+  State : Type u
+  Label : Type v
+  lts : LTS State Label
+
+/- Remark: I do not like the name 'bundled LTS'; and LTS is already the bundled notion. The name
+   `LTS` for the transition relation on a fixed set of states and labels is what is confusing here.
+    I propose to change that to `LTS-Structure` and call the above `LTS`.
```
ctchou: I hope not. `LTS` is used in many, many places already. We should keep its short name.

chenson2018: Rest assured we will not change this name. As I say above, in alignment with Mathlib it should be `LTSCat`. (And will not affect any of your existing work)

chenson2018: Also
> LTS is already the bundled notion

makes me think we have a mismatch in terminology that might be a Lean colloquialism. Could you clarify what you mean by this @ayberkt?

Cslib/Foundations/Semantics/LTS/Category.lean
fmontesi: Trying to find a pattern:

We could also call this `LTS.Bundled`, I guess? I do that in HML for `Satisfies` and judgements.
Then it could live in `Cslib/Foundations/Semantics/LTS/Bundled/Basic.lean`.

Cslib/Foundations/Semantics/LTS/LTSCat/Basic.lean
```diff
@@ -0,0 +1,109 @@
...
+
+* [N. Winskel and M. Nielsen, *Models for concurrency*][WinskelNielsen1995]
+-/
+
+/--
+We first define what is denoted Tran* in [WinskelNielsen1995]: the extension of
+a transition relation with idle transitions.
+-/
+def lift (trans : State → Label → State → Prop) :
+  State → (Option Label) → State → Prop :=
```
chenson2018: This signature fits on one line now.

Cslib/Foundations/Semantics/LTS/LTSCat/Basic.lean
```diff
@@ -0,0 +1,109 @@
...
+      have hp := p s s' l h
+      change ((μ l).bind ν).elim (g (f s) = g (f s')) _
+      cases hμ : μ l with
+      | none =>
+        rw [hμ] at hp
+        exact congrArg g hp
+      | some m =>
+        rw [hμ] at hp
+        exact q (f s) (f s') m hp
+    ⟨g ∘ f, μ >=> ν, r⟩
```
chenson2018: I think it is more idiomatic/clear to use `where` to split this up. I would write:

```suggestion
def LTS.Morphism.comp {lts₁ lts₂ lts₃} (f : LTS.Morphism lts₁ lts₂) (g : LTS.Morphism lts₂ lts₃) :
    LTS.Morphism lts₁ lts₃ where
  stateMap := g.stateMap ∘ f.stateMap
  labelMap := f.labelMap >=> g.labelMap
  labelMap_tr s s' l h := by
    obtain ⟨f, μ, p⟩ := f
    obtain ⟨g, ν, q⟩ := g
    change ((μ l).bind ν).elim (g (f s) = g (f s')) _
    cases hμ : μ l with grind [lift]
```

Cslib/Foundations/Semantics/LTS/LTSCat/Basic.lean
```diff
@@ -0,0 +1,109 @@
...
+
+/-- Finally, we prove that these form a category. -/
+instance : CategoryTheory.Category LTSCat where
+  Hom lts₁ lts₂ := LTS.Morphism lts₁ lts₂
+  id lts := LTS.Morphism.id lts
+  comp {lts₁} {lts₂} {lts₃} f g := @LTS.Morphism.comp lts₁ lts₂ lts₃ f g
```
chenson2018: Easier to read if you eta reduce

```suggestion
  Hom := LTS.Morphism
  id := LTS.Morphism.id
  comp := LTS.Morphism.comp
```

Cslib/Foundations/Semantics/LTS/LTSCat/Basic.lean
```diff
@@ -0,0 +1,109 @@
...
+  id_comp := by
+    intro _ _ f
+    cases f
+    simp only [LTS.Morphism.comp, LTS.Morphism.id, Function.comp_id]
+    congr 1
```
chenson2018: This direction can be inferred

```suggestion
```

Cslib/Foundations/Semantics/LTS/LTSCat/Basic.lean
```diff
@@ -0,0 +1,109 @@
+/-
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Ayberk Tosun
+-/
+
+module
+
+public import Mathlib.CategoryTheory.Category.Basic
+public import Cslib.Foundations.Semantics.LTS.Basic
+open Cslib
```
chenson2018: All modules should be in at least the `Cslib` namepsace, so this should move down below the `@[expose] public section` and become `namespace Cslib`. (As it is now below you are making `LTS.Morphism` etc. in a fresh namespace at the top level)

chenson2018 (changes requested):
This looks much better! The theory now all lines up, there are some additional small style/technical issues I have pointed out below. Please also see the failures in CI now that it has made it to the linting step.

Cslib/Foundations/Semantics/LTS/LTSCat/Basic.lean
```diff
@@ -0,0 +1,100 @@
...
+## References
+
+* [N. Winskel and M. Nielsen, *Models for concurrency*][WinskelNielsen1995]
+-/
+
+/--
+We first define what is denoted Tran* in [WinskelNielsen1995]: the extension of
+a transition relation with idle transitions.
+-/
+def lift (trans : State → Label → State → Prop) : State → (Option Label) → State → Prop :=
```
chenson2018: This is currently named `Cslib.lift`, which still isn't a great name/location combo. I'd say `LTS.lift`, except this is unbundled which makes that an odd name. Deferring to @fmontesi on this.

chenson2018 (approved):
Please see my naming comment above, but otherwise things look in pretty good shape to me.

(I pushed a few small commits, the most important of which is probably using some API from `Mathlib.Control` instead of breaking down notation with `change`)

Cslib/Foundations/Semantics/LTS/LTSCat/Basic.lean
fmontesi: Yes, we should namespace it under LTS. But `lift` looks too generic a name to me, this is doing something very specific, i.e., adding so-called 'idle' self-transitions. I'd also suggest to make it transform an LTS, so to enable stating properties about it without relying on knowledge of how LTS is defined.

How about `LTS.withIdle` or `LTS.withSelfLoops` or something like that?

In general, it would also make sense to move such operations on unbundled LTSs to other files, outside of category theory -- but we don't necessarily need to do this now. I see some operations in Winskel and Nielsen (like relabelling) that make sense to offer even to developments not interested in category theory.

chenson2018: `withIdle` seems fine, but it is maybe a bit odd for this to be in the `LTS` namespace when it doesn't actually include an `LTS` in the signature.

fmontesi: That's why I wrote

> I'd also suggest to make it transform an LTS

:-)

What I meant is to do:

`def LTS.withIdle (lts : LTS State Label) : LTS State (Option Label) := ...`

fmontesi:
There's a conflict. Also, please see the `withIdle` comment. Then I think we're good to go, thanks!

ayberkt:
> In general, it would also make sense to move such operations on unbundled LTSs to other files, outside of category theory -- but we don't necessarily need to do this now.

Okay, I’ll do this in the future as I continue to work on this.

> Also, please see the `withIdle` comment.

I renamed `lift` to `withIdle`.

I think this should be ready to merge now.

fmontesi (approved):
Thanks, LGTM now!

fmontesi:
Some check is failing. @ayberkt I think you need to run `lake exe mk_all --module`

ayberkt:
@fmontesi Done.

PR #417, opened 2026-03-11
refactor: change the simpNF for FreeM

eric-wieser:
This declares `liftBind` an implementation detail, encouraging users to work with `lift` and `bind` separately instead.

In particular:
* `>>=` is now the simp-normal form of `FreeM.bind`
* `<$>` is now the simp-normal form of `FreeM.map`
* `Pure.pure` is now the simp-normal form of `FreeM.pure`

Note that due to a failing in the design of monads in Lean around universes, it is not always possible to use the notation.
The advantage of using it when it _is_ available is that standard lemmas about lawful monads can apply.

---

Depends on:
* #439
* #490
* #525


Cslib/Foundations/Control/Monad/Free/Effects.lean
```diff
@@ -200,24 +213,49 @@ def run [Monoid ω] : FreeWriter ω α → α × ω
...
+@[simp]
+theorem _root_.WriterT.run_bind [Monoid ω] [Monad M] (x : WriterT ω M α) (f : α → WriterT ω M β) :
+    WriterT.run (x >>= f) = x.run >>= fun (a, w₁) => (fun (b, w₂) => (b, w₁ * w₂)) <$> (f a).run :=
+  rfl
+
+@[simp]
+theorem _root_.WriterT.run_tell [Monad M] (w : ω) :
+    WriterT.run (MonadWriter.tell w : WriterT ω M PUnit) = pure (.unit, w) := rfl
+
+end missing_from_mathlib
```
eric-wieser: https://github.com/leanprover-community/mathlib4/pull/36497

Cslib/Foundations/Control/Monad/Free/Effects.lean
```diff
@@ -286,6 +286,25 @@ instance [Monoid ω] : MonadWriter ω (FreeWriter ω) where
...
+
+@[simp]
+theorem _root_.ContT.run_pure (a : α) :
+    ContT.run (pure a : ContT r m α) k = k a := rfl
+
+@[simp]
+theorem _root_.ContT.run_bind (x : ContT r m α) (f : α → ContT r m β) :
+    ContT.run (x >>= f) k = ContT.run x fun x => (f x).run k := rfl
+
+end ForMathlib
```
eric-wieser: The linter rejects this, but I'm not entirely sure why.

Getting https://github.com/leanprover-community/mathlib4/pull/38379/ merged is another resolution.

chenson2018: This [Cslib.Lint.topNamespace](https://api.cslib.io/docs/Cslib/Foundations/Lint/Basic.html#Cslib.Lint.topNamespace) linter might be incorrect following the module system, please feel free to disable.

eric-wieser: Disabled.

eric-wieser: #522 removes the need for this.

chenson2018 (approved):
Thanks!

PR #439, opened 2026-03-18
fix: missing `pure`s in the `Id` monad

eric-wieser:
This fixes some statements that were type incorrect; their LHS was `Id X` and their RHS `X`. The latter can be cast to the former with `pure`.


eric-wieser:
@chenson2018, mind taking a look?

chenson2018 (approved):
This looks good to me, thanks for the patience and pinging me.

PR #446, opened 2026-03-19
feat(Algorithms): Prove insertion sort is stable

Arleee1:
Prove the insertionsort implementation is stable.
Aristotle AI was used for portions of this work.

Depends on #372. My contributions are only in my commits.


Cslib/AlgorithmsTheory/Algorithms/ListInsertionSort.lean
```diff
@@ -0,0 +1,158 @@
+/-
+Copyright (c) 2026 Shreyas Srinivas. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Shreyas Srinivas, Eric Wieser
```
Shreyas4991: you should append your name here to the authors list. You are adding new content to this file.

Cslib/AlgorithmsTheory/Models/ListComparisonSort.lean
```diff
@@ -145,6 +147,18 @@ def sortModelNat {α : Type*}
 
 end NatModel
 
+section SortStability
+
+/-- Definition of a stable sorting algorithm. -/
```
Shreyas4991: Definition of a stable *list* sorting algorithm. This definition uses List functions and is of course specific to lists. We should revisit the location of this definition at some point. This is a property of lists independent of Progs on the comparison query model.

Arleee1: Thank you for the review! I will update to specify that it is for lists. As for the location, I think it could make sense to define classes for list sorting algorithms in a separate file and another class that extends it for stable algorithms, what do you think of that idea?

Shreyas4991: That's a bigger change. I would leave it for a later PR. I just think a TODO comment would suffice. It should say that these definitions should be upstreamed or relocated

Arleee1:
Updated based on @Shreyas4991 's PR feedback, also golfed and cleaned up

Made similar updates in the merge sort stable PR #442

PR #458, opened 2026-03-25
feat(LambdaCalculus): basics of named representation (alpha equivalence and capture-avoiding substitution)

yinhaoxuan:
Named representation of untyped lambda calculus.

# Basic.lean
First, we define a naive variable renaming.
Second, we define alpha-equivalence according to "Gabbay and Pitts, A New Approach to Abstract Syntax with Variable Binding, 2002".
Third, we define capture-avoiding substitution.

# Properties.lean
AlphaEquiv.refl: reflexivity of alpha-equivalence
AlphaEquiv.symm: symmetry of alpha-equivalence
AlphaEquiv.trans: transitivity of alpha-equivalence
Subst.relation_iff_function: the relational and functional definition of capture-avoiding substitution are equivalent, modulo alpha-equivalence
subst.commutativity: commutativity of substitution, more commonly known as "substitution lemma (e.g. in Barendregt 1984)"


Cslib/Languages/LambdaCalculus/Named/Untyped/Basic.lean
```diff
@@ -91,25 +112,21 @@ def Term.subst [DecidableEq Var] [HasFresh Var] (m : Term Var) (x : Var) (r : Te
...
--- TODO
--- theorem Term.subst_comm
---   [DecidableEq Var] [HasFresh Var]
---   {m : Term Var} {x : Var} {n1 : Term Var} {y : Var} {n2 : Term Var} :
---   (m[x := n1])[y := n2] = (m[y := n2])[x := n1] := by
---   induction m
---   -- case var z =>
---   sorry
+/-- Allow grind to recognise the notation of substitution. -/
+@[grind ←]
```
chenson2018: It seems to generate the same `grind_pattern`, but I think it's more common to use a `grind =`  here and probably swap the sides of the equality.

Cslib/Languages/LambdaCalculus/Named/Untyped/Properties.lean
```diff
@@ -0,0 +1,658 @@
...
+Copyright (c) 2026 Haoxuan Yin. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Haoxuan Yin, Fabrizio Montesi
+-/
+
+module
+
+public import Cslib.Languages.LambdaCalculus.Named.Untyped.Basic
+
+public section
```
chenson2018, line 28: There is a lint here about a missing module docstring on merging main. (There was previously a bug in the Mathlib linter that made it not ever run that has since been fixed)

Cslib/Languages/LambdaCalculus/Named/Untyped/Properties.lean
```diff
@@ -0,0 +1,658 @@
...
+universe u
+
+variable {Var : Type u} [DecidableEq Var]
+
+namespace LambdaCalculus.Named.Untyped.Term
+
+/-- A variable in a term is either free or bound. -/
+theorem vars_either_fv_or_bv {m : Term Var} :
+    m.vars = m.fv ∪ m.bv := by
+  induction m <;> grind [fv, bv, vars]
```
chenson2018: Some theorem signatures like this fit on a single line

```suggestion
theorem vars_either_fv_or_bv {m : Term Var} : m.vars = m.fv ∪ m.bv := by
```

Cslib/Languages/LambdaCalculus/Named/Untyped/Properties.lean
```diff
@@ -0,0 +1,658 @@
...
+/-- Renaming a variable to itself has no effect. -/
+@[simp]
+theorem rename_same {m : Term Var} {x : Var} :
+  m.rename x x = m := by
+  induction m <;> grind [vars, rename]
+
+/-- Renaming a used variable changes the set of variables. -/
+@[simp]
+theorem rename_vars_used {m : Term Var} {x y : Var} :
+  x ∈ m.vars → (m.rename x y).vars = m.vars.erase x ∪ {y} := by
```
chenson2018: When the signature doesn't fit on one line, this gets an extra indentation:

```suggestion
    x ∈ m.vars → (m.rename x y).vars = m.vars.erase x ∪ {y} := by
```

These rules are all from the [Mathlib style guide](https://leanprover-community.github.io/contribute/style.html#structuring-definitions-and-theorems) that we try our best to follow, and soon there will be an autoformatter from the FRO that does this automatically.

Cslib/Languages/LambdaCalculus/Named/Untyped/Properties.lean
```diff
@@ -0,0 +1,658 @@
...
+
+/-- The set of variables after renaming. -/
+theorem rename_vars {m : Term Var} {x y : Var} :
+  (m.rename x y).vars = m.vars \ {x} ∪ (if x ∈ m.vars then {y} else ∅) := by
+  by_cases x ∈ m.vars <;> grind [vars, rename, rename_unused, rename_vars_used]
+
+/-- The set of free variables after renaming. -/
+theorem rename_fv {m : Term Var} {x y : Var} :
+  y ∉ m.vars → (m.rename x y).fv = m.fv \ {x} ∪ (if x ∈ m.fv then {y} else ∅) := by
+  induction m <;> grind [fv, vars, rename, vars_either_fv_or_bv]
```
chenson2018: Something we check for is that any usage of `grind` can be debugged with a successful `grind?`. This one fails for example. You can identify this by setting `set_option linter.tacticAnalysis.verifyGrindOnly true` in a file. In these case we ask that you make the proof a little more manual for maintenance reasons.

(This can happen when doing too much unfolding with `grind` that causes performance issues)

Cslib/Languages/LambdaCalculus/Named/Untyped/Basic.lean
```diff
@@ -37,52 +39,71 @@ inductive Term (Var : Type u) : Type u where
...
+    AlphaEquiv (m1.rename x1 y) (m2.rename x2 y) → AlphaEquiv (abs x1 m1) (abs x2 m2)
+  | app {m1 n1 m2 n2} : AlphaEquiv m1 n1 → AlphaEquiv m2 n2 → AlphaEquiv (app m1 m2) (app n1 n2)
+
+/-- Instance for the notation `m =α n`. -/
+instance instHasAlphaEquivTerm : HasAlphaEquiv (Term Var) where
+  AlphaEquiv := AlphaEquiv
+
+omit [HasFresh Var] in
+/-- Allow grind to recognise the notation of α-equivalence. -/
+@[grind ←]
```
chenson2018: We also try to use `scoped grind` whenever possible to keep annotations within an isolated namespace (for performance reasons).

Cslib/Languages/LambdaCalculus/Named/Untyped/Properties.lean
```diff
@@ -0,0 +1,658 @@
...
+  induction m <;> grind [vars, rename]
+
+/-- Commutativity of renaming, simpler version. -/
+theorem rename_comm {m : Term Var} {x y z w : Var} :
+  x ≠ z → y ∉ m.vars ∪ {x, z} → w ∉ m.vars ∪ {x, z} →
+  (m.rename x y).rename z w = (m.rename z w).rename x y := by
+  induction m <;> grind [vars, rename]
+
+/-- Commutativity of renaming, more general version. -/
+theorem rename_comm2 {m : Term Var} {x y z w : Var} :
```
chenson2018: If this is more general it should go first and then specialize into the more specific version. We are sometimes more loose about naming conventions than Mathlib, but we so try to avoid naming with numbers like this.

Cslib/Languages/LambdaCalculus/Named/Untyped/Properties.lean
```diff
@@ -0,0 +1,658 @@
...
+    have h2 : (m2.rename x2 y).fv \ {y} = m2.fv \ {x2} := by
+      grind [rename_fv, vars_either_fv_or_bv]
+    grind
+  | @app m1 n1 m2 n2 h1 h2 ih1 ih2 => grind [Term.fv]
+
+variable [HasFresh Var]
+
+/-- Reflexivity of α-equivalence. -/
+theorem AlphaEquiv.refl (m : Term Var) : m =α m := by
+  refine WellFounded.induction (C := fun m => m =α m) sizeOfWFRel.wf m ?_
```
chenson2018: This needs to be refactored into an induction principle rather than using this low level interface.

Cslib/Languages/LambdaCalculus/Named/Untyped/Properties.lean
```diff
@@ -0,0 +1,658 @@
...
+theorem rename_comm2 {m : Term Var} {x y z w : Var} :
+    y ∉ m.vars ∪ {x, z} → w ∉ m.vars ∪ {x, y, z} →
+    (m.rename x y).rename (if z = x then y else z) w = (m.rename z w).rename x y := by
+  intro hy hw
+  by_cases hzx : z = x
+  · grind [rename_same, rename_unused, rename_concat, rename_vars]
+  · grind [rename_comm]
+
+omit [DecidableEq Var] in
+@[grind norm↓← ]
```
chenson2018: I'm not very comfortable with the use of `grind norm` annotations, but I also suspect more generally that this shouldn't be exposed as a `grind` annotation at all.

Cslib/Languages/LambdaCalculus/Named/Untyped/Properties.lean
```diff
@@ -0,0 +1,658 @@
...
+  induction h with
+  | @var x => rfl
+  | @abs y x1 x2 m1 m2 hy h ih =>
+    rw [Term.fv, Term.fv]
+    have h1 : m1.fv \ {x1} = (m1.rename x1 y).fv \ {y} := by
+      grind [rename_fv, vars_either_fv_or_bv]
+    have h2 : (m2.rename x2 y).fv \ {y} = m2.fv \ {x2} := by
+      grind [rename_fv, vars_either_fv_or_bv]
+    grind
+  | @app m1 n1 m2 n2 h1 h2 ih1 ih2 => grind [Term.fv]
```
chenson2018: When you have several successive calls to grind like this, for readability and performance benefit you can use the interactive mode:

```suggestion
  | app => grind [fv]
  | @abs y x1 x2 m1 m2 hy h ih =>
    grind =>
      instantiate [fv, rename_fv, vars_either_fv_or_bv]
      have : m1.fv \ {x1} = (m1.rename x1 y).fv \ {y} 
      have : (m2.rename x2 y).fv \ {y} = m2.fv \ {x2}
```

(I had to swap the branches because of what I think is a core parsing bug)

Cslib/Languages/LambdaCalculus/Named/Untyped/Properties.lean
```diff
@@ -0,0 +1,658 @@
...
+variable [HasFresh Var]
+
+/-- Reflexivity of α-equivalence. -/
+theorem AlphaEquiv.refl (m : Term Var) : m =α m := by
+  refine WellFounded.induction (C := fun m => m =α m) sizeOfWFRel.wf m ?_
+  simp only; intro m ih
+  cases m with
+  | var x => apply AlphaEquiv.var
+  | abs x m =>
+    obtain ⟨z, hz⟩ := HasFresh.fresh_exists (m.vars ∪ {x})
```
chenson2018: When using `HasFresh` to select variables, the `free_union` elaborator (here are [the docs](https://api.cslib.io/docs/Cslib/Foundations/Data/HasFresh.html#Cslib.freeUnion)) is usually used so you don't have to specify every single variable manually:

```suggestion
    obtain ⟨z, hz⟩ := fresh_exists <| free_union [vars] Var
```

Cslib/Languages/LambdaCalculus/Named/Untyped/Properties.lean
```diff
@@ -0,0 +1,658 @@
...
+-/
+theorem Subst.function_to_relation {m r : Term Var} {x : Var} :
+    m.Subst x r (m[x := r]) := by
+  refine WellFounded.induction (C := fun m => m.Subst x r (m[x := r])) sizeOfWFRel.wf m ?_
+  simp only; intro m ih
+  cases m with
+  | var y =>
+    by_cases hyx : y = x
+    · subst y
+      simp only [← subst_def, subst.eq_1, ↓reduceIte]
```
chenson2018: There's some missing annotations here, I feel like you should be able to just write `grind [Subst.varMiss]`.

Cslib/Languages/LambdaCalculus/Named/Untyped/Properties.lean
```diff
@@ -0,0 +1,658 @@
...
+  | abs y m =>
+    by_cases hyx : y = x
+    · subst y
+      simp only [← subst_def, subst.eq_2, ↓reduceIte]
+      apply Subst.absShadow
+    · simp only [← subst_def, subst.eq_2, hyx, ↓reduceIte, Finset.union_insert]
+      by_cases hyr : y ∈ r.fv
+      · simp only [hyr]
+        have hz := fresh_notMem (insert x (insert y (m.vars ∪ r.vars)))
+        set z := fresh (insert x (insert y (m.vars ∪ r.vars)))
```
chenson2018: I'm not understanding why one of the `grind` proofs seems very fragile here, do you have any idea?

At minimum this works

```suggestion
        let s := {x, y} ∪ m.vars ∪ r.vars
        have hz := fresh_notMem s
        let z := fresh s
```
as being more readable, but I'd really like to understand why using `fresh_exists` or `free_union` makes this fail.

Cslib/Languages/LambdaCalculus/Named/Untyped/Properties.lean
```diff
@@ -0,0 +1,658 @@
...
+/-- Substituting a non-free variable has no effect. -/
+theorem subst.non_free {m r : Term Var} {x : Var} :
+    x ∉ m.fv → (m[x := r]) =α m := by
+  refine WellFounded.induction (C := fun m => x ∉ m.fv → (m[x := r]) =α m) sizeOfWFRel.wf m ?_
+  simp only; intro m ih hx
+  cases m with
+  | var y =>
+    have hyx : y ≠ x := by
+      grind [fv]
+    simp only [← subst_def, subst.eq_1, hyx, ↓reduceIte]
```
chenson2018: Again, `((var y)[x:=r])` is somehow missing from what grind

Cslib/Languages/LambdaCalculus/Named/Untyped/Properties.lean
```diff
@@ -0,0 +1,658 @@
...
+          apply ih <;> grind [fv, rename_fv, rename_eq_sizeOf]
+        · grind [rename_concat, AlphaEquiv.refl]
+      · simp only [← subst_def, subst.eq_2, hyx, ↓reduceIte, hyr, not_false_eq_true]
+        apply AlphaEquiv.context (c := Context.abs y Context.hole)
+        apply ih <;> grind [fv]
+  | app m1 m2 =>
+    simp only [← subst_def, subst.eq_3]
+    apply AlphaEquiv.app <;> apply ih <;> grind [fv]
+
+lemma subst.abs_fresh_helper {m r : Term Var} {x y z : Var} :
```
chenson2018: Maybe this should be `private`? I understand there is inherent complexity here, but I wonder if there's any meaningful way to split this up.

CslibTests/LambdaCalculus.lean
```diff
@@ -1,42 +1,49 @@
...
+  simp only [vars]
+  decide
 
--- section 5.3.4 of TAPL
-example : (abs y (app x y))[x := (app y z : Term ℕ)] = (abs w (app (app y z) w)) := by
-  simp +instances [subst, fv, bv, vars, rename, instHasFreshNat, HasFresh.ofSucc,
-    instHasSubstitutionTerm]
+example : (abs y (var x)).subst x (app (var y) (var z)) = (abs w (app (var y) (var z))) := by
+  simp only [rename, subst, fv, vars]
+  decide
```
chenson2018, line 36: Very nice that `decide` works here!

chenson2018 (changes requested):
I think the biggest things here are
- extracting out a well founded induction principle
- fixing the places where you have long `simp only [...]; grind`. To me these are mostly an indication of missing annotations
- some formatting style that I didn't comprehensively review and clutter up the PR with. It's mostly good, so if you could just take one more pass at thigns like indentation that would be appreciated.
- seeing if there's any way to split up some of these very large proofs, but maybe this is inherently difficult

It's also generally a bit hard to make sure there aren't errors with the substitution definition until this has more usage. I've had the thought that it would be nice to have a larger set of tests to safeguard around any incorrect definitions, based on something like https://github.com/sweirich/lambda-n-ways/. (Not asking you to do this, just a thought for a future PR)

Cslib/Languages/LambdaCalculus/Named/Untyped/Properties.lean
yinhaoxuan: This is because the definition of subst (the function, not the relation) uses the canonical fresh variable, i.e. the variable obtained from the operation "fresh {x, y} ∪ m.vars ∪ r.vars".

Cslib/Languages/LambdaCalculus/Named/Untyped/Properties.lean
yinhaoxuan: I don't think there's meaningful way to split this lemma up, as the two parts of this lemma depend on each other. But with more aggressive grinding I have been able to shorten the proof from 132 lines to 83 lines.

Cslib/Languages/LambdaCalculus/Named/Untyped/Properties.lean
yinhaoxuan: After we have proved the theorem subst.abs_fresh, we can then use any free variable in such scenarios.

chenson2018 (approved):
I think there's some subsequent cleaning and proving equivalent to other bindings that can happen here, but since Fabrizio has had a look I'm inclined to let this happen in further PRs.

PR #464, opened 2026-04-01
feat(Cryptography): formalise perfect secrecy and the one-time pad

SamuelSchlesinger:
Adds `Cslib.Cryptography.PerfectSecrecy` with information-theoretic private-key encryption schemes and perfect secrecy following Katz-Lindell, Chapter 2:

- `EncScheme`: private-key encryption (Definition 2.1)
- `PerfectlySecret`: perfect secrecy (Definition 2.3)
- `perfectlySecret_iff_ciphertextIndist`: ciphertext indistinguishability characterization (Lemma 2.5)
- `otp`: the one-time pad construction (Construction 2.9)
- `otp_perfectlySecret`: the OTP is perfectly secret (Theorem 2.10)
- `perfectlySecret_keySpace_ge`: Shannon's theorem, |K| ≥ |M| (Theorem 2.12)

For some context, Katz reviewed this in its original home: https://github.com/SamuelSchlesinger/introduction-to-modern-cryptography/pull/1.


SamuelSchlesinger:
I'd be quite interested in helping with future review in this `Cryptography` folder, if there is interest in including it.

Cslib/Cryptography/PerfectSecrecy/Internal/OneTimePad.lean
```diff
@@ -0,0 +1,36 @@
...
+/-!
+# One-Time Pad: Internal proofs
+
+The OTP ciphertext distribution is uniform regardless of message.
+-/
+
+namespace Cslib.Cryptography.PerfectSecrecy.OTP
+
+/-- The ciphertext distribution of the OTP is uniform, regardless of the message. -/
+theorem otp_ciphertextDist_eq_uniform (l : ℕ) (m : Fin l → Bool) :
```
wrenna-robson: As mention on Zulip BitVec might be better here - what is the "standard" cslib spelling of a bitvector?

Cslib/Cryptography/PerfectSecrecy/Internal/OneTimePad.lean
```diff
@@ -0,0 +1,36 @@
...
+    (PMF.uniformOfFintype (Fin l → Bool)).bind
+      (fun k => PMF.pure (fun i => k i ^^ m i)) =
+    PMF.uniformOfFintype (Fin l → Bool) := by
+  ext c
+  simp only [PMF.bind_apply, PMF.uniformOfFintype_apply, PMF.pure_apply]
+  have heq : ∀ a : Fin l → Bool,
```
wrenna-robson: This should probably be a lemma of some kind (might be easier in BitVec, not sure).

Cslib/Crypto/Protocols/PerfectSecrecy/Encryption.lean
```diff
@@ -0,0 +1,48 @@
...
+
+namespace Cslib.Cryptography.PerfectSecrecy
+
+/--
+A private-key encryption scheme over message space `M`, key space `K`,
+and ciphertext space `C` ([KatzLindell2021], Definition 2.1).
+-/
+structure EncScheme.{u} (M K C : Type u) where
+  /-- Probabilistic key generation. -/
+  gen : PMF K
```
wrenna-robson: We have briefly spoken about this, but it is worth considering whether rational probabilities might be desirable. The concern would be that sometimes you do want to be able to take limits of probabilities etc.

Cslib/Crypto/Protocols/PerfectSecrecy/Encryption.lean
```diff
@@ -0,0 +1,48 @@
...
+  /-- (Possibly randomized) encryption. -/
+  enc : K → M → PMF C
+  /-- Deterministic decryption. -/
+  dec : K → C → M
+  /-- Decryption inverts encryption for all keys in the support of `gen`. -/
+  correct : ∀ k, k ∈ gen.support → ∀ m c, c ∈ (enc k m).support → dec k c = m
```
wrenna-robson: This looks like the right statement of correctness to me. I wonder if "gen.support" (the set of valid keys) and (enc k m),support (the set of valid encryptions for a give k) want names/definitions (or at least abbreviations).

Cslib/Cryptography/PerfectSecrecy/PerfectSecrecy.lean
```diff
@@ -0,0 +1,51 @@
...
+namespace Cslib.Cryptography.PerfectSecrecy.EncScheme
+
+universe u
+variable {M K C : Type u}
+
+/-- A scheme is perfectly secret iff the ciphertext distribution is
+independent of the plaintext ([KatzLindell2021], Lemma 2.5). -/
+theorem perfectlySecret_iff_ciphertextIndist (scheme : EncScheme M K C) :
+    scheme.PerfectlySecret ↔
+    ∀ m₀ m₁ : M, scheme.ciphertextDist m₀ = scheme.ciphertextDist m₁ :=
```
wrenna-robson: This seems sensible to me and must be a pattern that occurs elsewhere.

Cslib/Crypto/Protocols/PerfectSecrecy/Defs.lean
```diff
@@ -0,0 +1,68 @@
...
+    (msgDist : PMF M) (c : C) (m : M) : ENNReal :=
+  scheme.jointDist msgDist (m, c) / scheme.marginalCiphertextDist msgDist c
+
+/-- An encryption scheme is perfectly secret if `Pr[M = m | C = c] = Pr[M = m]`
+for every prior, message, and ciphertext with positive probability
+([KatzLindell2021], Definition 2.3). -/
+def PerfectlySecret (scheme : EncScheme M K C) : Prop :=
+  ∀ (msgDist : PMF M) (m : M) (c : C),
+    c ∈ (scheme.marginalCiphertextDist msgDist).support →
+    scheme.posteriorMsgProb msgDist c m = msgDist m
```
wrenna-robson: This is very nearly equivalent to:
  ∀ (msgDist : PMF M) (c : C),
    c ∈ (scheme.marginalCiphertextDist msgDist).support →
    scheme.posteriorMsgProb msgDist c = msgDist

The issue is that the types don't match. But I think scheme.posteriorMsgProb msgDist c is a distribution!

Cslib/Crypto/Protocols/PerfectSecrecy/Defs.lean
```diff
@@ -0,0 +1,68 @@
...
+
+/-- Marginal ciphertext distribution given a message prior. -/
+noncomputable def marginalCiphertextDist (scheme : EncScheme M K C)
+    (msgDist : PMF M) : PMF C := do
+  let m ← msgDist
+  scheme.ciphertextDist m
+
+/-- Posterior probability `Pr[M = m | C = c]`. -/
+noncomputable def posteriorMsgProb (scheme : EncScheme M K C)
+    (msgDist : PMF M) (c : C) (m : M) : ENNReal :=
```
wrenna-robson: I think scheme.posteriorMsgProb msgDist c is a distribution. It would be good to define it in these terms. It will require a little proof (but do we not already have this in Mathlib? Maybe not).

Cslib/Cryptography/PerfectSecrecy/OneTimePad.lean
```diff
@@ -0,0 +1,53 @@
...
+## References
+
+* [J. Katz, Y. Lindell, *Introduction to Modern Cryptography*][KatzLindell2021]
+-/
+
+namespace Cslib.Cryptography.PerfectSecrecy
+
+/-- The one-time pad over `l`-bit strings. Encryption and decryption
+are pointwise XOR ([KatzLindell2021], Construction 2.9). -/
+noncomputable def otp (l : ℕ) : EncScheme (Fin l → Bool) (Fin l → Bool) (Fin l → Bool) where
```
wrenna-robson: I like this. However I do want to suggest that you could do this a different way - one could start with the (uniform) bernoulli distribution on Bool and show the most trivial case (when l = 1).  Then maybe there's a way to meaningfully take the "product". Thought I'm not sure the concept of a product of an encryption scheme exists in the literature (I would argue it's perfectly sensible though).

Cslib/Crypto/Protocols/PerfectSecrecy/Internal/PerfectSecrecy.lean
```diff
@@ -0,0 +1,164 @@
...
+
+namespace Cslib.Cryptography.PerfectSecrecy
+
+open PMF ENNReal
+
+universe u
+variable {M K C : Type u}
+
+/-- The joint distribution at `(m, c)` equals `msgDist m * ciphertextDist m c`. -/
+theorem jointDist_eq (scheme : EncScheme M K C) (msgDist : PMF M)
```
wrenna-robson, line 32: I really do feel this ought to be in Mathlib (sans the encryption context) because I can't see that anything here uses the particular properties of the encryption scheme?

Cslib/Cryptography/PerfectSecrecy/Internal/PerfectSecrecy.lean
```diff
@@ -0,0 +1,164 @@
...
+
+/-- Summing the joint distribution over messages gives the marginal ciphertext distribution. -/
+theorem jointDist_tsum_fst (scheme : EncScheme M K C) (msgDist : PMF M) (c : C) :
+    ∑' m, scheme.jointDist msgDist (m, c) = scheme.marginalCiphertextDist msgDist c := by
+  simp_rw [jointDist_eq]
+  change ∑' m, msgDist m * scheme.ciphertextDist m c =
+    PMF.bind msgDist (fun m => scheme.ciphertextDist m) c
+  rw [PMF.bind_apply]
+
+/-- Perfect secrecy is equivalent to message-ciphertext independence. -/
```
wrenna-robson: This proof seems real long to me. If the joint dist is the product of msgDist and the ciphertextDist, this is really just about cancelling on the left in ENNReal multiplication... isn't it?

Cslib/Cryptography/PerfectSecrecy/Internal/PerfectSecrecy.lean
```diff
@@ -0,0 +1,164 @@
...
+    have hne : (scheme.marginalCiphertextDist msgDist) c ≠ 0 :=
+      (PMF.mem_support_iff _ _).mp hc
+    have hne_top : (scheme.marginalCiphertextDist msgDist) c ≠ ⊤ :=
+      ne_top_of_le_ne_top one_ne_top (PMF.coe_le_one _ c)
+    simp only [EncScheme.posteriorMsgProb]
+    rw [h msgDist m c, ENNReal.mul_div_cancel_right hne hne_top]
+
+/-- Ciphertext indistinguishability implies message-ciphertext independence. -/
+theorem indep_of_ciphertextIndist (scheme : EncScheme M K C)
+    (h : ∀ m₀ m₁ : M, scheme.ciphertextDist m₀ = scheme.ciphertextDist m₁)
```
wrenna-robson: This predicate wants a name.

Cslib/Cryptography/PerfectSecrecy/Internal/PerfectSecrecy.lean
```diff
@@ -0,0 +1,164 @@
...
+  change scheme.ciphertextDist m c =
+    PMF.bind msgDist (fun m' => scheme.ciphertextDist m') c
+  rw [PMF.bind_apply]
+  have hd : ∀ m', scheme.ciphertextDist m' = scheme.ciphertextDist m :=
+    fun m' => h m' m
+  conv_rhs => arg 1; ext m'; rw [hd m']
+  rw [ENNReal.tsum_mul_right, PMF.tsum_coe, one_mul]
+
+/-- Ciphertext indistinguishability implies perfect secrecy. -/
+theorem backward (scheme : EncScheme M K C)
```
wrenna-robson: Not a good theorem name.

Cslib/Cryptography/PerfectSecrecy/Internal/PerfectSecrecy.lean
```diff
@@ -0,0 +1,164 @@
...
+    exact (ENNReal.mul_left_inj hm₀ hm₀_top).mp this
+  have h₁ : scheme.ciphertextDist m₁ c = scheme.marginalCiphertextDist μ c := by
+    have := key m₁; rw [mul_comm, mul_comm (μ m₁)] at this
+    exact (ENNReal.mul_left_inj hm₁ hm₁_top).mp this
+  exact h₀.trans h₁.symm
+
+/-- Perfect secrecy requires `|K| ≥ |M|` (Shannon's theorem). -/
+theorem shannonKeySpace [Fintype M] [Fintype K]
+    (scheme : EncScheme M K C) (h : scheme.PerfectlySecret) :
+    Fintype.card K ≥ Fintype.card M := by
```
wrenna-robson: Might be true with Nat.card?

Cslib/Cryptography/PerfectSecrecy/Internal/PerfectSecrecy.lean
```diff
@@ -0,0 +1,164 @@
...
+  · simp [Fintype.card_eq_zero]
+  · rw [not_isEmpty_iff] at hM
+    obtain ⟨m₀⟩ := hM
+    obtain ⟨c₀, hc₀⟩ := (scheme.ciphertextDist m₀).support_nonempty
+    have key_exists : ∀ m, ∃ k ∈ scheme.gen.support, c₀ ∈ (scheme.enc k m).support := by
+      intro m
+      have : c₀ ∈ (scheme.ciphertextDist m).support := by rw [hci m m₀]; exact hc₀
+      exact (PMF.mem_support_bind_iff _ _ _).mp this
+    choose f hf_mem hf_enc using key_exists
+    exact Fintype.card_le_of_injective f fun m₁ m₂ heq => by
```
wrenna-robson: I see an injectivity proof in a theorem, I think it should be a lemma (even if it is on the choice here).

linesthatinterlace (commented):
In general I think this is a good start on this stuff. My concern right now is about theorem organization and structure (including naming), and about whether any of the results you've proved should be Mathlib results or at least separated out (i.e. they are not really results about cryptographic concepts specifically).

SamuelSchlesinger:
> In general I think this is a good start on this stuff. My concern right now is about theorem organization and structure (including naming), and about whether any of the results you've proved should be Mathlib results or at least separated out (i.e. they are not really results about cryptographic concepts specifically).

Thanks for the feedback! I'll work on a follow up commit today that addresses each of these concerns. Re: Mathlib results, not totally sure how to approach it, I'll wait for more advice there.

fmontesi:
Regarding the directory name, is it open also to cryptology stuff in the future? If so, we could consider calling it simply 'Crypto'.

SamuelSchlesinger:
> Regarding the directory name, is it open also to cryptology stuff in the future? If so, we could consider calling it simply 'Crypto'.

Great idea. I was actually considering adding some of the ciphers from the first chapter of the book originally, that could be fun to have around pedagogically.

Cslib/Cryptography/PerfectSecrecy/Internal/PerfectSecrecy.lean
```diff
@@ -0,0 +1,164 @@
...
+
+@[expose] public section
+
+/-!
+# Perfect Secrecy: Internal proofs
+
+Auxiliary lemmas for perfect secrecy:
+- Equivalence of the conditional-probability and independence formulations
+- Both directions of the ciphertext indistinguishability characterization
+  ([KatzLindell2021], Lemma 2.5)
```
fmontesi: Missing citation in references.bib

Cslib/Crypto/Protocols/PerfectSecrecy/Defs.lean
fmontesi, line 1: Double directory name?

Cslib/Cryptography/PerfectSecrecy/Internal/PerfectSecrecy.lean
SamuelSchlesinger: Good catch. I also have the year wrong here.

Cslib/Cryptography/PerfectSecrecy/PerfectSecrecy.lean
fmontesi, line 1: PerfectSecrecy is used a lot as directory and file name. :-) 
Here it looks a bit generic. Could this be called something else?

SamuelSchlesinger: Yes, I'll fix that similarly to the double directory name. Comes from the copy-paste from the previous home.

Cslib/Cryptography/PerfectSecrecy.lean
fmontesi, line 1: What's the purpose of this file?

SamuelSchlesinger: Another copy-paste overlook, will reorganize.

SamuelSchlesinger: Ah no its not, its intentional. Its where we re-export all of these things from. I will add a doc comment.

chenson2018: I think it should just be removed completely, we avoid this pattern in general.

SamuelSchlesinger: Sounds good to me.

SamuelSchlesinger:
I believe all the current feedback should be addressed now, let me know if there are more nits or issues.

Cslib/Crypto/PerfectSecrecy/Internal/OneTimePad.lean
```diff
@@ -0,0 +1,45 @@
...
+
+/-!
+# One-Time Pad: Internal proofs
+
+The OTP ciphertext distribution is uniform regardless of message.
+-/
+
+namespace Cslib.Crypto.PerfectSecrecy.OTP
+
+noncomputable instance (n : ℕ) : Fintype (BitVec n) :=
```
wrenna-robson: No point having a Fintype instance be non-computable. Surprised this even needs to be. I would suggest actually FinEnum is most suitable here - and that could probably be added in Mathlib.

Cslib/Crypto/PerfectSecrecy/Internal/OneTimePad.lean
```diff
@@ -0,0 +1,45 @@
...
+  Fintype.ofEquiv (Fin (2 ^ n))
+    ⟨BitVec.ofFin, BitVec.toFin, fun x => by simp, fun x => by simp⟩
+
+/-- XOR by a fixed mask is self-inverse on `BitVec`: `c = k ^^^ m ↔ k = c ^^^ m`. -/
+lemma xor_right_eq_iff {l : ℕ} (c m k : BitVec l) :
+    (c = k ^^^ m) ↔ (k = c ^^^ m) := by
```
wrenna-robson: I think there's a special tactic to discharge bitvec goals. Ask around (I am not sure). This should definitely be upstream.

Cslib/Crypto/PerfectSecrecy/Internal/OneTimePad.lean
SamuelSchlesinger: Ah oops I was meaning to remove this before pushing. I'm nearly sure this is an unnecessary orphan, I bet its somewhere on Mathlib.

Cslib/Crypto/PerfectSecrecy/Internal/OneTimePad.lean
SamuelSchlesinger: Oh neat. I can grep around in Mathlib for it.

Cslib/Crypto/PerfectSecrecy/Internal/OneTimePad.lean
SamuelSchlesinger: Huh, it doesn't seem to be upstream AFAICT. I think you're right that we should upstream this as a FinEnum instance.

SamuelSchlesinger: I'm going to try to sidestep the need for it for now.

chenson2018: I'm not following the details here, but if you need an instance that should eventually be upstreamed, this is fine to have in CSLib. We just ask that you leave a `TODO` comment. (There's an action configured that opens issues for these with links to the source code.)

SamuelSchlesinger: Couldn't quite sidestep it, but at least made it local.

Cslib/Crypto/Protocols/PerfectSecrecy/Internal/OneTimePad.lean
```diff
@@ -0,0 +1,45 @@
...
+lemma xor_right_eq_iff {l : ℕ} (c m k : BitVec l) :
+    (c = k ^^^ m) ↔ (k = c ^^^ m) := by
+  constructor
+  · rintro rfl
+    rw [BitVec.xor_assoc, BitVec.xor_self, BitVec.xor_zero]
+  · rintro rfl
+    rw [BitVec.xor_assoc, BitVec.xor_self, BitVec.xor_zero]
+
+/-- The ciphertext distribution of the OTP is uniform, regardless of the message. -/
+theorem otp_ciphertextDist_eq_uniform (l : ℕ) (m : BitVec l) :
```
wrenna-robson, line 33:
```
module

public import Cslib.Init
public import Mathlib.Probability.Distributions.Uniform
public import Mathlib.Data.BitVec

@[expose] public section

/-!
# One-Time Pad: Internal proofs

The OTP ciphertext distribution is uniform regardless of message.
-/

namespace Cslib.Crypto.PerfectSecrecy.OTP

instance (n : ℕ) : Fintype (BitVec n) :=
  Fintype.ofEquiv (Fin (2 ^ n)) BitVec.equivFin.symm.toEquiv

/-- XOR by a fixed mask is self-inverse on `BitVec`: `c = k ^^^ m ↔ k = c ^^^ m`. -/
lemma xor_right_eq_iff {l : ℕ} (c m k : BitVec l) :
    (c = k ^^^ m) ↔ (k = c ^^^ m) := by grind

/-- The ciphertext distribution of the OTP is uniform, regardless of the message. -/
theorem otp_ciphertextDist_eq_uniform {l : ℕ} (m : BitVec l) :
    (PMF.uniformOfFintype _).bind (PMF.pure <| · ^^^ m) = PMF.uniformOfFintype _ := by
  simp [PMF.ext_iff, xor_right_eq_iff]

end Cslib.Crypto.PerfectSecrecy.OTP
```

linesthatinterlace (commented):
Posted some golfing.

Cslib/Crypto/PerfectSecrecy/Internal/OneTimePad.lean
SamuelSchlesinger: bv_decide doesn't seem to work, if that's what you meant

Cslib/Cryptography/PerfectSecrecy/Internal/OneTimePad.lean
SamuelSchlesinger: Switched to BitVec.

Cslib/Crypto/Protocols/PerfectSecrecy/Encryption.lean
SamuelSchlesinger: To me, support feels like the right phrasing, but I am open to suggestions.

Cslib/Crypto/Protocols/PerfectSecrecy/Encryption.lean
SamuelSchlesinger: I think given that this is really not going to be reused very much (the rest of the book defined encryption schemes totally differently), I feel we shouldn't define a custom notion of rational distributions just for this.

That said, I am very open to computable probability distributions being defined somewhere if they're going to be useful broadly.

ctchou (commented):
Just some general comments:

(1) I understand that K&L's textbook discusses only the bit vector and XOR version of one-time pad.  But the "alphabet" of a one-time pad scheme doesn't have to be of size 2^n, right?  Perhaps you can formulate a more general definition of one-time pad and prove its correct and then specialize it to the Vernam cipher.

(2) Shannon's theorem (sec.2.4 of K&L) would be a nice addition.  After all, it explains why the one-time pad is not practical in most use cases.

Shreyas4991:
> Just some general comments:
> 
> (1) I understand that K&L's textbook discusses only the bit vector and XOR version of one-time pad.  But the "alphabet" of a one-time pad scheme doesn't have to be of size 2^n, right?  Perhaps you can formulate a more general definition of one-time pad and prove its correct and then specialize it to the Vernam cipher.
> 
> (2) Shannon's theorem (sec.2.4 of K&L) would be a nice addition.  After all, it explains why the one-time pad is not practical in most use cases.

I think these are sufficiently big changes that they could go into a subsequent PR if required after discussion. The Katz book is a sufficiently strong foundation for all cryptography.

Cslib/Crypto/PerfectSecrecy/Internal/OneTimePad.lean
Shreyas4991: You could try Finset.sort with the order defined on naturals. Basically that's a computable way to get a list of elements out of a Finset where you have to provide a computable order relation.

SamuelSchlesinger:
> But the "alphabet" of a one-time pad scheme doesn't have to be of size 2^n, right?

Yes, in general any finite group works.

> Shannon's theorem (sec.2.4 of K&L) would be a nice addition

I agree with this.

> I think these are sufficiently big changes that they could go into a subsequent PR if required after discussion. The Katz book is a sufficiently strong foundation for all cryptography.

I am open to making these changes as a followup, but I planned to do some various cleanups here anyways so I may end up adding these things too.

Apologies for being silent here, I have been sick.

Cslib/Crypto/PerfectSecrecy/Internal/OneTimePad.lean
Shreyas4991: bv_decide extends the trusted code base beyond the kernel.

SamuelSchlesinger: Oh interesting, from the docs:

> bv_decide makes use of proof by reflection and adds the result of the compiled check as an axoim, thus adding the Lean compiler to the trusted code base.


TIL!

Shreyas4991: Concretely I don't think we allow these extended tcb tactics in mathlib or CSLib yet.

Cslib/Crypto/Protocols/PerfectSecrecy/Defs.lean
SamuelSchlesinger: To make `scheme.posteriorMsgProb` a distribution, I need to rely on stuff in `Internal` (in fact, I already prove its a distribution there). I put a TODO to reorganize things and do this.

Cslib/Crypto/Protocols/PerfectSecrecy/Internal/OneTimePad.lean
```diff
@@ -0,0 +1,39 @@
...
+
+-- TODO: upstream to Mathlib as a FinEnum instance for BitVec.
+@[reducible] noncomputable def bitVecFintype (n : ℕ) : Fintype (BitVec n) :=
+  Fintype.ofEquiv (Fin (2 ^ n))
+    ⟨BitVec.ofFin, BitVec.toFin, fun x => by simp, fun x => by simp⟩
+
+-- TODO: upstream to Mathlib — general BitVec XOR cancellation lemma.
+/-- XOR by a fixed mask is self-inverse on `BitVec`: `c = k ^^^ m ↔ k = c ^^^ m`. -/
+lemma xor_right_eq_iff {l : ℕ} (c m k : BitVec l) :
+    (c = k ^^^ m) ↔ (k = c ^^^ m) := by grind
```
eric-wieser: I think mathlib would usually write this as 
```suggestion
lemma eq_xor_iff_xor_eq {l : ℕ} (c m k : BitVec l) :
    (c = k ^^^ m) ↔ (x ^^^ m = k) := by grind
```
on the premise that the goal state is more static with fewer variables changing sides.

I've opened leanprover-community/mathlib4#37712 with some helper results to allow this to generalize.

Cslib/Crypto/Protocols/PerfectSecrecy/Internal/OneTimePad.lean
```diff
@@ -0,0 +1,39 @@
...
+
+/-- The ciphertext distribution of the OTP is uniform, regardless of the message. -/
+theorem otp_ciphertextDist_eq_uniform (l : ℕ) (m : BitVec l) :
+    haveI := bitVecFintype l
+    (PMF.uniformOfFintype (BitVec l)).bind
+      (fun k => PMF.pure (k ^^^ m)) =
+    PMF.uniformOfFintype (BitVec l) := by simp [PMF.ext_iff, xor_right_eq_iff]
```
eric-wieser: This is a general result about any bijective function, isn't it?

Cslib/Crypto/Protocols/PerfectSecrecy/Defs.lean
```diff
@@ -0,0 +1,80 @@
...
+
+namespace Cslib.Crypto.Protocols.PerfectSecrecy.EncScheme
+
+universe u
+variable {M K C : Type u}
+
+/-- The distribution of `Enc_K(m)` when `K ← Gen`. -/
+noncomputable def ciphertextDist (scheme : EncScheme M K C) (m : M) : PMF C := do
+  let k ← scheme.gen
+  scheme.enc k m
```
eric-wieser: Could abbreviate to
```suggestion
  scheme.enc (← scheme.gen) m
```
and throughout

SamuelSchlesinger: Oh very neat, I come from Haskell where such niceties aren't possible with do notation.

Cslib/Crypto/Protocols/PerfectSecrecy/Encryption.lean
```diff
@@ -0,0 +1,48 @@
...
+
+/--
+A private-key encryption scheme over message space `M`, key space `K`,
+and ciphertext space `C` ([KatzLindell2020], Definition 2.1).
+-/
+structure EncScheme.{u} (M K C : Type u) where
+  /-- Probabilistic key generation. -/
+  gen : PMF K
+  /-- (Possibly randomized) encryption. -/
+  enc : K → M → PMF C
```
eric-wieser: Better to name the arguments I think, as
```suggestion
  enc (key : K) (message : M) : PMF C
```
(mainly for documentation purposes)

Cslib/Crypto/Protocols/PerfectSecrecy/Internal/OneTimePad.lean
```diff
@@ -0,0 +1,39 @@
...
+
+The OTP ciphertext distribution is uniform regardless of message.
+-/
+
+namespace Cslib.Crypto.Protocols.PerfectSecrecy.OTP
+
+-- TODO: upstream to Mathlib as a FinEnum instance for BitVec.
+@[reducible] noncomputable def bitVecFintype (n : ℕ) : Fintype (BitVec n) :=
+  Fintype.ofEquiv (Fin (2 ^ n))
+    ⟨BitVec.ofFin, BitVec.toFin, fun x => by simp, fun x => by simp⟩
```
eric-wieser, line 25: This should surely just be an instance, and can be computable?

Cslib/Crypto/Protocols/PerfectSecrecy/OneTimePad.lean
```diff
@@ -0,0 +1,59 @@
...
+    EncScheme (BitVec l) (BitVec l) (BitVec l) :=
+  haveI := OTP.bitVecFintype l
+  { gen := PMF.uniformOfFintype _
+    enc := fun k m => PMF.pure (k ^^^ m)
+    dec := fun k c => k ^^^ c
+    correct := by
+      intro k _ m c hc
+      rw [PMF.mem_support_pure_iff] at hc
+      subst hc
+      rw [← BitVec.xor_assoc, BitVec.xor_self, BitVec.zero_xor] }
```
eric-wieser: This should generalize to any pure functions satisfing `Function.LeftInverse dec enc`; can you prove that and recover this as a special case?

Cslib/Crypto/Protocols/PerfectSecrecy/Internal/OneTimePad.lean
SamuelSchlesinger: Certainly can be computable, I originally had it as an instance but then figured it would be better to do it like this to avoid orphans.

Cslib/Crypto/Protocols/PerfectSecrecy/Encryption.lean
SamuelSchlesinger: How about renaming `K` to `Key` and `M` to `Message` instead? That way everything is self-documenting, including the output type of `gen`.

SamuelSchlesinger: I think I'll do both actually.

Cslib/Crypto/Protocols/PerfectSecrecy/Internal/OneTimePad.lean
SamuelSchlesinger: Fixed.

Cslib/Crypto/Protocols/PerfectSecrecy/OneTimePad.lean
SamuelSchlesinger: Added `ofPure` and derived otp as a special case.

Cslib/Crypto/Protocols/PerfectSecrecy/Encryption.lean
SamuelSchlesinger: Fixed.

Cslib/Crypto/Protocols/PerfectSecrecy/Defs.lean
SamuelSchlesinger: Fixed.

Cslib/Crypto/Protocols/PerfectSecrecy/Encryption.lean
wrenna-robson: Personally I prefer types not being words but it's to flavour really.

Cslib/Crypto/Protocols/PerfectSecrecy/Internal/OneTimePad.lean
wrenna-robson: It might be good to have the equiv that any given m defines (for a LawfulXor) - this is really a result *on equivs* after all.

Cslib/Crypto/Protocols/PerfectSecrecy/Internal/PerfectSecrecy.lean
wrenna-robson: Can you please extract the general result out of this? I don't mind if it isn't upstreamed straight away but I want to avoid specific theorems that are actually general in disguise.

Cslib/Crypto/Protocols/PerfectSecrecy/Internal/PerfectSecrecy.lean
```diff
@@ -0,0 +1,205 @@
...
+    congr 1
+    rw [tsum_eq_single c]
+    · simp [PMF.pure_apply]
+    · intro c' hc'; simp [PMF.pure_apply, hc'.symm]
+  · intro m' hm'
+    rw [PMF.bind_apply]
+    simp [PMF.pure_apply, hm'.symm]
+
+/-- Summing the joint distribution over messages gives the marginal ciphertext distribution. -/
+theorem jointDist_tsum_fst (scheme : EncScheme M K C) (msgDist : PMF M) (c : C) :
```
wrenna-robson, line 38: Similar.

Cslib/Crypto/Protocols/PerfectSecrecy/Internal/PerfectSecrecy.lean
```diff
@@ -0,0 +1,205 @@
...
+    ∑' m, scheme.jointDist msgDist (m, c) = scheme.marginalCiphertextDist msgDist c := by
+  simp_rw [jointDist_eq]
+  change ∑' m, msgDist m * scheme.ciphertextDist m c =
+    PMF.bind msgDist (fun m => scheme.ciphertextDist m) c
+  rw [PMF.bind_apply]
+
+/-- The posterior message probabilities sum to 1 when the ciphertext
+has positive probability. -/
+theorem posteriorMsgProb_hasSum (scheme : EncScheme M K C)
```
wrenna-robson: Similar.

Cslib/Crypto/Protocols/PerfectSecrecy/Internal/PerfectSecrecy.lean
```diff
@@ -0,0 +1,205 @@
...
+    rw [ENNReal.tsum_mul_right, jointDist_tsum_fst]
+    exact ENNReal.mul_inv_cancel hne hne_top
+  exact this ▸ ENNReal.summable.hasSum
+
+namespace EncScheme
+
+/-- The posterior message distribution `Pr[M | C = c]` as a probability
+distribution, given a message prior and a ciphertext in the support of
+the marginal distribution. -/
+noncomputable def posteriorMsgDist (scheme : EncScheme M K C)
```
wrenna-robson: Similar.

Cslib/Crypto/Protocols/PerfectSecrecy/Defs.lean
wrenna-robson: I am in two minds how to precede. I would really like these general results extracted more!

Cslib/Crypto/Protocols/PerfectSecrecy/Encryption.lean
wrenna-robson: I 100% agree that just for this is not so useful. But I could see a general utility (outside scope here though so don't worry about it).

Cslib/Crypto/Protocols/PerfectSecrecy/Encryption.lean
wrenna-robson: I think I agree for `gen.support`. For (enc k m).support there is a dependence so I could see a useful abbreviation being used. Don't worry about it for now.

Cslib/Crypto/Protocols/PerfectSecrecy/Internal/OneTimePad.lean
SamuelSchlesinger: It is, but when I made the theorem more general it wasn't particularly helpful for this result's clarity.

Cslib/Crypto/Protocols/PerfectSecrecy/Defs.lean
```diff
@@ -0,0 +1,84 @@
...
+variable {M K C : Type u}
+
+/-- The distribution of `Enc_K(m)` when `K ← Gen`. -/
+noncomputable def ciphertextDist (scheme : EncScheme M K C) (m : M) : PMF C := do
+  scheme.enc (← scheme.gen) m
+
+/-- Joint distribution of `(M, C)` given a message prior. -/
+noncomputable def jointDist (scheme : EncScheme M K C) (msgDist : PMF M) : PMF (M × C) := do
+  let m ← msgDist
+  return (m, ← scheme.ciphertextDist m)
```
eric-wieser, line 48:
```suggestion
  return (← msgDist, ← scheme.ciphertextDist m)
```
or if you prefer
```
  Prod.mk <$> msgDist <*> scheme.ciphertextDist m
```
which allows you to drop the `do`.

Cslib/Crypto/Protocols/PerfectSecrecy/Encryption.lean
```diff
@@ -0,0 +1,61 @@
...
+* [J. Katz, Y. Lindell, *Introduction to Modern Cryptography*][KatzLindell2020]
+-/
+
+namespace Cslib.Crypto.Protocols.PerfectSecrecy
+
+/--
+A private-key encryption scheme over message space `M`, key space `K`,
+and ciphertext space `C` ([KatzLindell2020], Definition 2.1).
+-/
+structure EncScheme.{u} (Message Key Ciphertext : Type u) where
```
eric-wieser:
```suggestion
structure EncScheme.{u} (Message Key Ciphertext : Type*) where
```
to allow the types to live in different universes.

Cslib/Crypto/Protocols/PerfectSecrecy/PMFUtilities.lean
```diff
@@ -0,0 +1,89 @@
...
+- `PMFUtilities.posterior_hasSum`: posterior probabilities sum to 1
+- `PMFUtilities.posteriorDist`: the posterior as a `PMF`
+-/
+
+namespace PMFUtilities
+
+open PMF ENNReal
+
+universe u
+variable {α β : Type u}
```
eric-wieser, line 39: Where possible these should be in different universes. You might need to specify them separately per theorem

Cslib/Crypto/Protocols/PerfectSecrecy/Defs.lean
SamuelSchlesinger: This doesn't work because `ciphertextDist` relies upon the `m`.

Cslib/Crypto/Protocols/PerfectSecrecy/PMFUtilities.lean
SamuelSchlesinger: Given that these are going to be upstreamed anyways, I'll make this change in those/that PR/s rather than making this one more wordy.

SamuelSchlesinger:
I think everyone's comments have been addressed. Let me know if there is any more feedback.

SamuelSchlesinger:
@linesthatinterlace I made a PR to my fork of Mathlib with the posterior definition and friends here: https://github.com/SamuelSchlesinger/mathlib4/pull/1. Hoping for some pre-review before bothering the mathlib reviewers for the first time :)

Cslib/Crypto/Protocols/PerfectSecrecy/PMFUtilities.lean
Shreyas4991: I think you want to have the most general version here. Upstreaming should not be anything more than a chore. Consider that the upatreaming process is a slow one and people will rely on this definition until upstreaming is complete.

fmontesi:
LGTM. @linesthatinterlace @eric-wieser can anybody give me a ping of approval? Then I'll merge.

Cslib/Crypto/Protocols/PerfectSecrecy/PMFUtilities.lean
SamuelSchlesinger: Reading through mathlib now, I feel like this might already be there in terms of kernels. I wonder if I can get rid of this entirely...

SamuelSchlesinger: Ah, no. I'd need to rephrase a lot more.

linesthatinterlace:
As long as there is a plan to back-port relevant things to Mathlib and improve proof later, I am happy for this to merge.

SamuelSchlesinger:
I have a PR to Mathlib to do so here: https://github.com/leanprover-community/mathlib4/pull/37938. I'm happy to make a followup to Cslib once we have access to these. Let me know what other aspects of these proofs needs to be improved in the meantime.

fmontesi (approved):

SamuelSchlesinger:
It got mad at my unicode caution symbols :) I'll delete those and run the lints locally to make sure everything works before I push back up.

SamuelSchlesinger:
@fmontesi I think you might need to turn back on auto-merge and trigger the ci-checks again.

chenson2018:
I've put it back in the queue.

fmontesi:
Whoops, concurrent PR management, sorry. Back in the queue. :-)

PR #470, opened 2026-04-06
feat: update operation for FinFun

fmontesi:
This PR adds `FinFun.update` and proves some basic results about the functional interface and support of the resulting function.


Cslib/Foundations/Data/FinFun/Update.lean
```diff
@@ -0,0 +1,78 @@
...
+/-! # Update for finite functions
+
+This module defines the update operation for finite functions, that is, the equivalent of
+`Function.update` for `FinFun`.
+-/
+
+namespace Cslib.FinFun
+
+/-- `FinFun` equivalent of `Function.update`. -/
+def update [Zero β] [DecidableEq α] [∀ b : β, Decidable (b = 0)] (f : α →₀ β) (a : α) (b : β) :
```
ctchou: I wonder if it is worthwhile to introduce a `DecidableEqZero` class, as the pattern `[∀ b : β, Decidable (b = 0)]` appears so many times in `FinFun`.

Cslib/Foundations/Data/FinFun/Update.lean
```diff
@@ -0,0 +1,78 @@
...
+/-- Updating a function with a nonempty element yields the same support with, possibly, the addition
+of the key. -/
+@[scoped grind .]
+theorem update_neq_zero_support_eq_union [Zero β] [DecidableEq α] [∀ b : β, Decidable (b = 0)]
+    (f : α →₀ β) (h : b ≠ 0) :
+    (f.update a b).support = f.support ∪ {a} := by grind
+
+/-- Updating a key in the support with a nonempty element preserves the support. -/
+theorem update_neq_zero_support_eq [Zero β] [DecidableEq α] [∀ b : β, Decidable (b = 0)]
+    (f : α →₀ β) (h₁ : f a ≠ 0) (h₂ : b ≠ 0) : (f.update a b).support = f.support := by grind
```
ctchou: Are these two theorems really necessary given `update_support` above?

Cslib/Foundations/Data/FinFun/Update.lean
fmontesi: Agreed, I'll introduce one.

Cslib/Foundations/Data/FinFun/Update.lean
fmontesi: Mmh, right; also, given the recent discussion on grind parsimony, probably better not to have them for now.

fmontesi:
All done!

Cslib/Foundations/Data/FinFun/Update.lean
```diff
@@ -0,0 +1,68 @@
...
+-/
+
+namespace Cslib.FinFun
+
+/-- `FinFun` equivalent of `Function.update`. -/
+def update [Zero β] [DecidableEq α] [DecidableEqZero β] (f : α →₀ β) (a : α) (b : β) :
+    α →₀ β where
+  fn := Function.update f.fn a b
+  support := if b = 0 then f.support \ {a} else f.support ∪ {a}
+  mem_support_fn := by grind
```
chenson2018: This seems problematic, as the `grind only` suggested by `grind?` fail. If you've not seen the Zulip discussion, the harm is us having to recover this by hand if there is a grind regression that breaks this proof. Could you try making it more explicit to avoid this?

Cslib/Foundations/Data/FinFun/Update.lean
```diff
@@ -0,0 +1,68 @@
...
+/-- `FinFun.update` is consistent with `Function.update`. -/
+@[simp]
+theorem update_coe [Zero β] [DecidableEq α] [DecidableEqZero β] (f : α →₀ β) :
+    ((f.update a b) : α → β) = Function.update f a b := by
+  grind [update]
+
+/-- Conditional characterisation of the functional interface of `FinFun.update`. -/
+theorem update_apply [Zero β] [DecidableEq α] [DecidableEqZero β] (f : α →₀ β) :
+    ((f.update a' b) a) = if a = a' then b else f a := by
+  simp only [update_coe, Function.update_apply]
```
chenson2018: I don't think the performance difference is big enough to squeeze this terminal simp?

```suggestion
  simp [Function.update_apply]
```

Cslib/Foundations/Data/FinFun/Update.lean
```diff
@@ -0,0 +1,68 @@
...
+def update [Zero β] [DecidableEq α] [DecidableEqZero β] (f : α →₀ β) (a : α) (b : β) :
+    α →₀ β where
+  fn := Function.update f.fn a b
+  support := if b = 0 then f.support \ {a} else f.support ∪ {a}
+  mem_support_fn := by grind
+
+/-- `FinFun.update` is consistent with `Function.update`. -/
+@[simp]
+theorem update_coe [Zero β] [DecidableEq α] [DecidableEqZero β] (f : α →₀ β) :
+    ((f.update a b) : α → β) = Function.update f a b := by
```
chenson2018:
```suggestion
    (f.update a b : α → β) = Function.update f a b := by
```

Cslib/Foundations/Data/FinFun/Update.lean
```diff
@@ -0,0 +1,68 @@
...
+This module defines the update operation for finite functions, that is, the equivalent of
+`Function.update` for `FinFun`.
+-/
+
+namespace Cslib.FinFun
+
+/-- `FinFun` equivalent of `Function.update`. -/
+def update [Zero β] [DecidableEq α] [DecidableEqZero β] (f : α →₀ β) (a : α) (b : β) :
+    α →₀ β where
+  fn := Function.update f.fn a b
```
chenson2018, line 28: Maybe preferred to write

```suggestion
  fn := Function.update f a b
```
since we usually use the coercion?

Cslib/Foundations/Data/FinFun/Update.lean
```diff
@@ -0,0 +1,68 @@
...
+/-- Conditional characterisation of the functional interface of `FinFun.update`. -/
+theorem update_apply [Zero β] [DecidableEq α] [DecidableEqZero β] (f : α →₀ β) :
+    ((f.update a' b) a) = if a = a' then b else f a := by
+  simp only [update_coe, Function.update_apply]
+
+/-- Conditional characterisation of the support of an updated `FinFun`. -/
+@[scoped grind =]
+theorem update_support [Zero β] [DecidableEq α] [DecidableEqZero β] (f : α →₀ β) :
+    (f.update a b).support = if b = 0 then f.support \ {a} else f.support ∪ {a} := by
+  simp only [update]
```
chenson2018: Same here

```suggestion
  simp [update]
```

Cslib/Foundations/Data/FinFun/Update.lean
```diff
@@ -0,0 +1,68 @@
...
+@[expose] public section
+
+/-! # Update for finite functions
+
+This module defines the update operation for finite functions, that is, the equivalent of
+`Function.update` for `FinFun`.
+-/
+
+namespace Cslib.FinFun
+
```
chenson2018, line 22: I'd use `variable [Zero β] [DecidableEq α] [DecidableEqZero β] (f : α →₀ β)` here.

Cslib/Foundations/Data/FinFun/Update.lean
```diff
@@ -0,0 +1,68 @@
...
+  mem_support_fn := by grind
+
+/-- `FinFun.update` is consistent with `Function.update`. -/
+@[simp]
+theorem update_coe [Zero β] [DecidableEq α] [DecidableEqZero β] (f : α →₀ β) :
+    ((f.update a b) : α → β) = Function.update f a b := by
+  grind [update]
+
+/-- Conditional characterisation of the functional interface of `FinFun.update`. -/
+theorem update_apply [Zero β] [DecidableEq α] [DecidableEqZero β] (f : α →₀ β) :
```
chenson2018: This seems like a reasonable `@[simp, grind =]` to me.

Cslib/Foundations/Data/FinFun/Update.lean
```diff
@@ -0,0 +1,68 @@
...
+/-- Updating a finite function on the same key is the same as doing the last update. -/
+@[scoped grind =, simp]
+theorem update_idem [Zero β] [DecidableEq α] [DecidableEqZero β] (f : α →₀ β) :
+    (f.update a b).update a b' = f.update a b' := by grind [update_apply]
+
+/-- Updates on different keys commute. -/
+@[scoped grind =]
+theorem update_comm [Zero β] [DecidableEq α] [DecidableEqZero β] (f : α →₀ β)
+    (h : a ≠ a') : (f.update a b).update a' b' = (f.update a' b').update a b := by
+  grind [update_apply]
```
chenson2018: Same issue with `grind?` here.

chenson2018 (changes requested):
Some small style comments and a couple of issues with `grind` maintainability.

ctchou (approved):
LGTM.

Cslib/Foundations/Data/FinFun/Update.lean
fmontesi: Interesting point.
I'm ok either way, but isn't it a bit weird that the way it's used (so the 'API', one might say) affects the way it's defined? This is the most natural definition since it's an operation to modify the structure, so it's ok to know about its internal implementation.

Cslib/Foundations/Data/FinFun/Update.lean
fmontesi: Can you give me a link? I can't seem to find the relevant Zulip discussion..

Cslib/Foundations/Data/FinFun/Update.lean
fmontesi: Agreed (I kept the explicit parameter for now, personal taste in how I read some theorems).

fmontesi:
@chenson2018 Should be fine now. I kept grind in some places but checked that grind? worked. Unless we want to preferrably use grind only now? (Will wait for the link to the discussion first.)

chenson2018 (approved):
Please merge #502 before this to make sure there are no text lints missed.

I pushed a commit with how I'd address the `grind?` issues. In cases where it's a issue, I replace with `grind only` and leave `set_option linter.tacticAnalysis.verifyGrindOnly false in` as appropriate. Elsewhere it is still preferable to use unsqueezed `grind`.

Cslib/Foundations/Data/FinFun/Update.lean
chenson2018: I think it's probably fine either way, I'm fine leaving it.

PR #472, opened 2026-04-06
feat(Protocols): Sigma protocols and Schnorr identification

ChristianoBraga:
## Summary
- Add a `SigmaProtocol` class capturing the three-move structure of Sigma protocols, with completeness, special soundness, and special honest-verifier zero-knowledge (SHVZK) as class fields.
- Add Schnorr's identification protocol as an instance of `SigmaProtocol`, over a cyclic group of prime order, with proofs of all three properties.

## New files
- `Cslib/Systems/Distributed/Protocols/Cryptographic/Sigma/Basic.lean` — Sigma protocol class
- `Cslib/Systems/Distributed/Protocols/Cryptographic/Sigma/SchnorrId.lean` — Schnorr instance

## References
- Boneh & Shoup, *A Graduate Course in Applied Cryptography*, Sections 19.1 and 19.4

## Test plan
- [ ] `lake build` passes with no errors
- [ ] CI workflow succeeds

### DIFF
--- a/Cslib/Systems/Distributed/Protocols/Cryptographic/Sigma/Basic.lean
+++ b/Cslib/Systems/Distributed/Protocols/Cryptographic/Sigma/Basic.lean
@@ -30,7 +30,7 @@ A Sigma Protocol for R is a pair (P, V).
 
 Reference:
 
-* [D. Boneh and V. Shoup,V., *A Graduate Course in Applied Cryptography*, One-time pad][BonehShoup],
+* [D. Boneh and V. Shoup, *A Graduate Course in Applied Cryptography*][BonehShoup],
   Section 19.4.
 
 Note: The notion of an _effective relation_ is implicit in the definition of a Sigma protocol.
--- a/Cslib/Systems/Distributed/Protocols/Cryptographic/Sigma/Basic.lean
+++ b/Cslib/Systems/Distributed/Protocols/Cryptographic/Sigma/Basic.lean
@@ -1,0 +1,69 @@
+/-
+Copyright (c) 2026 Christiano Braga. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Christiano Braga
+-/
+
+module
+
+public import Mathlib.Tactic
+
+@[expose] public section
+
+/-!
+A Sigma Protocol for R is a pair (P, V).
+
+- P is an interactive protocol algorithm called the prover, which takes as input a witness-statement
+  pair (x, y) ∈ R.
+- V is an interactive protocol algorithm called the verifier, which takes as input a statement
+  y ∈ Y, and which outputs accept or reject.
+- P and V are structured so that an interaction between them always works as follows:
+
+1. To start the protocol, P computes a message t, called the commitment, and sends t to V.
+2. Upon receiving P's commitment t, V chooses a challenge c at random from a finite challenge space
+   C, and sends c to P.
+3. Upon receiving V's challenge c, P computes a response z, and sends z to V.
+4. Upon receiving P's response z, V outputs either accept or reject, which must be computed strictly
+   as a function of the statement y and the conversation (t, c, z). In particular, V does not make
+   any random choices other than the selection of the challenge — all other computations are
+   completely deterministic.
+
+Reference:
+
+* [D. Boneh and V. Shoup,V., *A Graduate Course in Applied Cryptography*, One-time pad][BonehShoup],
+  Section 19.4.
+
+Note: The notion of an _effective relation_ is implicit in the definition of a Sigma protocol.
+      The relation R must be decidable, and the prover must be able to efficiently compute the
+      commitment and response functions, while the verifier must be able to efficiently compute
+      the verification function. In practice, these efficiency requirements are crucial for
+      the security and usability of the protocol, but they are not explicitly formalized in this
+      definition.
+-/
+
+universe u v w x y
+
+class SigmaProtocol
+    (Statement : Type u) (Witness : Type v)
+    (Commitment : Type w) (Challenge : Type x)
+    (Response : Type y) where
+  rel : Statement → Witness → Prop
+  commit : Statement → Witness → Commitment
+  respond : Statement → Witness → Commitment → Challenge → Response
+  verify : Statement → Commitment → Challenge → Response → Prop
+  extract : Statement → Commitment →
+            Challenge → Response →
+            Challenge → Response → Witness
+  -- Properties that every instance must satisfy
+  complete : ∀ (s : Statement) (w : Witness) (e : Challenge),
+               rel s w →
+               verify s (commit s w) e (respond s w (commit s w) e)
+  sound    : ∀ (s : Statement) (a : Commitment) (e e' : Challenge) (z z' : Response),
+               e ≠ e' →
+               verify s a e z →
+               verify s a e' z' →
+               rel s (extract s a e z e' z')
+  shvzk    : ∃ (simulate : Statement → Challenge → Commitment × Response),
+               ∀ (s : Statement) (e : Challenge),
+                 let (a, z) := simulate s e
+                 verify s a e z
--- a/Cslib/Systems/Distributed/Protocols/Cryptographic/Sigma/Basic.lean
+++ b/Cslib/Systems/Distributed/Protocols/Cryptographic/Sigma/Basic.lean
@@ -54,6 +54,10 @@ class SigmaProtocol
   extract : Statement → Commitment →
             Challenge → Response →
             Challenge → Response → Witness
+  simulate : Statement → Challenge → Commitment × Response
+  -- Non-degeneracy: the relation is inhabited for some statement-witness pair,
+  -- ensuring that completeness is not vacuously true.
+  nonDegenerate : ∃ (s : Statement) (w : Witness), rel s w
   -- Properties that every instance must satisfy
   complete : ∀ (s : Statement) (w : Witness) (e : Challenge),
                rel s w →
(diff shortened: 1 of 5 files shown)


Cslib/Systems/Distributed/Protocols/Cryptographic/Sigma/Basic.lean
```diff
@@ -0,0 +1,69 @@
...
+  -- Properties that every instance must satisfy
+  complete : ∀ (s : Statement) (w : Witness) (e : Challenge),
+               rel s w →
+               verify s (commit s w) e (respond s w (commit s w) e)
+  sound    : ∀ (s : Statement) (a : Commitment) (e e' : Challenge) (z z' : Response),
+               e ≠ e' →
+               verify s a e z →
+               verify s a e' z' →
+               rel s (extract s a e z e' z')
+  shvzk    : ∃ (simulate : Statement → Challenge → Commitment × Response),
```
SamuelSchlesinger: I think the simulator would be better as a field of the class, not an existential.

Cslib/Systems/Distributed/Protocols/Cryptographic/Sigma/Basic.lean
```diff
@@ -0,0 +1,69 @@
...
+      definition.
+-/
+
+universe u v w x y
+
+class SigmaProtocol
+    (Statement : Type u) (Witness : Type v)
+    (Commitment : Type w) (Challenge : Type x)
+    (Response : Type y) where
+  rel : Statement → Witness → Prop
```
SamuelSchlesinger, line 50: It is nice if SigmaProtocol instances are constrained to be non-degenerate. For instance, ensuring that `rel` is inhabited for some `statement`, `witness` pair. That way `complete` being satisfied actually means something interesting. Same for `verify`.

This is important if we want to make claims about Sigma Protocols existing with certain properties, it is helpful to defensively forbid empty types potentially giving vacuous proofs of such statements.

Cslib/Systems/Distributed/Protocols/Cryptographic/Sigma/SchnorrId.lean
```diff
@@ -0,0 +1,107 @@
...
+3. Response: The prover computes α_z := αₜ + α · c (mod q) and sends α_z to the verifier.
+4. Verification: The verifier accepts if and only if gᵅᶻ = uₜ · uᶜ.
+
+Reference:
+
+* [D. Boneh and V. Shoup,V., *A Graduate Course in Applied Cryptography*,
+  Schnorr Identification Protocol][BonehShoup], Section 19.1.
+-/
+
+section SchnorrHelpers
```
SamuelSchlesinger, line 38: These helpers should probably be moved near `ZMod` rather than living here, a TODO to this effect at least is a good idea.

Cslib/Systems/Distributed/Protocols/Cryptographic/Sigma/Basic.lean
ChristianoBraga: A nonDegenerate field was added to the class requiring the relation to be inhabited:
`nonDegenerate : ∃ (s : Statement) (w : Witness), rel s w`
In SchnorrId.lean, the proof is `⟨g ^ (0 : ZMod q).val, (0, 0), by simp⟩` — the generator itself with the zero exponent witnesses the relation.

Cslib/Systems/Distributed/Protocols/Cryptographic/Sigma/SchnorrId.lean
ChristianoBraga: A TODO comment was added above section SchnorrHelpers noting that `pow_mod_q`, `pow_val_mul`, and `pow_val_sub` are general lemmas and should eventually be moved near `ZMod`.

chenson2018 (changes requested):
Similar to my comment on the other PR, can you please add a description of AI use? We don't have a hard-line stance against it, but it is helpful for us in reviewing, especially for a PR adding new definitions.

Cslib/Systems/Distributed/Protocols/Cryptographic/Sigma/Basic.lean
```diff
@@ -0,0 +1,72 @@
...
+@[expose] public section
+
+/-!
+A Sigma Protocol for R is a pair (P, V).
+
+- P is an interactive protocol algorithm called the prover, which takes as input a witness-statement
+  pair (x, y) ∈ R.
+- V is an interactive protocol algorithm called the verifier, which takes as input a statement
+  y ∈ Y, and which outputs accept or reject.
+- P and V are structured so that an interaction between them always works as follows:
```
SamuelSchlesinger, line 20: I feel like this description is a bit odd. To me, a sigma protocol is an interactive protocol between two parties consisting of four algorithms: `commit`, `challenge`, `respond`, and `verify`. The protocol has two parties, the prover and the verifier. Then, proceed with the below.

Cslib/Systems/Distributed/Protocols/Cryptographic/Sigma/Basic.lean
```diff
@@ -0,0 +1,72 @@
...
+
+1. To start the protocol, P computes a message t, called the commitment, and sends t to V.
+2. Upon receiving P's commitment t, V chooses a challenge c at random from a finite challenge space
+   C, and sends c to P.
+3. Upon receiving V's challenge c, P computes a response z, and sends z to V.
+4. Upon receiving P's response z, V outputs either accept or reject, which must be computed strictly
+   as a function of the statement y and the conversation (t, c, z). In particular, V does not make
+   any random choices other than the selection of the challenge — all other computations are
+   completely deterministic.
```
SamuelSchlesinger, line 29: Lets just explicitly call P and V the verifier and the prover here, I feel it is clearer.

Cslib/Systems/Distributed/Protocols/Cryptographic/Sigma/Basic.lean
```diff
@@ -0,0 +1,72 @@
...
+
+* [D. Boneh and V. Shoup, *A Graduate Course in Applied Cryptography*][BonehShoup],
+  Section 19.4.
+
+Note: The notion of an _effective relation_ is implicit in the definition of a Sigma protocol.
+      The relation R must be decidable, and the prover must be able to efficiently compute the
+      commitment and response functions, while the verifier must be able to efficiently compute
+      the verification function. In practice, these efficiency requirements are crucial for
+      the security and usability of the protocol, but they are not explicitly formalized in this
+      definition.
```
SamuelSchlesinger, line 41: This speaks to the missing machinery we have, perhaps it should be added as a TODO. We really want to, long term, be discharging complexity requirements on our various pieces of schemes such as this one, computationally secure encryption schemes, etc.

Cslib/Systems/Distributed/Protocols/Cryptographic/Sigma/Basic.lean
```diff
@@ -0,0 +1,72 @@
...
+-/
+
+universe u v w x y
+
+class SigmaProtocol
+    (Statement : Type u) (Witness : Type v)
+    (Commitment : Type w) (Challenge : Type x)
+    (Response : Type y) where
+  rel : Statement → Witness → Prop
+  commit : Statement → Witness → Commitment
```
SamuelSchlesinger, line 51: `commit` must have randomized output, so this won't work. We need a PPT algorithm (I know we aren't discharging complexity requirements yet, but you can use PMF) that outputs instance-local secrets which are held by the prover, then those secrets must be used by `commit`. You can capture this in a `ProverState` if you'd like, this is common, but I don't have a strong preference for the spelling.

Cslib/Systems/Distributed/Protocols/Cryptographic/Sigma/Basic.lean
```diff
@@ -0,0 +1,72 @@
...
+  respond : Statement → Witness → Commitment → Challenge → Response
+  verify : Statement → Commitment → Challenge → Response → Prop
+  extract : Statement → Commitment →
+            Challenge → Response →
+            Challenge → Response → Witness
```
SamuelSchlesinger, line 56: I think the two Challenge, Response pairs being in tuples might read nicer down the line.

Cslib/Systems/Distributed/Protocols/Cryptographic/Sigma/Basic.lean
```diff
@@ -0,0 +1,72 @@
...
+  simulate : Statement → Challenge → Commitment × Response
```
SamuelSchlesinger, line 57: We need more randomness here as well, we'll randomly select our response in particular. You could just have `Response` as an argument I think?

Cslib/Systems/Distributed/Protocols/Cryptographic/Sigma/Basic.lean
```diff
@@ -0,0 +1,72 @@
...
+  -- Non-degeneracy: the relation is inhabited for some statement-witness pair,
+  -- ensuring that completeness is not vacuously true.
+  nonDegenerate : ∃ (s : Statement) (w : Witness), rel s w
+  -- Properties that every instance must satisfy
+  complete : ∀ (s : Statement) (w : Witness) (e : Challenge),
+               rel s w →
+               verify s (commit s w) e (respond s w (commit s w) e)
```
SamuelSchlesinger, line 64: We'll need to change this once we have the correct randomness in `commit`, this will have to apply for arbitrary client secrets.

Cslib/Systems/Distributed/Protocols/Cryptographic/Sigma/Basic.lean
```diff
@@ -0,0 +1,72 @@
...
+  sound    : ∀ (s : Statement) (a : Commitment) (e e' : Challenge) (z z' : Response),
+               e ≠ e' →
+               verify s a e z →
+               verify s a e' z' →
+               rel s (extract s a e z e' z')
+  shvzk    : ∀ (s : Statement) (e : Challenge),
+               let (a, z) := simulate s e
+               verify s a e z
```
SamuelSchlesinger, line 72: This is not what special HVZK is -- special HVZK says that the distribution of real protocol conversations matches the distribution of simulated protocol conversations. It is a statement about equality of PMFs.

SamuelSchlesinger (changes requested):
Just reviewed the sigma protocol file, but hopefully that feedback is helpful enough to make progress and those changes should flow through to the schnorr ID file, which I know needs changes given the changes needed in sigma.

PR #477, opened 2026-04-07
feat(Data/PFunctor): add free monad of a polynomial functor

quangvdao:
## Summary

- Port `PFunctor.FreeM` from [VCV-io](https://github.com/dtumad/VCV-io) (`ToMathlib/PFunctor/Free.lean`) to cslib.
- The free monad on a polynomial functor extends the W-type construction with an extra `pure` constructor, yielding a lawful monad that is free over `P : PFunctor`.

### Main definitions

- `PFunctor.FreeM`: inductive type with `pure` and `roll` constructors
- `FreeM.lift` / `FreeM.liftA`: lifting from the base polynomial functor
- `Monad` and `LawfulMonad` instances
- `FreeM.inductionOn` / `FreeM.construct`: propositional and dependent eliminators
- `FreeM.mapM`: canonical interpretation into any target monad, with `simp` lemmas for `bind`, `map`, `seq`, etc.

### Notes

- The `MonadHom`-related definitions (`mapMHom`, `mapMHom'`) from the original VCV-io source are omitted since cslib does not have `MonadHom` infrastructure. These can be added later if cslib gains monad homomorphism support.
- File placed at `Cslib/Foundations/Data/PFunctor/FreeM.lean` as a foundation for future polynomial functor work.
- Builds cleanly with no linter warnings.

Posted by Cursor assistant (model: claude-4.6-opus-high-thinking) on behalf of the user (Quang Dao) with approval.

Made with [Cursor](https://cursor.com)


eric-wieser:
Can you include an explanation of the difference with FreeM in the module docstring?

Cslib/Foundations/Data/PFunctor/FreeM.lean
```diff
@@ -0,0 +1,283 @@
...
+namespace PFunctor
+
+/-- The free monad on a polynomial functor.
+This extends the `W`-type construction with an extra `pure` constructor. -/
+inductive FreeM (P : PFunctor.{uA, uB}) : Type v → Type (max uA uB v)
+  /-- A leaf node wrapping a pure value. -/
+  | protected pure {α} (a : α) : FreeM P α
+  /-- A node with shape `a : P.A` and subtrees given by the continuation
+  `cont : P.B a → FreeM P α`. -/
+  | roll {α} (a : P.A) (cont : P.B a → FreeM P α) : FreeM P α
```
eric-wieser: This should probably be called `liftBind` to match the other FreeM

Cslib/Foundations/Data/PFunctor/FreeM.lean
```diff
@@ -0,0 +1,283 @@
...
+deriving Inhabited
+
+namespace FreeM
+
+variable {P : PFunctor.{uA, uB}} {α β γ : Type v}
+
+instance : Pure (FreeM P) where pure := .pure
+
+@[simp]
+theorem pure_eq_pure : (pure : α → FreeM P α) = FreeM.pure := rfl
```
eric-wieser: Note that #417, if it goes through, would advocate for flipping this lemma

Cslib/Foundations/Data/PFunctor/FreeM.lean
```diff
@@ -0,0 +1,283 @@
...
+      C (FreeM.roll a cont)) :
+    (x : FreeM P α) → C x
+  | FreeM.pure a => pure a
+  | FreeM.roll a cont => roll a _ (fun u ↦ FreeM.inductionOn pure roll (cont u))
+
+section construct
+
+/-- Dependent eliminator for `FreeM P`. -/
+@[elab_as_elim]
+protected def construct {C : FreeM P α → Type*}
```
eric-wieser: Isn't this the same as the autogenerated FreeM.rec?

Cslib/Foundations/Data/PFunctor/FreeM.lean
```diff
@@ -0,0 +1,283 @@
...
+
+end construct
+
+section mapM
+
+variable {m : Type uB → Type v} {α : Type uB}
+
+/-- Canonical mapping of `FreeM P` into any other monad, given an interpretation
+`interp : (a : P.A) → m (P.B a)`. -/
+protected def mapM [Pure m] [Bind m] (interp : (a : P.A) → m (P.B a)) : FreeM P α → m α
```
eric-wieser: I think we call this `run` in the other FreeM

Cslib/Foundations/Data/PFunctor/FreeM.lean
quangvdao: Yes, I flipped this

quangvdao:
Thanks @eric-wieser ! I believe I have addressed all of your comments, including the docstring explaining the difference with `FreeM`

Cslib/Foundations/Data/PFunctor/FreeM.lean
```diff
@@ -0,0 +1,271 @@
...
+
+## Comparison with `Cslib.FreeM`
+
+`Cslib.FreeM F` (in `Cslib.Foundations.Control.Monad.Free`) builds a free monad over an
+arbitrary type constructor `F : Type u → Type v`, using a single `liftBind` constructor that
+existentially quantifies the intermediate type:
+```
+| liftBind {ι : Type u} (op : F ι) (cont : ι → FreeM F α) : FreeM F α
+```
+Here the intermediate type `ι` is hidden, so `F` need not be a `Functor`.
```
eric-wieser: What does "hidden" mean?

Cslib/Foundations/Data/PFunctor/FreeM.lean
```diff
@@ -0,0 +1,271 @@
...
+
+We define the free monad on a **polynomial functor** (`PFunctor`), and prove some basic properties.
+
+The free monad `PFunctor.FreeM P` extends the W-type construction with an extra `pure`
+constructor, yielding a monad that is free over the polynomial functor `P`.
+
+## Comparison with `Cslib.FreeM`
+
+`Cslib.FreeM F` (in `Cslib.Foundations.Control.Monad.Free`) builds a free monad over an
+arbitrary type constructor `F : Type u → Type v`, using a single `liftBind` constructor that
```
eric-wieser: What is "single" signifying here? Both types have two constructors not one.

Cslib/Foundations/Data/PFunctor/FreeM.lean
quangvdao: thanks, addressed in latest commit

Cslib/Foundations/Data/PFunctor/FreeM.lean
quangvdao: thanks, addressed in latest commit

Cslib/Foundations/Data/PFunctor/FreeM.lean
```diff
@@ -0,0 +1,272 @@
...
+
+## Comparison with `Cslib.FreeM`
+
+`Cslib.FreeM F` (in `Cslib.Foundations.Control.Monad.Free`) builds a free monad over an
+arbitrary type constructor `F : Type u → Type v`.
+Its `liftBind` constructor existentially quantifies the intermediate type:
+```
+| liftBind {ι : Type u} (op : F ι) (cont : ι → FreeM F α) : FreeM F α
+```
+Because `ι` is an implicit (existentially bound) argument, the caller need not name it,
```
eric-wieser: "implicit" in lean usually means `{}`, but that's not relevent here. Similarly, "existentially-bound" would normally imply the use of `Exists`, which is also not present here. Could you clarify what you are trying say here?

Cslib/Foundations/Data/PFunctor/FreeM.lean
```diff
@@ -0,0 +1,272 @@
...
+so `F` need not be a `Functor`.
```
eric-wieser: Is the situation any different from `PFunctor`?

Cslib/Foundations/Data/PFunctor/FreeM.lean
```diff
@@ -0,0 +1,272 @@
...
+
+namespace PFunctor
+
+/-- The free monad on a polynomial functor.
+This extends the `W`-type construction with an extra `pure` constructor. -/
+inductive FreeM (P : PFunctor.{uA, uB}) : Type v → Type (max uA uB v)
+  /-- A leaf node wrapping a pure value. -/
+  | protected pure {α} (a : α) : FreeM P α
+  /-- Invoke the operation `a : P.A` with continuation `cont : P.B a → FreeM P α`. -/
+  | liftBind {α} (a : P.A) (cont : P.B a → FreeM P α) : FreeM P α
```
eric-wieser: nit:
```suggestion
  | liftBind {α} (a : P.A) (cont : P.B a → P.FreeM α) : P.FreeM α
```
and throughout

Cslib/Foundations/Data/PFunctor/FreeM.lean
```diff
@@ -0,0 +1,272 @@
...
+deriving Inhabited
+
+namespace FreeM
+
+variable {P : PFunctor.{uA, uB}} {α β γ : Type v}
```
eric-wieser: Does this work?
```suggestion
variable {P : PFunctor.{uA, uB}} {α β γ : Type*}
```

quangvdao:
Thanks, addressed those as well!

@fmontesi @chenson2018 could you take a look and approve if possible?

Cslib/Foundations/Data/PFunctor/Free.lean
```diff
@@ -0,0 +1,270 @@
...
+deriving Inhabited
+
+namespace FreeM
+
+variable {P : PFunctor.{uA, uB}} {α β γ : Type*}
+
+instance : Pure (P.FreeM) where pure := .pure
+
+@[simp]
+theorem pure_eq_pure : (FreeM.pure : α → P.FreeM α) = pure := rfl
```
SamuelSchlesinger, line 91: Is there a reason to differ from the convention in `Foundations/Control/Monad/Free.lean`?

Cslib/Foundations/Data/PFunctor/Free.lean
```diff
@@ -0,0 +1,270 @@
...
+
+@[simp]
+lemma liftM_pure' (a : α) : (FreeM.pure a : P.FreeM α).liftM interp = Pure.pure a := rfl
+
+@[simp]
+lemma liftM_liftBind (a : P.A) (cont : P.B a → P.FreeM α) :
+    (FreeM.liftBind a cont).liftM interp = interp a >>= fun u => (cont u).liftM interp := rfl
+
+@[simp]
+lemma liftM_pure (a : α) : (Pure.pure a : P.FreeM α).liftM interp = Pure.pure a := rfl
```
SamuelSchlesinger: Given that we have `pure_eq_pure`, I think we can remove this.

Cslib/Foundations/Data/PFunctor/Free.lean
```diff
@@ -0,0 +1,270 @@
...
+  (id_map := id_map)
+  (pure_bind := pure_bind)
+  (bind_assoc := FreeM.bind_assoc)
+
+lemma pure_inj (a b : α) : (pure a : P.FreeM α) = pure b ↔ a = b := by
+  constructor
+  · exact FreeM.pure.inj
+  · rintro rfl; rfl
+
+@[simp] lemma liftBind_inj (a a' : P.A)
```
SamuelSchlesinger: Because `liftBind.injEq` already exists as a simp lemma, I think we should remove `simp` here. A nicer proof might exist as well, this proof is weird. Try starting with `constructor`, the second branch becomes `rintro ⟨rfl, rfl⟩; rfl` and the first one goes `intro h`, `obtain ⟨rfl, h⟩ := FreeM.liftBind.inj h`, then `exact ⟨rfl, eq_of_heq h⟩`.

Cslib/Foundations/Data/PFunctor/Free.lean
```diff
@@ -0,0 +1,270 @@
...
+def liftA (a : P.A) : P.FreeM (P.B a) := lift ⟨a, id⟩
+
+instance : MonadLift P (P.FreeM) where
+  monadLift x := FreeM.lift x
+
+@[simp] lemma lift_ne_pure (x : P α) (y : α) :
+    (lift x : P.FreeM α) ≠ PFunctor.FreeM.pure y := by simp [lift]
+
+@[simp] lemma pure_ne_lift (x : P α) (y : α) :
+    PFunctor.FreeM.pure y ≠ (lift x : P.FreeM α) := by simp [lift]
```
SamuelSchlesinger: Lets make this and `lift_ne_pure` use `P.Obj` to match the rest of the file.

SamuelSchlesinger (changes requested):
Lets also add `Interprets` and  friends to match the existing free monad. Also, lets add docs there pointing over to this file.

SamuelSchlesinger:
If we want to leave `Interprets` stuff for a follow up I am happy to help there, but the rest seems necessary to address or at least talk through.

Cslib/Foundations/Data/PFunctor/Free.lean
eric-wieser: Yes, I asked for this change in https://github.com/leanprover/cslib/pull/477#discussion_r3048497919 in preparation for #417, which would align Free.lean with what is in this PR.

ChristianoBraga:
Free monads are beautiful structures: they cleanly separate syntax from semantics. The same program, written as data, can be given different interpretations, such as an interpreter, a compiler, or a static analysis algorithm, by means of handlers.

Cslib already offers a free monad in the so-called freer monad encoding, which pairs each operation with a continuation that keeps the recursion strictly positive, thus allowing it to be defined as an inductive type for any F. The price is a universe bump: because each operation hides an intermediate type (of Type u), the monad that stores it must live one universe higher.

This PR adds a complementary construction, PFunctor.FreeM P, parameterized by a polynomial functor, where operations are values and not quantified types. The universe bump is avoided, strict positivity is automatic, and the explicit data yields stronger eliminators.

All review comments have been addressed, and the module compiles cleanly against current main. @SamuelSchlesinger, could you re-review to clear your change request? @fmontesi, once cleared, I'd be grateful if you could approve and merge into main.

SamuelSchlesinger:
To @eric-wieser's point, @quangvdao can you restore the pure simp direction you reverted after I mistakenly asked for it to be changed back? I didn't previously see Eric's comment.

SamuelSchlesinger (approved):
LGTM after you revert the simp-normal form change in favor of @eric-wieser's original comment.

quangvdao:
Thanks for all the comments! Yes my latest commit reverted the simp direction

SamuelSchlesinger:
> Thanks for all the comments! Yes my latest commit reverted the simp direction

Thx! In that case LGTM, @fmontesi we can merge whenever.

Cslib/Foundations/Data/PFunctor/Free.lean
```diff
@@ -0,0 +1,305 @@
...
+theorem pure_eq_pure : (FreeM.pure : α → P.FreeM α) = pure := rfl
+
+attribute [-simp] FreeM.pure.injEq FreeM.pure.sizeOf_spec
+attribute [nolint simpNF] FreeM.pure.injEq FreeM.pure.sizeOf_spec
+
+/-- Lift an object of the base polynomial functor into the free monad. -/
+def lift (x : P.Obj α) : P.FreeM α := FreeM.liftBind x.1 (fun y ↦ FreeM.pure (x.2 y))
+
+/-- Lift a position of the base polynomial functor into the free monad. -/
+def liftA (a : P.A) : P.FreeM (P.B a) := lift ⟨a, id⟩
```
eric-wieser: I would guess this is more convenient as
```suggestion
abbrev liftA (a : P.A) : P.FreeM (P.B a) := lift ⟨a, id⟩
```

Cslib/Foundations/Data/PFunctor/Free.lean
```diff
@@ -0,0 +1,305 @@
...
+@[simp] lemma lift_ne_pure (x : P.Obj α) (y : α) :
+    (lift x : P.FreeM α) ≠ pure y := by simp [lift]
+
+@[simp] lemma pure_ne_lift (x : P.Obj α) (y : α) :
+    pure y ≠ (lift x : P.FreeM α) := by simp [lift]
+
+lemma monadLift_eq_lift (x : P.Obj α) : (x : P.FreeM α) = FreeM.lift x := rfl
+
+/-- Bind operator on `FreeM P` used in the monad definition. -/
+protected def bind : P.FreeM α → (α → P.FreeM β) → P.FreeM β
```
eric-wieser: Can you copy over the remark about `>>=` from the other `FreeM.bind`?

Cslib/Foundations/Data/PFunctor/Free.lean
```diff
@@ -0,0 +1,305 @@
...
+
+variable {P : PFunctor.{uA, uB}} {α β γ : Type*}
+
+instance : Pure (P.FreeM) where pure := .pure
+
+@[simp]
+theorem pure_eq_pure : (FreeM.pure : α → P.FreeM α) = pure := rfl
+
+attribute [-simp] FreeM.pure.injEq FreeM.pure.sizeOf_spec
+attribute [nolint simpNF] FreeM.pure.injEq FreeM.pure.sizeOf_spec
```
eric-wieser: In the other file we solve this with something else.

My understanding is that `-simp` only works for the current file, so this doesn't really work.

Cslib/Foundations/Data/PFunctor/Free.lean
```diff
@@ -0,0 +1,305 @@
...
+    simp [FreeM.bind] at *
+    simp [ih]
+
+instance : Bind (P.FreeM) where bind := .bind
+
+@[simp]
+theorem bind_eq_bind {α β : Type v} :
+    Bind.bind = (FreeM.bind : P.FreeM α → _ → P.FreeM β) := rfl
+
+/-- Map a function over a `FreeM` computation. -/
```
eric-wieser: Same here, let's copy the comment about `<$>`

Cslib/Foundations/Data/PFunctor/Free.lean
```diff
@@ -0,0 +1,305 @@
...
+
+@[simp]
+theorem pure_eq_pure : (FreeM.pure : α → P.FreeM α) = pure := rfl
+
+attribute [-simp] FreeM.pure.injEq FreeM.pure.sizeOf_spec
+attribute [nolint simpNF] FreeM.pure.injEq FreeM.pure.sizeOf_spec
+
+/-- Lift an object of the base polynomial functor into the free monad. -/
+def lift (x : P.Obj α) : P.FreeM α := FreeM.liftBind x.1 (fun y ↦ FreeM.pure (x.2 y))
+
```
eric-wieser, line 95: Let's add a copy of the `liftBind_eq` simp lemma, to normalize `liftBind` to `lift` + `bind`, as we do for the other `FreeM`.

You should also copy across the `theorem induction` from that file.

Cslib/Foundations/Data/PFunctor/Free.lean
```diff
@@ -0,0 +1,305 @@
...
+@[simp] lemma pure_eq_bind_iff (x : P.FreeM α) (f : α → P.FreeM β) (b : β) :
+    pure b = x.bind f ↔ ∃ a, x = pure a ∧ pure b = f a := by
+  cases x with
+  | pure a => exact ⟨fun h => ⟨a, rfl, h⟩, fun ⟨_, h, hf⟩ => by rwa [FreeM.pure.inj h]⟩
+  | liftBind a cont => simp [FreeM.bind]
+
+instance : LawfulFunctor (P.FreeM) where
+  map_const := rfl
+  id_map := id_map
+  comp_map _ _ := comp_map _ _
```
eric-wieser: I think yo ucan drop this entirely, it's implied by the result below.

Cslib/Foundations/Data/PFunctor/Free.lean
```diff
@@ -0,0 +1,305 @@
...
+
+section liftM
+
+variable {m : Type uB → Type v} {α : Type uB}
+
+/-- Interpret a `FreeM P` computation into any monad `m` by providing an interpretation
+`interp : (a : P.A) → m (P.B a)` for each operation. -/
+protected def liftM [Pure m] [Bind m] (interp : (a : P.A) → m (P.B a)) : P.FreeM α → m α
+  | .pure a => Pure.pure a
+  | .liftBind a cont => (interp a) >>= (fun u ↦ (cont u).liftM interp)
```
eric-wieser:
```suggestion
  | .liftBind a cont => interp a >>= fun u ↦ (cont u).liftM interp
```

Cslib/Foundations/Data/PFunctor/Free.lean
```diff
@@ -0,0 +1,305 @@
...
+# Free Monad of a Polynomial Functor
+
+We define the free monad on a **polynomial functor** (`PFunctor`), and prove some basic properties.
+
+The free monad `PFunctor.FreeM P` extends the W-type construction with an extra `pure`
+constructor, yielding a monad that is free over the polynomial functor `P`.
+
+## Comparison with `Cslib.FreeM`
+
+`Cslib.FreeM F` (in `Cslib.Foundations.Control.Monad.Free`) builds a free monad over an
```
eric-wieser:
```suggestion
`Cslib.FreeM F` (in `Cslib/Foundations/Control/Monad/Free.lean`) builds a free monad over an
```
I think has better docgen support

Cslib/Foundations/Data/PFunctor/Free.lean
```diff
@@ -0,0 +1,305 @@
...
+
+`PFunctor.FreeM P` instead takes a polynomial functor `P : PFunctor`, where the shapes
+`P.A` and positions `P.B a` are given explicitly.
+Its `liftBind` constructor uses the shape and continuation directly:
+```
+| liftBind (a : P.A) (cont : P.B a → P.FreeM α) : P.FreeM α
+```
+
+When the effect signature is naturally polynomial (a fixed set of operations, each with a
+known return type), `PFunctor.FreeM` gives stronger eliminators and avoids the universe
```
eric-wieser: Can you elaborate on what "stronger eliminator" means here?

Cslib/Foundations/Data/PFunctor/Free.lean
```diff
@@ -0,0 +1,305 @@
...
+    ∀ x : P.FreeM α, map (h ∘ g) x = map h (map g x)
+  | .pure a => rfl
+  | .liftBind a cont => by simp_all [map, comp_map]
+
+instance : Functor (P.FreeM) where
+  map := .map
+
+@[simp]
+theorem map_eq_map {α β : Type v} :
+    Functor.map = FreeM.map (P := P) (α := α) (β := β) := rfl
```
eric-wieser: This one is backwards compared to the other `FreeM`; you'll need to make changes below after fixing it.

eric-wieser:
@ChristianoBraga, could you clarify your involvement here? Without further context, it appears a little bit like you just asked a language model to post a summary comment.

fmontesi:
Thanks for the great work so far, @quangvdao and all reviewers!

@quangvdao, I see there are some last-minute but minor comments by @eric-wieser that look quick to address. Could you have a look? Please ping me here as soon as it's done and I'll merge.

eric-wieser:
@fmontesi, would you mind letting me give a final pass after @quangvdao addresses my comments? I think some of them might be somewhat tricky to address.

Cslib/Foundations/Data/PFunctor/Free.lean
```diff
@@ -160,22 +171,57 @@ lemma lift_bind (x : P.Obj α) (f : α → P.FreeM β) :
...
+    · intro h
+      cases h
+    · rintro ⟨_, h, _⟩
+      cases h
 
 @[simp] lemma pure_eq_bind_iff (x : P.FreeM α) (f : α → P.FreeM β) (b : β) :
     pure b = x.bind f ↔ ∃ a, x = pure a ∧ pure b = f a := by
   cases x with
-  | pure a => exact ⟨fun h => ⟨a, rfl, h⟩, fun ⟨_, h, hf⟩ => by rwa [FreeM.pure.inj h]⟩
-  | liftBind a cont => simp [FreeM.bind]
```
eric-wieser: Does this proof not essentially still work?

Cslib/Foundations/Data/PFunctor/Free.lean
```diff
@@ -0,0 +1,363 @@
...
+instance : Pure (P.FreeM) where pure := .pure
+
+@[simp]
+theorem pure_eq_pure : (FreeM.pure : α → P.FreeM α) = pure := rfl
+
+/-- Lift an object of the base polynomial functor into the free monad. -/
+def lift (x : P.Obj α) : P.FreeM α := FreeM.liftBind x.1 (fun y ↦ FreeM.pure (x.2 y))
+
+/-- Lift a position of the base polynomial functor into the free monad. -/
+abbrev liftA (a : P.A) : P.FreeM (P.B a) := lift ⟨a, id⟩
```
eric-wieser: I wonder if we want to swap these names; after all, `PFunctor.FreeM.liftA` here corresponds to `FreeM.lift`.

Cslib/Foundations/Data/PFunctor/Free.lean
```diff
@@ -0,0 +1,363 @@
...
+@[simp]
+lemma liftM_seqRight {α β : Type uB}
+    (interp : (a : P.A) → m (P.B a)) (x : P.FreeM α) (y : P.FreeM β) :
+    (x *> y).liftM interp = x.liftM interp *> y.liftM interp := by
+  simp [seqRight_eq_bind]
+
+@[simp]
+lemma liftM_lift (interp : (a : P.A) → m (P.B a)) (x : P.Obj α) :
+    (FreeM.lift x).liftM interp = x.2 <$> interp x.1 := by
+  change ((FreeM.liftA x.1).bind fun y ↦ pure (x.2 y)).liftM interp = x.2 <$> interp x.1
```
eric-wieser: This `change` feels like a code smell; can you rework the proof here and above to avoid it?

Cslib/Foundations/Data/PFunctor/Free.lean
```diff
@@ -0,0 +1,363 @@
...
+Its `liftBind` constructor uses the shape and continuation directly:
+```
+| liftBind (a : P.A) (cont : P.B a → P.FreeM α) : P.FreeM α
+```
+
+When the effect signature is naturally polynomial (a fixed set of operations, each with a
+known return type), `PFunctor.FreeM` avoids the universe bump that the abstract `ι` in
+`Cslib.FreeM` introduces. Its eliminators also expose the shape `a : P.A` and the
+fiber `P.B a` directly in the recursive case, rather than hiding them behind an
+arbitrary intermediate type `ι`.
```
eric-wieser: I still don't fully see the argument here; can you elaborate with an example? If you do induction for any concrete `F` when using the other FreeM, then you can instead induct on `op : F ι` to get the shape and fibers.

Perhaps the key thing is that this requires applying two induction principles and not one, and is somehow less powerful as a result? Unless you have a concrete example in mind which demonstrates this, I'd suggest dropping this claim.

Cslib/Foundations/Data/PFunctor/Free.lean
```diff
@@ -0,0 +1,363 @@
...
+
+universe u v uA uB
+
+namespace PFunctor
+
+-- Disable generation of unneeded lemmas which the simpNF linter would complain about.
+set_option genInjectivity false in
+set_option genSizeOfSpec false in
+/-- The free monad on a polynomial functor.
+This extends the `W`-type construction with an extra `pure` constructor. -/
```
eric-wieser:
```suggestion
This extends `WType` with an extra `pure` constructor. -/
```

Cslib/Foundations/Data/PFunctor/Free.lean
quangvdao: I dropped this comment

quangvdao:
@eric-wieser thanks for your thorough comments! all should be addressed now

ChristianoBraga:
Hi Erik, I did NOT use ANY AI here.

I am a Research Scientist with Beneficial AI Foundation and one of my
assignments is to help cslib community.

Regarding this PR, I am just interested in having VCV-io experience merged
into cslib. We at BAIF believe cslib could represent to Signal Shot what
mathlib did to Liquid Tensor experiment.

While talking to Alex Rademaker yesterday about several things including
this PR, he suggested I could try to help merging it given its advanced
review state.  He also suggested I could say a few words to help the
maintainers appreciate it and merge it.


Em qua., 27 de mai. de 2026 às 22:59, Eric Wieser ***@***.***>
escreveu:

Cslib/Foundations/Data/PFunctor/Free.lean
```diff
@@ -0,0 +1,358 @@
...
+`PFunctor.FreeM P` instead takes a polynomial functor `P : PFunctor`, where the shapes
+`P.A` and positions `P.B a` are given explicitly.
+Its `liftBind` constructor uses the shape and continuation directly:
+```
+| liftBind (a : P.A) (cont : P.B a → P.FreeM α) : P.FreeM α
+```
+
+When the effect signature is naturally polynomial (a fixed set of operations, each with a
+known return type), `PFunctor.FreeM` avoids the universe bump that the abstract `ι` in
+`Cslib.FreeM` introduces, and records operations using the explicit shape/fiber data of `P`.
```
eric-wieser: I believe you had an example that demonstrated why the universe bump matters; do you think it would be appropriate to include it here?

Cslib/Foundations/Data/PFunctor/Free.lean
```diff
@@ -0,0 +1,358 @@
...
+instance : Pure (P.FreeM) where pure := .pure
+
+@[simp]
+theorem pure_eq_pure : (FreeM.pure : α → P.FreeM α) = pure := rfl
+
+/-- Lift a shape of the base polynomial functor into the free monad. -/
+def lift (a : P.A) : P.FreeM (P.B a) := FreeM.liftBind a pure
+
+/-- Lift an object of the base polynomial functor into the free monad. -/
+def liftObj (x : P.Obj α) : P.FreeM α := FreeM.liftBind x.1 (fun y ↦ FreeM.pure (x.2 y))
```
eric-wieser: I'd perhaps suggest defiining this as `(lift x.1).map x.2`, or even `x.2 <$> lift x.1` if that doesn't cause universe issues. Then you could make it an `abbrev` to reduce the API surface needed.

Cslib/Foundations/Data/PFunctor/Free.lean
quangvdao: I put an example in the docstring (I feel that my particular motivating example is a bit too niche)

the point is that without an universe bump we could have `FreeM` objects be part of the return type for higher-order effects

quangvdao:
@fmontesi last CI run before merge?

fmontesi:
Run in progress. I've also relaxed the rules for running CI, this is unintentionally annoying, sorry. :-)

fmontesi (approved):

fmontesi:
Here we go, @quangvdao. :)

Cslib/Foundations/Data/PFunctor/Free.lean
fmontesi: @quangvdao Just wanna make sure this comment doesn't get lost. I can merge in the meantime, we can make this a future PR. (Want to avoid having conflicts again.)

quangvdao: thanks! I thought about this a bit and decided not to do it here, it seems the tradeoffs are nontrivial?

PR #483, opened 2026-04-10
feat(MachineLearning/PAC): PAC Learning Definitions and Sample Complexity Lower Bounds

SamuelSchlesinger:
This PR defines the Probably Approximately Correct (PAC) learning model. Then, it formalizes the information theoretic lower bound on sample complexity for PAC learning implemented in "A general lower bound on the number of examples needed for learning" (Andrzej Ehrenfeucht, David Haussler, Michael Kearns, Leslie Valiant, 1989).

Organized the PR such that it can be reviewed commit by commit.

### DIFF
--- a/Cslib/MachineLearning/PACLearning/Defs.lean
+++ b/Cslib/MachineLearning/PACLearning/Defs.lean
@@ -1,0 +1,152 @@
+/-
+Copyright (c) 2026 Samuel Schlesinger. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Samuel Schlesinger
+-/
+
+module
+
+public import Cslib.Init
+public import Mathlib.MeasureTheory.Measure.MeasureSpace
+public import Mathlib.MeasureTheory.Constructions.Pi
+public import Mathlib.Order.SymmDiff
+
+@[expose] public section
+
+/-! # PAC Learning
+
+This file defines the Probably Approximately Correct (PAC) learning model
+introduced by Valiant [Valiant1984]. A concept class `C` over a domain `α` is a
+collection of subsets of `α`. A learning algorithm receives a labeled sample
+drawn i.i.d. from an unknown distribution and must produce a hypothesis that,
+with high probability, has low error with respect to the true concept.
+
+## Main definitions
+
+- `ConceptClass`: a concept class over domain `α`, i.e., a set of subsets.
+- `LabeledSample`: a finite sequence of (point, label) pairs.
+- `sampleOf`: constructs a labeled sample from a sequence of points and a concept.
+- `hypothesisError`: the error of a hypothesis with respect to a concept under a
+  distribution, defined as the measure of the symmetric difference.
+- `Learner`: a function from labeled samples to hypotheses.
+- `IsPACLearner`: the property that a deterministic learner produces a hypothesis
+  with error at most `ε` with probability at least `1 - δ`, for every distribution
+  and concept from the class.
+- `IsRPACLearner`: the randomized variant, where the learner draws internal
+  randomness from a probability space `(Ω, Q)`.
+- `sampleComplexity`: the smallest sample size admitting a deterministic PAC learner.
+- `rsampleComplexity`: the smallest sample size admitting a randomized PAC learner.
+
+## Main statements
+
+- `IsPACLearner.toIsRPACLearner`: every deterministic PAC learner is in particular
+  a randomized PAC learner (with the trivial randomness space `PUnit`).
+
+## References
+
+* [L. G. Valiant, *A Theory of the Learnable*][Valiant1984]
+* [A. Ehrenfeucht, D. Haussler, M. Kearns, L. Valiant,
+  *A General Lower Bound on the Number of Examples Needed for Learning*][EHKV1989]
+-/
+
+open MeasureTheory Set
+open scoped ENNReal
+
+namespace Cslib.MachineLearning
+
+/-- A *concept class* over domain `α` is a collection of subsets of `α`. Each subset represents
+a concept (i.e., a binary classifier). -/
+abbrev ConceptClass (α : Type*) := Set (Set α)
+
+/-- A *labeled sample* of size `m` over domain `α` is a sequence of `(point, label)` pairs. -/
+abbrev LabeledSample (α : Type*) (m : ℕ) := Fin m → (α × Bool)
+
+open Classical in
+/-- Construct a labeled sample from a sequence of points `xs` and a concept `c`.
+Each point is labeled `true` if it belongs to the concept and `false` otherwise. -/
+noncomputable def sampleOf {α : Type*} {m : ℕ} (c : Set α) (xs : Fin m → α) :
+    LabeledSample α m :=
+  fun i => (xs i, decide (xs i ∈ c))
+
+/-- The *error* of a hypothesis `h` with respect to a target concept `c` under distribution `P`,
+defined as the measure of their symmetric difference `h ∆ c`. -/
+noncomputable def hypothesisError {α : Type*} [MeasurableSpace α] (P : Measure α)
+    (h c : Set α) : ℝ≥0∞ :=
+  P (symmDiff h c)
+
+/-- A learner using `m` samples is a function that takes a labeled sample of size `m` and produces
+a hypothesis (a subset of the domain). -/
+abbrev Learner (α : Type*) (m : ℕ) := LabeledSample α m → Set α
+
+variable {α : Type*} [MeasurableSpace α]
+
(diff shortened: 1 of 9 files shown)


SamuelSchlesinger:
If this is too much to review in one pass, I am happy to make a sequence of PRs to build it up. I think the review commit by commit works personally, but I am happy to oblige whatever folks would prefer.

Making a pass of proof golfing right now, I hope some of these lemmas can be simplified at least somewhat.

Cslib/MachineLearning/PACLearning/Defs.lean
```diff
@@ -0,0 +1,152 @@
...
+
+## Main statements
+
+- `IsPACLearner.toIsRPACLearner`: every deterministic PAC learner is in particular
+  a randomized PAC learner (with the trivial randomness space `PUnit`).
+
+## References
+
+* [L. G. Valiant, *A Theory of the Learnable*][Valiant1984]
+* [A. Ehrenfeucht, D. Haussler, M. Kearns, L. Valiant,
```
Shreyas4991, line 52: Textbook citations are better. Perhaps the Kearns and Vazirani book?

Cslib/MachineLearning/PACLearning/Defs.lean
```diff
@@ -0,0 +1,152 @@
...
+open scoped ENNReal
+
+namespace Cslib.MachineLearning
+
+/-- A *concept class* over domain `α` is a collection of subsets of `α`. Each subset represents
+a concept (i.e., a binary classifier). -/
+abbrev ConceptClass (α : Type*) := Set (Set α)
+
+/-- A *labeled sample* of size `m` over domain `α` is a sequence of `(point, label)` pairs. -/
+abbrev LabeledSample (α : Type*) (m : ℕ) := Fin m → (α × Bool)
```
Shreyas4991, line 66: You don't need a vector here. Labelled samples are traditionally defined as sets. Their order is only important in online learning. Even in that case, lists are better.

Cslib/MachineLearning/PACLearning/Defs.lean
```diff
@@ -0,0 +1,152 @@
...
+
+open Classical in
+/-- Construct a labeled sample from a sequence of points `xs` and a concept `c`.
+Each point is labeled `true` if it belongs to the concept and `false` otherwise. -/
+noncomputable def sampleOf {α : Type*} {m : ℕ} (c : Set α) (xs : Fin m → α) :
+    LabeledSample α m :=
+  fun i => (xs i, decide (xs i ∈ c))
```
Shreyas4991, line 73: If you still want to use Finvecs (which I think are not the best idea here), you should use [FinVec.map](https://leanprover-community.github.io/mathlib4_docs/find/?pattern=FinVec.map#doc). If you switch to lists, then `List.map`

Cslib/MachineLearning/PACLearning/Defs.lean
```diff
@@ -0,0 +1,152 @@
...
+
+/-- The *error* of a hypothesis `h` with respect to a target concept `c` under distribution `P`,
+defined as the measure of their symmetric difference `h ∆ c`. -/
+noncomputable def hypothesisError {α : Type*} [MeasurableSpace α] (P : Measure α)
```
Shreyas4991, line 77: You want to differentiate between positive and negative errors (false positives and false negatives resp), and then define error as the combination of them. This will matter in some problems. Even textbook exercises.

Shreyas4991 (changes requested):
Initial batch of review comments.

Cslib/MachineLearning/PACLearning/Defs.lean
SamuelSchlesinger: I think, given the entire formalization is around these papers, that citing these papers is highly appropriate.

SamuelSchlesinger: How about we add both?

Cslib/MachineLearning/PACLearning/Defs.lean
SamuelSchlesinger: The indexed-function representation `Fin m -> a` is precisely what makes the product measure construction work in Mathlib. Switching to Set or List would:

- Lose the fixed-size guarantee (m is the sample size parameter)
- Require a much more complex measure-theoretic setup (measures on List or Finset)
- Not play well with Measure.pi at all

I might not have as much experience as you with these APIs, so if you could share an alternative definition which works more cleanly I am happy to integrate it.

Cslib/MachineLearning/PACLearning/Defs.lean
SamuelSchlesinger: It would just add a `FinVec.map` invocation at the beginning. I think defining this as a function is cleaner. Again, please share a code snippet if you have a cleaner approach.

Cslib/MachineLearning/PACLearning/Defs.lean
SamuelSchlesinger: This is a great idea, I will do this.

Cslib/MachineLearning/PACLearning/Defs.lean
Shreyas4991: that's a better idea. Do recall that as this is a library, we want to use relatively standard definitions and give people a textbook they can consult for them. For example mathlib probability theory cites Kallenberg's textbook

Cslib/MachineLearning/PACLearning/Defs.lean
Shreyas4991: Lists would better than Sets. You can use https://leanprover-community.github.io/mathlib4_docs/find/?pattern=MeasureTheory.Measure.tprod#doc

Cslib/MachineLearning/PACLearning/Defs.lean
Shreyas4991: "It would just add a FinVec.map invocation" => Yes,here are API lemmas for FinVec.map

Another point. I think you are using classical PAC learning as the basis. In the interests of generality one must use agnostic PAC and specialize it to classical PAC (where the target concept comes from the concept class) and noisy PAC.

Cslib/MachineLearning/PACLearning/Defs.lean
SamuelSchlesinger: tprod constructs a product measure over a List of measurable spaces, I think it ends up being uglier than the product construction here.

Cslib/MachineLearning/PACLearning/Defs.lean
SamuelSchlesinger: > Yes,here are API lemmas for FinVec.map

What do you mean?

> In the interests of generality one must use agnostic PAC and specialize it to classical PAC

Is this really necessary for a first PR about this subject, to do the most general thing, or can we prove things about the simpler definitions to start and then generalize later? Given that they should provably reduce to these definitions in their specialized case, it should be simple to do later on.

Cslib/MachineLearning/PACLearning/Defs.lean
SamuelSchlesinger: Done in the last commit.

Cslib/MachineLearning/PACLearning/Defs.lean
Shreyas4991: CSLib inherits the mathlib tradition of "let's define it in the greatest possible generality". This allows the greatest possible extent of reuse of the same definition in many places. So generalisation where possible is very important.

SamuelSchlesinger: I believe we shouldn't preclude the more general definitions, but we don't need to include them right away. For me, a comparable instance is Lambda Calculus. We could have defined a Pure Type System right away and done things in full generality, but that isn't how it went AFAICT.

Shreyas4991: That's a nice idea in general, but in lean the library building philosophy is to build a unified library. This has a long history in the lean community at this point, and is a deeply entrenched principle for good reasons. So it is not for me to debate. You could raise it on Zulip if you like. This particular definition is not maximally general and will result in unnecessary API duplication.

SamuelSchlesinger: Alright, I'll give a more general definition a go in the definitions PR. I'll split that off tonight.

SamuelSchlesinger:
Created a new PR with just the definitions so we can merge things iteratively: https://github.com/leanprover/cslib/pull/492

SamuelSchlesinger:
Closing in favor of #492, lets move discussion to there. These proofs will hopefully be somewhat generalized anyways with the new definitions.

PR #495, opened 2026-04-16
feat(Cryptography/SecretSharing): Shamir's secret sharing

SamuelSchlesinger:
Added a general definition of a secret sharing protocol along with privacy definitions: view indistinguishability and perfect privacy. Implemented Shamir's secret sharing as an instance, then proved view indistinguishability and perfect privacy of translation invariant tail polynomial distributions. Specialized to the uniform tail polynomial distribution as that is the typical setting.

A few more PMF utilities were needed, I am planning to upstream those to Mathlib along with the existing ones from perfectly secret encryption schemes.


SamuelSchlesinger:
In https://github.com/leanprover-community/mathlib4/pull/37938#issuecomment-4258931020, I learned that in Mathlib they are deprecating PMF. I will change this to use measure theory and will swap out the work in Perfect Secrecy as well.

SamuelSchlesinger:
After some discussion on the Zulip, I am going to delay before figuring out exactly what to do. If someone ends up reviewing the cryptographic/mathematical content, it will likely still be productive as I will probably just change the spelling of the probability lemmas underneath and no more.

SamuelSchlesinger:
The discussion landed on: its fine to use PMF for now, so this PR is going to continue using it.

Cslib/Crypto/Protocols/SecretSharing/Internal/ShamirAlgebra.lean
```diff
@@ -0,0 +1,140 @@
...
+noncomputable section
+
+namespace Cslib.Crypto.Protocols.SecretSharing.Shamir.Internal
+
+open Polynomial
+
+variable {F : Type*} [Field F]
+
+/-- The secret carried by a sharing polynomial is its value at `0`. -/
+def secret (p : F[X]) : F :=
```
wrenna-robson: I feel a little like this could maybe be in the Polynomial namespace, if only to allow for dot notation.

Cslib/Crypto/Protocols/SecretSharing/Internal/ShamirAlgebra.lean
```diff
@@ -0,0 +1,140 @@
...
+  p.eval 0
+
+@[simp]
```
wrenna-robson: Hmm. This is quite an unusual simp lemma.

Cslib/Crypto/Protocols/SecretSharing/Internal/ShamirAlgebra.lean
```diff
@@ -0,0 +1,140 @@
...
+theorem secret_eq_coeff_zero (p : F[X]) : secret p = p.coeff 0 :=
+  (Polynomial.coeff_zero_eq_eval_zero p).symm
+
+/-- The tail polynomial determined by the first `n` coefficients. -/
+def tailPolynomial (n : ℕ) (coeffs : Fin n → F) : F[X] :=
```
wrenna-robson: Can't believe this doesn't already exist.

Cslib/Crypto/Protocols/SecretSharing/Internal/ShamirAlgebra.lean
```diff
@@ -0,0 +1,140 @@
...
+def sharingPolynomial (secretValue : F) (tail : F[X]) : F[X] :=
+  C secretValue + X * tail
+
+@[simp]
+theorem sharingPolynomial_eval (secretValue x : F) (tail : F[X]) :
+    (sharingPolynomial secretValue tail).eval x = secretValue + x * tail.eval x := by
+  simp [sharingPolynomial, mul_comm]
+
+@[simp]
+theorem coeff_zero_sharingPolynomial (secretValue : F) (tail : F[X]) :
```
wrenna-robson: My gut is this shouldn't itself be a simp lemma.

Cslib/Crypto/Protocols/SecretSharing/Internal/ShamirAlgebra.lean
```diff
@@ -0,0 +1,140 @@
...
+    (sharingPolynomial secretValue tail).coeff 0 = secretValue := by
+  simp [sharingPolynomial]
+
+theorem secret_sharingPolynomial (secretValue : F) (tail : F[X]) :
```
wrenna-robson: And this should be.

Cslib/Crypto/Protocols/SecretSharing/Internal/ShamirAlgebra.lean
```diff
@@ -0,0 +1,140 @@
...
+    secret (sharingPolynomial secretValue tail) = secretValue := by
+  simp [secret_eq_coeff_zero]
+
+/-- If the tail polynomial has degree `< n`, then the sharing polynomial has
+degree `< n + 1`. -/
+theorem degree_sharingPolynomial_lt_succ (secretValue : F) (tail : F[X]) {n : ℕ}
```
wrenna-robson: I wonder if starting with the natDegree would make this less overwrought (needs preconditions).

Cslib/Crypto/Protocols/SecretSharing/Internal/ShamirAlgebra.lean
```diff
@@ -0,0 +1,140 @@
...
+    (sharingPolynomial secretValue (tailPolynomial n coeffs)).degree <
+      (n + 1 : WithBot ℕ) :=
+  degree_sharingPolynomial_lt_succ secretValue (tailPolynomial n coeffs)
+    (tailPolynomial_degree_lt n coeffs)
+
+variable {ι : Type*} [DecidableEq ι]
+
+/-- Reconstruct the secret from a coalition's share values by interpolating the
+unique low-degree polynomial that matches them. -/
+def reconstruct (s : Finset ι) (x : ι → F) (σ : s → F) : F :=
```
wrenna-robson: This doesn't feel quite right to me. You only ever evaluate x on s in this definition, no? I think you want [Fintype ι] and then (x  σ : ι → F), take s as Finset.univ.

wrenna-robson: (I seem to recall when we redesigned this API, it was to allow for the fact that you might use some but not all of your nodes for interpolation - technically you could have an infinite number of nodes and just use a finite subset.)

Cslib/Crypto/Protocols/SecretSharing/Internal/ShamirAlgebra.lean
```diff
@@ -0,0 +1,140 @@
...
+  secret <| Lagrange.interpolate s.attach (fun i : s => x i) σ
+
+/-- If a polynomial of degree `< s.card` matches the observed share values on
+the coalition `s`, reconstruction recovers its secret. -/
+theorem reconstruct_eq_secret_of_eval_eq
+    {s : Finset ι} {x : ι → F} {σ : s → F} {p : F[X]}
+    (hx : Set.InjOn x s)
+    (hvalues : ∀ i : s, σ i = p.eval (x i))
```
wrenna-robson: surely this says σ = p.eval ∘ x ∘ Finset.val or something? Which I mean, you really want to just then get rid of sigma.

Cslib/Crypto/Protocols/SecretSharing/Internal/ShamirAlgebra.lean
```diff
@@ -0,0 +1,140 @@
...
+  simpa [reconstruct, secret] using congrArg (fun q : F[X] => q.eval 0) hp.symm
+
+/-- Reconstruction succeeds on the values of a Shamir sharing polynomial once
+the coalition is large enough. -/
+theorem reconstruct_sharingPolynomial_eq_secret
+    {s : Finset ι} {x : ι → F} {secretValue : F} {tail : F[X]}
+    (hx : Set.InjOn x s)
+    (hdeg : (sharingPolynomial secretValue tail).degree < s.card) :
+    reconstruct s x
+      (fun i : s => (sharingPolynomial secretValue tail).eval (x i)) = secretValue := by
```
wrenna-robson: Again I feel you are gaining little by s being a finset.

Cslib/Crypto/Protocols/SecretSharing/Basic.lean
```diff
@@ -0,0 +1,107 @@
...
+
+variable {Secret Randomness Party Share : Type*} [DecidableEq Party]
+
+/-- View indistinguishability implies posterior privacy. -/
+theorem perfectlyPrivate_of_viewIndist
+    (scheme : Scheme Secret Randomness Party Share)
+    (h : scheme.ViewIndist) :
+    scheme.PerfectlyPrivate := by
+  intro s hs secretDist v hv
+  exact Cslib.Crypto.Protocols.PerfectSecrecy.PMFUtilities.posteriorDist_eq_prior_of_outputIndist
```
wrenna-robson: I do think this namespaces are a bit long.

Cslib/Crypto/Protocols/SecretSharing/Internal/ShamirAlgebra.lean
SamuelSchlesinger: Addressed in 2f9d855. After switching reconstruction to `Polynomial.constantCoeff`, I added the semantic simp lemma `constantCoeff_sharingPolynomial`.

Cslib/Crypto/Protocols/SecretSharing/Internal/ShamirAlgebra.lean
SamuelSchlesinger: Addressed in 2f9d855 by deleting the local `secret_eq_coeff_zero` simp lemma together with the now-removed local `secret` accessor.

Cslib/Crypto/Protocols/SecretSharing/Internal/ShamirAlgebra.lean
SamuelSchlesinger: Addressed in 2f9d855 by replacing the arbitrary-`σ` theorem with `reconstruct_eq_constantCoeff_of_eval_eq`, stated directly for values `fun i => p.eval (x i)`.

Cslib/Crypto/Protocols/SecretSharing/Internal/ShamirAlgebra.lean
SamuelSchlesinger: Addressed in 2f9d855 by adding `natDegree_sharingPolynomial_le`; the `WithBot` degree theorem now converts from that helper instead of doing all the degree arithmetic inline.

Cslib/Crypto/Protocols/SecretSharing/Internal/ShamirAlgebra.lean
SamuelSchlesinger: Addressed in 2f9d855. The finite-subset use case is preserved by taking the finite index type to be the selected coalition subtype `s`, so interpolation can still use only the chosen nodes.

Cslib/Crypto/Protocols/SecretSharing/Internal/ShamirAlgebra.lean
SamuelSchlesinger: Addressed in 2f9d855 by confirming the upstream primitive here is `Polynomial.degreeLTEquiv`. I kept `tailPolynomial` as a tiny Shamir-local wrapper around it so the later Shamir code keeps the domain terminology.

Cslib/Crypto/Protocols/SecretSharing/Internal/ShamirAlgebra.lean
SamuelSchlesinger: Addressed in 2f9d855 by restating `reconstruct_sharingPolynomial_eq_secret` over an arbitrary finite index type. The scheme-level correctness proof still handles `Finset` coalitions by passing `ι := s`.

Cslib/Crypto/Protocols/SecretSharing/Internal/ShamirAlgebra.lean
SamuelSchlesinger: Addressed in 2f9d855 by refactoring the algebraic `reconstruct` primitive to use a finite index type with `(x σ : ι → F)` and `Finset.univ`; the scheme layer now instantiates it with the coalition subtype.

Cslib/Crypto/Protocols/SecretSharing/Basic.lean
SamuelSchlesinger: Addressed in 2f9d855 by shortening these references to `PerfectSecrecy.PMFUtilities.*`.

Cslib/Crypto/Protocols/SecretSharing/Internal/ShamirAlgebra.lean
SamuelSchlesinger: Addressed in 2f9d855 by removing the local Shamir `secret` accessor and using mathlib’s existing `Polynomial.constantCoeff` dot notation instead. I avoided adding a broad `Polynomial.secret` name since that terminology is protocol-specific.

Cslib/Crypto/Protocols/SecretSharing/Internal/ShamirAlgebra.lean
SamuelSchlesinger: Addressed in 2f9d855 by removing `@[simp]` from `coeff_zero_sharingPolynomial`.

fmontesi (approved):

fmontesi (approved):

PR #499, opened 2026-04-18
feat(Foundations/Data): Function view for Turing tapes

crei:
The central addition in this PR is `BiTape.get`: it allows (integer) index-based access to the Turing tape cells.

In addition, adds several lemmas that make it easier to work with `BiTape`s by viewing them as functions `ℤ → Option Symbol`. A sequence of moves and writes translates into a sequence of `Function.update` and thus maps everything into a domain that `simp` and `grind` can already nicely work with.


Cslib/Foundations/Data/BiTape.lean
```diff
@@ -78,6 +78,31 @@ def mk₁ (l : List Symbol) : BiTape Symbol :=
   | [] => ∅
   | h :: t => { head := some h, left := ∅, right := StackTape.map_some t }
 
+/-- Returns the tape symbol at positon `p` relative to the head, where
+positive numbers are right of the head and negative are left of the head. -/
+def get (t : BiTape Symbol) : ℤ → Option Symbol
+  | Int.ofNat 0 => t.head
+  | Int.ofNat (Nat.succ p') => t.right.toList[p']?.getD none
+  | Int.negSucc p' => t.left.toList[p']?.getD none
```
chenson2018: Usually the notation for these matches is used, right? I.e.

```suggestion
open scoped Int in
/-- Returns the tape symbol at positon `p` relative to the head, where
positive numbers are right of the head and negative are left of the head. -/
def get (t : BiTape Symbol) : ℤ → Option Symbol
  | 0 => t.head
  | (p' + 1 : Nat) => t.right.toList[p']?.getD none
  | -[p'+1] => t.left.toList[p']?.getD none
```

Cslib/Foundations/Data/BiTape.lean
```diff
@@ -78,6 +78,31 @@ def mk₁ (l : List Symbol) : BiTape Symbol :=
...
+
+/-- Two tapes are equal if and only if their `get` functions are equal. This allows to view
+tapes as functions `ℤ → Option Symbol`. -/
+lemma ext_get (t₁ t₂ : BiTape Symbol) (h_get_eq : ∀ p, t₁.get p = t₂.get p) :
+  t₁ = t₂ := by
+  obtain ⟨head₁, left₁, right₁⟩ := t₁
+  obtain ⟨head₂, left₂, right₂⟩ := t₂
+  have h_head : head₁ = head₂ := by simpa [get] using h_get_eq 0
+  have h_right : right₁ = right₂ := by
+    apply StackTape.ext_get
```
chenson2018: Should `StackTape.ext_get` and this theorem have `ext` attributes?

Cslib/Foundations/Data/BiTape.lean
```diff
@@ -78,6 +78,31 @@ def mk₁ (l : List Symbol) : BiTape Symbol :=
...
+    intro p
+    simpa [get] using h_get_eq p.succ
```
chenson2018: Again, is notation preferred?

```suggestion
    simpa [get] using h_get_eq (p + 1)
```

Cslib/Foundations/Data/BiTape.lean
```diff
@@ -78,6 +78,31 @@ def mk₁ (l : List Symbol) : BiTape Symbol :=
...
+  have h_left : left₁ = left₂ := by
+    apply StackTape.ext_get
+    intro p
+    simpa [get] using h_get_eq (Int.negSucc p)
+  simp only [h_head, h_left, h_right]
```
chenson2018: No need to squeeze terminal `simp`. Can also just be `grind` if you prefer.

Cslib/Foundations/Data/BiTape.lean
```diff
@@ -114,13 +139,83 @@ lemma move_left_move_right (t : BiTape Symbol) : t.move_left.move_right = t := b
...
+lemma get_move_left (t : BiTape Symbol) (p : ℤ) :
+    (t.move_left).get p = t.get (p - 1) := by
+  unfold move_left get
+  match p with
+  | Int.ofNat 0 =>
+    rw [show Int.ofNat 0 - 1 = Int.negSucc 0 from rfl]
+    simp [StackTape.head_eq_getD]
+  | Int.ofNat 1 => simp
+  | Int.ofNat (n + 2) =>
+    rw [show Int.ofNat (n + 2) - 1 = Int.ofNat (n + 1) from by grind]
```
chenson2018:
```suggestion
    rw [show Int.ofNat (n + 2) - 1 = Int.ofNat (n + 1) by lia]
```

Cslib/Foundations/Data/BiTape.lean
```diff
@@ -114,13 +139,83 @@ lemma move_left_move_right (t : BiTape Symbol) : t.move_left.move_right = t := b
...
+    simp
+  | Int.negSucc n => simp
+
+@[simp]
+lemma get_move_right (t : BiTape Symbol) (p : ℤ) :
+    (t.move_right).get p = t.get (p + 1) := by
+  unfold move_right get
+  match p with
+  | Int.ofNat n =>
+    rw [show Int.ofNat n + 1 = Int.ofNat (n + 1) from by grind]
```
chenson2018:
```suggestion
    rw [show Int.ofNat n + 1 = Int.ofNat (n + 1) by lia]
```

BoltonBailey:
Seems like a good addition! My only comment is: Should we add tags like @[ext] or `grind =` to some of these to improve integration with tactics?

Cslib/Foundations/Data/BiTape.lean
```diff
@@ -114,13 +139,83 @@ lemma move_left_move_right (t : BiTape Symbol) : t.move_left.move_right = t := b
...
+    simp
+
+@[simp]
+lemma get_optionMove (t : BiTape Symbol) (d : Option Dir) (p : ℤ) :
+    (t.optionMove d).get p = t.get (p + optionDirToInt d) := by
+  unfold optionMove optionDirToInt
+  cases d with
+  | none => simp
+  | some d =>
+    cases d <;> simp [move, show p + (-1 : ℤ) = p - 1 from by omega]
```
chenson2018: I'm not a fan of nested `simp` and `grind`. Here it seems particularly not needed because you can write

```suggestion
    cases d
    · grind [move, get_move_left]
    · grind [move, get_move_right]
```
maybe some of these should be `simp`/`grind =`?

Cslib/Foundations/Data/BiTape.lean
```diff
@@ -114,13 +139,83 @@ lemma move_left_move_right (t : BiTape Symbol) : t.move_left.move_right = t := b
...
+  | zero => simp
+  | succ n ih => simp [Function.iterate_succ_apply, ih, Int.add_assoc]
+
+@[simp]
+lemma get_move_left_iterate (t : BiTape Symbol) (n : ℕ) (p : ℤ) :
+    (move_left^[n] t).get p = t.get (p - n):= by
+  induction n generalizing t p with
+  | zero => simp
+  | succ n ih =>
+    have : p - n - 1 = p - (n + 1) := by grind
```
chenson2018:
```suggestion
    have : p - n - 1 = p - (n + 1) := by lia
```

Cslib/Foundations/Data/BiTape.lean
```diff
@@ -114,13 +139,83 @@ lemma move_left_move_right (t : BiTape Symbol) : t.move_left.move_right = t := b
...
 def write (t : BiTape Symbol) (a : Option Symbol) : BiTape Symbol := { t with head := a }
 
+@[simp]
+lemma get_write (t : BiTape Symbol) (a : Option Symbol) :
+    (t.write a).get = Function.update t.get 0 a := by
+  unfold write get Function.update
+  funext p
+  match p with
+  | Int.ofNat 0 => simp
+  | Int.ofNat (n + 1) => simp; grind
```
chenson2018: `simp; grind` is technically allowed, but not needed here
 
```suggestion
  | Int.ofNat (n + 1) => grind
```

chenson2018 (changes requested):
This looks straightforward, thanks! Just some style questions.

Cslib/Foundations/Data/BiTape.lean
crei: It seems that `grind [move]` solves the whole thing just after the `unfold` commands - maybe also because I added some more annotations.

Cslib/Foundations/Data/BiTape.lean
crei: After adding the attributes to `StackTape.ext` and `StackTape.ext_get`, I can use `ext1 p` here, but how can I be sure that the `ext` tactic uses `StackTape.ext_get` instead of `StackTape.ext`?

Cslib/Foundations/Data/BiTape.lean
```diff
@@ -78,6 +78,33 @@ def mk₁ (l : List Symbol) : BiTape Symbol :=
...
+  have h_head : head₁ = head₂ := by simpa [get] using h_get_eq 0
+  have h_right : right₁ = right₂ := by
+    apply StackTape.ext_get
+    intro p
+    simpa [get] using h_get_eq (p + 1)
+  have h_left : left₁ = left₂ := by
+    apply StackTape.ext_get
+    intro p
+    simpa [get] using h_get_eq (Int.negSucc p)
+  grind
```
chenson2018: It seems clearer to use `congr` here directly rather than have `grind` infer this:

```suggestion
  cases t₁
  congr
  · simpa [get] using h_get_eq 0
  · apply StackTape.ext_get
    intro p
    simpa [get] using h_get_eq (Int.negSucc p)
  · apply StackTape.ext_get
    intro p
    simpa [get] using h_get_eq (p + 1)
```

Cslib/Foundations/Data/BiTape.lean
```diff
@@ -114,13 +141,81 @@ lemma move_left_move_right (t : BiTape Symbol) : t.move_left.move_right = t := b
...
+  | some .right => 1
+
+@[simp, scoped grind =]
+lemma get_move_left (t : BiTape Symbol) (p : ℤ) :
+    (t.move_left).get p = t.get (p - 1) := by
+  unfold move_left get
+  match p with
+  | Int.ofNat 0 =>
+    rw [show Int.ofNat 0 - 1 = Int.negSucc 0 from rfl]
+    simp [StackTape.head_eq_getD]
```
chenson2018: For these where you have something like `rw [show ... from rfl[`, you can also just do

```suggestion
    simp [StackTape.head_eq_getD]
    rfl
```

Cslib/Foundations/Data/BiTape.lean
```diff
@@ -114,13 +141,81 @@ lemma move_left_move_right (t : BiTape Symbol) : t.move_left.move_right = t := b
...
+@[scoped grind]
+def optionDirToInt (d : Option Dir) : ℤ :=
+  match d with
+  | none => 0
+  | some .left => -1
+  | some .right => 1
+
+@[simp, scoped grind =]
+lemma get_move_left (t : BiTape Symbol) (p : ℤ) :
+    (t.move_left).get p = t.get (p - 1) := by
```
chenson2018: Fir on one line, and some parentheses can be removed:
```suggestion
lemma get_move_left (t : BiTape Symbol) (p : ℤ) : t.move_left.get p = t.get (p - 1) := by
```

Cslib/Foundations/Data/BiTape.lean
```diff
@@ -114,13 +141,81 @@ lemma move_left_move_right (t : BiTape Symbol) : t.move_left.move_right = t := b
...
+    simp [StackTape.head_eq_getD]
+  | Int.ofNat 1 => simp
+  | Int.ofNat (n + 2) =>
+    rw [show Int.ofNat (n + 2) - 1 = Int.ofNat (n + 1) by lia]
+    simp
+  | Int.negSucc n => simp
+
+@[simp, scoped grind =]
+lemma get_move_right (t : BiTape Symbol) (p : ℤ) :
+    (t.move_right).get p = t.get (p + 1) := by
```
chenson2018:
```suggestion
lemma get_move_right (t : BiTape Symbol) (p : ℤ) : t.move_right.get p = t.get (p + 1) := by
```

Cslib/Foundations/Data/BiTape.lean
```diff
@@ -114,13 +141,81 @@ lemma move_left_move_right (t : BiTape Symbol) : t.move_left.move_right = t := b
...
+  unfold move_right get
+  match p with
+  | Int.ofNat n =>
+    rw [show Int.ofNat n + 1 = Int.ofNat (n + 1) by lia]
```
chenson2018, line 183: Is this the preferred way to do these proofs in Mathlib? I rarely work with integers, but this seems very low-level.

chenson2018 (changes requested):
A couple more style comments.

Cslib/Foundations/Data/BiTape.lean
chenson2018: You're right, even if you remove both it picks up another usage of `ext`, so maybe this isn't helpful?

Cslib/Foundations/Data/BiTape.lean
crei: To be honest, I don't know. I think the conversions between natural numbers and integers makes this cumbersome (in combination with the matching). That's another reason why a view Int -> Option Symbol for tapes is so useful: It's much easier to combine with head movements (when they are converted to Int using optionDirToInt).

Cslib/Foundations/Data/BiTape.lean
```diff
@@ -54,54 +54,90 @@ structure BiTape (Symbol : Type) where
 
 namespace BiTape
 
-variable {Symbol : Type}
+open StackTape
 
-/-- The empty `BiTape` -/
-def nil : BiTape Symbol := ⟨none, ∅, ∅⟩
```
crei, line 60: I found this explicit `nil` definition a bit confusing and kept on unfolding it. I think it's better to use `∅` or `default`.

Cslib/Foundations/Data/BiTape.lean
```diff
@@ -54,54 +54,90 @@ structure BiTape (Symbol : Type) where
...
-lemma empty_eq_nil : (∅ : BiTape Symbol) = nil := rfl
+lemma empty_eq_default : (∅ : BiTape Symbol) = default := rfl
 
 /--
 Given a `List` of `Symbol`s, construct a `BiTape` by mapping the list to `some` elements
 and laying them out to the right side,
 with the head under the first element of the list if it exists.
 -/
 def mk₁ (l : List Symbol) : BiTape Symbol :=
-  match l with
```
crei, line 77: I found proofs to be easier with this definition. It helps if you are (which is most of the time) just interested in `.head`.

PR #503, opened 2026-04-19
feat: basic graph definitions

BasilRohner:
This PR introduces basic graph definitions: `Graph`, `SimpleGraph`, `DiGraph`, and `SimpleDiGraph`, along with coercions and notation that form the basis of algorithmic graph theory in CSLib.

**Design.** We intentionally diverge from [Mathlib's graph definitions](https://github.com/leanprover-community/mathlib4/blob/92a49c237223dbdc783596d5c3620b6eb1181bb2/Mathlib/Combinatorics/Graph/Basic.lean), prioritizing representations that support algorithmic reasoning. In graph algorithm design, it is common to manipulate graphs in many ways (e.g., dynamic graph operations such as adding or removing nodes/edges, contracting edges, and contracting vertex sets). As a result, we prioritize a minimal, simple definition of graphs that best aligns with how they are defined in the literature on algorithmic graph theory. In particular, we use set-based definitions of *both* the vertex and edge sets, along with minimal additional attributes, to reduce early proof obligations and select those whose proofs are closer to their textbook counterparts. 

**General structures.** Here we use `Set` everywhere, allowing statements about possibly infinite graphs uniformly without enforcing finiteness through extra axioms. We associate labels with edges for the more general multigraph notions (`Graph`, `DiGraph`) via the auxiliary `Edge` and `Arc` records, and abstract the label away for the more restricted notions (`SimpleGraph`, `SimpleDiGraph`), which use `Sym2 α` and `α × α` directly. The definitions have been tested in practice by further formalizing concepts such as walks (see the [GraphAlgorithms Walk module](https://github.com/sorrachai/GraphAlgorithms/blob/main/GraphAlgorithms/Core/Walk.lean)). However, to keep this PR concise, we defer discussion of the Walk design to a later PR.

**Comparison with PR [#427](https://github.com/leanprover/cslib/pull/427).** PR [#427](https://github.com/leanprover/cslib/pull/427) introduced `SimpleGraph` with `Finset`-based vertex and edge sets and edges via `Sym2`. Here, we use `Set` throughout, treating finiteness as an orthogonal concern rather than baking it into the type, and introduce a hierarchy comprising both general multigraph structures (`Graph`, `DiGraph`) with edge labels and specialized structures (`SimpleGraph`, `SimpleDiGraph`) without labels, since multi-edges are disallowed there.

**Main definitions.**
- `Graph`: (possibly infinite) undirected multigraph
- `SimpleGraph`: (possibly infinite) undirected graph without loops or multi-edges
- `DiGraph`: (possibly infinite) directed graph
- `SimpleDiGraph`: (possibly infinite) directed graph without loops or multi-edges

**Future work.** We plan to add fundamental graph algorithms building on these definitions.

Co-authored-by: Sorrachai Yingchareonthawornchai <sorrachai.cp@gmail.com>


ctchou:
I have some questions/comments:
(1) Does the `Graph` in this PR differ in a substantive way from the `Graph` in `Mathlib.Combinatorics.Graph.Basic`?  It seems to me that they are basically equivalent to each other.  If my understanding is correct, then can't you just define `Graph.endpoints` and `Graph.incidence` on mathlib's `Graph`, if they are the API you prefer?
(2) A `Set` that is `Set.Finite` can be converted into a `Finset` using:
https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Set/Finite/Basic.html#Set.Finite.toFinset
I know that this API is "noncomputable".  But once you obtain the `Finset` versions of `vertexSet` and `edgeSet`, you can operate on them in a completely "computable" manner.   By having both `Set` and `Finset` versions of basically the same definition, I'm afraid that you will have to repeat the statements and proofs of many theorems for the two versions.
(3) It seems to me that you can define `DiGraph` by extending `Quivers` in `Mathlib.Combinatorics.Quiver.Basic`, which is so general that I'm pretty sure it can accommodate any notion of "digraph" you have.

Shreyas4991:
Discussion thread on this PR : 
[Leanprover Zulip thread link](https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/New.20graph.20definitions/with/586402068)

Cslib/Algorithms/Lean/Graph/Graph.lean
```diff
@@ -0,0 +1,153 @@
...
+the vertex and edge sets may be infinite. -/
+structure Graph (α ε : Type*) where
+  /-- The set of vertices. -/
+  vertexSet : Set α
+  /-- The set of edges. -/
+  edgeSet : Set ε
+  /-- The endpoint map, sending each edge to its unordered pair of endpoints. -/
+  endpoints : ε → Sym2 α
+  /-- Every endpoint of an edge is a vertex. -/
+  incidence : ∀ e ∈ edgeSet, ∀ v ∈ endpoints e, v ∈ vertexSet
```
sorrachai: This may cause a junk value problem on endpoints where an endpoint outside the set must be specified. 

To avoid the junk value problem, we can define an edge object separately and define the graph directly over the edgeSet. Note that we use endpoints_id as the identifier for the multiplicity of an edge of the same endpoints 

```lean 4
structure UndirectedEdge (α β : Type*) where
  endpoints_id : β
  endpoints : Sym2 α
deriving DecidableEq

abbrev Edge := UndirectedEdge

structure UndirectedGraph (α β : Type*) where
  vertexSet : Set α
  edgeSet : Set (Edge α β)
  incidence : ∀ e ∈ edgeSet, ∀ v ∈ e.endpoints, v ∈ vertexSet
```

Cslib/Foundations/Combinatorics/Graph/Graph.lean
```diff
@@ -0,0 +1,153 @@
...
+
+/-- Typeclass for graph-like structures that have a vertex set. -/
+class HasVertexSet (G : Type*) (V : outParam Type*) where
+  /-- The vertex set of the graph. -/
+  vertexSet : G → V
+
+/-- Typeclass for graph-like structures that have an edge set. -/
+class HasEdgeSet (G : Type*) (E : outParam Type*) where
+  /-- The edge set of the graph. -/
+  edgeSet : G → E
```
eric-wieser, line 155: If you make these the canonical spelling then you need to restate all the `loopless` and `incidence` theorems in terms of them (and probably rename those fields to be `loopless'` to avoid stealing the preferred name)

eric-wieser:
> **Design.** We intentionally diverge from [Mathlib's graph definitions](https://github.com/leanprover-community/mathlib4/blob/92a49c237223dbdc783596d5c3620b6eb1181bb2/Mathlib/Combinatorics/Graph/Basic.lean), ...


Please record these design decisions in the module docstring for future readers, not just the git history / PR title.

Cslib/Algorithms/Lean/Graph/Graph.lean
```diff
@@ -0,0 +1,153 @@
...
+
+/-- A finite simple graph on `α`: finite vertex and edge sets, edges as unordered pairs of
+distinct vertices. -/
+@[grind]
+structure SimpleGraph (α : Type*) where
+  /-- The finite set of vertices. -/
+  vertexSet : Finset α
```
sorrachai: Inconsistency on the set definitions: SimpleGraph is defined using Finset, whereas Graph is defined using Set.

Cslib/Algorithms/Lean/Graph/Graph.lean
```diff
@@ -0,0 +1,153 @@
...
+  /-- The endpoint map, sending each edge to its ordered pair `(source, target)`. -/
+  endpoints : ε → α × α
+  /-- Both endpoints of every edge are vertices. -/
+  incidence : ∀ e ∈ edgeSet, (endpoints e).1 ∈ vertexSet ∧ (endpoints e).2 ∈ vertexSet
+
+/-- A finite simple directed graph on `α`: finite vertex and edge sets, edges as ordered
+pairs of distinct vertices. -/
+structure SimpleDiGraph (α : Type*) where
+  /-- The finite set of vertices. -/
+  vertexSet : Finset α
```
sorrachai: Finset vs. Set

Cslib/Algorithms/Lean/Graph/Graph.lean
```diff
@@ -0,0 +1,153 @@
...
+  loopless : ∀ e ∈ edgeSet, ¬ e.IsDiag
+
+/-- A directed graph on vertex type `α` with edges indexed by `ε`. The edge set is abstract;
+ordered endpoints are carried by a separate map. Parallel edges and loops are permitted,
+and both the vertex and edge sets may be infinite. -/
+structure DiGraph (α ε : Type*) where
+  /-- The set of vertices. -/
+  vertexSet : Set α
+  /-- The set of edges. -/
+  edgeSet : Set ε
```
sorrachai: Avoidable junk value problem using the suggested definitions of Arc (similarly to Edge).

Cslib/Algorithms/Lean/Graph/Graph.lean
BasilRohner: Thank you for pointing this out and for the discussion in the Zulip thread. I've updated the definitions in line with your proposed solution to avoid this issue.

Cslib/Algorithms/Lean/Graph/Graph.lean
BasilRohner: Thank you for pointing this out. This design decision was intentional, with the goal of arriving at a definition of `SimpleGraph` that serves as a canonical graph definition encompassing the most common usage. That said, I agree this introduces an inconsistency, and I've updated it to `Set α`. Finiteness can be supplied as a separate hypothesis.

Cslib/Foundations/Combinatorics/Graph/Graph.lean
BasilRohner: Thanks for pointing this out. I've restated the `loopless` and `incidence` theorems in terms of the new typeclasses and renamed the original fields to `loopless'` to free up the preferred name.

Cslib/Foundations/Combinatorics/Graph/Graph.lean
```diff
@@ -16,101 +16,134 @@ public import Mathlib.Data.Sym.Sym2
...
   /-- The set of edges. -/
-  edgeSet : Set ε
-  /-- The endpoint map, sending each edge to its unordered pair of endpoints. -/
-  endpoints : ε → Sym2 α
-  /-- Every endpoint of an edge is a vertex. -/
-  incidence : ∀ e ∈ edgeSet, ∀ v ∈ endpoints e, v ∈ vertexSet
-
-/-- A finite simple graph on `α`: finite vertex and edge sets, edges as unordered pairs of
-distinct vertices. -/
+  edgeSet : Set (Edge α β)
```
eric-wieser, line 87: Another option here would be `Sym2 a -> Set b`, which gives the set of edge labels for a given pair of vertices, and the empty set for disconnected nodes

BasilRohner: Thanks for the suggestion! We'd prefer to keep the set-based definition since it's closer to the informal `G = (V, E)` presentation, whereas your version is closer to the relational definition of `Mathlib.Combinatorics.Graph`.

sorrachai (approved):
I am happy with the current definitions now. After a long discussion in the Zulip chat, the design is not bad, but it is similar to Mathlib's definitions of simple undirected graphs, whose development of set-based definitions is still in its early stages. We are happy to collaborate as the library grows. 

Aside from the fragmentation of definitions in mathlib, one key principle of cslib is efficient computation, and we may redesign the definitions in the future to support efficient computation and reasoning about complexity. We prefer to have all graph definitions (undirected/directed and simple/non-simple) in one place, and to build other objects on top of the proposed graph definitions with efficient computation in mind.
 
P.S. This PR was a product of the Lean [hackathon](https://eth-its.ethz.ch/activities/lean-hackathon---eth-its-zurich.html) event at ETH Zurich. More basic primitives for graph algorithms will follow in subsequent PRs.

chenson2018 (dismissed):
I see that @sorrachai has approved this, but I really do not feel that the Zulip thread has converged on accepting this PR, unless I am missing some other discussion in the graph theory channel.

This is decidedly not my area of expertise, but I am very worried about such a large deviation from very active development in Mathlib. I'd rather take this slowly and have a definite consensus or lack thereof rather than merge something too quickly.

sorrachai:
> I see that @sorrachai has approved this, but I really do not feel that the Zulip thread has converged on accepting this PR, unless I am missing some other discussion in the graph theory channel.
> 
> This is decidedly not my area of expertise, but I am very worried about such a large deviation from very active development in Mathlib. I'd rather take this slowly and have a definite consensus or lack thereof rather than merge something too quickly.

I appreciate your concern. I would like to understand what happens if there is no consensus on this.

Shreyas4991:
I disagree that these definitions are computationally efficient, far from it. An efficient defintion would use lean arrays to represent neighbouring sets, better yet Vectors. This definition is merely a representational change which can be obtained by constructors on mathlib definitions.

BasilRohner:
**Summary.** This pulls together the points we've made in the Zulip thread and earlier reviews. The design has two main ideas:

1. **Embedded-set structure.** Vertex and edge sets are represented as `vertexSet : Set α` and `edgeSet : Set β`, and not using a type. A prime example of what this buys: subgraphs and graph operations stay inside the same type. A subgraph of `G : Graph α β` is itself a `Graph α β`, related by `≤`. Induced subgraph, edge and vertex deletion, and contraction are all maps `Graph α β → Graph α β`. No separate `Subgraph G` type, no `coe`/`spanningCoe` pair to choose between, no parallel lattice and API to keep in sync. The Mathlib workaround (`SimpleGraph.Subgraph` with its two coercions to `SimpleGraph ↥verts` and `SimpleGraph V`) has been a known pain point for years. 
  
2. **Consistency across the family.** All four graph types (`Graph`, `SimpleGraph`, `DiGraph`, `SimpleDiGraph`) use the same setup: embedded set-based vertex and edge sets, shared `HasVertexSet`/`HasEdgeSet` typeclasses, shared `V(G)`/`E(G)` notation, and forgetful maps (`SimpleGraph.toGraph`, `SimpleDiGraph.toDiGraph`) that carry properties from the general types down to the specialised ones. Walk machinery is a prime example: `VertexSeq`, `IsWalk`, `loopErase`, `takeUntil`, `dropUntil`, `rerootCycle` get defined once and reused across all four. Mathlib's `SimpleGraph.Walk` is indexed by a specific `SimpleGraph V` and doesn't transfer to `Graph α β`. Mathlib's graph ecosystem doesn't have this uniformity yet: `Mathlib.Combinatorics.Graph.Basic` (May 2025) uses embedded sets for multigraphs, while the older `SimpleGraph V` stays type-as-vertex-set, with no canonical map between them.
  
**On compatibility with Mathlib.** From what I understand, the concern raised so far is mostly not the design itself but the divergence from Mathlib's current API. We'd like eventual alignment too. Mathlib is already moving this way (`Graph α β` and a recent PR to add `vertexSet` to `Digraph` made by @Shreyas4991  are already steps in that direction), but extending the same approach to `SimpleGraph V`, for instance by adding a `vertexSet`, field may require serious refactoring. A lot of API rests on the type-as-vertex-set design, and every change has to be weighed against it. The new `Graph α β` is also still being filled in and refined for the same reason. 

Introducing the definitions prevents stalling and enables active development of graph theory on the CSLib side without waiting for Mathlib's API to adapt. We hope Mathlib will push its graph definitions in this direction over time, though that will mean significant adaptation of existing APIs given the accumulated technical debt. If it does, alignment should be straightforward. If it doesn't, we still think the embedded-set design is much better suited to algorithmic graph theory. Automata in CSLib are a close precedent: developed in parallel with Mathlib's, the two have coexisted productively, each handling the use cases best suited to its scope.

Cslib/Foundations/Combinatorics/Graph/Graph.lean
```diff
@@ -0,0 +1,212 @@
...
+The corresponding `Coe` instances are registered.
+-/
+
+namespace Cslib.Algorithms.Lean
+variable {α β : Type*}
+
+/-- An undirected edge with a label of type `β` and an unordered pair of endpoints. -/
+structure Edge (α β : Type*) where
+  /-- The edge label, used to distinguish parallel edges. -/
+  edgeLabel : β
```
sorrachai: edgeLabel is misleading. The labeling is intended for a specific pair of points.  I would name it "endpointsLabel".

Cslib/Foundations/Combinatorics/Graph/Graph.lean
```diff
@@ -0,0 +1,212 @@
...
+  /-- The unordered pair of endpoints. -/
+  endpoints : Sym2 α
+deriving DecidableEq
+
+/-- A directed edge with a label of type `β` and an ordered pair of endpoints. -/
+structure DiEdge (α β : Type*) where
+  /-- The edge label, used to distinguish parallel edges. -/
+  edgeLabel : β
```
sorrachai: Same comment here.

Cslib/Foundations/Combinatorics/Graph/Graph.lean
```diff
@@ -0,0 +1,212 @@
...
+/-- An undirected edge with a label of type `β` and an unordered pair of endpoints. -/
+structure Edge (α β : Type*) where
+  /-- The edge label, used to distinguish parallel edges. -/
+  edgeLabel : β
+  /-- The unordered pair of endpoints. -/
+  endpoints : Sym2 α
+deriving DecidableEq
+
+/-- A directed edge with a label of type `β` and an ordered pair of endpoints. -/
+structure DiEdge (α β : Type*) where
```
sorrachai: `DiEdge` should be `Arc`.

Cslib/Foundations/Combinatorics/Graph/Graph.lean
sorrachai:
```suggestion
  endpointsLabel : β
```

Cslib/Foundations/Combinatorics/Graph/Graph.lean
sorrachai:
```suggestion
  endpointsLabel : β
```

Cslib/Foundations/Combinatorics/Graph/Graph.lean
```diff
@@ -0,0 +1,212 @@
...
+prioritizing representations that support algorithmic reasoning. In graph algorithm design,
+it is common to manipulate graphs dynamically (adding or removing nodes/edges, contracting
+edges, etc.). We therefore use set-based definitions for both vertex and edge sets with
+minimal additional axioms, reducing early proof obligations and keeping proofs closer to
+their textbook counterparts.
+
+## Main definitions
+
+* `Edge α β`: an undirected edge with a label of type `β` and endpoints as a `Sym2 α`.
+* `DiEdge α β`: a directed edge with a label of type `β` and endpoints as `α × α`.
```
sorrachai: Use `Arc` instead of `DiEdge`

sorrachai:
Based on the online CSlib meeting today (28-05-2026), we agree to move forward with CSlib graphs using the current PRs, with a minor change to the naming. One important note is that we are happy to migrate/reuse appropriate definitions from Mathlib once their graph library (e.g., the graphLike class and the hierarchy of graph classes) becomes mature.  

One main question was whether the naming of graphs and digraphs in the PR was appropriate. 

There are two options:
1. UndiGraph, DiGraph (and the simple versions of SimpleUndiGraph, SimpleDiGraph). 
2. Graph, DiGraph (and the simple version of SimpleGraph, SimpleDiGraph). (The current choice in the PR)

Option 1 would be an unambiguous definition of what we mean by graphs. On the other hand, Option 2 is standard in a graph theory textbook. (For example, [Graphs & Digraphs - 7th Edition - Gary Chartrand](https://www.routledge.com/Graphs--Digraphs/Chartrand-Jordon-Vatter-Zhang/p/book/9781032606989)).  

In this case, I am inclining towards the second option. 

Other than my suggestions/comments in this PR, I don't have any other reservations.

chenson2018:
> Based on the online CSlib meeting today (28-05-2026), we agree to move forward with CSlib graphs using the current PRs, with a minor change to the naming. One important note is that we are happy to migrate/reuse appropriate definitions from Mathlib once their graph library (e.g., the graphLike class and the hierarchy of graph classes) becomes mature.

This seems very much at odds with what has been discussed on Zulip, and I find referring to Mathlib's graph library as not "mature" misleading.

What happened to discussion around points like [this comment](https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/New.20graph.20definitions/near/587764498) where @BasilRohner stated that they thought they could get what is desired here by defining new "alternative constructor" API that didn't require actually duplicating definitions or [this comment](https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/New.20graph.20definitions/near/586415086) about defining new types but proving a proof of an injective map into the existing `SimpleGraph`?

Shreyas4991:
@sorrachai and @fmontesi Notwithstanding other discussions, Might I also suggest moving this to a subfolder of foundations in this PR, since technically this PR is defining mathematical foundations for CS?

sorrachai:
Hi Shreyas4991, that's fine. We can discuss its home folder in a separate thread.

Hi @chenson2018, thanks for the comments. I am referring to the [roadmap](https://leanprover.zulipchat.com/#narrow/channel/252551-graph-theory/topic/Formalization.20of.20tree.20decompositions/near/593830719) for GraphLike, which may take some time until it fully develops. Currently, Mathlib has 4 definitions of graphs: [Quiver](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Combinatorics/Quiver/Basic.html#Quiver), [DiGraph](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Combinatorics/Digraph/Basic.html#Digraph), [Graph](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Combinatorics/Graph/Basic.html#Graph), [SimpleGraphs](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Combinatorics/SimpleGraph/Basic.html) (and maybe more in automata?). I know that they are working on having a more unified definition via the notion of [GraphLike]((https://leanprover.zulipchat.com/#narrow/channel/252551-graph-theory/topic/Formalization.20of.20tree.20decompositions/near/593830719)).   

I am happy to reuse whatever it has in Mathlib. However, as it stands, they only have a simple undirected multi-graph that is a set-based definition that we can potentially reuse. These suggestions from the [discussion](https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/New.20graph.20definitions/near/587764498) are valid. However, other than that, they do not have usable directed graphs, directed simple graphs, or undirected simple graphs, which they are working on right now. I think having all those definitions coherently in one place in CSLib is easier, especially for a new contributor. It is extremely simple. My working group downstream is quite happy with the definitions so far. The graph definitions in Mathlib are designed to maximize generality; they are currently quite complicated for newcomers who want to contribute to CSlib.  

My point is that I am willing to wait for Mathlib until all fragmentations in Mathlib are cleanly resolved before getting involved in the complicated process. I believe in the long run, the best version of the graph will definitely live in Mathlib, and we are happy to migrate to that. 

I encourage you to join us in the online cslib meeting for further elaboration and discussion.

chenson2018:
> I encourage you to join us in the online cslib meeting for further elaboration and discussion.

I did join last week, but these are pretty early for my current timezone. It would be helpful for these discussions to make their way back to Zulip, especially when there is already an ongoing thread. Would you be able to put a summary in the thread on graph definitions?

sorrachai:
Sure, I added the reference to the summary of today's discussion in that [thread](https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/New.20graph.20definitions/near/598397291)[#CSLib > New graph definitions @ 💬](https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/New.20graph.20definitions/near/598397291).

BasilRohner:
Renamed `DiEdge` to `Arc` and `edgeLabel` to `endpointsLabel`, and moved it to `Cslib/Foundations/Combinatorics/Graph`. 

On naming: I'd prefer to keep `Graph`/`SimpleGraph` over `UndiGraph`/`SimpleUndiGraph` since it's more concise and matches most of the literature. Given that the definitions live in one file and the directed variants are explicitly named, I don't think there's real ambiguity. I'm open to change it if the sentiment is very strong though.

sorrachai:
@chenson2018 I notice that you requested changes. Can you elaborate what changes you are looking for?

Shreyas4991:
I don't think any decision was made to merge this PR. Quite the opposite per the thread 

https://leanprover.zulipchat.com/#narrow/channel/605128-CSLib.3A-PR-reviews/topic/.23503.3A.20basic.20graph.20definitions/near/610631732

chenson2018 (changes requested):
I still hold the same objections to this PR. In short, I am unclear on: the motivation for duplication of Mathlib, what fundamentally differs in the formalization of graphs this is meant to support, and the responses or lack thereof to the many suggestions on Zulip for bridging this gap. I think this leads to an unnecessary fracture in the ecosystem for graph formalization, an area of active development in Mathlib that has formalized serious results.

If this is the final form of this PR, I am not sure that we'll reach a place of agreement. I'd like to hear from @fmontesi on this, perhaps on the Zulip where it can be easier to communicate.

Shreyas4991:
I second Chris. None of the critical points have received adequate response or been addressed so far. So I see no reason to move on. 

I would actually recommend closing this PR, abandoning the effort to needlessly duplicate Graphs, and instead working within mathlib's process. Graph algorithms can be built perfectly within the existing API and ongoing mathlib work.

sorrachai:
Hi @chenson2018, thanks for raising this — I appreciate the concern, and I want to be clear that I'm not proposing we avoid reusing Mathlib.

My worry is timing. Several of the pieces we'd be depending on are still WIP upstream, and I don't think it's realistic to block on them landing. I'd suggest we move forward with what we have here and reconcile with Mathlib once those pieces stabilize — that seems like a normal, healthy sequence to me, and it's easier to do once the library has grown enough to show us what the right shared abstractions actually are.

The practical concern is that this PR has been open for several months and we still don't have the definitions in. That cost is real, and it compounds — other work is waiting on these. Is there a specific risk you see in landing now and reconciling later that I might be underweighting? If it's something we could address with a TODO or a follow-up issue, I'd be glad to do that.

sorrachai:
I want to emphasize that there isn't much duplication at the moment, as Graph in Mathlib is undergoing a major refactoring.

Shreyas4991:
The issue is, going ahead still fractures the library ecosystem and makes repair harder for everyone, contributors, downstream users, and maintainers alike. Even if it takes time, it is best to wait for the refactor and build on what is available until then. Imho, the best thing to do is to close this PR, and speed up the Graph PRs and focus on the addition of the necessary combinatorial lemmas in mathlib. And I wish to emphasize again, that graph algorithms has absolutely no need for these changes.

chenson2018:
> I want to emphasize that there isn't much duplication at the moment, as Graph in Mathlib is undergoing a major refactoring.

This argument doesn't make sense to me. At the end of whatever refactoring is happening in Mathlib, there will still be these definitions in Mathlib, and they would still be duplicated here. If anything, this would be the best time for us to engage with the Mathlib folks and ensure that the refactor lands in a way that both libraries are happy with these definitions. Otherwise, it seems like a real possibility that we end up having the same conversation again in some months, with the difference being that Mathlib has already gone through one major refactoring and would have even less appetite to do so again.

sorrachai:
Sure, happy to engage with the Mathlib folks on this. The conclusion so far is to please wait for Mathlib's definitions until the refactoring is complete. One thing I'd ask us to weigh alongside that, though: there's a cost on the other side too. We've had nothing in CSLib on this for months, and that delay has its own price — other work is waiting on these definitions. I'd like us to factor that in rather than treating "wait for upstream" as the default with no downside. The design space is vast, and cslib focus might differ, and we inevitably diverge in the design at some point. We want to take advantage of the fact that we are relatively small and fast to develop graph algorithms independently.  I think automata theory has had this issue before, and cslib decided to use cs version, and it has been great so far.

Shreyas4991:
I think we can expect progress in CSLib on the algorithms side soon once the algorithms pr issue is resolved, for which a meeting will soon be arranged. There is 7000-8000 lines of Algolean waiting to be upstreamed. Making progress on CSLib algorithms is, therefore, orthogonal to this duplication.  

Splintering the graph library for the sake of short-term progress is not a wise choice. In fact the added burden of maintenance and back-and-forth with mathlib will hamper progress even more in the medium term.

chenson2018:
> I think automata theory has had this issue before, and cslib decided to use cs version, and it has been great so far.

As I've discussed before, I do not think this is analogous. This was a very inactive area of Mathlib, the introduction of `LTS` was strictly more general, and some Mathlib maintainers are of the opinion that this material should ultimately be downstreamed. None of these are true in in the case of graphs or this PR.

ctchou:
I don't have an opinion about the issues discussed above.  But I think it would help a lot if there is a summary of the differences and similarities of the various types of graphs proposed here and those existing in mathlib, perhaps with justifications for the proposed differences.

Shreyas4991:
> I don't have an opinion about the issues discussed above.  But I think it would help a lot if there is a summary of the differences and similarities of the various types of graphs proposed here and those existing in mathlib, perhaps with justifications for the proposed differences.

Unfortunately, in this particular PR,I don't think it helps at all. The duplication here is clear. Mathlib already has definitions for simple graphs, simple digraphs, and Graphs in general. This definition is easily subsumed within the mathlib definitions and working with them is the correct way forward even if it takes time. 


Further, I still don't think any justification has been offered for the claim that somehow this PR is the reason for lack of progress in algorithms. 

To clarify: Justification for this would usually take the form of a Zulip thread, where someone presents their attempt to define simple algorithms under existing definitions. Then respondents experiment with the code and explain the right way to fix the problems they face. If a change is needed in mathlib definitions, they are directed towards the right PRs. This process has simply not happened here, This is despite repeated requests on my part to show where the obstacle is in code.

sorrachai:
> I don't have an opinion about the issues discussed above. But I think it would help a lot if there is a summary of the differences and similarities of the various types of graphs proposed here and those existing in mathlib, perhaps with justifications for the proposed differences.

Thank you for the comment. Let's focus on the multi-graphs, which have the most critical design differences. We want to store and ensure computability of the graph definition. 

The PR's two multigraph structures, Graph α β and DiGraph α β, take the shape vertexSet: Set α plus edgeSet: Set (Edge α β) (resp. Set (Arc α β)) plus an incidence relation, where Edge/Arc are records bundling endpointsLabel : β with endpoints : Sym2 α (resp. α × α). 

This diverges from Mathlib's relational Graph α β, where IsLink : β → α → α → Prop plus eq_or_eq_of_isLink_of_isLink pins each edge name to at most one pair of ends.  An edge name is an opaque element of β, and recovering its endpoints means extracting a witness from the Prop-valued IsLink, which is erased at runtime.

The PR's version buys computable endpoint access, since e.endpoints is a projection that runs while recovering ends from a Mathlib edge name needs a noncomputable witness extraction. (For this discussion, let's assume that the vertex set is finite.) 

I am happy to extend simple graphs from Mathlib's definitions since they do not suffer from this issue.

Shreyas4991:
> Let's focus on the multi-graphs, which have the most critical design differences. We want to store and ensure computability of the graph definition.

This is a meaningless claim. See prior discussion on graph constructors in the Zulip. 

Further, this is an issue to be discussed and addressed in the context of mathlib's graph refactoring.

sorrachai:
Please focus on the technical discussion and give a technical comment. Why is it meaningless? If you want to point to a reference, please point to a specific statement that discusses this.

Shreyas4991:
> Please focus on the technical discussion. Why is it meaningless? If you want to point to a reference, please point to a specific statement that discusses this. 

I was trying to avoid a circular discussion here since the Zulip thread already included a long discussion on graph constructors. But since you insist 

1. You don't need defintions to be computable for algorithms on graphs. Extracting an adjacency list using suitable constructors  is sufficient. You only need to count the operations on elements of the list when you perform data structure (stacks/queues/heaps etc) insertions, reads, updates, and deletions. How you get the adjacency list is of no consequence since the adjacency list representation is given as input to algorithms. It's available for free as far as the algorithm is concerned. You can simply take Graph and extract an adj list by choice and just supply this list to the algorithm. Conversely you can trivially convert an adj list back to Graph using another constructor. 

2. This is not a valid reason to start this duplicating PR.

Shreyas4991:
Another point. In order to substantiate your point about the need for computability, I'd be grateful for a graph algorithms  example code that is currently failing in some way. I have been asking for this for months and haven't heard back. It would greatly help the discussion.

chenson2018:
> This is a meaningless claim. See prior discussion on graph constructors in the Zulip.

@Shreyas4991 When you phrase things so aggressively, it is counterproductive to having a discussion. I've consistently agreed with you on technical points here, but when you chose this style of communication, it derails the conversation.

chenson2018:
@sorrachai Has there been any thread you've started in the mathlib or graph theory channels to discuss this? I'd really appreciate if you can do this. My view is that while there's this refactoring of graphs happening in Mathlib is the perfect time for you to be included in the discussion. If already open Mathlib PRs address your needs, I'm happy for us to vendor them ahead of time temporarily if timeliness is the issue. If they still don't then let's get this sorted out now so that we can eventually use the Mathlib definitions. What I see is the worst case scenario if that this is merged and CSLib indefinitely has a duplicate definition of graphs. Now is the time to discuss what is needed, as after one large refactor there is very unlikely to be further appetite for changing definitions.

sorrachai:
Sure, another attempt [here](https://leanprover.zulipchat.com/#narrow/channel/252551-graph-theory/topic/A.20Set-based.20Multigraph.20Definition/near/616696846)[#graph theory > A Set-based Multigraph Definition @ 💬](https://leanprover.zulipchat.com/#narrow/channel/252551-graph-theory/topic/A.20Set-based.20Multigraph.20Definition/near/616696846).

sorrachai:
Based on the discussion, the current definition of Graph in Mathlib has noncomputable component (e.g., extracting an endpoint from an edge requires a choice function), whose fix requires modifying their definition. We are discussing if Mathlib is interested in fixing this or not.

Shreyas4991:
> Based on the discussion, the current definition of Graph in Mathlib has noncomputable component (e.g., extracting an endpoint from an edge requires a choice function), whose fix requires modifying their definition. We are discussing if Mathlib is interested in fixing this or not.

I don't think any agreement has been reached that `noncomputable` is a problem at all. People have offered solutions for how to make it computable if that were  to be a requirement.

Shreyas4991:
I made a demo PR to show that graph algorithms can indeed be formalized with Mathlib graph plus AdjList constructors. #804 . The goal is to present evidence against the claim that the duplication in this PR is somehow necessary for graph algorithms verification. I have done this in TimeM despite my well-known reservations about directly using writer monads for algorithms.

sorrachai:
Thanks for the demo. It is important to understand that here we are talking about a graph as a specification layer. What you show is an implementation of it that can be converted into a graph specification (with a non-computable specification signature). I am asking for a computable graph specification, which Mathlib currently does not support, and there is no good reason to sacrifice computability at the moment.

Shreyas4991:
> Thanks for the demo. It is important to understand that here we are talking about a graph as a specification layer. What you show is an implementation of it that can be converted into a graph specification (with a non-computable specification signature). I am asking for a computable graph specification, which Mathlib currently does not support, and there is no good reason to sacrifice computability at the moment. 


I am asking why you need this "computable"(in the lean sense) spec. I have shown that this doesn't block algorithms. Further in this PR (503) you are using `Set` and `Sym2` which are definitely not always computable (even in the Turing sense). 

EDIT: I have actually also demoed how to write adjlists explicitly as a structure and write constructors to and from Graph. #804 uses this API and everything stays lean computable. #805 works without any extra api but uses noncomputable. These are merely two design possibilities out of many. All of them work without duplicating graph API.

ctchou:
> The PR's two multigraph structures, Graph α β and DiGraph α β, take the shape vertexSet: Set α plus edgeSet: Set (Edge α β) (resp. Set (Arc α β)) plus an incidence relation, where Edge/Arc are records bundling endpointsLabel : β with endpoints : Sym2 α (resp. α × α).
> 
> This diverges from Mathlib's relational Graph α β, where IsLink : β → α → α → Prop plus eq_or_eq_of_isLink_of_isLink pins each edge name to at most one pair of ends. An edge name is an opaque element of β, and recovering its endpoints means extracting a witness from the Prop-valued IsLink, which is erased at runtime.

@sorrachai   I'm not sure I understand this point.  I think the definition of `Graph` in this PR allows the same edge label to be applied to two different edges, where "different" means that the edges have different sets of endpoints.  Here's a concrete example, where two different edges `e1` and `e2` share the same edge label `0`: 
```
def e1 : Edge (Fin 3) (Fin 1) where
  endpointsLabel := 0
  endpoints := Sym2.mk 0 1

def e2 : Edge (Fin 3) (Fin 1) where
  endpointsLabel := 0
  endpoints := Sym2.mk 1 2

def g : Graph (Fin 3) (Fin 1) where
  vertexSet := Set.univ
  edgeSet := {e1, e2}
  incidence' := by grind
```
I think that the purpose of `eq_or_eq_of_isLink_of_isLink` in mathlib's `Graph` is precisely to outlaw such cases.  Do you really want to allow them in your definition?

By the way, @BasilRohner  Please do not use the `main` branch for PR.  It can cause all sorts of problems for reviewers.

fmontesi:
> If this is the final form of this PR, I am not sure that we'll reach a place of agreement. I'd like to hear from @fmontesi on this, perhaps on the Zulip where it can be easier to communicate.

Dear all,
I had lost a piece of this discussion since there were parallel ones on Zulip. Sorrachai and I have been discussing whether what's needed can be covered by extending + introducing some abbreviations for mathlib's graph structures and CSLib's `LTS` (for the case of directed graphs with labelled edges, if there is no better Mathlib equivalent) -- I made a sketch. But this is still unclear. In particular, the bit about computability seems important to @sorrachai and I'd like to understand better what is meant and needed there.

fmontesi:
> > The PR's two multigraph structures, Graph α β and DiGraph α β, take the shape vertexSet: Set α plus edgeSet: Set (Edge α β) (resp. Set (Arc α β)) plus an incidence relation, where Edge/Arc are records bundling endpointsLabel : β with endpoints : Sym2 α (resp. α × α).
> > This diverges from Mathlib's relational Graph α β, where IsLink : β → α → α → Prop plus eq_or_eq_of_isLink_of_isLink pins each edge name to at most one pair of ends. An edge name is an opaque element of β, and recovering its endpoints means extracting a witness from the Prop-valued IsLink, which is erased at runtime.
> 
> @sorrachai I'm not sure I understand this point. I think the definition of `Graph` in this PR allows the same edge label to be applied to two different edges, where "different" means that the edges have different sets of endpoints. Here's a concrete example, where two different edges `e1` and `e2` share the same edge label `0`:
> 
> ```
> def e1 : Edge (Fin 3) (Fin 1) where
>   endpointsLabel := 0
>   endpoints := Sym2.mk 0 1
> 
> def e2 : Edge (Fin 3) (Fin 1) where
>   endpointsLabel := 0
>   endpoints := Sym2.mk 1 2
> 
> def g : Graph (Fin 3) (Fin 1) where
>   vertexSet := Set.univ
>   edgeSet := {e1, e2}
>   incidence' := by grind
> ```
> 
> I think that the purpose of `eq_or_eq_of_isLink_of_isLink` in mathlib's `Graph` is precisely to outlaw such cases. Do you really want to allow them in your definition?

Following up from my previous message, since this example nails it pretty nicely: if that's supposed to be allowed and two vertices can be connected by at most one edge for any single label, then I think we're looking at an extension of `LTS` or something similar.

Shreyas4991:
@fmontesi : the basic question is why is computability "needed". I have made demo PRs showing two different designs with and without computability for graph algorithms. This "need" for computability needs to be explained with strong reasons. 

Also multi edges often come up in algorithms when vertices are contracted. I don't understand why multi edges are an issue here.

sorrachai:
Based on the feedback so far, I have revised the definitions and opened a new [PR#810](https://github.com/leanprover/cslib/pull/810).

arademaker:
> > This is a meaningless claim. See prior discussion on graph constructors in the Zulip.
> 
> @Shreyas4991 When you phrase things so aggressively, it is counterproductive to having a discussion. I've consistently agreed with you on technical points here, but when you chose this style of communication, it derails the conversation.

@Shreyas4991 I don't think this message from @chenson2018 got a response — could you acknowledge it? The technical substance of your comments in this thread has been reasonable, but the tone has made the discussion harder than it needs to be for all of us. A brief acknowledgment and a more measured tone going forward would be very helpful.

Shreyas4991:
> > > This is a meaningless claim. See prior discussion on graph constructors in the Zulip.
> 
> > 
> 
> > @Shreyas4991 When you phrase things so aggressively, it is counterproductive to having a discussion. I've consistently agreed with you on technical points here, but when you chose this style of communication, it derails the conversation.
> 
> 
> 
> @Shreyas4991 I don't think this message from @chenson2018 got a response — could you acknowledge it? The technical substance of your comments in this thread has been reasonable, but the tone has made the discussion harder than it needs to be for all of us. A brief acknowledgment and a more measured tone going forward would be very helpful.

Sorry I didn't intend to be rude there. I could have been more specific and to the point about what I meant, which is that this was already previously discussed months ago.

PR #504, opened 2026-04-20
feat: recipe for logical fragments and multiplicative linear logic

fmontesi:
This PR:
- Adds multiplicative linear logic (MLL) as an example of how a fragment can be derived from a larger inference system (CLL), using Subtype and local recursive functions to eliminate irrelevant cases (see the example in the docstring).
- Puts CLL under the right `Cslib.Logic` namespace.
- Shortens a few simple definitions for CLL in the expected way.

### DIFF
--- a/Cslib/Logics/LinearLogic/CLL/Basic.lean
+++ b/Cslib/Logics/LinearLogic/CLL/Basic.lean
@@ -81,7 +81,7 @@ inductive Proposition.Context (Atom : Type u) : Type u where
 deriving DecidableEq, BEq
 
 /-- Replaces the hole in a propositional context with a propositions. -/
-@[scoped grind =]
+@[simp]
 def Proposition.Context.fill (c : Context Atom) (a : Proposition Atom) : Proposition Atom :=
   match c with
   | hole => a
@@ -143,8 +143,8 @@ def Proposition.dual : Proposition Atom → Proposition Atom
 theorem Proposition.top_eq_zero_dual : ⊤ = (0⫠ : Proposition Atom) := rfl
 
 /-- Duality preserves size. -/
-@[scoped grind _=_]
-theorem Proposition.dual_sizeOf (a : Proposition Atom) : sizeOf a = sizeOf a⫠ := by
+@[scoped grind =, simp]
+theorem Proposition.dual_sizeOf (a : Proposition Atom) : sizeOf a⫠ = sizeOf a := by
   induction a <;> simp [dual] <;> grind
 
 /-- No proposition is equal to its dual. -/
@@ -153,7 +153,7 @@ theorem Proposition.dual_neq (a : Proposition Atom) : a ≠ a⫠ := by
   cases a <;> simp [Proposition.dual]
 
 /-- Two propositions are equal iff their respective duals are equal. -/
-@[scoped grind =, simp]
+@[simp]
 theorem Proposition.dual_inj (a b : Proposition Atom) : a⫠ = b⫠ ↔ a = b := by
   refine ⟨fun h ↦ ?_, congrArg dual⟩
   induction a generalizing b <;> cases b <;> grind
@@ -227,24 +227,20 @@ def Proof.cut' (p : ⇓(a⫠ ::ₘ Γ)) (q : ⇓(a ::ₘ Δ)) : ⇓(Γ + Δ) :=
   p.cut r
 
 /-- Inversion of the ⅋ rule. -/
-@[scoped grind =]
 def Proof.parr_inversion {Γ : Sequent Atom} (h : ⇓((a ⅋ b) ::ₘ Γ)) : ⇓(a ::ₘ b ::ₘ Γ) :=
   show a ::ₘ b ::ₘ Γ = {a, b} + Γ by simp ▸
     cut' (show ({a, b} : Sequent Atom) = {a} + {b} by simp ▸ tensor ax' ax') h
 
 /-- Inversion of the ⊥ rule. -/
-@[scoped grind =]
 def Proof.bot_inversion {Γ : Sequent Atom} (h : ⇓(⊥ ::ₘ Γ)) : ⇓Γ := by
   convert Proof.cut' (a := ⊥) (Γ := {}) (Δ := Γ) Proof.one h
   simp
 
 /-- Inversion of the & rule, first component. -/
-@[scoped grind =]
 def Proof.with_inversion₁ {Γ : Sequent Atom} (h : ⇓((a & b) ::ₘ Γ)) : ⇓(a ::ₘ Γ) :=
   cut' (a := a & b) (oplus₁ ax') h
 
 /-- Inversion of the & rule, second component. -/
-@[scoped grind =]
 def Proof.with_inversion₂ {Γ : Sequent Atom} (h : ⇓((a & b) ::ₘ Γ)) : ⇓(b ::ₘ Γ) :=
   cut' (a := a & b) (oplus₂ ax') h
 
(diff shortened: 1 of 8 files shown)


thomaskwaring (approved):
This all looks very reasonable to me, it would be nice of course to see some theorems stress-testing the definitions, but that can happen down the track.

One suggestion (for here or in a follow-up) would be to define a custom cases / induction  principle for, say, `MLL.Proposition` — it would package up the process of unpacking the predicate and discharging unreachable cases, and might be a more idiomatic way to set up an API boundary.

Cslib/Logics/LinearLogic/CLL/Basic.lean
```diff
@@ -299,32 +279,32 @@ namespace Proposition
...
-@[symm, scoped grind →]
+@[symm]
 theorem Equiv.symm {a b : Proposition Atom} (h : a ≡ b) : b ≡ a := ⟨h.2, h.1⟩
 
 /-- Proof-irrelevant equivalence is transitive. -/
-@[scoped grind →]
 theorem Equiv.trans {a b c : Proposition Atom} (hab : a ≡ b) (hbc : b ≡ c) : a ≡ c :=
   equiv.trans (chooseEquiv hab) (chooseEquiv hbc)
 
+scoped grind_pattern Equiv.trans => a ≡ b, b ≡ c
```
chenson2018, line 306: This seems to be the same pattern as the removed annotation?

Cslib/Logics/LinearLogic/CLL/MLL.lean
```diff
@@ -0,0 +1,143 @@
...
+namespace Cslib.Logic.CLL
+
+/-- A proposition is in the multiplicative fragment of CLL. -/
+@[simp]
+def Proposition.IsMLL : Proposition Atom → Prop
+  | atom _ | atomDual _ | one | bot => True
+  | top | zero => False
+  | tensor a b | parr a b => a.IsMLL ∧ b.IsMLL
+  | oplus _ _ | .with _ _ => False
+  | bang _ | quest _ => False
```
chenson2018: Easier to read?

```suggestion
  | atom _ | atomDual _ | one | bot => True
  | tensor a b | parr a b => a.IsMLL ∧ b.IsMLL
  | _ => False
```

Cslib/Logics/LinearLogic/CLL/MLL.lean
```diff
@@ -0,0 +1,143 @@
...
+
+/-- Duality in MLL stays in MLL. -/
+@[scoped grind →]
+theorem Proposition.isMLL_dual {a : Proposition Atom} (ha : a.IsMLL) : a⫠.IsMLL := by
+  induction a <;> grind only [Proposition.dual, IsMLL]
```
chenson2018: The squeeze is not needed here, and let's be explicit about when we do or don't need the dual:

```suggestion
  induction a with
  | atom | atomDual | one | bot | tensor | parr => grind [dual, IsMLL]
  | _ => grind [IsMLL]
```

Cslib/Logics/LinearLogic/CLL/MLL.lean
```diff
@@ -0,0 +1,143 @@
...
+def Proposition.Context.IsMLL : Context Atom → Prop
+  | hole => True
+  | tensorL a b | tensorR a b | parrL a b | parrR a b => a.IsMLL ∧ b.IsMLL
+  | _ => False
+
+/-- Filling a multiplicative propositional context with a multiplicative proposition stays in MLL.
+-/
+theorem Proposition.Context.isMLL_fill {c : CLL.Proposition.Context Atom} {a : CLL.Proposition Atom}
+    (hc : c.IsMLL) : (c.fill a).IsMLL ↔ a.IsMLL := by induction c <;>
+      grind only [Context.fill, Proposition.IsMLL, Proposition.Context.IsMLL]
```
chenson2018: In general, the consensus is that you shouldn't construct a `induction <;> grind [...]` by taking the union of all the parameters used in each branch of the proof. I'd write:
   
```suggestion
    (hc : c.IsMLL) : (c.fill a).IsMLL ↔ a.IsMLL := by
  induction c with
  | hole => grind [fill]
  | tensorL | tensorR | parrL | parrR => grind [fill, IsMLL, Proposition.IsMLL]
  | _ => grind [IsMLL]
```

fmontesi:
@thomaskwaring 
I think that's a very good idea, I'll experiment with it when I find a moment.

Cslib/Logics/LinearLogic/CLL/MLL.lean
```diff
@@ -0,0 +1,143 @@
...
+open scoped Logic.InferenceSystem
+
+/-- A proof is in MLL. -/
+@[simp]
+def Proof.IsMLL {Γ : Sequent Atom} : ⇓Γ → Prop
+  | ax (a := a) => a.IsMLL
+  | cut p q | tensor p q => p.IsMLL ∧ q.IsMLL
+  | one => True
+  | bot p | parr p => p.IsMLL
+  | _ => False
```
chenson2018: I had no idea you could use do a pattern match with named paramaters like this. If you'd like, this can be combined with some below cases

```suggestion
  | ax (a := a) | bot a | parr a => a.IsMLL
  | cut p q | tensor p q => p.IsMLL ∧ q.IsMLL
  | one => True
  | _ => False
```

Cslib/Logics/LinearLogic/CLL/MLL.lean
```diff
@@ -0,0 +1,143 @@
...
+    grind [Proof.IsMLL, Multiset.insert_eq_cons, Multiset.mem_singleton]
+  case one =>
+    simp [Sequent.IsMLL, Proposition.IsMLL]
+  case parr | tensor | cut => grind [Proposition.IsMLL, Proof.IsMLL]
+  case bot Γ p ih =>
+    simp only [Proof.IsMLL] at hp
+    simp only [Sequent.IsMLL, Multiset.mem_cons, forall_eq_or_imp]
+    apply And.intro
+    · simp only [Proposition.IsMLL]
+    · grind only [Sequent.IsMLL]
```
chenson2018: This can be

```suggestion
    simp
    grind [Proof.IsMLL]
```
without any issues I can see.

Cslib/Logics/LinearLogic/CLL/MLL.lean
```diff
@@ -0,0 +1,143 @@
...
+/-- MLL sequents. -/
+abbrev Sequent (Atom : Type u) := {Γ : CLL.Sequent Atom // Γ.IsMLL}
+
+/-- MLL derivations. -/
+abbrev Proof {Atom : Type u} (Γ : Sequent Atom) := {p : CLL.Proof (Atom := Atom) Γ // p.IsMLL}
+
+/-- The sequent calculus of MLL. -/
+instance : Logic.InferenceSystem (Sequent Atom) := ⟨Proof⟩
+
+open scoped Logic.InferenceSystem
```
chenson2018: Why this open with no code following it?

Cslib/Logics/LinearLogic/CLL/MLL.lean
```diff
@@ -0,0 +1,143 @@
...
+
+namespace Cslib.Logic.MLL
+
+/-- MLL propositions. -/
+abbrev Proposition (Atom : Type u) := {a : CLL.Proposition Atom // a.IsMLL}
+
+/-- MLL propositional contexts. -/
+abbrev Proposition.Context (Atom : Type u) := {c : CLL.Proposition.Context Atom // c.IsMLL}
+
+open scoped CLL.Proposition CLL.Proposition.Context in
```
chenson2018: What is this `open scoped` for?

Cslib/Logics/LinearLogic/CLL/MLL.lean
```diff
@@ -0,0 +1,143 @@
...
+/-- Filling of an MLL propositional context. -/
+def Proposition.Context.fill (c : MLL.Proposition.Context Atom) (a : MLL.Proposition Atom) :
+    MLL.Proposition Atom :=
+  ⟨CLL.Proposition.Context.fill c a, (CLL.Proposition.Context.isMLL_fill c.property).2 a.property⟩
```
chenson2018: I might try to make this a bit more readable with a few formatting changes:

```suggestion
def Proposition.Context.fill (c : Proposition.Context Atom) (a : Proposition Atom) :
    Proposition Atom where
  val := CLL.Proposition.Context.fill c a
  property := (CLL.Proposition.Context.isMLL_fill c.property).mpr a.property
```

chenson2018 (changes requested):
A few small style questions. The removal of `grind` lints seems nice here too.

Cslib/Logics/LinearLogic/CLL/Basic.lean
fmontesi: Was it so smart already? I didn't check because I realised this is what they recommend in the grind documentation and I just switched. We can revert it I guess.

Cslib/Logics/LinearLogic/CLL/MLL.lean
fmontesi: Whoops, leftover, my bad. :)

Cslib/Logics/LinearLogic/CLL/MLL.lean
fmontesi: grind :-)

chenson2018: I'm not understanding, I don't see `grind` used anywhere in this definition? It seems fine without it.

Cslib/Logics/LinearLogic/CLL/Basic.lean
chenson2018: Yes if you look at the output of `@[grind?]`, it shows (in a somewhat difficult to read format...) the same pattern. I have a slight preference for `grind →` for readability, but they are equivalent.

chenson2018 (approved):
I'm approving but disabling auto-merge, as I think the `open scoped` I mentioned in my last comment can be removed.

Cslib/Logics/LinearLogic/CLL/MLL.lean
fmontesi: Sorry, I saw it at the wrong line! Removed.

PR #505, opened 2026-04-20
doc: add new reviewer

fmontesi:


chenson2018 (approved):
I think `doc` is the best label for PRs like these.

fmontesi:
> I think `doc` is the best label for PRs like these.

Agreed!

PR #518, opened 2026-04-24
feat: Tagged InferenceSystem

fmontesi:
Adds a tagged version of `InferenceSystem`, so that multiple inference systems for the same type can be instantiated and distinguished by means of a tag. Usage is exemplified with the multiplicative fragment of classical linear logic.


Cslib/Foundations/Logic/InferenceSystem.lean
```diff
@@ -14,38 +14,52 @@ namespace Cslib.Logic
 
 /--
 The notation typeclass for inference systems.
-This enables the notation `⇓a`, where `a : α` is a derivable value.
+This enables the notation `S⇓a`, where `S` is a tag for the inference system and `a : α`
+is a derivable value.
 -/
-class InferenceSystem (α : Type u) where
+class InferenceSystem (S : Type) (α : Type*) where
```
thomaskwaring: is there a reason this isn't `S : Type*`

thomaskwaring (approved):

Cslib/Foundations/Logic/InferenceSystem.lean
fmontesi: That's what it was originally, but it gave me type errors, I don't clearly remember where (maybe DerivationsIn).

thomaskwaring: okay, it does seem very restrictive to have it `Type 0` — can you fix it with explicit universes?

Cslib/Logics/LinearLogic/CLL/MLL.lean
```diff
@@ -97,11 +104,47 @@ def Proof.IsMLL {Γ : Sequent Atom} : ⇓Γ → Prop
...
+    (parr : ∀ {a b : Proposition Atom} {Γ : Sequent Atom}
+      (p : Proof (a ::ₘ b ::ₘ Γ)) {hp : p.IsMLL},
+      @motive (a ::ₘ b ::ₘ Γ) p hp → @motive ((a ⅋ b) ::ₘ Γ) (.parr p) (by simp [hp]))
+    (cut : ∀ {a : Proposition Atom} {Γ Δ : Sequent Atom}
+      (p : Proof (a ::ₘ Γ)) (q : Proof (a⫠ ::ₘ Δ)) {hp : p.IsMLL} {hq : q.IsMLL},
+      @motive (a ::ₘ Γ) p hp → @motive (a⫠ ::ₘ Δ) q hq →
+      @motive (Γ + Δ) (.cut p q) (by simp [hp, hq]))
+    {Γ : Sequent Atom} {p : Proof Γ} (h : p.IsMLL) : @motive Γ p h :=
+  match p, h with
+  | .ax (a := a), hp => @ax a (by simp only [IsMLL] at hp; exact hp)
```
chenson2018:
```suggestion
  | .ax (a := a), hp => @ax a (by simpa)
```

Cslib/Foundations/Logic/InferenceSystem.lean
```diff
@@ -14,38 +14,52 @@ namespace Cslib.Logic
...
   /--
-  `⇓a` is a derivation of `a`, that is, a witness that `a` is derivable.
+  `S⇓a` is a derivation of `a`, that is, a witness that `a` is derivable in the system `S`.
   The meaning of this notation is type-dependent.
   -/
-  derivation (s : α) : Sort v
+  derivation (a : α) : Sort v
+
+/-- Default tag for inference system instances. `⇓a` is short for `Default⇓a`. -/
+inductive InferenceSystem.Default
```
chenson2018: It took me a moment to understand why this was an empty inductive. I feel like I have more often seen something like
```lean
def InferenceSystem.Default := Unit
```
in these situations. I think this is functionally identical, but maybe easier to understand the intent for the next person like me who was confused.

Cslib/Logics/LinearLogic/CLL/MLL.lean
```diff
@@ -110,31 +153,57 @@ theorem Proof.isMLL_sequent {Γ : Sequent Atom} (p : ⇓Γ) (hp : p.IsMLL) : Γ.
...
 
-end Cslib.Logic.CLL
+/-- If a CLL derivation is cut-free and concludes an MLL sequent, then it is an MLL derivation. -/
+theorem Proof.isMLL_cutFree {Γ : Sequent Atom} (p : ⇓Γ) (hΓ : Γ.IsMLL)
+    (hp : p.cutFree) : p.IsMLL := by
+  induction p
+  case ax a =>
+    simp only [Sequent.IsMLL, Multiset.insert_eq_cons, Multiset.mem_cons, Multiset.mem_singleton,
+      forall_eq_or_imp, forall_eq] at hΓ
+    simp [hΓ]
```
chenson2018:
```suggestion
  case ax => simp_all
```

Cslib/Logics/LinearLogic/CLL/MLL.lean
```diff
@@ -110,31 +153,57 @@ theorem Proof.isMLL_sequent {Γ : Sequent Atom} (p : ⇓Γ) (hp : p.IsMLL) : Γ.
...
+  case one => simp
+  case bot _ _ ih =>
+    simp only [Sequent.IsMLL, Multiset.mem_cons, forall_eq_or_imp, Proposition.IsMLL,
+      true_and] at hΓ
+    specialize ih hΓ hp
+    grind only [IsMLL]
```
chenson2018:
```suggestion
    refine ih ?_ hp
    simpa using hΓ
```

Cslib/Logics/LinearLogic/CLL/MLL.lean
```diff
@@ -110,31 +153,57 @@ theorem Proof.isMLL_sequent {Γ : Sequent Atom} (p : ⇓Γ) (hp : p.IsMLL) : Γ.
...
+  case parr a b Γ p ih =>
+    simp only [Sequent.IsMLL, Multiset.mem_cons, forall_eq_or_imp, Proposition.IsMLL] at hΓ
+    specialize ih (by grind [Sequent.IsMLL]) hp
+    grind only [IsMLL]
```
chenson2018:
```suggestion
    refine ih ?_ hp
    simp at hΓ
    grind [Sequent.IsMLL]
```

Cslib/Logics/LinearLogic/CLL/MLL.lean
```diff
@@ -110,31 +153,57 @@ theorem Proof.isMLL_sequent {Γ : Sequent Atom} (p : ⇓Γ) (hp : p.IsMLL) : Γ.
...
+  case tensor a b Γ Δ p q ihp ihq =>
+    simp only [Sequent.IsMLL, Multiset.mem_cons, forall_eq_or_imp, Proposition.IsMLL] at hΓ
+    specialize ihp (by grind [Sequent.IsMLL])
+    specialize ihq (by grind [Sequent.IsMLL])
+    grind only [cutFree, IsMLL]
```
chenson2018:
```suggestion
    simp at hΓ
    refine ⟨ihp ?mll.p ?cut.p, ihq ?mll.q ?cut.q⟩
    case mll | mll => grind [Sequent.IsMLL]
    case cut | cut => grind [cutFree]
```

chenson2018 (changes requested):
Just a few style questions/suggestions.

Cslib/Foundations/Logic/InferenceSystem.lean
fmontesi: My purpose was to do the simplest possible thing, which is an empty type.

What's the point of making Default inhabited?

Cslib/Logics/LinearLogic/CLL/MLL.lean
fmontesi: I like the refine, but doesn't the simp at hΓ make the proof more brittle? Although I guess my simp only risks it as well if somebody changes the names of the lemmas...

Cslib/Foundations/Logic/InferenceSystem.lean
fmontesi: Oh, actually, I tried again and it worked now. It must've been something weird with a previous version.

Thanks for catching this.

Cslib/Logics/LinearLogic/CLL/MLL.lean
chenson2018: Yes, the general style here from Mathlib is that what `simp` and `grind` converge to should be more stable than a static `only` tactic.

Cslib/Foundations/Logic/InferenceSystem.lean
chenson2018: It doesn't matter that it's inhabited, you could use `Empty` just as easily. My point is that it doesn't really matter: we could pick any existing type and say that by convention this represents the default.

So I'd personally find `def InferenceSystem.Default := Empty` the more readable exact equivalent I suppose. Maybe a short comment is warranted in either case.

chenson2018 (approved):
Approving, as my last comment about `InferenceSystem.Default` seems fine either way, maybe a matter of personal taste.

chenson2018:
(I removed from the auto merge so you can decide what you prefer writing, and maybe leave a comment.)

fmontesi:
Thanks. I've made it an `opaque` alias for Empty, which should be equivalent to my original.

PR #522, opened 2026-04-27
chore: update mathlib

eric-wieser:
The motivation is to get some `ContT` lemmas to clean up the FreeM examples

Performed with `lake update mathlib`.


chenson2018 (approved):
Thanks!

PR #523, opened 2026-04-27
Feat(Cryptography): Signed Barrett reduction algorithm

atrieu:
This adds a formalisation of the signed variant of the Barrett reduction algorithm which is used in recent crypto schemes such as ML-KEM and ML-DSA.

### DIFF
--- a/Cslib/Crypto/Algorithms/BarrettReduction/Aux.lean
+++ b/Cslib/Crypto/Algorithms/BarrettReduction/Aux.lean
@@ -1,0 +1,199 @@
+/-
+Copyright (c) 2026 Alix Trieu. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Alix Trieu
+-/
+
+module
+
+public import Cslib.Init
+public import Mathlib.Data.Nat.Log
+public import Mathlib.Algebra.Order.Round
+public import Mathlib.Data.Rat.Floor
+public import Mathlib.Algebra.Order.Floor.Defs
+public import Mathlib.Data.Int.DivMod
+import Mathlib.Tactic
+
+/-
+# Auxiliary definitions and lemmas
+
+- Defines `clog2`, a base 2 upper logarithm and some associated lemmas
+- Additional facts about `bmod`, `floor` and `round`
+-/
+
+@[expose]
+public section
+
+namespace Nat
+
+def clog2 : ℕ → ℕ := Nat.clog 2
+
+lemma le_clog2_self (n : ℕ) :
+  n ≤ 2 ^ (n.clog2) := by
+  apply le_pow_clog (by simp) n
+
+lemma log2_le_clog2 (n : ℕ) :
+  n.log2 ≤ n.clog2 := by
+  rw [log2_eq_log_two]
+  apply Nat.log_le_clog 2 n
+
+lemma le_pow_iff_clog2_le {x y : ℕ} :
+  x ≤ 2 ^ y ↔ clog2 x ≤ y :=
+  by symm; apply Nat.clog_le_iff_le_pow; simp
+
+lemma clog2_le_log2 (n : ℕ) :
+  n.clog2 ≤ n.log2 + 1 := by
+  rw [log2_eq_log_two]
+  rw [← le_pow_iff_clog2_le]
+  apply le_of_lt
+  cases n with
+  | zero => simp
+  | succ n =>
+    rw [← log2_eq_log_two, ← Nat.log2_lt (by simp)]
+    simp
+
+lemma clog2_eq (n : ℕ) :
+  n.clog2 = if 2 ^ n.log2 < n then n.log2 + 1 else n.log2 := by
+  have H₀ := clog2_le_log2 n
+  have H₁ := log2_le_clog2 n
+  split_ifs with Hcond <;> rw [← Nat.lt_clog_iff_pow_lt (by simp), ← clog2] at Hcond <;> linarith
+
+end Nat
+
+namespace Int
+
+lemma abs_bmod_le (x : ℤ) (m : ℕ) (Hm : 0 < m) :
+  |x.bmod m| ≤ m / 2 := by
+  rw [abs_le]; apply And.intro
+  · apply Int.le_bmod Hm
+  · transitivity
+    · apply Int.bmod_le Hm
+    · omega
+
+lemma bmod_eq' (x : ℤ) (m : ℕ) :
+  x.bmod m = x - m * (round (x / (m: ℚ))) := by
+  rw [round_eq, Int.bmod]
+  have X: x % m < (m + 1) / 2 ↔ 2 * (x % m) < m := by omega
+  cases Nat.eq_zero_or_pos m with
+  | inl Hz => rw [Hz]; simp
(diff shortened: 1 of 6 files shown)


Cslib/Crypto/Algorithms/BarrettReduction/Aux.lean
```diff
@@ -0,0 +1,199 @@
...
+  |x.bmod m| ≤ m / 2 := by
+  rw [abs_le]; apply And.intro
+  · apply Int.le_bmod Hm
+  · transitivity
+    · apply Int.bmod_le Hm
+    · omega
+
+lemma bmod_eq' (x : ℤ) (m : ℕ) :
+  x.bmod m = x - m * (round (x / (m: ℚ))) := by
+  rw [round_eq, Int.bmod]
```
eric-wieser, line 74: Lean's indent style is
```suggestion
lemma bmod_eq' (x : ℤ) (m : ℕ) :
    x.bmod m = x - m * (round (x / (m: ℚ))) := by
  rw [round_eq, Int.bmod]
```
please apply this throughout

Cslib/Crypto/Algorithms/BarrettReduction/Signed.lean
```diff
@@ -0,0 +1,301 @@
...
+def is_approx (δ : ℚ) (α : ℚ → ℤ) : Prop :=
+  ∀ (x: ℚ), |(x - α x)| ≤ δ
+
+lemma round_is_approx : is_approx (1/2) round := by
+  intro x; apply abs_sub_round
+
+def round_to_even (x : ℚ) : ℤ :=
+  2 * ⌊(x / 2)⌉
+
+def mod_approx (α : ℚ → ℤ) (x : ℤ) (N : ℕ) : ℤ := x - ↑N * (α (x/N))
```
eric-wieser: Please look at the lean naming convention; these should be 
```suggestion
def IsApprox (δ : ℚ) (α : ℚ → ℤ) : Prop :=
  ∀ (x: ℚ), |(x - α x)| ≤ δ

lemma round_isApprox : IsApprox (1/2) round := by
  intro x; apply abs_sub_round

def roundToEven (x : ℚ) : ℤ :=
  2 * ⌊(x / 2)⌉

def modApprox (α : ℚ → ℤ) (x : ℤ) (N : ℕ) : ℤ := x - ↑N * (α (x/N))
``
etc: `def`s should never contain `_` in their name, and should be capitalized if they are `Prop`s

File: Cslib/Crypto/Algorithms/BarrettReduction/Aux.lean (hunk 1 above)
Comment (atrieu) [reply Cslib/Crypto/Algorithms/BarrettReduction/Aux.lean:74]: Thanks, fixed by [82f80d8](https://github.com/leanprover/cslib/pull/523/commits/82f80d8ef71a6a1f69daaf605b801e9a3de5f6f5)

File: Cslib/Crypto/Algorithms/BarrettReduction/Signed.lean (hunk 2 above)
Comment (atrieu) [reply Cslib/Crypto/Algorithms/BarrettReduction/Signed.lean]: Thanks, fixed by [82f80d8](https://github.com/leanprover/cslib/pull/523/commits/82f80d8ef71a6a1f69daaf605b801e9a3de5f6f5)

File: Cslib/Crypto/Algorithms/BarrettReduction/Signed.lean [hunk 3]
```diff
@@ -0,0 +1,298 @@
...
+  nth_rw 2 [← mul_one a]
+  rw [← barrettMul_spec a 1 M R k q] <;> try assumption
+  · rw [barrettReduce, barrettMul]
+    rw [mul_one]; simp
+  · simp only [Int.cast_one, one_mul, one_div]; simp at Hk; assumption
+  · simp only [abs_one]; refine one_le_pow₀ (by simp)
+
+end Cslib.Crypto.Algorithms.BarrettReduction.Signed
+
+section MLKEMExample
```
SamuelSchlesinger: This is polluting the namespace, but I think moving it to CslibTests/ should be sufficient to avoid. It probably belongs there anyways.

Cslib/Crypto/Algorithms/BarrettReduction/Signed.lean
```diff
@@ -0,0 +1,298 @@
...
+-/
+
+namespace Cslib.Crypto.Algorithms.BarrettReduction.Signed
+
+notation "⌊" x "⌉" => round (x : ℚ)
+
+def IsApprox (δ : ℚ) (α : ℚ → ℤ) : Prop :=
+  ∀ (x: ℚ), |(x - α x)| ≤ δ
+
+lemma round_isApprox : IsApprox (1/2) round := by
```
SamuelSchlesinger: This and IsApprox are dead code, but I'm guessing you plan to use them later?

Cslib/Crypto/Algorithms/BarrettReduction/Signed.lean
```diff
@@ -0,0 +1,298 @@
...
+
+module
+
+public import Cslib.Init
+public import Mathlib.Data.Nat.Log
+public import Mathlib.Data.Rat.Init
+public import Mathlib.Data.Rat.Floor
+public import Mathlib.Algebra.Order.Field.Rat
+public import Mathlib.Algebra.Order.Round
+import Mathlib.Tactic
```
SamuelSchlesinger: Try not to import this directly, Cslib.Init should have what you need and import directly otherwise.

Cslib/Crypto/Algorithms/BarrettReduction/Aux.lean
```diff
@@ -0,0 +1,199 @@
...
+module
+
+public import Cslib.Init
+public import Mathlib.Data.Nat.Log
+public import Mathlib.Algebra.Order.Round
+public import Mathlib.Data.Rat.Floor
+public import Mathlib.Algebra.Order.Floor.Defs
+public import Mathlib.Data.Int.DivMod
+import Mathlib.Tactic
+
```
SamuelSchlesinger, line 13: This file should have a note to upstream most of its contents to mathlib I think.

Cslib/Crypto/Algorithms/BarrettReduction/Signed.lean
atrieu: Thanks, moved to `CslibTests/`

Cslib/Crypto/Algorithms/BarrettReduction/Signed.lean
atrieu: Oops, this is leftover from when I was trying to prove a generic version of the reduction modular wrt the chosen approximation (round half up provides signed reduction, while floor provides unsigned reduction). I ended up proving this version which provides better bounds iirc.

I removed it for now.

Cslib/Crypto/Algorithms/BarrettReduction/Signed.lean
atrieu: Done

Cslib/Crypto/Algorithms/BarrettReduction/Aux.lean
atrieu: I added a note, thanks

Issue #23, opened 2025-07-28
Notation and term elaborators


fmontesi:
In logic, semantics, and type systems (at least!), we encounter often the situation where we'd like to use important Lean symbols for notation. Examples: `T → T` for arrow types (functions) in an embedded type system, `A → B` for implication in an embedded logic, `M → N` for a reduction, `x : T` for x has type T, etc.

We will also need to combine notations in nontrivial ways. For example, this is the shape of reductions between typing judgements used in plenty of papers: `(Γ ⊢ M : T) → (Γ ⊢ M' : T)`.

In a Zulip discussion ([#computer science > Overloading Lean's notation](https://leanprover.zulipchat.com/#narrow/channel/516743-computer-science/topic/Overloading.20Lean's.20notation)), we discussed the possibility of using term elaborators. This is an example from Iris: https://github.com/leanprover-community/iris-lean/blob/master/src/Iris/BI/Notation.lean.
Thus we could write things like `red(ty(Γ ⊢ M : T) → ty(Γ ⊢ M' : T))` (the nested `ty` might be unnecessary).

The design of these elaborators will need some thinking, so I'm opening this issue to keep record.

chenson2018:
I'd like to start with something small to kick this off. Occasionally when applying a lemma or constructor, a `ReductionSystem` is exposed in the infoview e.g. a term like `rs.Red M N` instead of the notation we have defined. I think this is a use case for a pretty printer? 

Again though, it would need to be a family of pretty printers. I think this is exactly what is happening in this example Kyle  provided:

https://github.com/UCSCFormalMethods/LeanLTL/blob/ef8d4c9dd94d3ae85e76328549b70e35551891f2/LeanLTL/TraceSet/Notation.lean#L268-L290

chenson2018:
In light of https://github.com/cs-lean/cslib/pull/25#issuecomment-3149603164 about relying on `notation3`, I thought I should leave a link to [this Zulip thread](https://leanprover.zulipchat.com/#narrow/channel/239415-metaprogramming-.2F-tactics/topic/writing.20a.20command.20to.20define.20.28a.20family.20of.29.20unexpanders). If at some later point we want/need to remove this dependency, that thread has the information on how to do something similar ourselves.

chenson2018:
[...]

PR #528, opened 2026-04-29
feat: Modal Logic

fmontesi:
Adds Modal Logic and the Modal Cube, including all the 15 well-known modal logics (K, D, T, S5, etc.) and their relationships.

The notation that overlaps with propositional logic is consistent with the existing module.

Note for future work (or here, if somebody can see an easy fix that doesn't compromise readability for modal logicians): I'm not very satisfied by the notation for judgements, which includes a ',' so it can lead to some parsing trouble.

### DIFF
--- a/Cslib/Foundations/Data/Relation.lean
+++ b/Cslib/Foundations/Data/Relation.lean
@@ -77,13 +77,8 @@ class RightEuclidean (r : α → α → Prop) where
 class Serial (r : α → α → Prop) where
   serial a : Reducible r a
 
-@[scoped grind →]
-lemma refl_serial (r : α → α → Prop) (h : Std.Refl r) : Relation.Serial r := by
-  constructor
-  intro a
-  obtain ⟨h⟩ := h
-  exists a
-  exact h a
+instance [Std.Refl r] : Relation.Serial r where
+  serial a := ⟨a, Std.Refl.refl a⟩
 
 /-- A relation has the diamond property when all reductions with a common origin are joinable -/
 abbrev Diamond (r : α → α → Prop) := ∀ {a b c : α}, r a b → r a c → Join r b c
--- a/Cslib/Foundations/Data/Relation.lean
+++ b/Cslib/Foundations/Data/Relation.lean
@@ -156,8 +156,11 @@ abbrev Reducible (r : α → α → Prop) (x : α) : Prop := ∃ y, r x y
 class Serial (r : α → α → Prop) where
   serial a : Reducible r a
 
-instance [Std.Refl r] : Relation.Serial r where
-  serial a := ⟨a, Std.Refl.refl a⟩
+@[scoped grind →]
+lemma refl_serial (r : α → α → Prop) (h : Std.Refl r) : Relation.Serial r where
+  serial a := ⟨a, h.refl a⟩
+
+instance [instRefl : Std.Refl r] : Relation.Serial r := refl_serial r instRefl
 
 /-- An element is normal if it is not reducible. -/
 abbrev Normal (r : α → α → Prop) (x : α) : Prop := ¬ Reducible r x
--- a/Cslib/Foundations/Data/Relation.lean
+++ b/Cslib/Foundations/Data/Relation.lean
@@ -73,13 +73,6 @@ def UpTo (r s : α → α → Prop) : α → α → Prop := Comp s (Comp r s)
 class RightEuclidean (r : α → α → Prop) where
   rightEuclidean : r a b → r a c → r b c
 
-/-- A relation `r` is serial if every element is `Reducible`. -/
-class Serial (r : α → α → Prop) where
-  serial a : Reducible r a
-
-instance [Std.Refl r] : Relation.Serial r where
-  serial a := ⟨a, Std.Refl.refl a⟩
-
 /-- A relation has the diamond property when all reductions with a common origin are joinable -/
 abbrev Diamond (r : α → α → Prop) := ∀ {a b c : α}, r a b → r a c → Join r b c
 
@@ -158,6 +151,13 @@ theorem Confluent_of_unique_end {x : α} (h : ∀ y : α, ReflTransGen r y x) : 
 
 /-- An element is reducible with respect to a relation if there is a value it is related to. -/
 abbrev Reducible (r : α → α → Prop) (x : α) : Prop := ∃ y, r x y
+
+/-- A relation `r` is serial if every element is `Reducible`. -/
+class Serial (r : α → α → Prop) where
+  serial a : Reducible r a
+
+instance [Std.Refl r] : Relation.Serial r where
+  serial a := ⟨a, Std.Refl.refl a⟩
 
 /-- An element is normal if it is not reducible. -/
 abbrev Normal (r : α → α → Prop) (x : α) : Prop := ¬ Reducible r x
--- a/Cslib/Foundations/Data/Relation.lean
+++ b/Cslib/Foundations/Data/Relation.lean
@@ -73,9 +73,9 @@ def UpTo (r s : α → α → Prop) : α → α → Prop := Comp s (Comp r s)
 class RightEuclidean (r : α → α → Prop) where
   rightEuclidean : r a b → r a c → r b c
 
-/-- A relation `r` is serial if every element is related to some element. -/
+/-- A relation `r` is serial if every element is `Reducible`. -/
 class Serial (r : α → α → Prop) where
-  serial : ∀ a, ∃ b, r a b
+  serial a : Reducible r a
 
 @[scoped grind →]
 lemma refl_serial (r : α → α → Prop) (h : Std.Refl r) : Relation.Serial r := by
--- a/Cslib/Foundations/Data/Relation.lean
+++ b/Cslib/Foundations/Data/Relation.lean
@@ -69,6 +69,22 @@ theorem MJoin.single (h : ReflTransGen r a b) : MJoin r a b := by
 /-- The relation `r` 'up to' the relation `s`. -/
 def UpTo (r s : α → α → Prop) : α → α → Prop := Comp s (Comp r s)
 
+/-- A relation `r` is (right) Euclidean if `r a b` and `r a c` guarantee `r b c`. -/
+class RightEuclidean (r : α → α → Prop) where
+  rightEuclidean : r a b → r a c → r b c
+
+/-- A relation `r` is serial if every element is related to some element. -/
+class Serial (r : α → α → Prop) where
+  serial : ∀ a, ∃ b, r a b
+
+@[scoped grind →]
+lemma refl_serial (r : α → α → Prop) (h : Std.Refl r) : Relation.Serial r := by
+  constructor
+  intro a
+  obtain ⟨h⟩ := h
+  exists a
+  exact h a
+
 /-- A relation has the diamond property when all reductions with a common origin are joinable -/
 abbrev Diamond (r : α → α → Prop) := ∀ {a b c : α}, r a b → r a c → Join r b c
 
--- a/Cslib/Logics/Modal/Basic.lean
+++ b/Cslib/Logics/Modal/Basic.lean
@@ -115,7 +115,7 @@ theorem neg_satisfies : ⇓Modal[m,w ⊨ ¬φ] ↔ ¬⇓Modal[m,w ⊨ φ] := by
 Disjunction is defined in terms of the more primitive connectives given in `Proposition`.
 This result proves that the definition is correct. -/
 @[scoped grind =]
-theorem Satisfies.or_char {m : Model World Atom} :
+theorem Satisfies.or_iff_or {m : Model World Atom} :
     ⇓Modal[m,w ⊨ φ₁ ∨ φ₂] ↔ ⇓Modal[m,w ⊨ φ₁] ∨ ⇓Modal[m,w ⊨ φ₂] := by grind [Proposition.or]
 
 /-- Characterisation of the `→` connective.
@@ -124,7 +124,7 @@ Implication is defined in terms of the more primitive connectives given in `Prop
 This result proves that the definition is correct.
 -/
 @[scoped grind =]
-theorem Satisfies.impl_char {m : Model World Atom} :
+theorem Satisfies.impl_iff_impl {m : Model World Atom} :
     ⇓Modal[m,w ⊨ φ₁ → φ₂] ↔ (⇓Modal[m,w ⊨ φ₁] → ⇓Modal[m,w ⊨ φ₂]) := by grind [Proposition.impl]
 
 /-- Characterisation of the `□` modality.
@@ -132,7 +132,7 @@ theorem Satisfies.impl_char {m : Model World Atom} :
 Necessity is defined in terms of the more primitive connectives given in `Proposition`.
 This result proves that the definition is correct. -/
 @[scoped grind =]
-theorem Satisfies.box_char {m : Model World Atom} :
+theorem Satisfies.box_iff_forall {m : Model World Atom} :
     ⇓Modal[m,w ⊨ □φ] ↔ ∀ w', m.r w w' → ⇓Modal[m,w' ⊨ φ] := by grind [Proposition.box]
 
 /-- The theory of a world in a model is the set of all propositions that it satifies. -/
@@ -211,7 +211,7 @@ theorem Satisfies.b_symm
 /-- The 4 axiom, valid for all transitive models. -/
 theorem Satisfies.four {m : Model World Atom} [instTrans : IsTrans World m.r] {w : World}
     (φ : Proposition Atom) : ⇓Modal[m,w ⊨ ⋄⋄φ → ⋄φ] := by
-  simp only [impl_char]
+  simp only [impl_iff_impl]
   intro h
   rcases h with ⟨w', h₁, w'', h₂, hs⟩
   exists w''
--- a/Cslib/Logics/Modal/Basic.lean
+++ b/Cslib/Logics/Modal/Basic.lean
@@ -1,0 +1,284 @@
+/-
+Copyright (c) 2026 Fabrizio Montesi. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Fabrizio Montesi, Marianna Girlando
+-/
+
+module
+
+public import Cslib.Init
+public import Cslib.Foundations.Logic.InferenceSystem
+public import Mathlib.Data.Set.Basic
+public import Mathlib.Order.Defs.Unbundled
+public import Cslib.Foundations.Data.Relation
+
+/-! # Modal Logic
+
+Modal logic is a logic for reasoning about relational structures, studying statements about
+necessity (`□φ`) and possibility `⋄φ`.
+
+## References
+
+* [P. Blackburn, M. de Rijke, Y. Venema, *Modal Logic*][Blackburn2001]
+* The definitions of theory equivalence and the denotational semantics of worlds are inspired by
+  the development of `Cslib.Logic.HML`.
+-/
+
+@[expose] public section
+
+namespace Cslib.Logic.Modal
+
+/-- A model consists of a relation between worlds `r` and a valuation `v`. -/
+structure Model (World : Type*) (Atom : Type*) where
+  /-- World accessibility relation. -/
+  r : World → World → Prop
+  /-- Valuation of atoms at a world. -/
+  v : World → Atom → Prop
+
+/-- Propositions. -/
+inductive Proposition (Atom : Type u) : Type u where
+  /-- Atomic proposition. -/
+  | atom (p : Atom)
+  /-- Negation. -/
+  | neg (φ : Proposition Atom)
+  /-- Conjunction. -/
+  | and (φ₁ φ₂ : Proposition Atom)
+  /-- Possibility. -/
+  | diamond (φ : Proposition Atom)
+
+@[inherit_doc] scoped prefix:40 "¬" => Proposition.neg
+@[inherit_doc] scoped infix:36 " ∧ " => Proposition.and
+@[inherit_doc] scoped prefix:40 "⋄" => Proposition.diamond
+
+/-- Disjunction. -/
+def Proposition.or (φ₁ φ₂ : Proposition Atom) : Proposition Atom := ¬(¬φ₁ ∧ ¬φ₂)
+
+@[inherit_doc] scoped infix:35 " ∨ " => Proposition.or
+
+/-- Implication. -/
+def Proposition.impl (φ₁ φ₂ : Proposition Atom) : Proposition Atom := ¬φ₁ ∨ φ₂
+
+@[inherit_doc] scoped infix:30 " → " => Proposition.impl
+
+/-- Bi-implication. -/
+@[simp, scoped grind =]
+def Proposition.biImpl (φ₁ φ₂ : Proposition Atom) : Proposition Atom := (φ₁ → φ₂) ∧ (φ₂ → φ₁)
+
+@[inherit_doc] scoped infix:30 " ↔ " => Proposition.biImpl
+
+/-- Necessity. -/
+def Proposition.box (φ : Proposition Atom) : Proposition Atom := ¬⋄¬φ
+
+@[inherit_doc] scoped prefix:40 "□" => Proposition.box
+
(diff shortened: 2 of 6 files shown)


Cslib/Foundations/Data/Relation.lean
```diff
@@ -69,6 +69,22 @@ theorem MJoin.single (h : ReflTransGen r a b) : MJoin r a b := by
 /-- The relation `r` 'up to' the relation `s`. -/
 def UpTo (r s : α → α → Prop) : α → α → Prop := Comp s (Comp r s)
 
+/-- A relation `r` is (right) Euclidean if `r a b` and `r a c` guarantee `r b c`. -/
+class RightEuclidean (r : α → α → Prop) where
+  rightEuclidean : r a b → r a c → r b c
+
+/-- A relation `r` is serial if every element is related to some element. -/
+class Serial (r : α → α → Prop) where
+  serial : ∀ a, ∃ b, r a b
```
chenson2018: We have an `abbrev` called `Reducible` a bit further down that I would use to write

```suggestion
/-- A relation `r` is serial if every element is `Reducible`. -/
class Serial (r : α → α → Prop) where
  serial a : Reducible r a
```

Cslib/Foundations/Data/Relation.lean
```diff
@@ -69,6 +69,22 @@ theorem MJoin.single (h : ReflTransGen r a b) : MJoin r a b := by
...
+
+@[scoped grind →]
+lemma refl_serial (r : α → α → Prop) (h : Std.Refl r) : Relation.Serial r := by
+  constructor
+  intro a
+  obtain ⟨h⟩ := h
+  exists a
+  exact h a
```
chenson2018: This should be

```suggestion
instance [Std.Refl r] : Relation.Serial r where
  serial a := ⟨a, Std.Refl.refl a⟩
```

Cslib/Logics/Modal/Basic.lean
```diff
@@ -0,0 +1,284 @@
...
+  /-- Constructs a judgement. -/
+  mk ::
+  /-- Model. -/
+  m : Model World Atom
+  /-- The world satisfying the proposition `φ`. -/
+  w : World
+  /-- The proposition satisfied by the world `w`. -/
+  φ : Proposition Atom
+
+@[inherit_doc] scoped notation m "," w " ⊨ " φ => Judgement.mk m w φ
```
chenson2018: You mentioned some issues using a comma in notation. There are some unicode alternatives. I see that `‚`  for instance already has an abbreviation `\glq` but doesn't seem to be used in Mathlib.

chenson2018 (changes requested):
Starting with just a couple of comments about the additions in `Cslib.Foundations.Data.Relation`

Cslib/Logics/Modal/Basic.lean
```diff
@@ -0,0 +1,284 @@
...
+* The definitions of theory equivalence and the denotational semantics of worlds are inspired by
+  the development of `Cslib.Logic.HML`.
+-/
+
+@[expose] public section
+
+namespace Cslib.Logic.Modal
+
+/-- A model consists of a relation between worlds `r` and a valuation `v`. -/
+structure Model (World : Type*) (Atom : Type*) where
```
thomaskwaring, line 33: Is there scope for generalizing this outside the `Modal` directory? i have a development of Kripke models for propositional logic which is effectively the same

Cslib/Logics/Modal/Basic.lean
```diff
@@ -0,0 +1,284 @@
...
+    Std.Refl r where
+  refl := by
+    intro w
+    rcases instAtomNonempty with ⟨a⟩
+    let v := fun (w' : World) (a : Atom) => w' = w
+    let h' := h (v := v) (w := w) (φ := .atom a)
+    grind
+
+/-- In any reflexive model, `□φ → φ` is equivalent to `φ → ⋄φ`. -/
+theorem Satisfies.t_box_diamond [instRefl : Std.Refl m.r] : ⇓(m,w ⊨ □φ → φ) ↔ ⇓(m,w ⊨ φ → ⋄φ) := by
```
thomaskwaring: is it worth generalising this slightly to be modus tollens + duality?

Cslib/Logics/Modal/Basic.lean
```diff
@@ -0,0 +1,284 @@
...
+/-- Any model that admits the axiom T is reflexive. -/
+theorem Satisfies.t_refl
+    {r : World → World → Prop}
+    [instAtomNonempty : Nonempty Atom]
+    (h : ∀ {v : World → Atom → Prop} {w : World} {φ : Proposition Atom},
+      ⇓((Model.mk r v),w ⊨ φ → ⋄φ)) :
+    Std.Refl r where
+  refl := by
+    intro w
+    rcases instAtomNonempty with ⟨a⟩
```
thomaskwaring: this can be [Classical.arbitrary](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Logic/Nonempty.html#Classical.arbitrary), if you want

Cslib/Logics/Modal/Cube.lean
```diff
@@ -0,0 +1,138 @@
...
+@[simp, scoped grind =]
+def Four World Atom := logic {m : Model World Atom | IsTrans World m.r}
+
+/-- The modal logic 5. -/
+@[simp, scoped grind =]
+def Five World Atom := logic {m : Model World Atom | Relation.RightEuclidean m.r}
+
+/-- The modal logic K45. -/
+@[simp, scoped grind =]
+def K45 World Atom := (K World Atom) ⊔ ((Four World Atom) ⊔ (Five World Atom))
```
thomaskwaring: is the canonical spelling of this to use `∪` for `Set.union` rather than "sup"? also are the parentheses (ie associating to the right) meaningful? ditto below

Cslib/Logics/Modal/Cube.lean
```diff
@@ -0,0 +1,138 @@
...
+
+This section proves the essential inclusions of modal logics.
+
+The other inclusions in the Modal Cube can be derived from the properties of `≤` and `⊔`, as shown
+in `k_leq_t`.
+-/
+
+open scoped Proposition
+
+theorem k_leq_d : (K World Atom ≤ D World Atom) := by
```
thomaskwaring: similarly as above, `≤` unfolds to `⊆`

Cslib/Logics/Modal/Basic.lean
fmontesi: Thanks, I didn't know, I'll add it to my list of experiments to try.
A problem I'm considering is that whatever character we choose, we'll likely have many other notations for logic and type systems where commas and semicolons are gonna be used. We're gonna run out of unicode symbols fast. :-\

Cslib/Foundations/Data/Relation.lean
chenson2018: 1. Yes, I'd just not really known which of these to really commit to
2. I don't understand the proposal, could you clarify what you would want `Reducible` and `Serial` as they currently exist in this PR to be named?

Cslib/Logics/Modal/Basic.lean
fmontesi: Maybe? I'd need to see what you need, can you give me an example? If we can make it useful outside of here, we can of course do it.

Cslib/Logics/Modal/Basic.lean
fmontesi: I don't know. How would that look like? Why would it be better?

Cslib/Foundations/Data/Relation.lean
fmontesi: > This should be

Mmh, I get an error in Cube.lean because of this now, how would you fix that?

Cslib/Logics/Modal/Cube.lean
fmontesi: Changing parentheses breaks proofs, yes (associativity changes).

I'm not sure about `∪`, since it reveals we're dealing with sets, but maybe I'm being overzealous in avoiding defeq abuse? @chenson2018 any thoughts on this?

Cslib/Foundations/Data/Relation.lean
fmontesi: 2 was wrong (I misread a def), sorry, I didn't edit my comment in time.

Cslib/Foundations/Data/Relation.lean
fmontesi: Fixed by having both versions (the instance now builds on the lemma).

Cslib/Logics/Modal/Cube.lean
thomaskwaring: i see, currently the terms have type `Set` (i think?) so it's not abuse i would have thought — if it were an alias specific to this situation then i see why you would use `sup`

Cslib/Logics/Modal/Basic.lean
thomaskwaring: actually ignore me i misread the statement

Cslib/Logics/Modal/Basic.lean
chenson2018: Unfortunately, I think we have to make some preliminary decision now. Having this notation around makes even properly reviewing this PR difficult.

Cslib/Logics/Modal/Cube.lean
chenson2018: I'm not sure about the defeq abuse question, but when I remove these parentheses I'm not seeing any errors?

Cslib/Logics/Modal/Cube.lean
```diff
@@ -0,0 +1,138 @@
...
+  intro φ; grind
+
+theorem k_leq_four : (K World Atom ≤ Four World Atom) := by
+  intro φ; grind
+
+theorem k_leq_five : (K World Atom ≤ Five World Atom) := by
+  intro φ; grind
+
+theorem d_leq_t : (D World Atom ≤ T World Atom) := by
+  intro φ; grind [Relation.refl_serial]
```
chenson2018: You should open the `Relation` scope (just for this theorem if you prefer) instead of specifying `Relation.refl_serial` manually.

(I'm still a bit skeptical of the `grind` annotation, but I really need the comma notation issue to be fixed to be able to properly look at this)

Cslib/Logics/Modal/Basic.lean
```diff
@@ -0,0 +1,285 @@
...
+
+@[inherit_doc] scoped infix:35 " ∨ " => Proposition.or
+
+/-- Implication. -/
+def Proposition.impl (φ₁ φ₂ : Proposition Atom) : Proposition Atom := ¬φ₁ ∨ φ₂
+
+@[inherit_doc] scoped infix:30 " → " => Proposition.impl
+
+/-- Bi-implication. -/
+@[simp, scoped grind =]
```
kmill: I would suggest not making this `simp`, since bi-implication is a basic connective — and it's extensionally the equality relation for propositions.

There will also be theorems about iffs, and we don't want `simp` to make them become accidentally inapplicable.

For naming, I think it would be reasonable to call this `Proposition.iff` so that it mirrors the global `Iff`. In general, I would suggest mirroring the Lean names — for example, `not` instead of `neg`.

Cslib/Logics/Modal/Basic.lean
```diff
@@ -0,0 +1,285 @@
...
+/-- Propositions. -/
+inductive Proposition (Atom : Type u) : Type u where
+  /-- Atomic proposition. -/
+  | atom (p : Atom)
+  /-- Negation. -/
+  | neg (φ : Proposition Atom)
+  /-- Conjunction. -/
+  | and (φ₁ φ₂ : Proposition Atom)
+  /-- Possibility. -/
+  | diamond (φ : Proposition Atom)
```
kmill, line 48: One thing about this is that to have `false` you need `Atom` to be nonempty, which proves to make the formalizations more complicated than necessary, with all the `[Nonempty Atom]` hypotheses.

I would strongly suggest adding `false` as an extra constructor. I would even suggest adding the rest of the operations here (`box`, `or`, and `impl`) rather than make them be derived notions. Conceptually there's something nice about starting with a minimal set of primitive propositions, but practically I believe it will be cleaner overall if we rely on the semantics given on `Satisfies`, and have some associated lemmas for relating the connectives.

This might take a bit extra boilerplate, to enable rewriting inside `Judgment.mk` using iffs, but I think overall it's worth the expense.

Cslib/Logics/Modal/Basic.lean
kmill: I'm worried about overriding core notations, even if it's scoped. It's "supported", but it can lead to confusing situations, and it also triggers a different style of elaboration that's more aggressive with giving error messages and aborting, rather than continuing with `sorry`.

Something that can work is to create a macro that translates core propositional notations into corresponding `Proposition`s. We did this with LeanLTL for LTL, and I'm happy to help develop one for modal logic. I'm imagining syntax like `Modal[φ₁ ∨ φ₂]` to stand for `Proposition.or φ₁ φ₂` and `Modal[m,w ⊨ φ₁ ∨ φ₂]` for `Judgement.mk m w (Proposition.or φ₁ φ₂)`. There can be a pretty printer to use `Modal[...]` as well.

One benefit is then you can be sure the precedences are all exactly the same as the core notations — because it *is* the core notations.

Cslib/Logics/Modal/Cube.lean
fmontesi: We had proofs not going through. I'll double check.

Cslib/Logics/Modal/Basic.lean
fmontesi: I fully agree, this slipped through a refactoring (the others don't simp). I've also called it `iff`.

Cslib/Logics/Modal/Basic.lean
fmontesi: The meanings of the derived connectives are in the respective `_char` theorems, like `impl_char`.

I'd like to take the discussion of making the inductive larger to Zulip and do it in a separate PR, because this goes in the opposite direction of comments we got for HML, so I'd like to delve into the pros and cons of making the inductive larger.

Cslib/Logics/Modal/Basic.lean
fmontesi: @chenson2018, I see (and I agree).

I'd rather put a prefix then instead of using a mystic comma, so I've gone ahead and done that. As concrete token, I've adopted the 'Modal' suggested by @kmill.

@kmill, some macro/elaborator would be very nice and in general is what we need I think, see: https://github.com/leanprover/cslib/issues/23

Let's connect about that on Zulip. For now just adding a prefix to the notation seems to work.

Cslib/Logics/Modal/Cube.lean
fmontesi: Please try again, I've updated the notation. :-)

Cslib/Logics/Modal/Basic.lean
fmontesi: I've checked things that were giving trouble because of the notation (that I was aware of, at least) now work properly.

Cslib/Logics/Modal/Cube.lean
fmontesi: For some reason the default associativity now works (I guess it's because in a previous version I was doing something weird with the World and Atom params), I'm removing the parentheses. Thanks for noticing!

Cslib/Logics/Modal/Basic.lean
thomaskwaring: cool — this is my current [Kripke model development](https://github.com/thomaskwaring/cslib_SKI/blob/kripke/Cslib/Logics/Propositional/Semantics/Kripke.lean#L21), here i've taken the partial order as an assumption on the worlds, which makes sense for the intuitionistic case, and likewise folded in the assumption that the valuation is monotonic, which is as i understand it orthogonal to the concerns of modal logic.

thomaskwaring: i think i'd be happy for this to be a refactor down the track — i'll define first algebraic semantics etc so Kripke models aren't really a priority (i want them eventually for [the disjunction property](https://github.com/thomaskwaring/cslib_SKI/blob/8e38de4b1bbf2d873c2dded1457efb429be2b313/Cslib/Logics/Propositional/Semantics/Kripke.lean#L204), but that is also waiting for some mathlib results), but perhaps we can just keep this in mind, & have some broader discussion about unifying "semantics for logic" on zulip.

fmontesi: Ah, I see what you mean. Yes, let's discuss it on Zulip, I'll post my response there.

Cslib/Logics/Modal/Basic.lean
```diff
@@ -0,0 +1,285 @@
...
+/-- If two worlds are not theory equivalent, there exists a distinguishing proposition. -/
+lemma not_theoryEq_satisfies (h : ¬TheoryEq m w₁ w₂) :
+    ∃ φ, (⇓Modal[m,w₁ ⊨ φ] ∧ ¬⇓Modal[m,w₂ ⊨ φ]) := by grind [=_ neg_satisfies]
+
+/-- If two worlds are theory equivalent and the former satisfies a proposition, the latter does as
+well. -/
+theorem theoryEq_satisfies {m : Model World Atom} (h : TheoryEq m w₁ w₂)
+    (hs : Satisfies m w₁ φ) : ⇓Modal[m,w₂ ⊨ φ] := by
+  unfold TheoryEq theory at h
+  rw [Set.ext_iff] at h
```
chenson2018: Seems like this should be extracted as a `TheoryEq.ext` lemma?

Cslib/Logics/Modal/Basic.lean
```diff
@@ -0,0 +1,285 @@
...
+  exact (h φ).mp hs
+
+/-- The K axiom, valid for all models. -/
+theorem Satisfies.k : ⇓Modal[m,w ⊨ □(φ₁ → φ₂) → (□φ₁ → □φ₂)] := by grind
+
+/-- The dual axiom, valid for all models. -/
+theorem Satisfies.dual : ⇓Modal[m,w ⊨ ⋄φ ↔ ¬□¬φ] := by
+  constructor <;> grind
```
chenson2018: The second proof here has a failing `grind?`, so I would squeeze into

```suggestion
  constructor
  · grind
  · grind only [→ satisfies_theory, usr Set.mem_setOf_eq, = impl_iff_impl, = derivation_def,
    = neg_satisfies, Satisfies, = box_iff_forall, = Set.setOf_true]
```

Cslib/Logics/Modal/Basic.lean
```diff
@@ -0,0 +1,285 @@
...
+
+/-- In any reflexive model, `□φ → φ` is equivalent to `φ → ⋄φ`. -/
+theorem Satisfies.t_box_diamond [instRefl : Std.Refl m.r] :
+    ⇓Modal[m,w ⊨ □φ → φ] ↔ ⇓Modal[m,w ⊨ φ → ⋄φ] := by
+  grind [instRefl.refl]
+
+/-- The B axiom, valid for all symmetric models. -/
+theorem Satisfies.b {m : Model World Atom} [instSymm : Std.Symm m.r] {w : World}
+    (φ : Proposition Atom) : ⇓Modal[m,w ⊨ φ → □⋄φ] := by
+  grind [instSymm.symm w]
```
chenson2018: For `grind?` reasons, let's write

```suggestion
  have := instSymm.symm w
  grind
```

Cslib/Logics/Modal/Basic.lean
```diff
@@ -0,0 +1,285 @@
...
+
+/-- Any model that admits the axiom B is symmetric. -/
+theorem Satisfies.b_symm
+    {r : World → World → Prop}
+    [instAtomNonempty : Nonempty Atom]
+    (h : ∀ {v : World → Atom → Prop} {w : World} {φ : Proposition Atom},
+      ⇓Modal[⟨r, v⟩,w ⊨ φ → □⋄φ]) :
+    Std.Symm r where
```
chenson2018: Let's format this a bit nicer:

```suggestion
theorem Satisfies.b_symm {World Atom} {r : World → World → Prop} [Nonempty Atom]
    (h : ∀ {v} {w} {φ : Proposition Atom}, ⇓Modal[⟨r, v⟩,w ⊨ φ → □⋄φ]) : Std.Symm r where
```

Cslib/Logics/Modal/Basic.lean
```diff
@@ -0,0 +1,285 @@
...
+    grind
+
+/-- The 4 axiom, valid for all transitive models. -/
+theorem Satisfies.four {m : Model World Atom} [instTrans : IsTrans World m.r] {w : World}
+    (φ : Proposition Atom) : ⇓Modal[m,w ⊨ ⋄⋄φ → ⋄φ] := by
+  simp only [impl_iff_impl]
+  intro h
+  rcases h with ⟨w', h₁, w'', h₂, hs⟩
+  exists w''
+  grind [IsTrans]
```
chenson2018: `grind` is problematic here, so let's just be explicit since it's a simple proof

```suggestion
theorem Satisfies.four {m : Model World Atom} [IsTrans World m.r] {w : World}
    (φ : Proposition Atom) : ⇓Modal[m,w ⊨ ⋄⋄φ → ⋄φ] := by
  simp only [impl_iff_impl]
  intro h
  rcases h with ⟨w', h₁, w'', h₂, hs⟩
  exact ⟨w'', IsTrans.trans _ _ _ h₁ h₂, hs⟩
```

Cslib/Logics/Modal/Basic.lean
```diff
@@ -0,0 +1,285 @@
...
+    intro w₁ w₂ w₃ h₁ h₂
+    have a := Classical.arbitrary Atom
+    let v := fun (w' : World) (a : Atom) => w' = w₃
+    let h' := h (v := v) (w := w₁) (φ := .atom a)
+    grind
+
+/-- The 5 axiom, valid for all Euclidean models. -/
+theorem Satisfies.five {m : Model World Atom} [instRightEuclidean : Relation.RightEuclidean m.r]
+    {w : World}
+    (φ : Proposition Atom) : ⇓Modal[m,w ⊨ ⋄φ → □⋄φ] := by grind [Relation.RightEuclidean]
```
chenson2018: Having `grind` work with a typeclass like this is problematic, and I'm not sure that you make use of `Relation.RightEuclidean` as a class anywhere? I think the right thing to do is split this up [like in FFL](https://github.com/FormalizedFormalLogic/Foundation/blob/master/Foundation/Vorspiel/Rel/Euclidean.lean#L12-L13) into a proposition and a class, and use the proposition for theorems like these.

Cslib/Logics/Modal/Basic.lean
```diff
@@ -0,0 +1,285 @@
...
+    [instAtomNonempty : Nonempty Atom]
+    (h : ∀ {v : World → Atom → Prop} {w : World} {φ : Proposition Atom},
+      ⇓Modal[⟨r, v⟩,w ⊨ φ → □⋄φ]) :
+    Std.Symm r where
+  symm := by
+    intro w₁
+    have a := Classical.arbitrary Atom
+    let v₁ := fun (w' : World) (a : Atom) => w' = w₁
+    let h₁ := h (v := v₁) (w := w₁) (φ := .atom a)
+    grind
```
chenson2018, line 207: This prevents any `grind?` issues:
```suggestion
    simp [impl_iff_impl] at h₁
    grind
```

chenson2018 (changes requested):
I've pointed out some places below where the issue of failing `grind?` comes up. I realize this is a bit of an annoyance for otherwise okay proofs, but it's important for maintainability (especially because I will be a bit less available in the near future).

My general recommendation is during development to use `set_option linter.tacticAnalysis.verifyGrindOnly true` to catch these.

Cslib/Logics/Modal/Basic.lean
fmontesi: I'd rather stay in line with core for now (where this kind of stuff is a class), but we can discuss it for later (this sounds relevant in general). For now I've updated the proof so that I don't pass that argument to grind.

fmontesi:
All done, thanks. :)

chenson2018 (approved):
I also pushed a couple of small style commits, so please check that these look fine.

fmontesi:
Looks good, thank you very much!

PR #529, opened 2026-04-29
ci: automatically open a PR bumping mathlib to its breaking change

marcelolynch:
The `track-incompatibility` action can open a PR pointing to a breaking change, so it can be worked on immediately by checking out the created branch


chenson2018 (approved):
Thanks!

PR #536, opened 2026-05-02
refactor(Logics/Propositional): classical and intuitionistic inference systems

thomaskwaring:
We amend the definitions of `IsClassical` and `IsIntuitionistic` in `Logics.Propositional.Defs` to refer to an inference system, rather than a theory. This makes inhabitation of these typeclasses independent of the chosen axiomatisation, so, for instance, we can define `instance instIsIntuitionisticOfIsClassical [IsClassical Atom T] : IsIntuitionistic Atom T`, which before was impossible. We describe some common alternative axiom systems for classical logic, and introduce some derived rules.


fmontesi (approved):
LGTM. Could you fix the merge conflicts, @thomaskwaring?

I'm not so sure about the naming dne, efq, etc... they are rather arcane. Any better suggestions? But we can revisit this choice later, since we're working on introducing logical operators.

thomaskwaring:
@fmontesi done, thanks! i see your point about the names, i was going for brevity, but something like `derBotImpl` and `derNotNotImpl` could work — what do you think?

PR #549, opened 2026-05-06
feat(Foundations/Data/Relation): strongly normalising elements of a type equipped with a relation

thomaskwaring:
This PR defines a predicate `SN r x` (better naming suggestions welcome) expressing that there is no infinite chain of `r`-related elements starting with `x`. This extends the definition of `Terminating` using `WellFounded` to individual elements, using `Acc`.

---

I'm very open to being told this is not a useful generalisation / abstraction, but the existing API for `Acc` does seem to make certain proofs easier. For instance, `sn_app_left` in `LambdaCalculus/LocallyNameless/Untyped/StrongNorm` falls out as reduction on `M` is a subrelation of reduction on `M.app N`.

### DIFF
--- a/Cslib/Foundations/Data/Relation.lean
+++ b/Cslib/Foundations/Data/Relation.lean
@@ -205,7 +205,7 @@ theorem Confluent.equivalence_join_reflTransGen (h : Confluent r) :
 
 /-- An element `x` is `SN` (for strongly-normalising) for a relation `r` if it is accesible under
 the inverse of `r`. -/
-abbrev SN (r : α → α → Prop) (x : α) := Acc (fun a b => r b a) x
+abbrev SN (r : α → α → Prop) := Acc (fun a b => r b a)
 
 lemma SN_iff_SN_of_rel (x : α) : SN r x ↔ ∀ y, r x y → SN r y := by grind [Acc]
 
@@ -222,15 +222,16 @@ lemma SN.of_rel_reflTransGen (hx : SN r x) (h : ReflTransGen r x y) : SN r y := 
 lemma SN.transGen (hx : SN r x) : SN (TransGen r) x := by
   have eq : TransGen (Function.swap r) = (fun a b => TransGen r b a) := by
     ext
+    exact transGen_swap
   simpa [eq] using Acc.transGen hx
 
-lemma SN.subrelation {r' : α → α → Prop} (hx : SN r x) (h : Subrelation r' r) : SN r' x := by
+lemma SN.of_le {r' : α → α → Prop} (hx : SN r x) (h : r' ≤ r) : SN r' x := by
   refine Subrelation.accessible ?_ hx
-  grind [Subrelation]
+  exact subrelation_iff_le.mpr fun {x y} => h y x
 
 @[simp]
 lemma SN.iff_transGen (x : α) : SN (TransGen r) x ↔ SN r x :=
-  ⟨fun hx => hx.subrelation TransGen.single, transGen⟩
+  ⟨fun hx => hx.of_le <| fun _ _ => TransGen.single, transGen⟩
 
 /-- `SN r x` is equivalent to the more elementary definition, that there is no infinite sequence
 of reductions starting with `x`. -/
@@ -267,10 +268,10 @@ theorem Terminating.iff_isEmpty_chain :
     Terminating r ↔ IsEmpty {f : ℕ → α // ∀ n, r (f n) (f (n + 1))} :=
   wellFounded_iff_isEmpty_descending_chain
 
-theorem Terminating.subrelation {r' : α → α → Prop} (hr : Terminating r) (h : Subrelation r' r) :
+theorem Terminating.of_le {r' : α → α → Prop} (hr : Terminating r) (h : r' ≤ r) :
     Terminating r' := by
   rw [iff_forall_sn] at hr ⊢
-  exact fun x => (hr x).subrelation h
+  exact fun x => (hr x).of_le h
 
 lemma Terminating.subtype_sn (r : α → α → Prop) :
     Terminating (α := {x // SN r x}) (fun a b => r a b) :=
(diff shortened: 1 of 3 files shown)


Cslib/Foundations/Data/Relation.lean
```diff
@@ -203,38 +203,90 @@ theorem Confluent.equivalence_join_reflTransGen (h : Confluent r) :
...
+  induction h with
+  | refl => exact hx
+  | tail _ h ih => exact ih.of_rel h
+
+lemma SN.transGen (hx : SN r x) : SN (TransGen r) x := by
+  suffices _ : (fun a b => TransGen r b a) = TransGen (Function.swap r) by
+    grind [Acc.transGen]
+  grind [transGen_swap]
+
+lemma SN.subrelation {r' : α → α → Prop} (hx : SN r x) (h : Subrelation r' r) : SN r' x := by
```
chenson2018: The consensus on Zulip seems to be that `≤` is the preferred spelling, see https://leanprover.zulipchat.com/#narrow/channel/144837-PR-reviews/topic/.2330526.20use.20Subrelation.20in.20Mathlib.2ELogic.2ERelation/with/546572621

Cslib/Foundations/Data/Relation.lean
```diff
@@ -203,38 +203,90 @@ theorem Confluent.equivalence_join_reflTransGen (h : Confluent r) :
...
+@[grind →]
+lemma SN.of_rel_reflTransGen (hx : SN r x) (h : ReflTransGen r x y) : SN r y := by
+  induction h with
+  | refl => exact hx
+  | tail _ h ih => exact ih.of_rel h
+
+lemma SN.transGen (hx : SN r x) : SN (TransGen r) x := by
+  suffices _ : (fun a b => TransGen r b a) = TransGen (Function.swap r) by
+    grind [Acc.transGen]
+  grind [transGen_swap]
```
chenson2018: I dislike nesting `suffices` and `grind` for readability reasons. I think I might forgo `grind` altogether here and just write

```suggestion
  have eq : TransGen (Function.swap r) = (fun a b => TransGen r b a) := by
    ext
  simpa [eq] using Acc.transGen hx
```

Cslib/Foundations/Data/Relation.lean
```diff
@@ -203,38 +203,90 @@ theorem Confluent.equivalence_join_reflTransGen (h : Confluent r) :
   apply equivalence_join
   grind
 
+/-- An element `x` is `SN` (for strongly-normalising) for a relation `r` if it is accesible under
+the inverse of `r`. -/
+abbrev SN (r : α → α → Prop) (x : α) := Acc (fun a b => r b a) x
```
chenson2018: I think for consistency with `Terminating` I'd write 

```suggestion
abbrev SN (r : α → α → Prop) := Acc (fun a b => r b a)
```

Do you have an opinion about us writing `fun a b => r b a` versus using `Function.swap`? It seems a bit painful either way.

Cslib/Languages/LambdaCalculus/LocallyNameless/Untyped/StrongNorm.lean
```diff
@@ -24,59 +24,52 @@ namespace LambdaCalculus.LocallyNameless.Untyped.Term
...
 
 /-- The left side of a strongly normalizing application is strongly normalizing. -/
-lemma sn_app_left (M N : Term Var) (lc_N : Term.LC N) (sn_MN : SN (M.app N)) :
-    SN M := by
-  generalize Heq : M.app N = P
-  rw [Heq] at sn_MN
-  induction sn_MN generalizing M N with grind
+lemma sn_app_left (M N : Term Var) (lc_N : Term.LC N) (sn_MN : SN FullBeta (M.app N)) :
+    SN FullBeta M := by
+  refine sn_MN.onFun_of_image (f := fun (M : Term Var) => M.app N) |>.subrelation ?_
```
chenson2018:
```suggestion
  apply sn_MN.onFun_of_image (f := (·.app N)) |>.subrelation
```

Cslib/Languages/LambdaCalculus/LocallyNameless/Untyped/StrongNorm.lean
```diff
@@ -24,59 +24,52 @@ namespace LambdaCalculus.LocallyNameless.Untyped.Term
...
 
 /-- The right side of a strongly normalizing application is strongly normalizing. -/
-lemma sn_app_right (M N : Term Var) (lc_N : Term.LC M) (sn_MN : SN (M.app N)) :
-    SN N := by
-  generalize Heq : M.app N = P
-  rw [Heq] at sn_MN
-  induction sn_MN generalizing M N with grind
+lemma sn_app_right (M N : Term Var) (lc_M : Term.LC M) (sn_MN : SN FullBeta (M.app N)) :
+    SN FullBeta N := by
+  refine sn_MN.onFun_of_image (f := fun (N : Term Var) => M.app N) |>.subrelation ?_
```
chenson2018:
```suggestion
  apply sn_MN.onFun_of_image (f := (M.app ·)) |>.subrelation
```

chenson2018 (changes requested):
I think this makes sense and is consistent with how we defined `Terminating`. I had in mind something pretty similar and hadn't found the time, so thanks!

Just a few mostly style reviews:

Cslib/Foundations/Data/Relation.lean
thomaskwaring: yep that's much nicer for sure

Cslib/Foundations/Data/Relation.lean
thomaskwaring: i agree, no particular opinion i went with what you had for terminating (`flip` is also an option)

Cslib/Foundations/Data/Relation.lean
thomaskwaring: ok thanks

chenson2018 (approved):
Looks good, thanks for doing this!

Cslib/Foundations/Data/Relation.lean
chenson2018: Yeah, let's leave this for now. It is somewhat dissatisfying that we do all this work just to change the direction that the relation goes...

PR #550, opened 2026-05-06
Start general definitions for computation models

kesslermaximilian:
This is a draft so far and should accompany the discussion on zulip.
If we think this design pattern is useful and should be part of cslib, I'll extend these definitiions and also add appropriate documentation etc.

### DIFF
--- a/Cslib/Computability/Machines/ComputationModel/Basic.lean
+++ b/Cslib/Computability/Machines/ComputationModel/Basic.lean
@@ -75,6 +75,7 @@ and taking output from this computation context.
 -/
 class TransitionMachine (τ : Type u) (Γᵢ Γₒ : outParam (Type v)) extends TransitionSystem τ where
   init {t : τ} : List Γᵢ → cfg t
+  -- TODO: Potentially work with a partial function here instead of `Option`?
   output {t : τ} : cfg t → Option (List Γₒ)
 
 namespace TransitionSystem
@@ -84,13 +85,16 @@ namespace TransitionSystem
 variable {τ : Type u} [TransitionSystem τ]
 
 def EvalsTo (t : τ) (a b : cfg t) := Relation.ReflTransGen red a b
+
 /--
-A "proof" that `t` reaches `b` from `a` in at most `n` steps, remembering the specific number
-of steps.
+The machine `t` reaches `b` from `a` in at most `n` steps.
 -/
 def EvalsToInTime (t : τ) (a b : cfg t) (n : ℕ) := Relation.RelatesWithinSteps red a b n
 
 variable {t : τ} {a b c : cfg t} {n n₁ n₂ : ℕ}
+
+-- TODO: Potentially get rid of these definitions and work with `RelatesWithinSteps` directly
+-- in proof contexts?
 
 lemma EvalsTo.refl : EvalsTo t a a := Relation.ReflTransGen.refl
 
@@ -143,6 +147,8 @@ def OutputsInTime.of_le {t : τ} {n m : ℕ} {l : List Γᵢ} {l' : List Γₒ} 
   output_eq := hv.output_eq
 
 
+-- TODO: This is only true for deterministic machines.
+-- Add the corresponding typeclass and then proof this statement
 /-
 The output of any computation of transition machines is unique.
 lemma OutputsInTime.output_unique {t : τ} {n₁ n₂ : ℕ} {l : List Γᵢ} {l'₁ l'₂ : List Γₒ}
@@ -171,7 +177,8 @@ lemma OutputsInTime.output_unique {t : τ} {n₁ n₂ : ℕ} {l : List Γᵢ} {l
 variable (τ) in
 /--
 "Proof" that `f` is computable by the system `τ`.
-A witness machine is bundled as part of this structure.
+A witness machine is bundled as part of this structure,
+as well as a function bounding execution time.
 -/
 structure TimeComputable (f : List Γᵢ → List Γₒ) where
   t : τ
@@ -181,7 +188,7 @@ structure TimeComputable (f : List Γᵢ → List Γₒ) where
 variable (τ) in
 /--
 "Proof" that `f` is computable in polynomial time by the system `τ`.
-A witness machine and polynomial are bundled as part of this structure.
+A witness machine and polynomial which bounds runtime are bundled as part of this structure.
 -/
 structure PolyTimeComputable (f : List Γᵢ → List Γₒ) extends TimeComputable τ f where
   poly : Polynomial ℕ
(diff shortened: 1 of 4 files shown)


kesslermaximilian:
THe corresponding topic on zulip is [#CSLib > Typeclasses for computation models](https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/Typeclasses.20for.20computation.20models/with/593270174)

Cslib/Computability/Machines/ComputationModel/Basic.lean
```diff
@@ -0,0 +1,124 @@
...
+
+namespace Turing
+
+/-- Bundle of a type `cfg` with a step function `cfg → Option cfg`. -/
+class TransitionSystem (τ : Type*) where
+  cfg (a : τ) : Type*
+  step {a : τ} : cfg a → Option (cfg a)
+
+/-- An abstract version of a turing machine with input alphabet `Γ₀` and output alphabet `Γ₁`. -/
+class Transducer (τ : Type*) (Γᵢₙ Γₒᵤₜ : Type) extends TransitionSystem τ where
```
crei: At least on github web, `Γₒᵤₜ`  renders super weirdly (at least in the monospace font). So I take my suggestion from zulip back a bit. Maybe `Γᵢ` and `Γₒ` ? Please also update the comment above.

kesslermaximilian: Yes, that is true, and I changed it accordingly

ctchou (commented):
In general, I find it difficult to review basic definitions like these in isolation.  We need to see them "in action" before we can judge their quality.  I would recommend that they be used to at least express some nontrivial results before we accept them.

Cslib/Computability/Machines/ComputationModel/Basic.lean
```diff
@@ -0,0 +1,195 @@
...
+
+namespace Computation
+
+/--
+For each element `(t : τ)`, there is a bundle of a type `cfg t` with a step / transition function
+`cfg t → Option (cfg t)`.
+-/
+class TransitionSystem (τ : Type u) where
+  cfg (t : τ) : Type*
+  step {t : τ} : cfg t → Option (cfg t)
```
ctchou: First, I don't understand why you need the `cfg` field.  What's wrong with dropping `cfg` and replacing every `cfg t` by `τ`?  In general, it's more hassle dealing with a family of types than just a single type.

Second, there is already a theory of partial function is mathlib:
https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/PFun.html
Any reason you want to develop your own theory of partial functions?

kesslermaximilian:
> In general, I find it difficult to review basic definitions like these in isolation. We need to see them "in action" before we can judge their quality. I would recommend that they be used to at least express some nontrivial results before we accept them.

Yes, that is very fair. These definitions stem from a previous project of mine (which was based on mathlib and I'm trying to adapt to CSLib while reworking / polishing), and you can find it here: https://git.abstractnonsen.se/max/mathlib4

In particular, examples of the code in action are the instances in this file here: https://git.abstractnonsen.se/max/mathlib4/src/commit/0f4c50edbe7ac8b5f92fee9f514b4385446c4bd6/Mathlib/Computability/TM0Computable.lean#L82-L86, https://git.abstractnonsen.se/max/mathlib4/src/commit/0f4c50edbe7ac8b5f92fee9f514b4385446c4bd6/Mathlib/Computability/TM0Computable.lean#L164-L168, which describe how different (pre-existing) variants of turing machines can be viewed as these typeclasses.
This allows to state things like https://git.abstractnonsen.se/max/mathlib4/src/commit/0f4c50edbe7ac8b5f92fee9f514b4385446c4bd6/Mathlib/Computability/TM0Computable.lean#L294-L297 (that the two turing models are equivalent up to polynomial time factors), and use it for things like stating the definition of P (https://git.abstractnonsen.se/max/mathlib4/src/commit/0f4c50edbe7ac8b5f92fee9f514b4385446c4bd6/Mathlib/Computability/NP.lean#L166-L167) and deducing that it is independent of the specific turing model because of the equivalence: https://git.abstractnonsen.se/max/mathlib4/src/commit/0f4c50edbe7ac8b5f92fee9f514b4385446c4bd6/Mathlib/Computability/NP.lean#L228-L230

I will try to port more of my code in the near future and also demonstrate how I would envision these definitions being used in CSLib.

Cslib/Computability/Machines/ComputationModel/Basic.lean
kesslermaximilian: The idea is to think of `τ` as, for example, "the type of single tape turing machines". Then `cfg t` is the "type of configurations of the specific turing machine t`, which is (potentially) different for any specific one-tape turing machine, since they can have different state types.
So with `cfg` I wanted to capture that the "computation context / configuration" is actually dependent on the machine / system you started with.

Regarding the theory of partial functions: I don't really want to work with PFun, because what we're dealing here is a special case of it, namely that we have a well-defined domain (where we return `some` values). For me, this seemed easier to work with than `PFun`.

ctchou: The distinction between `t ` and `cfg t` is irrelevant to the theory you are developing here.  So far as I can tell, you can define and prove all you have here by using `step : T → Option T`.  As a concrete example, take a look at how I use LTS to model the transition relation of a distributed algorithm:
https://github.com/ctchou/cslib/blob/flp-work/Cslib/Computability/Distributed/FLP/Algorithm.lean
where for each distributed algorithm `a`, I define an LTS `a.LTS` representing the transition relation of `a`.  I'm sure you can do the same for Turing machines.

For that matter, a partial transition function is equivalent to a deterministic LTS:
https://api.cslib.io/docs/Cslib/Foundations/Semantics/LTS/Basic.html#Cslib.LTS.Deterministic
in which the label type is a singleton.  What you define and prove here can be defined and proved in terms of deterministic LTS.  For example, `EvalsToInTime t a b n` can be stated as something like `t.LTS.MTr a xs b ∧ xs.length ≤ n`, where `t.LTS` is the LTS representing the transition relation of the Turing machine `t` and `a` and `b` are configurations of `t`.  Note that we don't need to "lift" `a` and `b` to their `Option`-ed versions.  (Indeed, by using relational statements, I think we suspect `Option` only rarely.) For another example, the theorem `EvalsToInTime.trans` is an easy consequence of `LTS.MTr.comp`:
https://api.cslib.io/docs/Cslib/Foundations/Semantics/LTS/Basic.html#Cslib.LTS.MTr.comp
The theorem `OutputsInTime.output_unique` doesn't have a counterpart in LTS yet, but that's only because not many theorems have been proved about deterministic LTS, which can be easily remedied.

kesslermaximilian:
I will be closing this for now and re-opening a PR once I have more code examples / discussion has proceded.

PR #557, opened 2026-05-10
feat: some lemmas about Euclidean relations

chenson2018:
Some lemmas on Euclidean relations, including defining `LeftEuclidean` to accompany the existing `RightEuclidean`. I erred on the side of leaving most of these as theorems, as these are a bit problematic if all made instances.

I will note again that these existed in FFL, but this adds a bit more than is proven there.

### DIFF
--- a/Cslib/Foundations/Data/Relation.lean
+++ b/Cslib/Foundations/Data/Relation.lean
@@ -69,9 +69,116 @@ theorem MJoin.single (h : ReflTransGen r a b) : MJoin r a b := by
 /-- The relation `r` 'up to' the relation `s`. -/
 def UpTo (r s : α → α → Prop) : α → α → Prop := Comp s (Comp r s)
 
+section euclidean
+
 /-- A relation `r` is (right) Euclidean if `r a b` and `r a c` guarantee `r b c`. -/
 class RightEuclidean (r : α → α → Prop) where
   rightEuclidean : r a b → r a c → r b c
+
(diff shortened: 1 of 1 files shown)


Cslib/Foundations/Data/Relation.lean
```diff
@@ -73,6 +73,109 @@ def UpTo (r s : α → α → Prop) : α → α → Prop := Comp s (Comp r s)
...
+theorem antisymm_rightUnique [Std.Antisymm r] : Relator.RightUnique r := by
+  intros a b c ab ac
+  exact antisymm (rightEuclidean ab ac) (rightEuclidean ac ab)
+
+theorem rightUnique_antisymm (h : Relator.RightUnique r) : Std.Antisymm r where
+  antisymm _ _ ab ba := h ba (refl_range ab)
+
+end RightEuclidean
+
+namespace LeftEuclidean
```
ctchou, line 110: I wonder if the theorems in this section can be derived from those in the section `RightEucliean` using the duality between left and right Euclidean relations (namely, via `leftEuclidean_swap`).  As more theorems are added, going through the duality will probably save some work.

ctchou (approved):
LGTM, except for the comment below.

Cslib/Foundations/Data/Relation.lean
chenson2018: I wasn't sure if `to_dual` supported this use case, but if possible maybe good for a follow-up PR.

fmontesi (approved):

PR #558, opened 2026-05-11
feat(data structures): simple functional queue from Okasaki

c-cube:
Hi, this is my first attempt at contributing Lean code.

I dug out the old Okasaki, and started off with the simple queue. This should
contain a correctness proof, a complexity proof, and a mapping to a ghost list
of elements. `TimeM` doesn't seem to be enough for amortized complexity so
I added a module for that, and I also removed the `ZeroAdd` typeclass constraint
because `\N` doesn't implement it(!).

AI disclosure: all code is mine, but I had some assistance from GLM 5.1 to
tidy up proof terms and get unstuck.

### DIFF
--- a/Cslib/Algorithms/Lean/Amortized.lean
+++ b/Cslib/Algorithms/Lean/Amortized.lean
@@ -29,8 +29,6 @@ class Potential φ α extends CommRing φ, LinearOrder φ, IsStrictOrderedRing �
   /-- [Okasaki, *Purely Functional Data Structures*, 1996][okasaki1996] -/
   potential : α → φ
 
-  potentialNonNegative : forall (x : α), (potential x : φ) ≥ 0
-
 class Op α o where
   applyOp : α → o → TimeM ℕ α
 
@@ -77,7 +75,7 @@ theorem constantAmortizedCostL {α o φ : Type*}
     set potX := (Potential.potential x : φ)
     set potOpX := (Potential.potential applyOpX.ret : φ)
     set potOps2 := (Potential.potential applyOps2.ret : φ)
-    have potOpXPos := (Potential.potentialNonNegative (φ := φ) applyOpX.ret)
+    /- have potOpXPos := (Potential.potentialNonNegative (φ := φ) applyOpX.ret) -/
     ring_nf
     have jfdoit := add_le_add bound1 bound2
     ring_nf at jfdoit
--- a/Cslib/Algorithms/Lean/Amortized.lean
+++ b/Cslib/Algorithms/Lean/Amortized.lean
@@ -26,9 +26,13 @@ This complements `TimeM` in the cases where amortized costs are necessary.
 namespace Cslib.Algorithms.Lean.Amortized
 
 /-- Physicist method: a potential (lower bound on savings) defined on a
-    data structure -/
-class Potential φ α [CommRing φ] [LinearOrder φ] [IsStrictOrderedRing φ] where
-  /-- [Okasaki, *Purely Functional Data Structures*, 1996][okasaki1996] -/
+    data structure.
+    [Okasaki, *Purely Functional Data Structures*, 1996][okasaki1996] -/
+class Potential (φ α : Type*) [CommRing φ] [LinearOrder φ] [IsStrictOrderedRing φ] where
+  /-- The potential, a representation of savings accumulated (just like
+      potential energy), to be released later. In some functional data
+      structures, amortized costs allow some operations to be more expensive
+      by "using" potential previously accumulated in cheaper operations. -/
   potential : α → φ
 
 class Op α o where
--- a/Cslib/Algorithms/Lean/Amortized.lean
+++ b/Cslib/Algorithms/Lean/Amortized.lean
@@ -23,8 +23,7 @@ namespace Cslib.Algorithms.Lean.Amortized
 /-- Physicist method: a potential (lower bound on savings) defined on a
     data structure -/
 class Potential α where
-  /-- Initial potential should be 0.
-      [Okasaki, *Purely Functional Data Structures*, 1996][okasaki1996] -/
+  /-- [Okasaki, *Purely Functional Data Structures*, 1996][okasaki1996] -/
   potential : α → Nat
 
 class Op α o where
@@ -33,14 +32,6 @@ class Op α o where
 @[simp] def applyOps {α o : Type*} [Op α o] (x : α) (ops : List o)
     : TimeM ℕ α :=
   List.foldlM (fun x op => Op.applyOp x op) x ops
-
-def triples {α o : Type*} [Op α o] (x : α) (ops : List o) : List (α × o) × α :=
-  match ops with
-  | [] => ([], x)
-  | o :: ops2 =>
-    let ⟨ x2, _cost ⟩ := Op.applyOp x o
-    let ⟨ pairs, final ⟩ := triples x2 ops2
-    ((x2, o) :: pairs, final)
 
 /-- Amortized cost with the physicist's method,
     following Okasaki, chapter 5 -/
--- a/Cslib/Algorithms/Lean/Amortized.lean
+++ b/Cslib/Algorithms/Lean/Amortized.lean
@@ -25,7 +25,7 @@ namespace Cslib.Algorithms.Lean.Amortized
 
 /-- Physicist method: a potential (lower bound on savings) defined on a
     data structure -/
-class Potential φ α extends CommRing φ, PartialOrder φ, IsStrictOrderedRing φ where
+class Potential φ α extends CommRing φ, LinearOrder φ, IsStrictOrderedRing φ where
   /-- [Okasaki, *Purely Functional Data Structures*, 1996][okasaki1996] -/
   potential : α → φ
 
@@ -51,7 +51,7 @@ def amortizedCost {α o φ : Type*}
   cost over a series of operations is bounded by `k * ops.length`. -/
 theorem constantAmortizedCostL {α o φ : Type*}
     [h_op : Op α o] [h_pot : Potential φ α]
-    (k : φ) (h_k_pos : k > 0) (h_bounded : ∀ (x : α) (op : o), amortizedCost x op ≤ k)
+    (k : φ) (h_bounded : ∀ (x : α) (op : o), amortizedCost x op ≤ k)
     (x : α) (ops : List o)
     : (applyOps x ops).time
         + Potential.potential (applyOps x ops).ret - Potential.potential x
@@ -81,7 +81,6 @@ theorem constantAmortizedCostL {α o φ : Type*}
     ring_nf
     have jfdoit := add_le_add bound1 bound2
     ring_nf at jfdoit
-    exact jfdoit
     linarith
 
 end Cslib.Algorithms.Lean.Amortized
--- a/Cslib/Algorithms/Lean/Amortized.lean
+++ b/Cslib/Algorithms/Lean/Amortized.lean
@@ -35,29 +35,28 @@ class Potential (φ α : Type*) [CommRing φ] [LinearOrder φ] [IsStrictOrderedR
       by "using" potential previously accumulated in cheaper operations. -/
   potential : α → φ
 
-class Op α o where
-  applyOp : α → o → TimeM ℕ α
+abbrev Op α := α → TimeM ℕ α
 
-@[simp] def applyOps {α o : Type*} [Op α o] (x : α) (ops : List o)
+@[simp] def applyOps {α : Type*} (x : α) (ops : List (Op α))
     : TimeM ℕ α :=
-  List.foldlM (fun x op => Op.applyOp x op) x ops
+  List.foldlM (fun x op => op x) x ops
 
 /-- Amortized cost with the physicist's method,
     following Okasaki, chapter 5 -/
-def amortizedCost {α o φ : Type*}
-    [Op α o] [CommRing φ] [LinearOrder φ] [IsStrictOrderedRing φ] [Potential φ α]
-    (x : α) (op : o) : φ :=
-  Nat.cast (Op.applyOp x op).time
-    + Potential.potential (Op.applyOp x op).ret
+def amortizedCost {α φ : Type*}
+    [CommRing φ] [LinearOrder φ] [IsStrictOrderedRing φ] [Potential φ α]
+    (x : α) (op : Op α) : φ :=
+  Nat.cast (op x).time
+    + Potential.potential (op x).ret
     - Potential.potential x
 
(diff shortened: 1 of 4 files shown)


Cslib/Algorithms/Lean/FunctionalQueue/FunctionalQueue.lean
```diff
@@ -0,0 +1,237 @@
...
+
+/-!
+# Functional Queue
+
+A classic two-list queue with amortized O(1) `push` and `pop`.
+
+The representation uses two lists: a front list (for dequeue) and a back list
+(for enqueue). When the front list becomes empty, the back list is reversed
+and becomes the new front. This yields amortized O(1) operations.
+
```
sorrachai, line 24: Please define the cost model that TimeM tracks for your list algorithm.

Cslib/Algorithms/Lean/Amortized.lean
```diff
@@ -0,0 +1,43 @@
...
+This complements `TimeM` in the cases where amortized costs are necessary.
+-/
+
+@[expose] public section
+
+namespace Cslib.Algorithms.Lean.Amortized
+
+/-- Physicist method: a potential (lower bound on savings) defined on a
+    data structure -/
+class Potential α where
```
sorrachai: In general, potential function arguments do not require non-negativity of the potential.

Cslib/Algorithms/Lean/FunctionalQueue/FunctionalQueue.lean
```diff
@@ -0,0 +1,237 @@
...
+    : Amortized.amortizedCost q op ≤ 2 := by
+  simp only [Amortized.amortizedCost, Amortized.Potential.potential, Nat.sub_le_iff_le_add]
+  cases op with
+  | push x =>
+    simp only [Amortized.Op.applyOp, applyOp, potential]
+    cases h_front : q.front <;> (rw [Raw.push, Raw.rebalance, h_front] at ⊢; grind)
+  | pop =>
+    simp only [Amortized.Op.applyOp, applyOp, potential]
+    cases h_front : q.front <;> (rw [Raw.pop, h_front] at ⊢; grind [Raw.rebalance])
+
```
sorrachai, line 202: Amortized cost is nice, but not immediately usable. To provide a good API for users, can you add a theorem about the total cost of a sequence of m operations on the queue of n elements? This should be at most 2m+n.

Cslib/Algorithms/Lean/Amortized.lean
```diff
@@ -0,0 +1,43 @@
...
+
+@[simp] def applyOps {α o : Type u} [Op α o] (ops : List o) (x : α)
+    : TimeM ℕ α :=
+  List.foldlM (fun x op => Op.applyOp x op) x ops
+
+/-- Amortized cost with the physicist's method,
+    following Okasaki, chapter 5 -/
+def amortizedCost {α o : Type*} [Op α o] [Potential α] (x : α) (op : o) : ℕ :=
+  (Op.applyOp x op).time + Potential.potential (Op.applyOp x op).ret - Potential.potential x
+
```
sorrachai, line 91: Some theorems/APIs about amortized cost imply that the total cost would be nice to have.

Cslib/Algorithms/Lean/Amortized.lean
```diff
@@ -0,0 +1,43 @@
...
+@[expose] public section
+
+namespace Cslib.Algorithms.Lean.Amortized
+
+/-- Physicist method: a potential (lower bound on savings) defined on a
+    data structure -/
+class Potential α where
+  /-- non-negative potential. Initial potential should be 0.
+      [Okasaki, *Purely Functional Data Structures*, 1996][okasaki1996] -/
+  potential : α → Nat
```
sorrachai: Generic type is better. I wouldn't use Nat, as some applications may use real numbers. In some cases, we allow negative numbers.

Cslib/Algorithms/Lean/FunctionalQueue/FunctionalQueue.lean
c-cube: done, although it's probably not the most elegant way :)

Cslib/Algorithms/Lean/Amortized.lean
c-cube: can I put some sort of associated type in the potential class? like

```
class Potential a where
  domain : Zero + Add
  potential : a -> domain
```
?

or at least something that works with `linarith`/`grind`?

eric-wieser: That would be `class Potential (α) [ZeroAdd α] where`

Cslib/Algorithms/Lean/Amortized.lean
```diff
@@ -22,21 +23,58 @@ namespace Cslib.Algorithms.Lean.Amortized
...
+    + Potential.potential (applyOps x ops).ret
+    - Potential.potential x
+
+/-- If each operation's cost is bounded by `k`, then the amortized
+  cost over a series of operations is bounded by `k * ops.length`. -/
+theorem constantAmortizedCostL {α o : Type*}
+    [h_op : Op α o] [h_pot : Potential α]
+    (k : ℕ) (h_bounded : ∀ (x : α) (op : o), amortizedCost x op ≤ k)
+    (x : α) (ops : List o)
+    : amortizedCostL x ops ≤ k * ops.length
```
sorrachai: Thanks for the edit. I would state the theorem this way:

(applyOps x ops).time <= k * ops.length + Potential.potential x

This formulation explicitly relates the total cost (LHS) to the amortized cost. Users only care about the LHS, and the initial potential can be bounded depending on the application.

Cslib/Algorithms/Lean/Amortized.lean
```diff
@@ -22,21 +23,58 @@ namespace Cslib.Algorithms.Lean.Amortized
...
     following Okasaki, chapter 5 -/
-def amortizedCost {α o : Type*} [Op α o] [Potential α] (x : α) (op : o) : ℕ :=
-  (Op.applyOp x op).time + Potential.potential (Op.applyOp x op).ret - Potential.potential x
+def amortizedCost {α o : Type*} [Op α o] [Potential α]
+    (x : α) (op : o) : ℕ :=
+  (Op.applyOp x op).time
+    + Potential.potential (Op.applyOp x op).ret
+    - Potential.potential x
+
+def amortizedCostL {α o : Type*} [Op α o] [Potential α]
```
sorrachai: Not sure if this definition is needed.

Cslib/Algorithms/Lean/Amortized.lean
```diff
@@ -0,0 +1,88 @@
...
+This complements `TimeM` in the cases where amortized costs are necessary.
+-/
+
+@[expose] public section
+
+namespace Cslib.Algorithms.Lean.Amortized
+
+/-- Physicist method: a potential (lower bound on savings) defined on a
+    data structure -/
+class Potential φ α extends CommRing φ, LinearOrder φ, IsStrictOrderedRing φ where
```
eric-wieser: This should almost surely be
```suggestion
class Potential (φ α) [CommRing φ] [LinearOrder φ] [IsStrictOrderedRing φ] where
```

c-cube: ok, that's fixed.

Cslib/Algorithms/Lean/Amortized.lean
c-cube: I kept the more accurate bound but otherwise `amortizedCostL` is gone.

Cslib/Algorithms/Lean/Amortized.lean
c-cube: I needed a lot more in the end, but it's now generic.

c-cube:
this is ready for re-review

Cslib/Algorithms/Lean/FunctionalQueue/FunctionalQueue.lean
```diff
@@ -0,0 +1,256 @@
...
+  front : List α
+  back : List α
+
+/-- Well-formedness: if front is empty, back must be empty too. -/
+def invariant {α : Type u} (q : Raw.FunctionalQueue α) : Prop :=
+  q.front = [] → q.back = []
+
+/-- The logical contents of the queue: `front ++ back.reverse`. -/
+def ghostList {α : Type u} (q : FunctionalQueue α) : List α :=
+  List.append q.front q.back.reverse
```
eric-wieser: more idiomatic as
```suggestion
  q.front ++ q.back.reverse
```

Cslib/Algorithms/Lean/FunctionalQueue/FunctionalQueue.lean
```diff
@@ -0,0 +1,256 @@
...
+structure FunctionalQueue (α : Type u) where
+  front : List α
+  back : List α
+
+/-- Well-formedness: if front is empty, back must be empty too. -/
+def invariant {α : Type u} (q : Raw.FunctionalQueue α) : Prop :=
+  q.front = [] → q.back = []
+
+/-- The logical contents of the queue: `front ++ back.reverse`. -/
+def ghostList {α : Type u} (q : FunctionalQueue α) : List α :=
```
eric-wieser: Why not just call this `toList`?

Cslib/Algorithms/Lean/Amortized.lean
```diff
@@ -0,0 +1,89 @@
...
+This complements `TimeM` in the cases where amortized costs are necessary.
+-/
+
+@[expose] public section
+
+namespace Cslib.Algorithms.Lean.Amortized
+
+/-- Physicist method: a potential (lower bound on savings) defined on a
+    data structure -/
+class Potential φ α [CommRing φ] [LinearOrder φ] [IsStrictOrderedRing φ] where
```
eric-wieser:
```suggestion
class Potential (φ α : Type*) [CommRing φ] [LinearOrder φ] [IsStrictOrderedRing φ] where
```
is good style

Cslib/Algorithms/Lean/Amortized.lean
```diff
@@ -0,0 +1,89 @@
...
+  /-- [Okasaki, *Purely Functional Data Structures*, 1996][okasaki1996] -/
+  potential : α → φ
```
eric-wieser: This docstring should be readable without having to read the reference.

Cslib/Algorithms/Lean/Amortized.lean
```diff
@@ -0,0 +1,89 @@
...
+
+class Op α o where
+  applyOp : α → o → TimeM ℕ α
```
eric-wieser: This has no docstring; and this only makse sense if you are saying "for any given pair of types `α o`, there is exactly one possible operation".

Cslib/Algorithms/Lean/FunctionalQueue/FunctionalQueue.lean
```diff
@@ -0,0 +1,256 @@
...
+    ghostList (rebalance q).ret = ghostList q := by
+  obtain ⟨f, b⟩ := q
+  simp [rebalance, ghostList]
+  split <;> grind [List.reverse_append]
+
+/-- Enqueue an element. -/
+def push {α : Type u} (x : α) (q : FunctionalQueue α)
+    : TimeM ℕ (FunctionalQueue α) := do
+  TimeM.tick 1
+  rebalance ⟨ q.front, x :: q.back ⟩
```
eric-wieser, line 98: In my opinion, the `TimeM` stuff clutters the API here; especially if you find yourself wanting to execute this code, and now end up computing a bunch of counters that you have no interest in. I'd suggest something like
```suggestion
/-- Enqueue an element. -/
def push {α : Type u} (x : α) (q : FunctionalQueue α) : FunctionalQueue α :=
  rebalance ⟨q.front, x :: q.back⟩

/-- Push as an algorithm
def pushAlg {α : Type u} (x : α) (q : FunctionalQueue α)
    : TimeM ℕ (FunctionalQueue α) := do
  TimeM.tick 1
  rebalanceAlg ⟨ q.front, x :: q.back ⟩
  
@[simp] theorem run_pushAlg : (pushAlg x q).ret = pushAlg x q := sorry
```
and similar for the others.

Perhaps this needs discussion on Zulip.

Cslib/Algorithms/Lean/Amortized.lean
c-cube: how else should I write this? no typeclass at all?

eric-wieser: Yes, I think drop the typeclass entirely.

Cslib/Algorithms/Lean/FunctionalQueue/FunctionalQueue.lean
```diff
@@ -0,0 +1,256 @@
...
+universe u
+
+namespace Raw
+
+structure FunctionalQueue (α : Type u) where
+  front : List α
+  back : List α
+
+/-- Well-formedness: if front is empty, back must be empty too. -/
+def invariant {α : Type u} (q : Raw.FunctionalQueue α) : Prop :=
```
eric-wieser:
```suggestion
def Invariant {α : Type u} (q : Raw.FunctionalQueue α) : Prop :=
```
since this is a `Prop`

Cslib/Algorithms/Lean/FunctionalQueue/FunctionalQueue.lean
```diff
@@ -0,0 +1,256 @@
...
+  simp [rebalance, toList]
+  split <;> grind [List.reverse_append]
+
+/-- Enqueue an element. -/
+def push {α : Type u} (x : α) (q : FunctionalQueue α)
+    : TimeM ℕ (FunctionalQueue α) := do
+  TimeM.tick 1
+  rebalance ⟨ q.front, x :: q.back ⟩
+
+theorem pushInvariant {α : Type u} (x : α) (q : FunctionalQueue α)
```
eric-wieser: This should be called `Invariant.push`

Cslib/Algorithms/Lean/Amortized.lean
```diff
@@ -0,0 +1,93 @@
...
+-/
+
+@[expose] public section
+
+namespace Cslib.Algorithms.Lean.Amortized
+
+/-- Physicist method: a potential (lower bound on savings) defined on a
+    data structure.
+    [Okasaki, *Purely Functional Data Structures*, 1996][okasaki1996] -/
+class Potential (φ α : Type*) [CommRing φ] [LinearOrder φ] [IsStrictOrderedRing φ] where
```
eric-wieser, line 31: Note that you can change this to
```suggestion
class Potential (φ α : Type*) where
```

PR #570, opened 2026-05-17
feat(algorithms): add bubble, insertion, counting, and quicksort

Robertboy18:
I wanted to contribute more of the standard sorting algorithms alongside the existing MergeSort development. This PR adds stable bubble sort, stable insertion sort, counting sort for natural-number lists, and a duplicate-safe randomized quicksort, all using the existing TimeM style so the algorithms come with executable definitions, correctness theorems, and comparison-cost bounds.

For stability, I added a shared StableByValue predicate. The idea is that if we filter the input and output by any fixed value, we should get the same subsequence. This captures the usual list-level meaning of stability: equal elements appear in the same relative order after sorting. Bubble sort is stable because it swaps only when the next element is strictly smaller, never when two elements are equal. Insertion sort is stable because it processes the input from right to left and inserts a new element before equal elements already in the sorted tail. Counting sort is stable for List ℕ because each value’s filtered subsequence is completely determined by how many copies of that value occur.

The randomized quicksort part adds a small RandomTimeM abstraction: a PMF over timed computations. I used this because randomized algorithms do not produce one timed result; they produce a distribution over possible timed runs. The quicksort implementation samples a pivot index uniformly, removes that pivot occurrence, and partitions the rest into values below, equal to, and above the pivot. The equal bucket is not recursively sorted, so duplicates are handled directly instead of relying on a distinctness assumption.

For time complexity, the deterministic algorithms prove their expected comparison bounds directly in TimeM. For randomized quicksort, this PR proves sortedness and permutation correctness for every supported run, a quadratic worst-case comparison bound for every supported run, and the corresponding expected-time bound. It also adds reusable expected-time algebra for RandomTimeM, including the monad law for ENNReal expectations and a formal one-step expected-time recurrence for the actual quicksort implementation. I included the standard harmonic O(n log n) bound for the all-distinct rank recurrence as the next bridge toward the sharper average-case theorem.

AI usage: I used GPT-5.5 as an assistant while working on this PR, mainly for help navigating longer Lean proofs, trying proof refactors using some of the more advanced automation tactics. I reviewed the generated suggestions, ran the relevant CSLib checks locally, and take responsibility for the final code. Let me know if something looks bad or unnecessary. 

Concretely, this PR adds:

- StableByValue as a shared stability predicate for list sorting.
- Stable bubble sort with sortedness, permutation, stability, and quadratic time proofs.
- Stable insertion sort with sortedness, permutation, stability, and quadratic time proofs.
- Counting sort on List ℕ with count preservation, sortedness, permutation, stability, and exact time proofs.
- RandomTimeM, with expected-value and expected-time lemmas for randomized timed computations.
- Randomized quicksort with PMF semantics, duplicate-safe three-way partitioning, correctness for every supported run, quadratic worst-case time, expected-time bounds, and an expected-time recurrence lemma.

Validation run locally:

- lake build Cslib
- lake test
- lake lint
- lake exe checkInitImports
- lake exe lint-style
- lake exe mk_all --module

### DIFF
--- a/Cslib.lean
+++ b/Cslib.lean
@@ -1,6 +1,12 @@
 module  -- shake: keep-all
 
+public import Cslib.Algorithms.Lean.Sorting
+public import Cslib.Algorithms.Lean.BubbleSort.BubbleSort
+public import Cslib.Algorithms.Lean.CountingSort.CountingSort
+public import Cslib.Algorithms.Lean.InsertionSort.InsertionSort
 public import Cslib.Algorithms.Lean.MergeSort.MergeSort
+public import Cslib.Algorithms.Lean.QuickSort.QuickSort
+public import Cslib.Algorithms.Lean.RandomTimeM
 public import Cslib.Algorithms.Lean.TimeM
 public import Cslib.Computability.Automata.Acceptors.Acceptor
 public import Cslib.Computability.Automata.Acceptors.OmegaAcceptor
--- a/Cslib.lean
+++ b/Cslib.lean
@@ -1,12 +1,12 @@
 module  -- shake: keep-all
 
-public import Cslib.Algorithms.Lean.Sorting
 public import Cslib.Algorithms.Lean.BubbleSort.BubbleSort
 public import Cslib.Algorithms.Lean.CountingSort.CountingSort
 public import Cslib.Algorithms.Lean.InsertionSort.InsertionSort
 public import Cslib.Algorithms.Lean.MergeSort.MergeSort
 public import Cslib.Algorithms.Lean.QuickSort.QuickSort
 public import Cslib.Algorithms.Lean.RandomTimeM
+public import Cslib.Algorithms.Lean.Sorting
 public import Cslib.Algorithms.Lean.TimeM
 public import Cslib.Computability.Automata.Acceptors.Acceptor
 public import Cslib.Computability.Automata.Acceptors.OmegaAcceptor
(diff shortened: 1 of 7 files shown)


sorrachai:
The code quality is not very good. In particular, quicksort formalization is low quality.  I recommend splitting into sub PRs to review for each sorting algorithm.

Robertboy18:
Thank you for the comment, I am wondering what the quality should be like? I followed how merge sort was written for most of it, so a bit taken aback by the comment. I can split it into multiple PRs then to make it easier to review especially since for quicksort I also introduced the randomTime

sorrachai:
Let's split into one algorithm per PR so I can give specific comments in each PR. 

For quicksort, your formulation is fine, but can be improved further.  There are many ways to design a framework for analyzing randomized algorithms.  I see that you wrap TimeM inside PMF, and your quicksort algorithm starts with explicit time tracking. I recommend using the Monad transformer instead (you can wait a bit until TimeT is ready). The benefit of a monad transformer is that the algorithm statement will look identical to a normal implementation, and you can also execute it (assuming IO is good randomness), as you can see in the ongoing effort on another ongoing effort on [quicksort](https://github.com/AntoineduFresne/Master-Thesis/blob/main/ARA/QuickSort.lean).

We need a careful framework for analyzing randomized algorithms. Since there are now two working proposals for quicksort. I suggest we discuss the design choice for randomized algorithms in Zulip before the PR.

Robertboy18 (commented):
Thanks, that makes sense. I’ll split this into one PR per deterministic sorting algorithm so each can be reviewed independently.

For quicksort/randomized algorithms, I’ll hold off on including `RandomTimeM` in this PR series and start a Zulip discussion about the right framework, especially whether to use `TimeT` / a monad-transformer approach instead.

PR #572, opened 2026-05-19
feat: decidable `LcAt` and `LC`

lengyijun:
This PR makes `LcAt` and `LC` decidable.

Before:

```lean
def Term_LA0L1 : Term String := .abs (.app (.bvar 0) (.abs (.bvar 1)))

theorem LA0L1_lc : Term_LA0L1.LC := by
unfold Term_LA0L1
apply LC.abs
intros
simp [open', openRec]
apply LC.app
grind
apply LC.abs
intros
simp [open', openRec]
grind
exact ∅
exact ∅
```

After: 

```lean
theorem LA0L1_lc : LcAt 0 Term_LA0L1 := by decide
```

### DIFF
--- a/Cslib/Languages/LambdaCalculus/LocallyNameless/Untyped/LcAt.lean
+++ b/Cslib/Languages/LambdaCalculus/LocallyNameless/Untyped/LcAt.lean
@@ -27,10 +27,10 @@ variable {Var : Type u}
 
 /-- `LcAt k M` is satisfied when all bound indices of M are smaller than `k`. -/
 @[simp, scoped grind =]
-def LcAt (k : ℕ) : Term Var → Prop
+def LcAt (k : ℕ) : Term Var → Bool
 | bvar i => i < k
-| fvar _ => True
-| app t₁ t₂ => LcAt k t₁ ∧ LcAt k t₂
+| fvar _ => true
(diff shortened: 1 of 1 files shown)


lengyijun:
I don't understand what's problem of pr title

chenson2018:
> I don't understand what's problem of pr title

It had some leading whitespace.

chenson2018:
If making `LcAt` decidable, we should also use `lcAt_iff_LC` (later in this file) to also get a `Decidable` instance for `LC`.

chenson2018 (changes requested):
(see my above comments)

chenson2018:
I see you requested my review again, but I don't see a new commit. In case I was unclear, what I intended was for you to add something like
```lean
instance [HasFresh Var] [DecidableEq Var] (t : Term Var) : Decidable t.LC := by
  rw [← lcAt_iff_LC]
  infer_instance
```
in this file.

lengyijun:
@chenson2018  `DecidableEq Var` looks unnecessary  ?

chenson2018:
> @chenson2018 `DecidableEq Var` looks unnecessary ?

You're right, thanks for catching that!

chenson2018 (approved):
Looks good, thanks!

PR #574, opened 2026-05-19
feat: more lemmas on Euclidean relations

chenson2018:
A few more lemmas on Euclidean relations. Some other changes to note:
- The addition of `Relation.{cod,dom}`, mirroring what is done in Mathlib with `SetRel` but for unbundled relations. I am mulling over making a dedicated definition for restrictions of an unbundled relation, but wanted to have a few theorems to test with first.
- I change the name `RightEuclidean.refl_range` to `RightEuclidean.refl_cod` for naming consistency
- Using `Relator.LeftTotal` for the definition of `Serial`. None of the modal logic proofs change at all, as these are identical.


Cslib/Foundations/Data/Relation.lean
```diff
@@ -33,6 +37,66 @@ theorem WellFounded.iff_transGen : WellFounded (Relation.TransGen r) ↔ WellFou
 
 namespace Relation
 
+/-- Domain of a relation. -/
+def dom (r : α → α → Prop) : Set α := {a | ∃ b, r a b}
+
+/-- Codomain of a relation, aka range. -/
+def cod (r : α → α → Prop) : Set α := {b | ∃ a, r a b}
```
ctchou: Any reason to limit the relation `r` to have the same type for its first and second arguments?  None of the definitions and theorems seems to require that restriction, nor does `SetRel` make that assumption.

Cslib/Foundations/Data/Relation.lean
```diff
@@ -33,6 +37,66 @@ theorem WellFounded.iff_transGen : WellFounded (Relation.TransGen r) ↔ WellFou
...
+    ext a b
+    simp
+    grind => have : b ∈ cod r; finish
+  mpr h := by grind
+
+@[simp]
+lemma cod_inv : cod (fun a b => r b a) = dom r := rfl
+
+@[simp]
+lemma dom_inv : dom (fun a b => r b a) = cod r := rfl
```
ctchou, line 92: Perhaps we can defined something like:
```
def RelInv r a b := r b a
```
Then the `cod` versions of many of these theorems can be derived from their `dom` versions plus properties of `RelInv`.

Cslib/Foundations/Data/Relation.lean
ctchou, line 244: I have made this comment before: one should be able to derive the results about left euclidean relations from the corresponding results about right euclidean relations via duality.  I think the number of results are getting large enough for this approach to be seriously considered.

ctchou (commented):
An overall comment is that this file is getting large and probably should be split.

Cslib/Foundations/Data/Relation.lean
chenson2018: I agree, and after your previous comment I opened [this thread](https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/to_dual.20for.20left.2Fright.20Euclidean.20relations/with/594650073). There's nothing ready made for this, but I have a rough idea of how to make this work.

chenson2018:
> An overall comment is that this file is getting large and probably should be split.

Agreed. The limit in Mathlib is 1500 lines. I'll think about where best to split this as we approach that size in future PRs.

Cslib/Foundations/Data/Relation.lean
chenson2018: I originally did this because unbundled relations typically have the same type, but since this is easy to change and more general I've made this change.

Cslib/Foundations/Data/Relation.lean
chenson2018: If taking this approach, I would have used the existing `flip` or `Function.swap`. They cause some problems that I attribute to there being a mismatch between core and Mathlib API. (Though maybe there's something more technical I've missed.) In a [recent poll](https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/definitions.20for.20termination.20and.20strong.20normalization/with/593370737) this form had more support, so I've tried to standardize on this approach, though I agree it is a bit ugly looking.

fmontesi (approved):
LGTM (I shared @ctchou's concerns initially but the reasoning behind choices seems ok).

PR #575, opened 2026-05-19
feat(algorithms): add stable bubble sort

Robertboy18:
This PR adds the stable bubble sort! The algorithm works in the usual way: each pass bubbles one maximal element to the end, then recursively sorts the remaining prefix. The stability proof follows the implementation pretty directly. We only swap when the next element is strictly smaller, so equal elements are never exchanged and keep their original order.

The PR proves that `bubbleSort` returns a sorted permutation of the input, preserves equal-value order via `StableByValue`, and runs within a quadratic comparison bound.

Validation:

- `lake build Cslib`

### DIFF
--- a/Cslib.lean
+++ b/Cslib.lean
@@ -1,6 +1,8 @@
 module  -- shake: keep-all
 
+public import Cslib.Algorithms.Lean.BubbleSort.BubbleSort
 public import Cslib.Algorithms.Lean.MergeSort.MergeSort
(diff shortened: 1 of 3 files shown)


Robertboy18:
@sorrachai hi! I split #570 into #575, #576, and #577 like you suggested, and left out the randomized quicksort/RandomTimeM part. Whenever you have time, I’d appreciate a review :)

PR #582, opened 2026-05-20
feat(Foundations/Logic): Notation typeclasses and models

thomaskwaring:
A proposal for typeclasses expressing that a type of "propositions" has semantics in a type of "models".

## Main definitions

- `ModelClass α β` : objects of type `β` carry a predicate `Forces` on objects of type `α`.
- `ParamModelClass α β` : objects `M : β` carry a forcing relation parametrised over some auxiliary `Param M : Type*`
- `InterpModelClass α β` : each `M : β` has an associated ground type `Ground M`, an interpretation `interp  : α → Ground M` and a `filter : Set (Ground β)`, which induces a forcing relation as `Forces M a := interp a ∈ filter M`.
- Soundness and completeness wrt an inference system

I've tried to add enough examples to show how this might be used, but the design ought to be discussed before we settle on anything this general.

### DIFF
--- a/Cslib/Foundations/Logic/Model.lean
+++ b/Cslib/Foundations/Logic/Model.lean
@@ -1,0 +1,98 @@
+/-
+Copyright (c) 2025 Thomas Waring. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Thomas Waring
+-/
+
+module
+
+public import Cslib.Foundations.Logic.Connectives
+public import Cslib.Foundations.Logic.InferenceSystem
+public import Cslib.Logics.Modal.Basic
+
+public section
+
+namespace Cslib.Logic
+
+/-- Objects of type `β` carry a forcing relation with the proposition type `α`. -/
+class ModelClass (α : outParam (Type*)) (β : Type*) where
+  /-- `Forces b a` has the intended semantics "`a` is valid in the model `b`". -/
+  Forces : β → α → Prop
+
+scoped notation "⊨[" b "] " a => ModelClass.Forces b a
+
+/-- Objects of type `β` carry a forcing relation worlds of type `γ` and the proposition type `α`. -/
+class ParamModelClass (α : outParam (Type*)) (β : Type*) where
+  Param : β → Type*
+  /-- Forcing relation associated to each parameter. -/
+  ForcesAt (b : β) : (Param b) → α → Prop
+
+scoped notation w " ⊨[" b "] " a => ParamModelClass.ForcesAt b w a
+
+instance ParamModelClass.toModelClass {α β : Type*} [ParamModelClass α β] : ModelClass α β where
+  Forces M A := ∀ w : ParamModelClass.Param M, w ⊨[M] A
+
+/-- Bundled interpretation function. -/
+structure Interp (α β : Type*) where
+  /-- Interpret a proposition in a model. -/
+  interp : α → β
+
+/-- An `InterpModel` consists of an interpretation function, and a set specifying which
+  interpretations are considered valid. -/
+structure InterpModel (α β : Type*) extends Interp α β where
+  /-- The set of valid interpretations. -/
+  filter : Set β
+
+instance {α β : Type*} : ModelClass α (InterpModel α β) where
+  Forces M a := M.interp a ∈ M.filter
+
+namespace Interp
+
+class AlgebraicAnd {α β : Type*} [HasAnd α] [Min β] (M : Interp α β) where
+  interp_and_eq (x y : α) : M.interp (x ∧ y) = M.interp x ⊓ M.interp y
+
+class AlgebraicOr {α β : Type*} [HasOr α] [Max β] (M : Interp α β) where
+  interp_or_eq (x y : α) : M.interp (x ∨ y) = M.interp x ⊔ M.interp y
+
+class AlgebraicImpl {α β : Type*} [HasImpl α] [HImp β] (M : Interp α β) where
+  interp_impl_eq (x y : α) : M.interp (x → y) = M.interp x ⇨ M.interp y
+
+class AlgebraicNot {α β : Type*} [HasNot α] [Compl β] (M : Interp α β) where
+  interp_not_eq (x : α) : M.interp (¬ x) = (M.interp x)ᶜ
+
+end Interp
+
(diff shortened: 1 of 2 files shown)


Cslib/Foundations/Logic/Model.lean
```diff
@@ -0,0 +1,170 @@
...
+/-! # Semantics for logical systems -/
+
+public section
+
+namespace Cslib.Logic
+
+/-- Objects of type `β` carry a forcing relation with the proposition type `α`. -/
+class ModelClass (α : outParam Type*) (β : Type*) where
+  /-- `Forces b a` has the intended semantics "`a` is valid in the model `b`". -/
+  Forces : β → α → Prop
```
ctchou, line 23: I think `Entails` or `Satisfies` is a better name than `Forces`:
https://en.wikipedia.org/wiki/Double_turnstile

Cslib/Foundations/Logic/Model.lean
```diff
@@ -0,0 +1,170 @@
...
+public import Cslib.Logics.Propositional.NaturalDeduction.Basic
+
+/-! # Semantics for logical systems -/
+
+public section
+
+namespace Cslib.Logic
+
+/-- Objects of type `β` carry a forcing relation with the proposition type `α`. -/
+class ModelClass (α : outParam Type*) (β : Type*) where
```
ctchou: Do we really need `Class` in the name?

Cslib/Foundations/Logic/Model.lean
```diff
@@ -0,0 +1,170 @@
...
+  /-- `Forces b a` has the intended semantics "`a` is valid in the model `b`". -/
+  Forces : β → α → Prop
+
+scoped notation "⊨[" b "] " a => ModelClass.Forces b a
+
+/-- Objects of type `β` carry a forcing relation worlds of type `γ` and the proposition type `α`. -/
+class ParamModelClass (α : outParam Type*) (β : Type*) where
+  Param : β → Type*
+  /-- Forcing relation associated to each parameter. -/
+  ForcesAt (b : β) : (Param b) → α → Prop
```
ctchou, line 31: Same comment about `Forces` as above.

Cslib/Foundations/Logic/Model.lean
```diff
@@ -0,0 +1,170 @@
...
+
+scoped notation w " ⊨[" b "] " a => ParamModelClass.ForcesAt b w a
```
ctchou, line 33: I find the order in which the arguments appear in the notation unintuitive: `a` is a "world" in model `b`, and yet `w` appears before `b` in the notation.  Wouldn't it be better to write something like `b, w ⊨ a`?

Cslib/Foundations/Logic/Model.lean
```diff
@@ -0,0 +1,170 @@
...
+  /-- Type carrying interpretation. -/
+  Ground : β → Type*
+  /-- Interpret a proposition in a model. -/
+  interp : (b : β) → α → Ground b
+
+/-- An `InterpModel` consists of an interpretation function, and a set specifying which
+  interpretations are considered valid. -/
+class InterpModelClass (α : outParam (Type*)) (β : Type*) extends HasInterp α β where
+  /-- The set of valid interpretations. -/
+  filter (b : β) : Set (Ground b)
```
ctchou, line 49: Rather than `Set`, perhaps something like `CompleteAtomicBooleanAlgebra` will provide more generality?

Cslib/Foundations/Logic/Model.lean
```diff
@@ -0,0 +1,170 @@
...
+  interp_and_eq (M : β) (x y : α) : interp M (x ∧ y) = interp M x ⊓ interp M y
+
+class AlgebraicOr (α β : Type*) [HasInterp α β] [HasOr α] [∀ b : β, Max (Ground b)] where
+  interp_or_eq (M : β) (x y : α) : interp M (x ∨ y) = interp M x ⊔ interp M y
+
+class AlgebraicImpl (α β : Type*) [HasInterp α β] [HasImpl α] [∀ b : β, HImp (Ground b)] where
+  interp_impl_eq (M : β) (x y : α) : interp M (x → y) = interp M x ⇨ interp M y
+
+class AlgebraicNot (α β : Type*) [HasInterp α β] [HasNot α] [∀ b : β, Compl (Ground b)] where
+  interp_not_eq (M : β) (x y : α) : interp M (¬ x) = (interp M x)ᶜ
```
ctchou, line 66: I don't understand these definitions.  What are those lattice-theoretic operations like `⊓` and `⊔`?  What sort of algebraic structures do you assume `Ground b` to have?  How are `Ground b` related to `filter b`?

ctchou (commented):
General comments:
* Could you remove the prime "'" in this PR branch's name?  The prime symbol is a quote for shell and needs to be escaped in shell commands like `git checkout`.
* Why do many class names contain the word `Class`, which seems completely unnecessay?

This is a partial review.  More comments will be added later.

Cslib/Foundations/Logic/Model.lean
```diff
@@ -0,0 +1,170 @@
...
+public section
+
+namespace Cslib.Logic
+
+/-- Objects of type `β` carry a forcing relation with the proposition type `α`. -/
+class ModelClass (α : outParam Type*) (β : Type*) where
+  /-- `Forces b a` has the intended semantics "`a` is valid in the model `b`". -/
+  Forces : β → α → Prop
+
+scoped notation "⊨[" b "] " a => ModelClass.Forces b a
```
ctchou, line 25: Why not write `b ⊨ a`?  See also my comment about the notation for `ForcesAt`.

Cslib/Foundations/Logic/Model.lean
```diff
@@ -0,0 +1,170 @@
...
+
+/-- Objects of type `β` carry a forcing relation worlds of type `γ` and the proposition type `α`. -/
+class ParamModelClass (α : outParam Type*) (β : Type*) where
+  Param : β → Type*
```
ctchou, line 29: It is not clear to me why we need this generality.  Can you give a natural example of the use of `ParamModelClass`?

Note also that `ParamModelClass` can be implemented using `ModelClass` by substituting a dependent sum type for `β`, namely, `Σ b : β, Param b`, which means that the only thing that `ParamModelClass` gives you is the bundling of `Param`.

PR #584, opened 2026-05-20
refactor: use `Option State` rather than `Sum State Unit` in `LTS.totalize` construction

ctchou:
The `LTS.totalize` construction adds a "sink state" and transitions to the sink state in order to make the LTS total.  Originally the sink state is represented by the `()` in the RHS of a direct sum `State ⊕ Unit`.  This PR changes the state space of `LTS.totalize` to `Option State` instead, where the sink state is represented by `none`.  This equivalent representation is easier to read and, more importantly, avoids the confusing "sum of sums" types in NA/Concat.lean and NA.Loop.lean, where another level of constructions also use the direct sum types.


fmontesi (approved):
Makes good sense to me!

PR #587, opened 2026-05-21
feat(Foundations/Logic): Notation typeclasses and models

thomaskwaring:
A proposal for typeclasses expressing that a type of "propositions" has semantics in a type of "models".

## Main definitions

- `Models α β` : objects of type `β` carry a predicate `Satisfies` on objects of type `α`.
- `ParamModels α β` : objects `M : β` carry a satisfaction relation parametrised over some auxiliary `Param M : Type*`
- `InterpModels α β` : each `M : β` has an associated ground type `Ground M`, an interpretation `interp  : α → Ground M` and a `filter : Set (Ground β)`, which induces a forcing relation as `Satisfies M a := interp a ∈ filter M`.
- Soundness and completeness wrt an inference system

I've tried to add enough examples to show how this might be used, but the design ought to be discussed before we settle on anything this general.

### DIFF
--- a/Cslib/Foundations/Logic/Model.lean
+++ b/Cslib/Foundations/Logic/Model.lean
@@ -1,0 +1,98 @@
+/-
+Copyright (c) 2025 Thomas Waring. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Thomas Waring
+-/
+
+module
+
+public import Cslib.Foundations.Logic.Connectives
+public import Cslib.Foundations.Logic.InferenceSystem
+public import Cslib.Logics.Modal.Basic
+
+public section
+
+namespace Cslib.Logic
+
+/-- Objects of type `β` carry a forcing relation with the proposition type `α`. -/
+class ModelClass (α : outParam (Type*)) (β : Type*) where
+  /-- `Forces b a` has the intended semantics "`a` is valid in the model `b`". -/
+  Forces : β → α → Prop
+
+scoped notation "⊨[" b "] " a => ModelClass.Forces b a
+
+/-- Objects of type `β` carry a forcing relation worlds of type `γ` and the proposition type `α`. -/
+class ParamModelClass (α : outParam (Type*)) (β : Type*) where
+  Param : β → Type*
+  /-- Forcing relation associated to each parameter. -/
+  ForcesAt (b : β) : (Param b) → α → Prop
+
+scoped notation w " ⊨[" b "] " a => ParamModelClass.ForcesAt b w a
+
+instance ParamModelClass.toModelClass {α β : Type*} [ParamModelClass α β] : ModelClass α β where
+  Forces M A := ∀ w : ParamModelClass.Param M, w ⊨[M] A
+
+/-- Bundled interpretation function. -/
+structure Interp (α β : Type*) where
+  /-- Interpret a proposition in a model. -/
+  interp : α → β
+
+/-- An `InterpModel` consists of an interpretation function, and a set specifying which
+  interpretations are considered valid. -/
+structure InterpModel (α β : Type*) extends Interp α β where
+  /-- The set of valid interpretations. -/
+  filter : Set β
+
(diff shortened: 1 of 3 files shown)


Cslib/Foundations/Logic/Model.lean
```diff
@@ -0,0 +1,170 @@
...
+
+scoped notation "⊨[" b "] " a => Models.Satisfies b a
+
+/-- Objects of type `β` carry a forcing relation worlds of type `γ` and the proposition type `α`. -/
+class ParamModels (α : outParam Type*) (β : Type*) where
+  Param : β → Type*
+  /-- Forcing relation associated to each parameter. -/
+  SatisfiesAt (b : β) : (Param b) → α → Prop
+
+scoped notation w " ⊨[" b "] " a => ParamModels.SatisfiesAt b w a
```
ctchou: I still find this notation unintuitive: `w` depends on `b` and yet it appears to the left of `b`.

For that matter, why does the model `b` appear in a bracket to the right of the turnstile `⊨` in this and the notation for `Models/Satisfies`?  It seems to me standard for the model to appear to the left of the `⊨`.

Cslib/Foundations/Logic/Model.lean
```diff
@@ -0,0 +1,183 @@
...
+  sorry -- i have this in a branch
+
+def Theory.lindenbaum [DecidableEq Atom] (T : Theory Atom) : HeytingModel Atom :=
+  sorry -- usual Heyting-algebra of propositions modulo equivalence
+
+theorem Theory.lindenbaum_complete [DecidableEq Atom] {T : Theory Atom} :
+    IsCompleteModel (Proposition Atom) (HeytingModel Atom) T T.lindenbaum :=
+  sorry -- also in a branch
+
+abbrev Valuation (Atom : Type*) := Atom → Prop
```
ctchou, line 191: You have the notion of a `HeytingModel` above.  Is there a corresponding notion for classical logic?

Conversely, is `HeytingModel` really necessary?  Would an abbreviation for `Atom → H`, where `H` is a `GeneralizedHeytingAlgebra`, suffice?

I'm basically asking for a consistent treatment of the two.

ctchou (commented):
Two questions/comments:
(1) What is the distinction between `Models` and `InterpModels`?  I know the latter is a special case of the former.  But is there an example of the former which is not in the form of the latter?
(2) There is still no concrete example of how `ParamModels` is actually used.

Cslib/Foundations/Logic/Model.lean
thomaskwaring: I bundled the algebra `H` into the type because I wanted a single type `HeytingModel Atom` to capture possibly varying Heyting algebras — eg we might want a theorem saying semantics in finite Heyting algebras are complete, which is not true if you restrict to a single algebra (different to classical logic, where the single Boolean algebra `Prop` / `Bool` is enough).

In any case, these are mostly there as examples to illustrate the general design, a full treatment of semantics for propositional logic will be it's own development & certainly these questions require discussion.

benbrastmckie:
Hi @thomaskwaring — following our [Zulip discussion](https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/Propositional.20Logic/with/604219492) in the Propositional Logic thread:

I noticed this PR creates `Foundations/Logic/Connectives.lean` with semantic typeclasses (Models, ParamModels, InterpModels), where PR #648 creates the same file path with syntactic connective typeclasses (HasBot, HasImp, etc.). PR #648's description already flags this.

The content is complementary rather than conflicting — yours addresses the semantic/model side, ours addresses the syntactic/connective side. They could coexist in the same file or be split into `Connectives.lean` (syntax) and `Models.lean` (semantics). Happy to coordinate on the file organization.

Your algebraic completeness theorem over `GeneralizedHeytingAlgebra` with the `v |= T` framing is exactly the right generalization — as I mentioned on Zulip, that architecture is orthogonal to the formula type design and we should adopt it. I'd be happy to collaborate on integrating the algebraic semantics layer.

Issue #590, opened 2026-05-21
feat(Computability): Framework for encoding arbitrary types on Turing machines


Sfgangloff:
### Description:

This issue addresses the first TODO in Cslib/Computability/Turing/SingleTapeTM.lean:

"Encoding of types in lists to represent computations on arbitrary types."

Currently, TimeComputable and PolyTimeComputable are restricted to functions of type List Symbol → List Symbol. To expand this to standard computable functions (e.g., over Nat, Bool, pairs, and eventually Rat for Computable Analysis), we need a standardized way to represent arbitrary types on the TM tape.

### Proposed Architecture

I propose introducing a TapeEncodable typeclass that bundles the encoding function, the decoding function, and the proof of round-trip correctness:

```Lean
class TapeEncodable (α : Type) (Symbol : Type) where
  encode : α → List Symbol
  decode : List Symbol → Option α
  decode_encode_eq : ∀ (a : α), decode (encode a) = some a
```

### Implementation Plan

I would like to break this into a few manageable PRs:

- Phase 1 (This week): Introduce the TapeEncodable typeclass and provide the trivial instance for List Symbol.

- Phase 2: Provide basic instances (e.g., for Bool, Nat, and α × β).

- Phase 3: Generalize TimeComputable and PolyTimeComputable to accept arbitrary types α and β given [TapeEncodable α Symbol] and [TapeEncodable β Symbol].

I will start with PR 1.

PR #591, opened 2026-05-21
feat(Computability): introduce TapeEncodable class for Turing machine types

Sfgangloff:
### Description

This PR addresses phase 1 of #590 (and resolves the first TODO in `Cslib/Computability/Turing/SingleTapeTM.lean`).

It introduces the `TapeEncodable` typeclass to establish a standardized framework for encoding arbitrary types onto a `SingleTapeTM` tape. Currently, computability in CSLib is restricted to functions of type `List Symbol → List Symbol`. This typeclass is the required foundational step to allow `TimeComputable` and `PolyTimeComputable` to operate on standard computable types (like `Nat`, `Bool`, pairs, and eventually `Rat`).

### Changes

- Created `Cslib/Computability/Turing/Encoding.lean`.
- Defined `class TapeEncodable (α : Type) (Symbol : Type)`.
- Provided the trivial instance `TapeEncodable (List Symbol) Symbol` to ensure backwards compatibility with existing machines.
- Added `public import Cslib.Computability.Turing.Encoding` to `SingleTapeTM.lean`.

### Next Steps (Future PRs)
- Provide basic `TapeEncodable` instances (`Bool`, `Nat`, `α × β`).
- Refactor `TimeComputable` to accept arbitrary types `α` and `β` given `[TapeEncodable α Symbol]` and `[TapeEncodable β Symbol]`.


Cslib/Computability/Machines/SingleTapeTuring/Encoding.lean
```diff
@@ -0,0 +1,49 @@
+Here is the exact code for the new file.
```
crei: Please clean up these AI leftovers.

Cslib/Computability/Machines/SingleTapeTuring/Encoding.lean
```diff
@@ -0,0 +1,49 @@
...
+
+You can create a new file named Encoding.lean in the Cslib/Computability/Turing/ directory and copy-paste this directly into it. I've included the standard copyright header that matches the rest of the files in that folder.
+
+Lean
+/-
+Copyright (c) 2026 Silvère Gangloff. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Bolton Bailey, Pim Spelier, Daan van Gent
```
crei: I think this should just be your name.

Cslib/Computability/Machines/SingleTapeTuring/Encoding.lean
```diff
@@ -0,0 +1,49 @@
...
+-/
+
+import Mathlib.Tactic.Basic
+
+/-!
+# Turing Machine Tape Encodings
```
crei: I wonder if this should not be a bit more generic, unrelated to Turing machines. Also the class could be called "StringEncodable" (in contrast to the generic Mathlib Encodable that encodes to Nat).

crei (changes requested):
While I think this is something that is definitely needed, I'm wondering a bit how we would actually encode e.g. a pair.

What I'm getting at is that in order to make this work in a generic way, we need things like "commas" and "parentheses" and a promise that the inner types respect the parentheses.

Of course we could see this as specific to the machine type, and then it would not matter too much if it is not generic.

Cslib/Computability/Machines/SingleTapeTuring/Encoding.lean
```diff
@@ -0,0 +1,49 @@
...
+Provides a canonical way to translate back and forth between a type `α` 
+and a `List Symbol`, alongside a proof that decoding an encoded value succeeds.
+-/
+class TapeEncodable (α : Type) (Symbol : Type) where
+  /-- Translates the type into a tape-compatible list of symbols -/
+  encode : α → List Symbol
+  /-- Attempts to parse a list of symbols back into the type -/
+  decode : List Symbol → Option α
+  /-- Proof that decoding a freshly encoded value yields the original value -/
+  decode_encode_eq : ∀ (a : α), decode (encode a) = some a
```
SamuelSchlesinger: I wonder if `encode_inj : encode.Injective` would be nicer here?

Cslib/Computability/Machines/SingleTapeTuring/Encoding.lean
```diff
@@ -0,0 +1,49 @@
...
+
+/-- 
+The trivial encoding for `List Symbol` itself. 
+This ensures backward compatibility with machines that already operate 
+directly on tape strings.
+-/
+instance : TapeEncodable (List Symbol) Symbol where
+  encode := id
+  decode := some
+  decode_encode_eq _ := rfl
```
SamuelSchlesinger: I think it would be nice if we had instances for common types, which I understand is hard to do with a generic Symbol. Feels not so useful without that.

Cslib/Computability/Machines/SingleTapeTuring/Encoding.lean
Sfgangloff: @crei Thanks, that makes sense. It is true that it may be usable in other contexts.

Cslib/Computability/Machines/SingleTapeTuring/Encoding.lean
Sfgangloff: @SamuelSchlesinger I think for the case of Turing machines we want both an encoding and decoding procedure, to interpret the results. In other cases, it is true that we may want some weaker notions of encoding (say, for indexing), but I think that should be introduced at the same time as these use cases.

SamuelSchlesinger: Yes, Injective implies the existence of such a decoding, though making it explicit is fine too.

PR #596, opened 2026-05-23
feat(Foundations/Combinatorics/BooleanFunctions): Fourier expansion, Plancherel, and Parseval

SamuelSchlesinger:
Formalizes Chapter 1 of O'Donnell's "Analysis of Boolean Functions" through Plancherel and Parseval.

Defines `HammingCube n = Fin n → ZMod 2` and the space `BooleanFunction n` of real-valued functions on the cube, carrying the uniform-measure L² inner product `⟪f, g⟫ = 𝔼[f·g]` as an `InnerProductSpace ℝ`.

Defines the parity functions `χ S` and Fourier coefficients `𝓕 f S = ⟪f, χ S⟫`, then proves the Fourier expansion theorem (Theorem 1.1, existence and uniqueness) and orthonormality of the parity functions (Theorem 1.5). Plancherel and Parseval follow from packaging the parity functions as an `OrthonormalBasis` and applying Mathlib's `sum_inner_mul_inner`.

Ported from my `SamuelSchlesinger/analysis-of-boolean-functions` repo. The rest of Chapter 1 from that repository is left for a follow-up, this first PR felt rightsized.

Claude Code (with Opus 4.7 xHigh) was used as a coding assistant in the original repository, and in the port as well.

### DIFF
--- a/Cslib/Foundations/Combinatorics/BooleanFunctions/FourierExpansion.lean
+++ b/Cslib/Foundations/Combinatorics/BooleanFunctions/FourierExpansion.lean
@@ -1,0 +1,412 @@
+/-
+Copyright (c) 2026 Samuel Schlesinger. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Samuel Schlesinger
+-/
+
+module
+
+public import Cslib.Init
+public import Mathlib.Algebra.BigOperators.Expect
+public import Mathlib.Analysis.InnerProductSpace.PiL2
+public import Mathlib.Data.ZMod.Basic
+
+/-!
+# Fourier expansion of Boolean functions
+
+The cube `Fin n → ZMod 2` is paired with the `±1` encoding
+`chi : ZMod 2 → ℝ`, `b ↦ (-1)^b`; a Boolean-valued function is one with
+range `{-1, +1} ⊂ ℝ`.
+
+Real-valued functions on the cube carry the uniform-measure `L²` inner
+product `⟪f, g⟫ = 𝔼 x, f x * g x`. The parity functions
+`χ_S(x) = ∏_{i ∈ S} (-1)^(xᵢ)` form an orthonormal basis, and every function
+has a unique Fourier expansion `f(x) = ∑_S 𝓕 f S · χ_S(x)` with
+`𝓕 f S = ⟪f, χ_S⟫`.
+
+This file formalizes Chapter 1 of [ODonnell2014] up to Parseval/Plancherel.
+
+## Main definitions
+
+- `HammingCube`: the Hamming cube `Fin n → ZMod 2`.
+- `BooleanFunction`: real-valued functions on the Hamming cube, equipped with
+  the uniform-measure `L²` inner product as an `InnerProductSpace ℝ`.
+- `chi`: the `±1` encoding `ZMod 2 → ℝ` sending `0 ↦ 1` and `1 ↦ -1`.
+- `parityFun`: the parity function `χ_S : 𝔽₂ⁿ → ℝ`,
+  `(χ_S)(x) = ∏_{i ∈ S} χ(xᵢ)`.
+- `fourierCoeff`: Fourier coefficient `𝓕 f S = ⟪f, χ_S⟫`.
+- `IsBooleanValued`: the predicate that `f` takes values in `{-1, 1}`.
+
+## Main results
+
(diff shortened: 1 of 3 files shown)


Cslib/Foundations/Combinatorics/BooleanFunctions/FourierExpansion.lean
```diff
@@ -0,0 +1,412 @@
...
+
+instance : FunLike (BooleanFunction n) (HammingCube n) ℝ where
+  coe f := f
+  coe_injective' _ _ h := h
+
+@[ext]
+theorem ext {f g : BooleanFunction n} (h : ∀ x, f x = g x) : f = g :=
+  DFunLike.ext f g h
+
+@[simp] theorem zero_apply (x : HammingCube n) : (0 : BooleanFunction n) x = 0 := rfl
```
chenson2018, line 95: I think the new way to get these kind of theorems is via the classes added in https://github.com/leanprover-community/mathlib4/pull/37779/. I've just merged a Mathlib bump so that you can try this out.

Cslib/Foundations/Combinatorics/BooleanFunctions/FourierExpansion.lean
```diff
@@ -0,0 +1,412 @@
...
+
+open Finset
+open scoped BigOperators
+
+variable {n : ℕ}
+
+/-! ### The Hamming cube and the space of Boolean functions -/
+
+/-- The Hamming cube `𝔽₂ⁿ`, modelled as `Fin n → ZMod 2`. -/
+abbrev HammingCube (n : ℕ) := Fin n → ZMod 2
```
chenson2018, line 70: Could you explain why this representation is the most convenient? Both for my own understanding reviewing and the sake of documentation.

chenson2018 (commented):
A couple of small questions about definitions before I look closer at the proofs that follow.

Cslib/Foundations/Combinatorics/BooleanFunctions/FourierExpansion.lean
SamuelSchlesinger: Another natural representation would be Fin n -> Bool, which I initially used in my first draft. However, the book uses F_2 and the additive group structure on F_2 is the right one to be thinking about when we think about parities, etc. Thus, you end up translating back and forth anyways.

Further, when you get to the later chapters, we begin to generalize to arbitrary characters of abelian groups, so starting there is more illustrative for when we go to lift some of these definitions and theorems to characters of arbitrary Abelian groups.

Cslib/Foundations/Combinatorics/BooleanFunctions/FourierExpansion.lean
SamuelSchlesinger: Ooh neat, thanks for pointing it out.

SamuelSchlesinger:
> A couple of small questions about definitions before I look closer at the proofs that follow.

Responded! My one thought is that as a definition in the standalone formalization, HammingCube using F_2 makes sense, but perhaps in CSLib we want HammingCube to use Bool, define it in a more general place, and eat the translation costs here.

SamuelSchlesinger:
I'm realizing now that a lot of this stuff is in Mathlib under names I didn't know to look for, already in the arbitrary Abelian group generality. I think it would still be nice to have the O'Donnell interface, but maybe re-deriving it from the Mathlib work is a better approach.

Going to make a PR targeting this one in my personal repo to try and understand how much better that ends up being.

Edit: didn't end up helping, maybe a bridge to `AddChar` would be useful for folks downstream though.

Cslib/Foundations/Combinatorics/BooleanFunctions/FourierExpansion.lean
Shreyas4991: Can't you actually write the whole theory for arbitrary abelian groups? I always thought the book takes this path for ease of presentation and because they don't want to deal with Fourier transforms on abelian groups from the get go. See how Laci Babai introduces it here :

https://people.cs.uchicago.edu/~laci/reu02/fourier.pdf


I think for CSLib you want the fully general version.

chenson2018: There is a comment below where Samuel notes this and that at least some of this appears in Mathlib.

PR #599, opened 2026-05-26
feat: subst_refl

lengyijun:
This pr 

1. adds a new lemma: `subst_refl`

```
lemma subst_refl (m : Term Var) (x : Var) :
  m[x := fvar x] = m
```

This states that substituting a variable with its own free-variable form leaves the term unchanged.  
The proof is implemented using structural induction and `grind` tactics.  

---

2. Refactors the `close_open` lemma

The lemma:

```
lemma close_open (x : Var) (t : Term Var) (k : ℕ) (t_lc : LC t) :
  t⟦k ↜ x⟧⟦k ↝ fvar x⟧ = t
```

is rewritten with ` grind [subst_refl]`

PR summary generated by copilot


chenson2018 (approved):
Thanks!

PR #600, opened 2026-05-26
chore: use `tfae_*` tactics

chenson2018:
I wasn't aware previously that `TFAE` came with these tactics. This is a bit nicer since all these theorems are `private`.


ctchou (approved):
LGTM.

Any idea why the TFAE goals are numbered starting from 1 rather than 0 by the tactic?  It seems inconsistent with the usual numbering of `List` elements?

fmontesi (approved):

PR #602, opened 2026-05-27
chore: refactor proofs where `grind?` fails

chenson2018:
These are sources of technical debt as now reported in the [weekly linting report](https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/Weekly.20linting.20log/with/563737370). The idea is that a successful grind proof can fail to report the theorems it used via `grind?`, which means that if these proofs break across toolchains that it becomes significantly harder to repair.

Most of these are fixed by squeezing the call to grind and unsetting `linter.tacticAnalysis.verifyGrindOnly` so they no longer appear in the weekly report. Unfortunately, this can't be on by default for performance reasons, but I highly encourage using this linter when adding any grind proofs.


fmontesi:
LGTM. @ctchou can you have a quick look at the automata-related changes?

fmontesi (approved):

ctchou (approved):
LGTM.

But I do wonder if `grind` will be stable enough in the future so that we don't need to revisit this issue again.

chenson2018:
> LGTM.
> 
> But I do wonder if `grind` will be stable enough in the future so that we don't need to revisit this issue again.

The FRO has encouraged minimizing and reporting these, so it is on their radar and I think will improve over time. My suggestion in the meantime is to use the `linter.tacticAnalysis.verifyGrindOnly` option while developing and reviewing to prevent more of these from being added.

PR #603, opened 2026-05-28
feat(LTS): labeled transition systems as coalgebras

tannerduve:
This PR adds `Cslib/Foundations/Semantics/LTS/LTSCat/Coalgebra.lean`, which defines labelled transition systems with a fixed label set `L` as coalgebras of the endofunctor `F_L X = Set (L × X)` on `Type u`.

`LTS.Endo Label` is the endofunctor `F_L : Type u ⥤ Type u`. `LTS.Coalgebra Label` is the category of `F_L`-coalgebras, defined as `Endofunctor.Coalgebra (Endo Label)`, using Mathlib's `Coalgebra`. `LTS.Coalgebra.toLTSCat` is a faithful functor from the coalgebra category to `LTSCat`. 

The main result of this file is that coalgebra morphisms coincide with functional bisimulations. More precisely, we have an equivalence `graphBisimEquiv`, stating that morphisms `V ⟶ W` are in bijection with the state maps `V.V → W.V` whose graph is a functional bisimulation between the underlying LTSs

This construction does not give an equivalence between categories, as `LTSCat` is strictly larger. Its objects can have different label sets and its morphisms can rename labels via `labelMap`, drop them to silent steps, and need only be one-way simulations, while coalgebra morphisms fix the label set.

### DIFF
--- a/Cslib/Foundations/Semantics/LTS/LTSCat/Coalgebra.lean
+++ b/Cslib/Foundations/Semantics/LTS/LTSCat/Coalgebra.lean
@@ -1,0 +1,193 @@
+/-
+Copyright (c) 2026 Tanner Duve (Logical Intelligence). All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Tanner Duve
+-/
+
+module
+
+public import Cslib.Foundations.Semantics.LTS.Bisimulation
+public import Cslib.Foundations.Semantics.LTS.LTSCat.Basic
+public import Mathlib.CategoryTheory.Endofunctor.Algebra
+public import Mathlib.CategoryTheory.Types.Basic
+public import Mathlib.CategoryTheory.Functor.FullyFaithful
+
+/-!
+# Labelled transition systems as coalgebras
+
+For a label set `L`, a labelled transition system on `X` is the same datum as a
+coalgebra `α : X → Set (L × X)` for the `Set`-endofunctor `F_L X = Set (L × X)`.
+This file defines `F_L` as an endofunctor of `Type u`, exhibits LTSs as
+coalgebras via Mathlib's `CategoryTheory.Endofunctor.Coalgebra`, and proves
+that Mathlib's coalgebra morphisms coincide with functional bisimulations on
+the underlying LTSs.
+
+## Main definitions
+
+- `LTS.Endo`: the endofunctor `F_L X = Set (L × X)` on `Type u`.
+- `LTS.Coalgebra.toLTSCat`: the forgetful functor from coalgebras of `F_L` to
+  `LTSCat` that maps a functional bisimulation to its underlying simulation.
+
+## Main results
+
+- `LTS.Coalgebra.instFaithful`: `toLTSCat` is faithful.
+- `LTS.Coalgebra.graphBisimEquiv`: coalgebra morphisms `V ⟶ W` are in bijection
+  with the state maps `V.V → W.V` whose graph is a functional bisimulation
+  between the underlying LTSs.
+
+## References
+
+* [J. Rutten, *Universal coalgebra: a theory of systems*][Rutten2000]
+* [G. Winskel and M. Nielsen, *Models for concurrency*][WinskelNielsen1995]
+-/
+
+@[expose] public section
+
+namespace Cslib
+
+universe u
+
+open CategoryTheory TypeCat Endofunctor
+
+namespace LTS
+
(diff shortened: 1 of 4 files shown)


Cslib/Foundations/Semantics/LTS/LTSCat/Coalgebra.lean
```diff
@@ -0,0 +1,193 @@
...
+namespace Endo
+
+variable {Label : Type u} {X Y Z : Type u}
+
+@[simp] lemma mem_mapFn (f : X → Y) (S : Set (Label × X)) (p : Label × Y) :
+    p ∈ mapFn Label f S ↔ ∃ x : X, (p.1, x) ∈ S ∧ f x = p.2 := Iff.rfl
+
+lemma mapFn_id : mapFn Label (_root_.id : X → X) = _root_.id := by
+  funext S
+  ext ⟨l, x⟩
```
thomaskwaring, line 69:
```suggestion
  ext S ⟨l, x⟩
```
also below

Cslib/Foundations/Semantics/LTS/LTSCat/Coalgebra.lean
```diff
@@ -0,0 +1,193 @@
...
+
+universe u
+
+open CategoryTheory TypeCat Endofunctor
+
+namespace LTS
+
+/-- Action on morphisms of the LTS endofunctor: image of a set under
+`(l, x) ↦ (l, f x)`. -/
+def Endo.mapFn (Label : Type u) {X Y : Type u} (f : X → Y) (S : Set (Label × X)) :
```
thomaskwaring, line 56: a more informative name for this functor might be nice, but one doesn't immediately come to mind. the idea (as i understand it) is that the structure map `State ⟶ Endo Label State` takes a state to its one-step descendants (paired with the label for that step) — if you can find a pithy way of stating that that would be great.

Cslib/Foundations/Semantics/LTS/LTSCat/Coalgebra.lean
```diff
@@ -0,0 +1,193 @@
...
+    · exact (fish_pure pure).symm
+
+/-- `toLTSCat` is faithful: a functional bisimulation is determined by its
+underlying state map. -/
+instance : (toLTSCat Label).Faithful where
+  map_injective {_ _} _ _ h := by
+    apply Endofunctor.Coalgebra.Hom.ext
+    apply Hom.ext
+    apply Fun.ext
+    funext x
```
thomaskwaring, line 139:
```suggestion
    ext x
```

Cslib/Foundations/Semantics/LTS/LTSCat/Coalgebra.lean
```diff
@@ -0,0 +1,193 @@
...
+underlying state map of a (unique) coalgebra morphism. -/
+def homOfGraphBisim {V W : Coalgebra Label} (f : V.V → W.V)
+    (h : LTS.IsBisimulation (toLTS V) (toLTS W) (fun s t => f s = t)) : V ⟶ W where
+  f := ofHom f
+  h := by
+    apply Hom.ext
+    apply Fun.ext
+    funext s
+    ext ⟨l, t'⟩
+    simp only [Endo]
```
thomaskwaring, line 167:
```suggestion
    ext s ⟨l, t'⟩
```

Cslib/Foundations/Semantics/LTS/LTSCat/Coalgebra.lean
```diff
@@ -0,0 +1,193 @@
...
+def graphBisimEquiv {V W : Coalgebra Label} :
+    (V ⟶ W) ≃
+      { f : V.V → W.V //
+          LTS.IsBisimulation (toLTS V) (toLTS W) (fun s t => f s = t) } where
+  toFun φ := ⟨φ.f, graph_isBisimulation φ⟩
+  invFun p := homOfGraphBisim p.1 p.2
+  left_inv φ := by
+    apply Endofunctor.Coalgebra.Hom.ext
+    rfl
+  right_inv _ := rfl
```
thomaskwaring, line 187:
```suggestion
```
i think the default tactics can handle this, if you want

thomaskwaring (changes requested):
this mostly looks pretty good to me — i'm not super familiar with mathlib's category theory library, so it's possible there are more further golfs / automation / use of tactics in the style of the suggestions i have made.

PR #604, opened 2026-05-28
feat(Control/Monad): Weakest preconditions and mcvgen for free monads

tannerduve:
This PR adds `Cslib/Foundations/Control/Monad/Free/WP.lean`, which gives `FreeM` programs a `Std.Do.WP` interpretation by lifting logical handlers into the predicate-transformer monad `PredTrans`.

A logical handler `{ι : Type u} → F ι → PredTrans ps ι` assigns each operation of the effect signature `F` a `PredTrans ps` specification. Since `PredTrans ps` is a monad and `FreeM` is free, `FreeM.liftM` extends such a handler to a monad morphism `FreeM F → PredTrans ps`; this is the weakest precondition and is compositional wrt the monadic structure.

A `FreeM.WP F ps` typeclass selects a default handler for `F` and induces `WP (FreeM F) ps` and `WPMonad (FreeM F) ps` instances, so that `Std.Do.Triple` and `mvcgen` apply to `FreeM F` programs. The main coherence lemma is `FreeM.liftM_wp_eq_wp_liftM`, and `FreeM.wp_lift` exposes the one-operation case.

Handlers are provided for cslib's existing `FreeState` and `FreeReader`, along with `@[spec]` lemmas `Spec.get_FreeState`, `Spec.set_FreeState`, and `Spec.read_FreeReader` that let `mvcgen` step through state and reader programs.

`CslibTests/FreeMonadWP.lean` contains example `Triple`s discharged by `mvcgen`, covering the supplied state and reader handlers, custom effects, combined effects, and demonic choice.

The design follows Vistrup, Sammler, Jung, *Program Logics à la Carte* (POPL 2025), which uses coinductive free monads (ITrees) and Iris separation logic. Here we use inductive free monads and `Std.Do` program logic.

### DIFF
--- a/Cslib/Foundations/Control/Monad/Free/Effects.lean
+++ b/Cslib/Foundations/Control/Monad/Free/Effects.lean
@@ -7,6 +7,7 @@ Authors: Tanner Duve
 module
 
 public import Cslib.Foundations.Control.Monad.Free
+public import Cslib.Foundations.Control.Monad.Free.WP
 public import Mathlib.Control.Monad.Cont
 
 /-!
@@ -39,9 +40,13 @@ Free monad, state monad, writer monad, continuation monad
 
 @[expose] public section
 
+set_option mvcgen.warning false
+
 namespace Cslib
 
 namespace FreeM
+
+open Std.Do
 
 universe u v w w' w''
 
@@ -156,6 +161,32 @@ lemma run'_bind (x : FreeState σ α) (f : α → FreeState σ β) (s₀ : σ) :
   congr_arg Prod.fst <| run_bind _ _ _
 
 end FreeState
+
+/-- Logical handler for the state effect, induced by `Std.Do`'s `WP (StateM σ)`. -/
+def StateF.handler {σ : Type u} : LHandler (StateF σ) (.arg σ .pure) :=
+  LHandler.ofInterp (m := StateM σ) (fun _ op => FreeState.stateInterp op)
+
+instance StateF.instHasHandler {σ : Type u} :
+    HasHandler (StateF σ) (.arg σ .pure) where
+  handler := StateF.handler
+
+/-- WP of a `FreeState` program matches WP of its `StateM` interpretation. -/
+theorem StateF.wp_FreeState_eq_wp_toStateM {σ α : Type u} (comp : FreeState σ α) :
+    wp comp = wp (FreeState.toStateM comp) :=
+  wpH_ofInterp_eq_wp_liftM (m := StateM σ)
+    (fun _ op => FreeState.stateInterp op) comp
+
+/-- Hoare spec for `get` on `FreeState`. -/
+@[spec]
+theorem Spec.get_FreeState {σ : Type u} {Q : PostCond σ (.arg σ .pure)} :
+    Triple (MonadStateOf.get : FreeState σ σ) (spred(fun s => Q.1 s s)) Q := by
+  mvcgen
+
+/-- Hoare spec for `set` on `FreeState`. -/
+@[spec]
+theorem Spec.set_FreeState {σ : Type u} (s : σ) {Q : PostCond PUnit (.arg σ .pure)} :
+    Triple (MonadStateOf.set s : FreeState σ PUnit) (spred(fun _ => Q.1 ⟨⟩ s)) Q := by
+  mvcgen
 
 /-! ### Writer Monad via `FreeM` -/
 
@@ -452,6 +483,26 @@ instance instMonadWithReaderOf : MonadWithReaderOf σ (FreeReader σ) where
     simpa [withTheReader, instMonadWithReaderOf, run] using! (ih (f s) s)
 
 end FreeReader
+
+/-- Logical handler for the reader effect, induced by `Std.Do`'s `WP (ReaderM σ)`. -/
+def ReaderF.handler {σ : Type u} : LHandler (ReaderF σ) (.arg σ .pure) :=
+  LHandler.ofInterp (m := ReaderM σ) (fun _ op => FreeReader.readInterp op)
+
+instance ReaderF.instHasHandler {σ : Type u} :
+    HasHandler (ReaderF σ) (.arg σ .pure) where
+  handler := ReaderF.handler
+
+/-- WP of a `FreeReader` program matches WP of its `ReaderM` interpretation. -/
+theorem ReaderF.wp_FreeReader_eq_wp_toReaderM {σ α : Type u} (comp : FreeReader σ α) :
+    wp comp = wp (FreeReader.toReaderM comp) :=
+  wpH_ofInterp_eq_wp_liftM (m := ReaderM σ)
+    (fun _ op => FreeReader.readInterp op) comp
+
+/-- Hoare spec for `read` on `FreeReader`. -/
+@[spec]
+theorem Spec.read_FreeReader {ρ : Type u} {Q : PostCond ρ (.arg ρ .pure)} :
+    Triple (MonadReaderOf.read : FreeReader ρ ρ) (spred(fun r => Q.1 r r)) Q := by
+  mvcgen
 
 end FreeM
 
--- a/Cslib/Foundations/Control/Monad/Free/Effects.lean
+++ b/Cslib/Foundations/Control/Monad/Free/Effects.lean
@@ -163,17 +163,18 @@ lemma run'_bind (x : FreeState σ α) (f : α → FreeState σ β) (s₀ : σ) :
 end FreeState
 
 /-- Logical handler for the state effect, induced by `Std.Do`'s `WP (StateM σ)`. -/
-def StateF.handler {σ : Type u} : LHandler (StateF σ) (.arg σ .pure) :=
-  LHandler.ofInterp (m := StateM σ) (fun _ op => FreeState.stateInterp op)
+def StateF.handler {σ : Type u} :
+    {ι : Type u} → StateF σ ι → PredTrans (.arg σ .pure) ι :=
+  fun {ι} op => wp (FreeState.stateInterp (σ := σ) (α := ι) op)
 
-instance StateF.instHasHandler {σ : Type u} :
-    HasHandler (StateF σ) (.arg σ .pure) where
+instance StateF.instWP {σ : Type u} :
+    FreeM.WP (StateF σ) (.arg σ .pure) where
   handler := StateF.handler
 
 /-- WP of a `FreeState` program matches WP of its `StateM` interpretation. -/
 theorem StateF.wp_FreeState_eq_wp_toStateM {σ α : Type u} (comp : FreeState σ α) :
     wp comp = wp (FreeState.toStateM comp) :=
-  wpH_ofInterp_eq_wp_liftM (m := StateM σ)
+  liftM_wp_eq_wp_liftM (m := StateM σ)
     (fun _ op => FreeState.stateInterp op) comp
 
 /-- Hoare spec for `get` on `FreeState`. -/
@@ -485,17 +486,18 @@ instance instMonadWithReaderOf : MonadWithReaderOf σ (FreeReader σ) where
 end FreeReader
 
 /-- Logical handler for the reader effect, induced by `Std.Do`'s `WP (ReaderM σ)`. -/
-def ReaderF.handler {σ : Type u} : LHandler (ReaderF σ) (.arg σ .pure) :=
-  LHandler.ofInterp (m := ReaderM σ) (fun _ op => FreeReader.readInterp op)
+def ReaderF.handler {σ : Type u} :
+    {ι : Type u} → ReaderF σ ι → PredTrans (.arg σ .pure) ι :=
+  fun {ι} op => wp (FreeReader.readInterp (σ := σ) (α := ι) op)
 
-instance ReaderF.instHasHandler {σ : Type u} :
-    HasHandler (ReaderF σ) (.arg σ .pure) where
+instance ReaderF.instWP {σ : Type u} :
+    FreeM.WP (ReaderF σ) (.arg σ .pure) where
   handler := ReaderF.handler
 
 /-- WP of a `FreeReader` program matches WP of its `ReaderM` interpretation. -/
 theorem ReaderF.wp_FreeReader_eq_wp_toReaderM {σ α : Type u} (comp : FreeReader σ α) :
     wp comp = wp (FreeReader.toReaderM comp) :=
-  wpH_ofInterp_eq_wp_liftM (m := ReaderM σ)
+  liftM_wp_eq_wp_liftM (m := ReaderM σ)
     (fun _ op => FreeReader.readInterp op) comp
 
 /-- Hoare spec for `read` on `FreeReader`. -/
--- a/Cslib/Foundations/Control/Monad/Free/Effects.lean
+++ b/Cslib/Foundations/Control/Monad/Free/Effects.lean
@@ -163,16 +163,12 @@ lemma run'_bind (x : FreeState σ α) (f : α → FreeState σ β) (s₀ : σ) :
 end FreeState
 
 /-- Logical handler for the state effect, induced by `Std.Do`'s `WP (StateM σ)`. -/
-def StateF.handler {σ : Type u} :
-    {ι : Type u} → StateF σ ι → PredTrans (.arg σ .pure) ι :=
-  fun {ι} op => wp (FreeState.stateInterp (σ := σ) (α := ι) op)
-
 instance StateF.instWP {σ : Type u} :
     FreeM.WP (StateF σ) (.arg σ .pure) where
-  handler := StateF.handler
+  handler op := wp (FreeState.stateInterp op)
 
 /-- WP of a `FreeState` program matches WP of its `StateM` interpretation. -/
-theorem StateF.wp_FreeState_eq_wp_toStateM {σ α : Type u} (comp : FreeState σ α) :
+theorem StateF.wp_freeState_eq_wp_toStateM {σ α : Type u} (comp : FreeState σ α) :
     wp comp = wp (FreeState.toStateM comp) :=
   liftM_wp_eq_wp_liftM (m := StateM σ)
     (fun _ op => FreeState.stateInterp op) comp
@@ -486,16 +482,12 @@ instance instMonadWithReaderOf : MonadWithReaderOf σ (FreeReader σ) where
 end FreeReader
 
 /-- Logical handler for the reader effect, induced by `Std.Do`'s `WP (ReaderM σ)`. -/
-def ReaderF.handler {σ : Type u} :
-    {ι : Type u} → ReaderF σ ι → PredTrans (.arg σ .pure) ι :=
-  fun {ι} op => wp (FreeReader.readInterp (σ := σ) (α := ι) op)
-
 instance ReaderF.instWP {σ : Type u} :
     FreeM.WP (ReaderF σ) (.arg σ .pure) where
-  handler := ReaderF.handler
+  handler op := wp (FreeReader.readInterp op)
 
 /-- WP of a `FreeReader` program matches WP of its `ReaderM` interpretation. -/
-theorem ReaderF.wp_FreeReader_eq_wp_toReaderM {σ α : Type u} (comp : FreeReader σ α) :
+theorem ReaderF.wp_freeReader_eq_wp_toReaderM {σ α : Type u} (comp : FreeReader σ α) :
     wp comp = wp (FreeReader.toReaderM comp) :=
   liftM_wp_eq_wp_liftM (m := ReaderM σ)
     (fun _ op => FreeReader.readInterp op) comp
--- a/Cslib/Foundations/Control/Monad/Free/WP.lean
+++ b/Cslib/Foundations/Control/Monad/Free/WP.lean
@@ -10,6 +10,7 @@ public import Cslib.Foundations.Control.Monad.Free
 public import Std.Do.PredTrans
 public import Std.Do.WP.Basic
 public import Std.Do.WP.Monad
+public import Std.Do.WP.Adequate
 public import Std.Do.Triple
 
 /-!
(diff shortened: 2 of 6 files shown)


Cslib/Foundations/Control/Monad/Free/WP.lean
```diff
@@ -0,0 +1,190 @@
...
+  run {ι : Type u} : F ι → PredTrans ps ι
+
+namespace LHandler
+
+/-- Sum of handlers; the counterpart of the paper's `H₁ ⊕ H₂`. -/
+def sum (H₁ : LHandler F ps) (H₂ : LHandler G ps) :
+    LHandler (fun α => F α ⊕ G α) ps where
+  run op := match op with
+    | .inl x => H₁.run x
+    | .inr y => H₂.run y
```
eric-wieser:
```suggestion
  run := Sum.elim H₁.run H₂.run
```
(or maybe there are dependent types here)

Cslib/Foundations/Control/Monad/Free/WP.lean
```diff
@@ -0,0 +1,178 @@
...
+
+namespace FreeM
+
+universe u v w
+
+variable {F G : Type u → Type v} {ps : PostShape.{u}} {α β : Type u}
+
+/-- A logical effect handler: an effect handler from `F` into the predicate-transformer monad
+`PredTrans ps`. Wrapped in a structure to isolate the implicit `ι`-binder from class resolution
+at use sites. -/
```
eric-wieser: What does this last sentence mean?

tannerduve: It's a workaround for an instance-synthesis issue that came up. I think making LHandler a struct instead of a Π-type hides the implicit {ι} inside the field so typeclass search doesn't see it. I will drop this from the doctstring though it's not helpful.

Cslib/Foundations/Control/Monad/Free/WP.lean
```diff
@@ -0,0 +1,175 @@
...
+
+universe u v w
+
+variable {F G : Type u → Type v} {ps : PostShape.{u}} {α β : Type u}
+
+/-- A logical handler: an effect handler from `F` into the predicate-transformer monad
+`PredTrans ps`. -/
+structure LHandler (F : Type u → Type v) (ps : PostShape.{u}) : Type (max (u + 1) v) where
+  /-- The assignment from operations to predicate transformers. -/
+  run {ι : Type u} : F ι → PredTrans ps ι
```
eric-wieser: Now that the docstring is shorter; what's the motivation for making this a structure? What goes wrong if you make it an `abbrev`?

Cslib/Foundations/Control/Monad/Free/WP.lean
```diff
@@ -0,0 +1,175 @@
...
+  induction x with
+  | pure a => simp [wpH, FreeM.liftM, WPMonad.wp_pure]
+  | lift_bind op k ih =>
+    simp [WPMonad.wp_bind, ih]
+
+/-- Records a default logical handler for `F` at shape `ps`, enabling the global
+`WP (FreeM F) ps` instance and any `Triple`/`mvcgen` reasoning over `FreeM F`. -/
+class HasHandler (F : Type u → Type v) (ps : outParam (PostShape.{u})) where
+  /-- The default logical handler for `F`. -/
+  handler : LHandler F ps
```
eric-wieser: Concretely, I'm curious whether this extra abstraction is useful, or if this would be better as
```suggestion
  handler {ι : Type u} : F ι → PredTrans ps ι
```
replacing `wpH` with `liftM` below.

Cslib/Foundations/Control/Monad/Free/WP.lean
```diff
@@ -0,0 +1,175 @@
...
+@[spec]
+theorem Spec.get_FreeState {σ : Type u} {Q : PostCond σ (.arg σ .pure)} :
+    Triple (MonadStateOf.get : FreeState σ σ) (spred(fun s => Q.1 s s)) Q := by
+  mvcgen
+
+/-- Hoare spec for `set` on `FreeState`. -/
+@[spec]
+theorem Spec.set_FreeState {σ : Type u} (s : σ) {Q : PostCond PUnit (.arg σ .pure)} :
+    Triple (MonadStateOf.set s : FreeState σ PUnit) (spred(fun _ => Q.1 ⟨⟩ s)) Q := by
+  mvcgen
```
eric-wieser: I'd be inclined to rehome these (and the ones below) into the `Effects.lean` file, since that file seems largely to be a demo of how to use the FreeM API, and WP stuff seems like it would be a nice part of that demo.

Cslib/Foundations/Control/Monad/Free/WP.lean
tannerduve: looks like this does work as an abbrev and can just be inlined instead. i guess the structure was not necessary, I did have a confusing typeclass resolution issue previously but it was solved elsewhere. will make changes momentarily

tannerduve: I think using the abbrev is more convenient than inlining everywhere, do you disagree?

eric-wieser: I would lean towards inlining, as I suspect you can just drop a lot of theorems and definitions once you do so.

eric-wieser: But feel free to push the `abbrev` version first as an intermediate step, and I can take a look before you go all the way.

Cslib/Foundations/Control/Monad/Free/WP.lean
```diff
@@ -0,0 +1,127 @@
...
+@[simp] theorem ofInterp_apply {m : Type u → Type w} [WP m ps]
+    (interp : ∀ ι : Type u, F ι → m ι) {ι : Type u} (op : F ι) :
+    LHandler.ofInterp interp op = wp (interp ι op) := rfl
+
+end LHandler
+
+/-- Weakest-precondition interpretation of a `FreeM F α` program against a logical handler `H`.
+Defined as `FreeM.liftM` instantiated at `PredTrans ps`, the unique monad morphism
+`FreeM F → PredTrans ps` extending `H` per the universal property of `FreeM`. -/
+def wpH (H : LHandler F ps) (x : FreeM F α) : PredTrans ps α :=
```
eric-wieser: I'd suggest just inlining this? I don't think we need a second name for liftM

Cslib/Foundations/Control/Monad/Free/WP.lean
```diff
@@ -0,0 +1,127 @@
...
+    (interp : ∀ ι : Type u, F ι → m ι) (x : FreeM F α) :
+    wpH (LHandler.ofInterp interp) x = wp (x.liftM (fun {_} => interp _)) := by
+  induction x with
+  | pure a => simp [wpH, FreeM.liftM, WPMonad.wp_pure]
+  | lift_bind op k ih =>
+    simp [WPMonad.wp_bind, ih]
+
+/-- Records a default logical handler for `F` at shape `ps`, enabling the global
+`WP (FreeM F) ps` instance and any `Triple`/`mvcgen` reasoning over `FreeM F`. -/
+class HasHandler (F : Type u → Type v) (ps : outParam (PostShape.{u})) where
```
eric-wieser: I think that lean is moving away from using `Has` in class names. Perhaps this should be called `FreeM.WP` instead, since it's really a special case of WP that is more convenient for FreeM?

Cslib/Foundations/Control/Monad/Free/WP.lean
```diff
@@ -0,0 +1,127 @@
...
+universe u v w
+
+variable {F G : Type u → Type v} {ps : PostShape.{u}} {α β : Type u}
+
+/-- A logical handler: an effect handler from `F` into the predicate-transformer monad
+`PredTrans ps`. -/
+abbrev LHandler (F : Type u → Type v) (ps : PostShape.{u}) : Type (max (u + 1) v) :=
+  ∀ {ι : Type u}, F ι → PredTrans ps ι
+
+namespace LHandler
```
eric-wieser: I would be very tempted to just drop this whole section and the abbrev above, and again inline things.

Cslib/Foundations/Control/Monad/Free/WP.lean
```diff
@@ -0,0 +1,127 @@
...
+    {ι : Type u} (x : F ι) :
+    LHandler.sum H₁ H₂ (Sum.inl x : F ι ⊕ G ι) = H₁ x := rfl
+
+@[simp] theorem sum_inr (H₁ : LHandler F ps) (H₂ : LHandler G ps)
+    {ι : Type u} (y : G ι) :
+    LHandler.sum H₁ H₂ (Sum.inr y : F ι ⊕ G ι) = H₂ y := rfl
+
+/-- Derive a logical handler from an effect handler into any `[WP m ps]` monad, by composing
+with `m`'s WP. -/
+def ofInterp {m : Type u → Type w} [WP m ps]
```
eric-wieser: Finally, I'd be very tempted to inline this too

Cslib/Foundations/Control/Monad/Free/WP.lean
```diff
@@ -0,0 +1,127 @@
...
+    wpH H (x >>= f) = wpH H x >>= fun a => wpH H (f a) :=
+  liftM_bind _ x f
+
+/-- Adequacy theorem: WP via `FreeM` against an `ofInterp`-derived handler agrees with
+`Std.Do`'s WP of the `liftM` interpretation. Equivalently, two monad morphisms
+`FreeM F → PredTrans ps` extending the same handler are equal. -/
+theorem wpH_ofInterp_eq_wp_liftM
+    {m : Type u → Type w} [Monad m] [LawfulMonad m] [WPMonad m ps]
+    (interp : ∀ ι : Type u, F ι → m ι) (x : FreeM F α) :
+    wpH (LHandler.ofInterp interp) x = wp (x.liftM (fun {_} => interp _)) := by
```
eric-wieser:
```suggestion
    x.liftM (fun {ι} op => wp (interp ι op)) = wp (x.liftM (fun {_} => interp _)) := by
```
after inlining

Cslib/Foundations/Control/Monad/Free/WP.lean
```diff
@@ -0,0 +1,75 @@
...
+    x.liftM (fun {ι} op => wp (interp ι op)) =
+      wp (x.liftM (fun {_} => interp _)) := by
+  induction x with
+  | pure a => simp [WPMonad.wp_pure]
+  | lift_bind op k ih =>
+    simp [WPMonad.wp_bind, ih]
+
+/-- Records a default logical handler for `F` at shape `ps`, enabling the global
+`WP (FreeM F) ps` instance and any `Triple`/`mvcgen` reasoning over `FreeM F`. -/
+class WP (F : Type u → Type v) (ps : outParam (PostShape.{u})) where
```
eric-wieser:
```suggestion
protected class WP (F : Type u → Type v) (ps : outParam (PostShape.{u})) where
```
so that you don't need `Do.WP` below.

Cslib/Foundations/Control/Monad/Free/WP.lean
```diff
@@ -0,0 +1,75 @@
...
+variable {F : Type u → Type v} {ps : PostShape.{u}} {α β : Type u}
+
+/-- Adequacy theorem: WP via `FreeM` against a WP-derived handler agrees with
+`Std.Do`'s WP of the `liftM` interpretation. Equivalently, two monad morphisms
+`FreeM F → PredTrans ps` extending the same handler are equal. -/
+theorem liftM_wp_eq_wp_liftM
+    {m : Type u → Type w} [Monad m] [WPMonad m ps]
+    (interp : ∀ ι : Type u, F ι → m ι) (x : FreeM F α) :
+    x.liftM (fun {ι} op => wp (interp ι op)) =
+      wp (x.liftM (fun {_} => interp _)) := by
```
eric-wieser: This lemma is good, but doesn't look like the adequacy lemmas for other monads, like [Std.Do.Id.of_wp_run_eq](https://leanprover-community.github.io/mathlib4_docs/Std/Do/WP/Basic.html#Std.Do.Id.of_wp_run_eq), which are about `entails`. Are you sure we're not missing the true adequacy lemma?

Cslib/Foundations/Control/Monad/Free/WP.lean
```diff
@@ -0,0 +1,75 @@
...
+  /-- The default logical handler for `F`. -/
+  handler {ι : Type u} : F ι → PredTrans ps ι
+
+instance instStdDoWP [WP F ps] : Std.Do.WP (FreeM F) ps where
+  wp x := x.liftM WP.handler
+
+instance instWPMonadFreeM [WP F ps] : WPMonad (FreeM F) ps where
+  wp_pure _ := rfl
+  wp_bind x f := liftM_bind _ x f
+
```
eric-wieser, line 81: I think we're perhaps missing a lemma here about `FreeM.lift`? That is, `wp (FreeM.lift op) = WP.handler op`

CslibTests/FreeMonadWP.lean
```diff
@@ -0,0 +1,230 @@
...
+`a : α`. Verification must consider all possible values of `a`. -/
+inductive DemonicF : Type → Type 1 where
+  /-- Choose an element of `α`. -/
+  | choice (α : Type) : DemonicF α
+
+/-- Logical handler for `DemonicF`: the predicate transformer for `choice α` is universal
+quantification over `α`. Conjunctivity of `∀` (i.e. `∀ a, P a ∧ Q a ⊣⊢ (∀ a, P a) ∧ (∀ a, Q a)`)
+is what makes this admissible in `PredTrans`. -/
+def DemonicF.handler {ps : PostShape} : {ι : Type} → DemonicF ι → PredTrans ps ι :=
+  fun op => match op with
```
eric-wieser:
```suggestion
```
probably works fine here?

Cslib/Foundations/Control/Monad/Free/Effects.lean
```diff
@@ -157,6 +162,33 @@ lemma run'_bind (x : FreeState σ α) (f : α → FreeState σ β) (s₀ : σ) :
...
 end FreeState
 
+/-- Logical handler for the state effect, induced by `Std.Do`'s `WP (StateM σ)`. -/
+def StateF.handler {σ : Type u} :
+    {ι : Type u} → StateF σ ι → PredTrans (.arg σ .pure) ι :=
+  fun {ι} op => wp (FreeState.stateInterp (σ := σ) (α := ι) op)
+
+instance StateF.instWP {σ : Type u} :
+    FreeM.WP (StateF σ) (.arg σ .pure) where
+  handler := StateF.handler
```
eric-wieser: May as well inline things here:
```suggestion
/-- Logical handler for the state effect, induced by `Std.Do`'s `WP (StateM σ)`. -/
instance StateF.instWP {σ : Type u} :
    FreeM.WP (StateF σ) (.arg σ .pure) where
  handler op := wp (FreeState.stateInterp (σ := σ) (α := ι) op)
```
or similar. Same throoughout

Cslib/Foundations/Control/Monad/Free/WP.lean
```diff
@@ -0,0 +1,75 @@
...
+effect handler `F ι → PredTrans ps ι` to a unique monad morphism `liftM H`,
+so weakest preconditions are compositional in `FreeM`'s monadic structure. A
+`[FreeM.WP F ps]` instance plugs `FreeM F` into `Std.Do`'s `WP`/`WPMonad`/`Triple`
+infrastructure.
+
+The WP's structural rules are immediate from `liftM` being a monad morphism; the
+adequacy theorem `liftM_wp_eq_wp_liftM` — that WP-via-handler agrees with
+`Std.Do`'s WP of the `liftM` interpretation — is the same statement of uniqueness.
+
+The design follows [Vistrup, Sammler, Jung. *Program Logics à la Carte.* POPL 2025], adapted
```
eric-wieser: This is missing a link to the bibtex file.

Cslib/Foundations/Control/Monad/Free/WP.lean
```diff
@@ -0,0 +1,75 @@
...
+  | lift_bind op k ih =>
+    simp [WPMonad.wp_bind, ih]
+
+/-- Records a default logical handler for `F` at shape `ps`, enabling the global
+`WP (FreeM F) ps` instance and any `Triple`/`mvcgen` reasoning over `FreeM F`. -/
+class WP (F : Type u → Type v) (ps : outParam (PostShape.{u})) where
+  /-- The default logical handler for `F`. -/
+  handler {ι : Type u} : F ι → PredTrans ps ι
+
+instance instStdDoWP [WP F ps] : Std.Do.WP (FreeM F) ps where
```
eric-wieser:
```suggestion
instance WP.toDoWP [WP F ps] : Std.Do.WP (FreeM F) ps where
```

Cslib/Foundations/Control/Monad/Free/WP.lean
```diff
@@ -0,0 +1,75 @@
...
+  wp x := x.liftM WP.handler
+
+instance instWPMonadFreeM [WP F ps] : WPMonad (FreeM F) ps where
```
eric-wieser:
```suggestion
instance [WP F ps] : WPMonad (FreeM F) ps where
```

Cslib/Foundations/Control/Monad/Free/Effects.lean
```diff
@@ -157,6 +162,33 @@ lemma run'_bind (x : FreeState σ α) (f : α → FreeState σ β) (s₀ : σ) :
...
+
+/-- WP of a `FreeState` program matches WP of its `StateM` interpretation. -/
+theorem StateF.wp_FreeState_eq_wp_toStateM {σ α : Type u} (comp : FreeState σ α) :
```
eric-wieser: The naming convention is
```suggestion
theorem StateF.wp_freeState_eq_wp_toStateM {σ α : Type u} (comp : FreeState σ α) :
```
(and throughout)

CslibTests/FreeMonadWP.lean
```diff
@@ -0,0 +1,230 @@
...
+`a : α`. Verification must consider all possible values of `a`. -/
+inductive DemonicF : Type → Type 1 where
+  /-- Choose an element of `α`. -/
+  | choice (α : Type) : DemonicF α
+
+/-- Logical handler for `DemonicF`: the predicate transformer for `choice α` is universal
+quantification over `α`. Conjunctivity of `∀` (i.e. `∀ a, P a ∧ Q a ⊣⊢ (∀ a, P a) ∧ (∀ a, Q a)`)
+is what makes this admissible in `PredTrans`. -/
+def DemonicF.handler {ps : PostShape} : {ι : Type} → DemonicF ι → PredTrans ps ι :=
+  fun
```
eric-wieser:
```suggestion
def DemonicF.handler {ps : PostShape} : {ι : Type} → DemonicF ι → PredTrans ps ι
```
perhaps

CslibTests/FreeMonadWP.lean
```diff
@@ -0,0 +1,230 @@
...
+  | _, .tick => modify (· + 1)
+  | _, .read => MonadStateOf.get
+
+/-- Logical handler for `CounterF` induced by `interp` and `Std.Do`'s `WP (StateM Nat)`
+instance. -/
+def handler : {ι : Type} → CounterF ι → PredTrans (.arg Nat .pure) ι :=
+  fun {ι} op => wp (CounterF.interp ι op)
+
+instance : FreeM.WP CounterF (.arg Nat .pure) where
+  handler := CounterF.handler
```
eric-wieser: Another one to inline

Cslib/Foundations/Control/Monad/Free/WP.lean
```diff
@@ -0,0 +1,107 @@
...
+  liftM_lift (m := PredTrans ps) FreeM.WP.handler op
+
+/-- If `wp⟦x⟧` proves postcondition `P`, then `x` interpreted into any `WPAdequate` monad `m`
+via `interp` satisfies `Internal.Ensures P`. Follows from `liftM_wp_eq_wp_liftM` and `m`'s
+adequacy. -/
+theorem ensures_liftM_of_wp
+    {m : Type u → Type w} [Monad m] [WPMonad m ps] [WPAdequate m ps]
+    (interp : ∀ ι : Type u, F ι → m ι) (x : FreeM F α) {P : α → Prop}
+    (hwp : ⊢ₛ (x.liftM (fun {ι} op => wp (interp ι op))).apply (⇓? a => ⌜P a⌝)) :
+    Internal.Ensures P (x.liftM (fun {_} => interp _)) :=
```
eric-wieser: `Internal` makes me pause here; is this really supposed to be used in API?

eric-wieser:
Please update the PR description to reflect what's currently here.

Cslib/Foundations/Control/Monad/Free/WP.lean
```diff
@@ -0,0 +1,91 @@
...
+effect handler `F ι → PredTrans ps ι` to a unique monad morphism `liftM H`,
+so weakest preconditions are compositional in `FreeM`'s monadic structure. A
+`[FreeM.WP F ps]` instance plugs `FreeM F` into `Std.Do`'s `WP`/`WPMonad`/`Triple`
+infrastructure.
+
+The WP's structural rules are immediate from `liftM` being a monad morphism. The
+coherence lemma `liftM_wp_eq_wp_liftM` states that WP-via-handler agrees with `Std.Do`'s WP of the
+`liftM` interpretation, i.e. uniqueness of the lifted monad morphism.
+
+The design follows [VistrupSammlerJung2025], adapted from coinductive ITrees to inductive `FreeM`
```
eric-wieser: I believe you have to include a title for references to work, as `[The title of the paper][bibtexkey]`

Cslib/Foundations/Control/Monad/Free/WP.lean
```diff
@@ -0,0 +1,91 @@
...
+- `wp_lift`: `wp` of a single lifted operation is the handler applied to it.
+
+## References
+
+See [VistrupSammlerJung2025] for the original account on ITree program logics.
+-/
+
+@[expose] public section
+
+set_option mvcgen.warning false
```
eric-wieser: I assume this is no longer needed?

Cslib/Foundations/Control/Monad/Free/WP.lean
```diff
@@ -0,0 +1,90 @@
...
+  wp x := x.liftM FreeM.WP.handler
+
+instance [FreeM.WP F ps] : WPMonad (FreeM F) ps where
+  wp_pure _ := rfl
+  wp_bind x f := liftM_bind _ x f
+
+/-- `wp` of a single lifted operation is the handler applied to that operation. -/
+@[simp]
+theorem wp_lift [FreeM.WP F ps] (op : F α) :
+    wp (lift op : FreeM F α) = FreeM.WP.handler op :=
```
eric-wieser, line 85: Should we have a version of this stated with the `wp` notation? Would the lemmas be identical, or would that one be further-applied?

PR #607, opened 2026-05-29
feat(Logic): logical operators

fmontesi:
This PR introduces typeclasses for logical operators (connectives and modalities) and refactors modal and propositional logics with appropriate instances to these.

Co-authored-by: Thomas Waring <thomas.waring99@gmail.com>


Cslib/Logics/Modal/Basic.lean
```diff
@@ -164,13 +191,9 @@ theorem theoryEq_satisfies {m : Model World Atom} (h : TheoryEq m w₁ w₂)
...
 theorem Satisfies.k : ⇓Modal[m,w ⊨ □(φ₁ → φ₂) → (□φ₁ → □φ₂)] := by grind
 
-set_option linter.tacticAnalysis.verifyGrindOnly false in
 /-- The dual axiom, valid for all models. -/
 theorem Satisfies.dual : ⇓Modal[m,w ⊨ ◇φ ↔ ¬□¬φ] := by
-  constructor
-  · grind
-  · grind only [→ satisfies_theory, usr Set.mem_setOf_eq, = impl_iff_impl, = derivation_def,
-    = neg_satisfies, Satisfies, = box_iff_forall, = Set.setOf_true]
+  constructor <;> grind
```
chenson2018: This should not be undone, `grind?` still fails here.

Cslib/Logics/Modal/Basic.lean
```diff
@@ -6,7 +6,13 @@ Authors: Fabrizio Montesi, Marianna Girlando
...
 module
 
-public import Cslib.Init
+public import Cslib.Foundations.Logic.Operators.And
+public import Cslib.Foundations.Logic.Operators.Or
+public import Cslib.Foundations.Logic.Operators.Impl
+public import Cslib.Foundations.Logic.Operators.Not
+public import Cslib.Foundations.Logic.Operators.Box
+public import Cslib.Foundations.Logic.Operators.Diamond
+public import Cslib.Foundations.Logic.Operators.Iff
```
chenson2018: Would it be better to just have one file for these? If they're just notation typeclasses it seems unlikely to be heavyweight and they're likely to be used together.

Cslib/Logics/Modal/Basic.lean
```diff
@@ -47,29 +53,51 @@ inductive Proposition (Atom : Type u) : Type u where
...
 
-@[inherit_doc] scoped prefix:40 "¬" => Proposition.neg
-@[inherit_doc] scoped infix:36 " ∧ " => Proposition.and
-@[inherit_doc] scoped prefix:40 "◇" => Proposition.diamond
+instance : HasNot (Proposition Atom) := {not := Proposition.neg}
+instance : HasAnd (Proposition Atom) := {and := Proposition.and}
+instance : HasDiamond (Proposition Atom) := {diamond := Proposition.diamond}
+
+@[simp, scoped grind =]
+lemma Proposition.not_def (φ : Proposition Atom) : (¬φ) = φ.neg := rfl
```
chenson2018: Are these `simp`/`grind =` lemmas all backwards? I thought they should simplify into the notation like [List.append_eq](https://leanprover-community.github.io/mathlib4_docs/Init/Data/List/Basic.html#List.append_eq), [Nat.add_eq](https://leanprover-community.github.io/mathlib4_docs/Init/Data/Nat/Basic.html#Nat.add_eq), etc. If this causes failures, I think it means some lemmas are not properly using the notation.

chenson2018 (changes requested):

Cslib/Logics/Modal/Basic.lean
thomaskwaring: this was my sense as well, but i couldn't make the `grind` see through the notation / typeclass correctly in the proofs of `Satisfies.or_iff_or` & relatives without explicitly rewriting back to `φ.neg` or whatever (undoubtedly due to my not understanding transparency / notation / grind / something properly)

Cslib/Logics/Modal/Basic.lean
fmontesi: I don't know yet. We will expand these files with extended classes for expected properties about these operators, but you're right that some will require importing more than one.. I think we'll know more once we get there.

Cslib/Logics/Modal/Basic.lean
chenson2018: I spent some time with this. I think that if you add lemmas that make sure to use the notation for judgements, e.g. `⇓Modal[m,w ⊨ φ₁ ∧ φ₂] = (⇓Modal[m,w ⊨ φ₁] ∧ ⇓Modal[m,w ⊨ φ₂])` that this becomes a lot easier. As it is the `grind` annotation being on `Satisfies` means unfolding this is always required.

Cslib/Logics/Modal/Basic.lean
ctchou: I propose 3 files:
* `Modal` containing box and diamond.
* `Tensor` by itself.
* `Propositional` for the reset.

BTW, do we need parameterized box and diamond for HML?

Cslib/Logics/Modal/Basic.lean
fmontesi: Should be ok now.

benbrastmckie:
Hi @fmontesi — I wanted to flag some overlap with PR #648 so we can coordinate.

PR #648 includes `Connectives.lean` with a single-file approach to connective typeclasses (`HasBot`, `HasImp`, `HasAnd`, `HasOr`), following @chenson2018's consolidation suggestion. A planned Modal follow-up would extend it with `HasBox`/`ModalConnectives`.

The main naming difference is `HasImpl`/`impl` here vs `HasImp`/`imp` in #648. I went with `imp` for consistency with FormalizedFormalLogic and the rule name prefix convention (`impI`/`impE`), but happy to align whichever way you and other reviewers prefer.

A follow-up to #648 would refactor the Modal constructors from `{atom, not, and, diamond}` to `{atom, bot, imp, box}`. The case for primitive `bot` is in [this Zulip message](https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/Propositional.20Logic/near/604219492) (substitution invariance, free algebra property). Related reasoning favors primitive `box`: necessitation becomes a pure rule — `⊢ φ → ⊢ □φ` — stated in terms of `box` alone, rather than an interaction rule requiring `dia` and `neg`, or `dia`, `bot`, and `imp`. This would affect how typeclass instances are written. I'd welcome your input on this issue and will post on the [Modal Zulip](https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/Modal.20Logic/with/602410242) thread.

Cslib/Foundations/Logic/Operators/And.lean
eric-wieser, line 1: I'd suggest merging all these operators into a single `LogicOperators` file, since then you can give one docstring that summarizes the conventions of their meanings.

Cslib/Logics/Propositional/Defs.lean
```diff
@@ -67,10 +73,14 @@ instance instTopProposition [Inhabited Atom] : Top (Proposition Atom) := ⟨.top
 
 example [Bot Atom] : (⊤ : Proposition Atom) = Proposition.impl ⊥ ⊥ := rfl
 
-@[inherit_doc] scoped infix:36 " ∧ " => Proposition.and
-@[inherit_doc] scoped infix:35 " ∨ " => Proposition.or
-@[inherit_doc] scoped infix:30 " → " => Proposition.impl
-@[inherit_doc] scoped prefix:40 " ¬ " => Proposition.neg
+instance : HasAnd (Proposition Atom) := {and := Proposition.and}
```
eric-wieser: A little cleaner as
```suggestion
instance : HasAnd (Proposition Atom) where and := .and
```
etc

Cslib/Logics/LinearLogic/CLL/Basic.lean
```diff
@@ -650,7 +650,7 @@ instance : Congruence (Proposition Atom) Proposition.Equiv where
     intro ctx a b hab
     induction ctx <;> grind [= Context.fill]
 
-noncomputable instance : LogicalEquivalence (Proposition Atom) (Sequent Atom) Proof where
+noncomputable instance : HasLogicalEquivalence (Proposition Atom) (Sequent Atom) where
```
eric-wieser, line 653: Note that the `Has` prefix is largely a Lean 3-ism.

benbrastmckie:
Following up on my comment above: the overlap I flagged is now resolved on #648's side — I've removed the connective typeclasses from #648, so it no longer duplicates the classes in this PR. #648 now contributes the five-primitive `Proposition` (primitive `⊥`), its natural deduction, and an `MPL`/`IPL`/`CPL`/`IsIntuitionistic` theory layer.

That leaves one spot where the two PRs still meet: your head has since its own `Proposition` and a near-identical `MPL`/`IPL`/`CPL`/`IsIntuitionistic` development in the same file, `Propositional/Defs.lean`. To stay coordinated we'd need a single `Proposition` and one copy of that theory layer. The one substantive decision is how `⊥` is represented — here it's `Proposition = {atom, and, or, impl}` with `⊥ = atom ⊥` gated on `[Bot Atom]`; in #648 `⊥` is a primitive constructor.

I'd favor the primitive `⊥`, mainly for substitution-invariance ([Zulip](https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/Propositional.20Logic/near/604219492)): when `⊥ = atom ⊥`, falsum lies in the image of `atom`, so a substitution `Atom → Proposition` can rewrite it and the type is no longer the free algebra over the signature. A primitive `bot` keeps `⊥` fixed under substitution, drops the `[Bot Atom]`/`[Inhabited Atom]` side-conditions (`not_eq : (A → ⊥) = ¬A` becomes an unconditional `rfl`, and minimal logic works over any atom type), and gives `⊥` its own case in every recursion rather than folding it into `atom`. The gap is narrow either way — #607 already reuses Mathlib `Bot`/`Top` and has `not_eq := rfl` — so whichever we pick, that theory layer should live in one place; happy to do the merge, or to adapt #648 to whatever you settle on.

Two smaller notes: I'd keep the `Has*` prefix (bare `And`/`Or`/`Iff` clash with core, and CSLib already uses `HasSubstitution`/`HasFresh`), and the connective precedences are worth recording in `NOTATION.md`. The file consolidation and the grind-into-notation point already look covered.

Cslib/Logics/LinearLogic/CLL/Basic.lean
fmontesi: We use `HasX` for 'canonical' versions of `X` for a type, whenever `X` is instead parametrised over the target type, also in other typeclasses.

Cslib/Logics/Modal/Basic.lean
fmontesi: I'm working on making a single file for now -- it does work well for documenting this. But I'm keeping the thought of splitting into logic-specific files if it becomes too unwieldy.

Cslib/Foundations/Logic/Operators.lean
```diff
@@ -0,0 +1,96 @@
...
+  /-- `¬a` is the negation of `a`. -/
+  not (a : α) : α
+
+@[inherit_doc] scoped notation:max "¬" p:40 => HasNot.not p
+
+end Propositional
+
+section Modal
+
+-- ## Modalities
```
eric-wieser:
```suggestion
/-! ## Modalities -/
```
will make this a heading in the docs, same throughout.

Cslib/Foundations/Logic/Operators.lean
```diff
@@ -0,0 +1,96 @@
...
+  and (a b : α) : α
+
+@[inherit_doc] scoped infixr:36 " ∧ " => HasAnd.and
+
+/-- The type `α` has an or connective (`∨`). -/
+class HasOr (α : Type*) where
+  /-- `a ∨ b` is the disjunction of `a` and `b`. -/
+  or (a b : α) : α
+
+@[inherit_doc] scoped infixr:30 " ∨ " => HasOr.or
```
eric-wieser, line 38: I think I'd still strongly recommend something like
```suggestion
/-- And connectives; an alternate notation for `⊓` when working with logical connectives. -/
scoped infixr:36 " ∧ " => Max.max

/-- Or connectives; an alternate notation for `⊔` when working with logical connectives. -/
scoped infixr:30 " ∨ " => Min.min
```
to save reinventing the lemmas already in mathlib.

eric-wieser: (similarly below for `HImp`, `bihimp`, and `hnot`)

thomaskwaring: do you have an example of lemmas we would duplicate? eg you would almost never have `a ∧ b = b ∧ a` as an equality, so none of the lattice lemmas would apply

Cslib/Logics/Propositional/Defs.lean
fmontesi: I'd even use

```
instance : HasAnd (Proposition Atom) := ⟨.and⟩
```

What do you think, @eric-wieser, @thomaskwaring?

Cslib/Foundations/Logic/Operators.lean
fmontesi: We'd need to quotient things up to `LogicalEquivalence`, which we're not ready for yet and I'm not sure would be useful/easy to use (I actually don't have an opinion yet, gotta get there to test..). Let's rediscuss this once we get nearer.

Cslib/Logics/Modal/Basic.lean
fmontesi: (Fixed)

PR #613, opened 2026-06-03
chore(Semantics/LTS/Bisimulation): golf and simplify

thomaskwaring:
Some of the arguments in `Bisimulation` can be simplified by symmetry and use of API lemmas.


fmontesi (approved):
This looks great, thanks!

PR #615, opened 2026-06-04
feat(Automata, LTS, and Turing): saturated multistep transitions, mapLabel for LTS, and Nondeterministic Turing Machines (NTM)

fmontesi:
This PR:
1. Fixes a bug in the formulation of EpsilonNA by adopting a new saturated multistep transition `LTS.SMTr` for LTS, which correctly invokes tau-closure for empty strings.
2. Adds nondeterministic Turing machines (NTM), as an example of how Turing machines can be formulated on top of our current library for transition systems.

Both changes have been discussed at the last CSLib General Meeting and are part of a larger discussion towards integrating machine models and transition systems (postponed to later work).


Cslib/Computability/Machines/Turing/SingleTape/Defs.lean
```diff
@@ -0,0 +1,41 @@
...
+public import Cslib.Foundations.Data.BiTape
+
+/-! # Basic definitions for Turing Machines (TMs) -/
+
+@[expose] public section
+
+namespace Cslib.Computability.Turing.SingleTape
+
+/-- The transition labels used by a single-tape Turing Machine. -/
+structure TrLabel (State Symbol : Type*) where
```
Shreyas4991, line 18: This could be an inductive type. We can replace the transition function with the basic operations on a TM. For more discussion [see Zulip thread for this PR](https://leanprover.zulipchat.com/#narrow/channel/605128-CSLib.3A-PR-reviews/topic/.23615.3A.20NTM.20and.20saturated.20multistep.20transitions/near/600110047)

Shreyas4991 (commented):
This PR seems fine to me. The only question is whether we shouldn't just define the basic operations as move, read, write, etc, as opposed to a single step transition.

### Pro:
The TM can be programmed by a lean function that chains these operations 

###  Possible Con:
We will have to count read, write, and move as separate operations. However this is a small constant factor blow up which has no meaningful consequences in complexity theory or computability theory. 

### Discussion Thread

[Link](https://leanprover.zulipchat.com/#narrow/channel/605128-CSLib.3A-PR-reviews/topic/.23615.3A.20NTM.20and.20saturated.20multistep.20transitions/near/600110047)

Cslib/Computability/Machines/Turing/SingleTape/Defs.lean
```diff
@@ -0,0 +1,41 @@
...
+  /-- Tape of the machine (memory). -/
+  tape : Turing.BiTape Symbol
+
+/-- Helper builder for a configuration with a given tape content. -/
+def Cfg.mk₁ (s : State) (xs : List Symbol) : Cfg State Symbol where
+  state := s
+  tape := Turing.BiTape.mk₁ xs
+
+/-- The space used by a configuration is the space used by its tape. -/
+def Cfg.spaceUsed (cfg : Cfg State Symbol) : ℕ := cfg.tape.spaceUsed
```
crei, line 39: I think this definition might be misleading because it implement a rather uncommon notion of space (although it might be equivalent in the end). This definition is mainly used to prove that the  output length is bounded by the number of steps. So maybe we should rather rename it to "tapeSize" or something like that.
But we can also leave that for a later change when we introduce a notion of space.

Cslib/Computability/Machines/Turing/SingleTape/Deterministic.lean
```diff
@@ -62,11 +62,12 @@ We also provide ways of constructing polynomial-runtime TMs
 
 @[expose] public section
 
-open Cslib Relation
+open Relation
 
-namespace Turing
+namespace Cslib.Turing
```
crei, line 67: Is the idea here that we don't want to extend Mathlib's Turing namespace?

Cslib/Computability/Machines/Turing/SingleTape/NonDeterministic.lean
```diff
@@ -0,0 +1,59 @@
...
+  accept : Set State
+  /-- Proof that all accepting states are halting states. -/
+  accept_halting (hmem : s ∈ accept) : ¬∃ μ s', Tr s μ s'
+
+variable {State Symbol : Type*}
+
+/-- An NTM yields a small-step operational semantics on configurations, which codifies an execution
+step. -/
+@[scoped grind =]
+def SingleTapeNTM.Red (m : SingleTapeNTM State Symbol)
```
crei, line 40: This might be common for automata, but I myself would be able to understand this better if the documentation string would reference the abbreviation "Red" in some way ;)

fmontesi:
Closed in favour of #683

PR #617, opened 2026-06-05
chore(LTS): add API for TraceEq and relatives

thomaskwaring:
Follow-up to cslib#613 adding API & simplifying arguments related to trace equivalence.

---

- [ ] depends on: cslib#613


Cslib/Foundations/Semantics/LTS/Basic.lean
```diff
@@ -94,6 +94,12 @@ theorem MTr.single {s1 : State} {μ : Label} {s2 : State} :
   · exact h
   · apply MTr.refl
 
+lemma MTr.cons_iff {lts : LTS State Label} :
+    lts.MTr s1 (μ :: μs) s2 ↔ ∃ s, lts.Tr s1 μ s ∧ lts.MTr s μs s2 := by
+  constructor
+  · rintro (_ | ⟨htr, hmtr⟩); exact ⟨_, htr, hmtr⟩
+  · intro ⟨s, htr, hmtr⟩; exact .stepL htr hmtr
```
ctchou: I think we prefer to use newlines over `;`.

Cslib/Foundations/Semantics/LTS/Basic.lean
```diff
@@ -94,6 +94,12 @@ theorem MTr.single {s1 : State} {μ : Label} {s2 : State} :
   · exact h
   · apply MTr.refl
 
+lemma MTr.cons_iff {lts : LTS State Label} :
```
ctchou: There seems no good reason to call this theorem `lemma`.

The same comment for other `lemma`s below.

Cslib/Foundations/Semantics/LTS/Basic.lean
```diff
@@ -167,6 +192,19 @@ class Deterministic (lts : LTS State Label) where
   deterministic (s1 : State) (μ : Label) (s2 s3 : State) :
     lts.Tr s1 μ s2 → lts.Tr s1 μ s3 → s2 = s3
 
+lemma Tr.eq_of_tr_of_deterministic {lts : LTS State Label} [lts.Deterministic]
+    (htr : lts.Tr s1 μ s2) (htr' : lts.Tr s1 μ s2') : s2 = s2' :=
+  Deterministic.deterministic s1 μ s2 s2' htr htr'
+
+lemma MTr.eq_of_mtr_of_deterministic {lts : LTS State Label} [lts.Deterministic]
```
ctchou: Perhaps these two theorems should be called `Deterministic.eq_of_{tr,mTr}`?

ctchou (changes requested):
It is painful to review a PR that depends on another PR.  I think you should just merge this PR and #613 into a single PR.

chenson2018:
> It is painful to review a PR that depends on another PR. I think you should just merge this PR and #613 into a single PR.

This is very standard to do and helpful for seeing larger changes. Typically more detailed reviewing of the dependent PR is just held off until the first is merged.

Cslib/Foundations/Semantics/LTS/Basic.lean
thomaskwaring: i wrote it this way for dot notation, which i think is useful no?

thomaskwaring:
@ctchou thanks for your comments, i'll update #613 in response — feel free to hold off looking at this one until the other is merged

Cslib/Foundations/Semantics/LTS/Basic.lean
ctchou: A few lines down, the expression `htr.eq_of_tr_of_deterministic htr'` in your code would become `Deterministic.eq_of_tr htr htr'` in my proposal.  I'm not sure the the former is better than the latter.  The latter gives equal footing to `htr` and `htr'`, which seems more appropriate than the asymmetry between them in the former.  The latter is slightly shorter, too.

thomaskwaring: okay fair point i'm convinced :)

Cslib/Foundations/Semantics/LTS/Basic.lean
```diff
@@ -128,12 +136,31 @@ theorem MTr.single_invert (s1 : State) (μ : Label) (s2 : State) :
...
+@[simp] theorem MTr.singleton_iff (s1 : State) (μ : Label) (s2 : State) :
+  lts.MTr s1 [μ] s2 ↔ lts.Tr s1 μ s2 := ⟨MTr.single_invert lts s1 μ s2, MTr.single lts⟩
+
 /-- In any zero-steps multistep transition, the origin and the derivative are the same. -/
 @[scoped grind .]
 theorem MTr.nil_eq (h : lts.MTr s1 [] s2) : s1 = s2 := by
   cases h
   rfl
 
+theorem MTr.split {lts : LTS State Label} (h : lts.MTr s1 (μs ++ μs') s2) :
```
ctchou, line 155: Please keep the original doc-string from `Execution.lean`.

Cslib/Foundations/Semantics/LTS/Basic.lean
```diff
@@ -167,6 +194,19 @@ class Deterministic (lts : LTS State Label) where
   deterministic (s1 : State) (μ : Label) (s2 s3 : State) :
     lts.Tr s1 μ s2 → lts.Tr s1 μ s3 → s2 = s3
 
+theorem Deterministic.eq_of_tr {lts : LTS State Label} [lts.Deterministic]
+    (htr : lts.Tr s1 μ s2) (htr' : lts.Tr s1 μ s2') : s2 = s2' :=
+  Deterministic.deterministic s1 μ s2 s2' htr htr'
+
+lemma Deterministic.eq_of_mtr {lts : LTS State Label} [lts.Deterministic]
```
ctchou: I think the naming convention should result in `Deterministic.eq_of_mTr`.

Cslib/Foundations/Semantics/LTS/Bisimulation.lean
```diff
@@ -412,30 +402,28 @@ theorem IsBisimulation.deterministic_traceEq_isBisimulation
...
     (s₁ ~[lts₁,lts₂] s₂) := by
-  exists TraceEq lts₁ lts₂
-  constructor
-  case left =>
-    exact h
-  case right =>
-    apply IsBisimulation.deterministic_traceEq_isBisimulation
+  use TraceEq lts₁ lts₂, h, IsBisimulation.deterministic_traceEq_isBisimulation
+
+theorem Deterministic.bisim_tfae {lts₁ : LTS State₁ Label} {lts₂ : LTS State₂ Label}
```
ctchou, line 416: Please add a doc-string.

Cslib/Foundations/Semantics/LTS/TraceEq.lean
```diff
@@ -26,22 +26,65 @@ Definitions and results on trace equivalence for `LTS`s.
...
+  ext μs
+  constructor
+  · intro ⟨s'', hmtr⟩
+    use s'', MTr.stepL h hmtr
+  · intro ⟨s'', hmtr⟩
+    rcases hmtr with (_ | ⟨htr, hmtr⟩)
+    rw [←deterministic _ _ _ _ h htr] at hmtr
+    exact ⟨s'', hmtr⟩
+
+theorem MTr.traces_eq_of_deterministic {lts : LTS State Label} [lts.Deterministic]
```
ctchou: Name suggestions: `Deterministic.traces_of_{tr,mTr}`.

Cslib/Foundations/Semantics/LTS/TraceEq.lean
```diff
@@ -91,50 +129,39 @@ theorem HomTraceEq.eqv : Equivalence (· ~tr[lts] ·) where
 instance : Trans (TraceEq lts₁ lts₂) (TraceEq lts₂ lts₃) (TraceEq lts₁ lts₃) where
   trans := TraceEq.trans
 
+lemma TraceEq.exists_mtr_of_mtr {lts₁ : LTS State₁ Label} {lts₂ : LTS State₂ Label}
+    (h : s₁ ~tr[lts₁,lts₂] s₂) (htr : lts₁.MTr s₁ μs s₁') : ∃ s₂', lts₂.MTr s₂ μs s₂' := by
+  rw [←mem_traces_iff, ←h]
+  exact ⟨s₁', htr⟩
+
+lemma TraceEq.exists_tr_of_tr {lts₁ : LTS State₁ Label} {lts₂ : LTS State₂ Label}
```
ctchou: Is the `exists` in the names really necessary?

Cslib/Foundations/Semantics/LTS/TraceEq.lean
```diff
@@ -91,50 +129,39 @@ theorem HomTraceEq.eqv : Equivalence (· ~tr[lts] ·) where
...
+    (h : s₁ ~tr[lts₁,lts₂] s₂) (htr : lts₁.Tr s₁ μ s₁') : ∃ s₂', lts₂.Tr s₂ μ s₂' := by
+  rw [←mem_traces_singleton_iff, ←h, mem_traces_singleton_iff]
+  exact ⟨s₁', htr⟩
+
+lemma TraceEq.traceEq_of_tr_of_tr {lts₁ : LTS State₁ Label} {lts₂ : LTS State₂ Label}
```
ctchou: Do we need the second `of` in the name?

ctchou (changes requested):
Overall comments:
* Please add doc-strings for new theorems.
* Is there a meaningful distinction between your `lemma`s and `theorem`s?  Why not call them `theorem` uniformly?

Cslib/Foundations/Semantics/LTS/Basic.lean
thomaskwaring: good catch, thanks

Cslib/Foundations/Semantics/LTS/TraceEq.lean
thomaskwaring: thanks! that's definitely better

Cslib/Foundations/Semantics/LTS/TraceEq.lean
thomaskwaring: i think i prefer it to capture faithfully the content of the result, but you have more stake in this side of the library so happy to defer to your judgement

Cslib/Foundations/Semantics/LTS/TraceEq.lean
thomaskwaring: are you suggesting `traceEq_of_tr` or `traceEq_of_tr_tr` ? i think i prefer the current to both but again happy to defer to your preference.

thomaskwaring:
thanks again @ctchou — you're right i wasn't being particularly systematic with `lemma` vs `theorem`, i've updated it to be `theorem` unless something is unambiguously a supporting result.

ctchou (approved):
LGTM.

fmontesi (approved):

PR #619, opened 2026-06-05
feat(FLP): show that asynchronous distributed consensus is possible when there is no fault

ctchou:
This PR presents an asynchronous distributed consensus algorithm and proves that it achieves consensus when there is no fault.  This result is not needed for proving the FLP impossibility result, but is included to show that the notion of an algorithm defined in `Algorithm.lean` is not vacuous, in the sense that it does allow a working asynchronous consensus algorithm when there is no fault.

Zulip discussion:
https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/Impossibility.20of.20distributed.20consensus/with/592462001


Shreyas4991:
I recommend editing the title to "asynchronous distributed systems".  In synchronous and partially synchronous systems, consensus is achievable with for a fraction of nodes being Byzantinely faulty nodes.

ctchou:
Good point.  I have updated the doc-strings, commit message, and this PR's title accordingly.

Cslib/Foundations/Semantics/LTS/Basic.lean
```diff
@@ -170,6 +170,22 @@ theorem MTr.append_iff : lts.MTr s1 (μs ++ μs') s2 ↔ ∃ s, lts.MTr s1 μs s
...
+/-- Single-step invariant. -/
+@[scoped grind =]
+def TrInv (p : State → Prop) : Prop :=
+  ∀ s1 μ s2, lts.Tr s1 μ s2 → p s1 → p s2
+
+/-- Multistep invariant. -/
+@[scoped grind =]
+def MTrInv (p : State → Prop) : Prop :=
+  ∀ s1 μs s2, lts.MTr s1 μs s2 → p s1 → p s2
+
```
fmontesi, line 182: Really like this, we'll be able to use it also for expressing safety properties like 'lack of communication errors' in process calculi.

fmontesi (approved):
LGTM

PR #632, opened 2026-06-11
chore: split the `Relation` module

chenson2018:
This was already at ~1000 lines and growing. I've split this without making any other changes with one exception: the `CoeDep` instances previously specific to `dom` and `cod` are generalized as this makes imports less awkward (and is also useful for another PR I will make soon).


arademaker:
> the CoeDep instances previously specific to dom and cod are generalized as this makes imports less awkward

Can you elaborate on that?

chenson2018:
> > the CoeDep instances previously specific to dom and cod are generalized as this makes imports less awkward
> 
> Can you elaborate on that?

The current `CoeDep` instances are here: https://api.cslib.io/docs/Cslib/Foundations/Data/Relation.html#Relation.instCoeDepForallForallPropForallElemDomForall

At the top of the new `Defs.lean` file you can see I generalize this to allow any set. There was no need to be specific to (co)domain in the first place, as the coercion makes sense generally.

ctchou (changes requested):
The main doc-string of each new file should be updated to give some information of the file's contents.

ctchou (approved):
LGTM.

chenson2018:
(I merged main to pick up some deprecation adjustments)

fmontesi (approved):

PR #661, opened 2026-06-18
feat: basic defs of collections of labelled transition systems

Shreyas4991:
I introduce a general framework for reasoning about collections of labelled transition systems. Such systems arise in a number of models of concurrent and distributed systems as semantic models. Here, we don't aim to provide a specific semantic model. We instead provide LTS API that can be used for manipulating collections of LTSes when defining such models in subsequent PRs.

Context : This was discussed in the online CSLib meeting on 18th June 2026.

### DIFF
--- a/Cslib/Foundations/Semantics/LTS/VectorLTS.lean
+++ b/Cslib/Foundations/Semantics/LTS/VectorLTS.lean
@@ -37,7 +37,7 @@ namespace Cslib
 An indexed collection of LTSes. Each component LTS represents
 the local view of an entity.
 -/
-abbrev LTSVec (Entity : Type u) (State Label : Entity → Type*) :=
+def LTSVec (Entity : Type u) (State Label : Entity → Type*) :=
   (e : Entity) → LTS (State e) (Label e)
 
 /--
@@ -46,7 +46,7 @@ which represents the global view of a collection of LTSes, collectively
 transition for each input vector label to their next states. This
 matches how synchronous distributed systems behave
 -/
-abbrev SynchronousVecLTS (Entity : Type*) (State Label : Entity → Type*) :=
+def SynchronousVecLTS (Entity : Type*) (State Label : Entity → Type*) :=
   LTS ((e : Entity) → State e) ((e : Entity) → Label e)
 
 
@@ -71,7 +71,7 @@ def AsynchronousVecLTS (Entity) (State Label : Entity → Type*) :=
 A model of an asynchronously advancing collection of LTSes. This abstracts
 the behaviour of asynchronous distributed systems.
 -/
-abbrev LTSVec.toAsynchronousVecLTS {Entity}
+def LTSVec.toAsynchronousVecLTS {Entity}
     (State Label : Entity → Type*)
     (l : LTSVec Entity State Label) :
     AsynchronousVecLTS Entity State Label where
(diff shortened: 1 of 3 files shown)


CslibTests.lean
```diff
@@ -1,13 +1,15 @@
...
+public import CslibTests.DFA
+public import CslibTests.FreeMonad
+public import CslibTests.GrindLint
+public import CslibTests.HML
+public import CslibTests.HasFresh
+public import CslibTests.ImportWithMathlib
+public import CslibTests.LTS
+public import CslibTests.LambdaCalculus
+public import CslibTests.MLL
+public import CslibTests.Reduction
```
ctchou: The change to `CslibTests.lean` is not necessary; see:
[#CSLib > CslibTests.lean change in the last mathlib bump](https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/CslibTests.2Elean.20change.20in.20the.20last.20mathlib.20bump/with/602649742)

Cslib/Foundations/Semantics/LTS/VectorLTS.lean
```diff
@@ -0,0 +1,69 @@
...
+
+@[expose] public section
+
+namespace Cslib
+
+/--
+An indexed collection of LTSes. Each component LTS represents
+the local view of an entity.
+-/
+abbrev LTSVec (Entity : Type u) (State Label : Entity → Type*) :=
```
ctchou: Any reason `Entity` has type `Type u` rather than `Type*`?

Same comment for occurrences of `Entity` below.

Shreyas4991: Universe resolution only succeeds with at least one explicit universe parameter. That's basically the reason in both instances, but I recall the error being thrown up only in the second occurrence.

CslibTests.lean
Shreyas4991: Oh this is new. Thanks I'll fix it

Cslib/Foundations/Semantics/LTS/VectorLTS.lean
ctchou: I just tried replacing every `Type u` by `Type*`.  It didn't seem to break anything.

Cslib/Foundations/Semantics/LTS/VectorLTS.lean
```diff
@@ -0,0 +1,69 @@
...
+/--
+A model of an asynchronously advancing collection of LTSes. This abstracts
+the behaviour of asynchronous distributed systems.
+-/
+abbrev LTSVec.toAsynchronousVecLTS (Entity : Type u)
+    (State Label : Entity → Type*)
+    (l : LTSVec Entity State Label) :
+    AsynchronousVecLTS Entity State Label where
+  Tr := fun s ⟨e, μe⟩ s' =>
+    (l e).Tr (s e) μe (s' e) ∧ ∀ e' ≠ e, s e = s' e
```
ctchou: Why are some definitions `abbrev` while others are `def`?  Is there a consistent rationale?

Cslib/Foundations/Semantics/LTS/VectorLTS.lean
Shreyas4991: Even in line 56? That's the definition I got an error in.

Cslib/Foundations/Semantics/LTS/VectorLTS.lean
Shreyas4991: The `abbrev`s are both type synonyms. They are intended to give appropriate names and make them convenient to use since the type signature is non-trivial. They necessarily need to be defeq to the types they name.

Cslib/Foundations/Semantics/LTS/VectorLTS.lean
Shreyas4991: Removed all needless `Type u` and `Type*` annotations now.

PR #663, opened 2026-06-19
feat: `Set.ReflOn`, `Set.SymmOn`

chenson2018:
A first pass at adding definitions corresponding to the concept of having some property of a (homogeneous) relation over a set restriction, as discussed in [this thread](https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/unbundled.20relations.20with.20restricted.20.28co.29domain/with/596236853). I also add lemmas `of_{cod,dom}` for convenience of working in the "constructive" case where we explicitly are given the relation as evidence of the (co)domain.

### DIFF
--- a/Cslib/Foundations/Relation/Defs.lean
+++ b/Cslib/Foundations/Relation/Defs.lean
@@ -138,15 +138,19 @@ namespace Set
 
 open Relation
 
+/-- `ReflOn s r` is true when a relation `r` is reflexive on its restriction to a set `s`. -/
 def ReflOn (s : Set α) (r : α → α → Prop) : Prop :=
   ∀ a ∈ s, r a a
 
 -- these names are used in the literature, so we provide them as `abbrev`
 
+/-- `LeftQuasiRefl r` is true when a relation `r` is reflexive on its domain. -/
 abbrev LeftQuasiRefl (r : α → α → Prop) := (dom r).ReflOn r
 
+/-- `RightQuasiRefl r` is true when a relation `r` is reflexive on its codomain. -/
 abbrev RightQuasiRefl (r : α → α → Prop) := (cod r).ReflOn r
 
+/-- `SymmOn s r` ts true when a relation `r` is symmetric on its restriction to a set `s`. -/
 def SymmOn (s : Set α) (r : α → α → Prop) : Prop :=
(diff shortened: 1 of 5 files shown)


Cslib/Foundations/Relation/Defs.lean
```diff
@@ -133,3 +133,25 @@ def transRight (s r : α → α → Prop) [IsTrans α r] (h : s ≤ r) : Trans r
...
+
+-- these names are used in the literature, so we provide them as `abbrev`
+
+/-- `LeftQuasiRefl r` is true when a relation `r` is reflexive on its domain. -/
+abbrev LeftQuasiRefl (r : α → α → Prop) := (dom r).ReflOn r
+
+/-- `RightQuasiRefl r` is true when a relation `r` is reflexive on its codomain. -/
+abbrev RightQuasiRefl (r : α → α → Prop) := (cod r).ReflOn r
+
+/-- `SymmOn s r` ts true when a relation `r` is symmetric on its restriction to a set `s`. -/
```
fmontesi: ts -> is.

Rest looks good to me.

fmontesi (approved):

Cslib/Foundations/Relation/Defs.lean
chenson2018: Thanks. I'll wait a bit to merge in case there's any comments on Zulip.

Cslib/Foundations/Relation/Domain.lean
```diff
@@ -34,6 +34,9 @@ variable {β : Type*} {r : α → β → Prop}
 @[simp, grind =] lemma mem_dom : a ∈ dom r ↔ ∃ b, r a b := .rfl
 @[simp, grind =] lemma mem_cod : b ∈ cod r ↔ ∃ a, r a b := .rfl
 
+theorem of_dom (hab : r a b) : a ∈ dom r := by grind
+theorem of_cod (hab : r a b) : b ∈ cod r := by grind
```
thomaskwaring, line 38: this naming is a little confusing to me — doesn't the convention usually read "of_{hypothesis}" where here it's "of_{conclusion}" — otherwise lgtm :)

thomaskwaring (approved):

PR #672, opened 2026-06-21
refactor(LocallyNameless/Untyped): Rename redex_abs_close to steps_abs_close

lengyijun:


chenson2018:
I'm not sure the naming here is better. I think it should really be named after the inductive the notation is for. I think it would be better to more systematically address the naming in the lambda calculus modules rather than one particular theorem.

Issue #638, opened 2026-06-12
TimeM's `✓` tick breaks when HasWellFormed is imported


FrankieNC:
Cslib's own `✓` notations conflict with each other: `TimeM`'s tick, a leading do-element macro, is captured by `HasWellFormed`'s global postfix whenever both are imported. The practical effect is that the documented `TimeM` syntax works when importing the module on its own and silently breaks for anyone who imports Cslib as a whole. This is the same class of problem as #631 (a global notation without a whitespace guard reaching across line breaks).

## The problem

Two notations share the `✓` token:

```lean
-- Cslib/Algorithms/Lean/TimeM.lean
macro "✓" body:doElem : doElem => `(doElem| ✓[1] $body)

-- Cslib/Foundations/Syntax/HasWellFormed.lean
notation x:max "✓" => HasWellFormed.wf x
```

The second one is global and has no whitespace guard, so with both in scope the parser does not stop at the end of the do item *before* the tick: it sees the `✓` on the next line and reads it as the postfix applied to the previous expression. The tick is consumed, and the errors land on the line above it, which makes this quite hard to diagnose.

## MWE

```lean
import Cslib.Algorithms.Lean.TimeM

open Cslib.Algorithms.Lean in
def f (n : Nat) : TimeM Nat Nat := do
  let m := n
  ✓ return m
```

This compiles. Add `import Cslib.Foundations.Syntax.HasWellFormed` (or just use `import Cslib`) and it fails with

```
error: failed to synthesize instance of type class Cslib.HasWellFormed ℕ   -- on the `let m := n` line
error: Application type mismatch: m has type Prop but is expected to have type ℕ
```

because `let m := n` has become `let m := n✓`, i.e. `HasWellFormed.wf n : Prop`.

## How I encountered this

I ran into this with the treap exercise from the [Theoretical Computer Science Proving in Lean challenge](https://www.codabench.org/competitions/16161/) in the ICML 2026 [AI for Math workshop](https://ai4math2026.github.io/). The challenge's `TreapLogic` task file implements treap split and merge, instrumented with `TimeM` to count operations, and uses the tick exactly as documented:

```lean
def mergeT (l r : TreapNode Key Prio) : TimeM ℕ (TreapNode Key Prio) := do
  match l, r with
  ...
  | BinaryTree.node kp_1 l_1 r_1, BinaryTree.node kp_2 l_2 r_2 =>
    if kp_1.prio ≥ kp_2.prio then
      ...
    else
[...]

PR #675, opened 2026-06-22
fix(Foundations): add noWs guard to well-formed postfix notation

VitaliPath:
Addresses #638. Converts the global well-formed notation to a macro with a `noWs` whitespace guard. This explicitly prevents the postfix parser from greedily consuming the `✓` token across line breaks, resolving the syntax collision with TimeM's prefix tick notation when both modules are imported simultaneously.


chenson2018:
Could you please add a test in `CSLibTests` for this?

VitaliPath:
Added the regression test in `CslibTests/HasWellFormed.lean` and verified the fix resolves both the prefix syntax collision and preserves the expected postfix evaluation. Let me know if you need anything else.

VitaliPath:
Added the `#guard_msgs` wrapper to the test to prevent the compiler info trace from failing the strict CI checks. Should be totally green now

chenson2018 (approved):
Thanks!

VitaliPath:
Hey @chenson2018, I just rebased to resolve the import conflict. The history is clean, and the CI checks are green again. Ready for the merge queue whenever you have a moment

chenson2018 (approved):

PR #683, opened 2026-06-26
feat(Machines): Single-Tape Nondeterministic Turing Machines (NTMs)

fmontesi:
- Adds NTMs, defined as nondeterministic automata over `TrLabel` and the usual derived 'yields' relation (`Red`).
- Instantiates Acceptor and Transducer for NTM.
- Proves characterisation theorems that connect reductions on configurations to single- and multistep transitions of the underlying machine.

### DIFF
--- a/Cslib/Computability/Machines/Turing/SingleTape/Defs.lean
+++ b/Cslib/Computability/Machines/Turing/SingleTape/Defs.lean
@@ -25,16 +25,32 @@ inductive TrLabel (Symbol : Type*)
   /-- Do nothing. -/
   | skip
 
-/-- Applies a transition label to a tape, returning `none` if it is not possible. -/
-def TrLabel.applyToTape [DecidableEq Symbol] (μ : TrLabel Symbol) (tape : Turing.BiTape Symbol) :
+/-- Applies a transition label to a tape, returning `none` if it is not possible.
+The input is taken as an `Option` to make the function composable. -/
+def TrLabel.applyToTape [DecidableEq Symbol]
+    (otape : Option (Turing.BiTape Symbol)) (μ : TrLabel Symbol) :
     Option (Turing.BiTape Symbol) :=
-  match μ with
-  | read x => if x = tape.head then some tape else none
-  | write x => some (tape.write x)
-  | move d => some (tape.move d)
-  | skip => some tape
+  match μ, otape with
+  | read x, some tape => if x = tape.head then some tape else none
+  | write x, some tape => some (tape.write x)
+  | move d, some tape => some (tape.move d)
+  | skip, some tape => some tape
+  | _, _ => none
+
+@[scoped grind →]
+theorem TrLabel.applyToTape_isSome [DecidableEq Symbol] {μ : TrLabel Symbol}
+    {ot : Option (Turing.BiTape Symbol)} (h : (μ.applyToTape ot).isSome) : ot.isSome := by
+  have ⟨t', ht'⟩ := Option.isSome_iff_exists.mp h
+  simp only [applyToTape] at ht'
+  grind [applyToTape]
+
+@[scoped grind →]
+theorem TrLabel.applyToTape_foldl_isSome [DecidableEq Symbol] {μs : List (TrLabel Symbol)}
+    {ot : Option (Turing.BiTape Symbol)} (h : (μs.foldl applyToTape ot).isSome) : ot.isSome := by
+  induction μs generalizing ot <;> grind
 
 /-- Configuration of a single-tape Turing machine. -/
+@[ext]
 structure Cfg (State Symbol : Type*) where
   /-- The state that the machine is in. -/
   state : State
--- a/Cslib/Computability/Machines/Turing/SingleTape/Defs.lean
+++ b/Cslib/Computability/Machines/Turing/SingleTape/Defs.lean
@@ -1,0 +1,52 @@
+/-
+Copyright (c) 2026 Fabrizio Montesi. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Fabrizio Montesi, Bolton Bailey
+-/
+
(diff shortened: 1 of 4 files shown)


Cslib/Computability/Machines/Turing/SingleTape/Defs.lean
```diff
@@ -0,0 +1,68 @@
...
+  /-- Tape of the machine (memory). -/
+  tape : Turing.BiTape Symbol
+
+/-- Helper builder for a configuration with a given tape content. -/
+def Cfg.mk₁ (s : State) (xs : List Symbol) : Cfg State Symbol where
+  state := s
+  tape := Turing.BiTape.mk₁ xs
+
+/-- The space used by a configuration is the space used by its tape. -/
+def Cfg.spaceUsed (cfg : Cfg State Symbol) : ℕ := cfg.tape.spaceUsed
```
crei, line 66: Moving my comment from the other PR:

I think this definition might be misleading because it implement a rather uncommon notion of space (although it might be equivalent in the end). This definition is mainly used to prove that the output length is bounded by the number of steps. So maybe we should rather rename it to "tapeSize" or something like that.
But we can also leave that for a later change when we introduce a notion of space.

Cslib/Computability/Machines/Turing/SingleTape/Defs.lean
```diff
@@ -0,0 +1,68 @@
...
+/-! # Basic definitions for Turing Machines (TMs) -/
+
+@[expose] public section
+
+namespace Cslib.Computability.Turing.SingleTape
+
+/-- The transition labels used by a single-tape Turing Machine. -/
+inductive TrLabel (Symbol : Type*)
+  /-- Read `x` from the tape. -/
+  | read (x : Symbol)
```
Shreyas4991, line 20: Why does read need a parameter. Shouldn't read just return whatever symbol is on the thread?

Cslib/Computability/Machines/Turing/SingleTape/NonDeterministic.lean
```diff
@@ -0,0 +1,110 @@
...
+
+variable {State Symbol : Type*}
+
+namespace SingleTapeNTM
+
+variable [DecidableEq Symbol]
+
+/-- An NTM yields a small-step operational semantics on configurations, which codifies an execution
+step. -/
+def Red (m : SingleTapeNTM State Symbol)
```
crei: Could you add a small comment that explains why it is called `Red`?

Cslib/Computability/Machines/Turing/SingleTape/Defs.lean
```diff
@@ -0,0 +1,68 @@
...
+  /-- Do nothing. -/
+  | skip
+
+/-- Applies a transition label to a tape, returning `none` if it is not possible.
+The input is taken as an `Option` to make the function composable. -/
+def TrLabel.applyToTape [DecidableEq Symbol]
+    (otape : Option (Turing.BiTape Symbol)) (μ : TrLabel Symbol) :
+    Option (Turing.BiTape Symbol) :=
+  match μ, otape with
+  | read x, some tape => if x = tape.head then some tape else none
```
Shreyas4991, line 34: Why is the tape an option type? We always operate on the tape don't we?

Shreyas4991 (commented):
Design comments

Cslib/Computability/Machines/Turing/SingleTape/NonDeterministic.lean
```diff
@@ -0,0 +1,110 @@
...
+instance : Transducer (SingleTapeNTM State Symbol) Symbol Symbol where
+    Translates (m : SingleTapeNTM State Symbol) (xs ys : List Symbol) :=
+  ∃ s ∈ m.start, ∃ c',
+    c'.state ∈ m.accept ∧
+    m.MRed (Cfg.mk₁ s xs) c' ∧
+    /- The following condition on output deviates from textbooks, in order to have the same
+    criterion as for deterministic TMs. We might want to revisit this in the future.
+    -/
+    c'.tape = Turing.BiTape.mk₁ ys
+
```
crei, line 120: Could you add another Prop that states that the NTM accepts an input in at most t steps?

Cslib/Computability/Machines/Turing/SingleTape/NonDeterministic.lean
fmontesi: It was short for reduction semantics, but in this context the 'yields' relation is used in textbooks, so I'm switching to that.

Cslib/Computability/Machines/Turing/SingleTape/Defs.lean
fmontesi: I'll leave it out of this PR (I inherited this IIRC), but I agree we should discuss this.

Cslib/Computability/Machines/Turing/SingleTape/NonDeterministic.lean
fmontesi: Done. See `AcceptsInAtMostSteps`.

Cslib/Computability/Machines/Turing/SingleTape/Defs.lean
fmontesi: Because I need to use this with `List.foldl` in the NTM defs, but if you have a better suggestion I'm happy to hear it.

Cslib/Computability/Machines/Turing/SingleTape/Defs.lean
fmontesi: This codifies the transition where the machine needs to read the symbol `x` to change states according to its transition relation. It's the usual thing in textbooks on TMs and automata, as well as labelled transition systems with 'observable actions'.

crei (approved):

PR #686, opened 2026-07-01
chore: disable .github/workflows/lake-update.yml on personal forks

ctchou:
Make sure this job is run only on the main cslib repo `leanprover/cslib'` and not on personal forks.


chenson2018 (approved):
Thanks!

PR #679, opened 2026-06-24
feat(untyped): standardization theorem for the lambda calculus

m-ow:
This PR proves the standardization theorem: if `M` beta-reduces to `N` in any number of steps, then `N` is reachable from `M` by a standard reduction.

Builds on #671.

PR #700, opened 2026-07-08
feat(untyped): the Leftmost Reduction Theorem

m-ow:
Introduces leftmost reduction `⭢ℓ`, a beta-step contracting the redex at position 0, together with normal forms, and proves the Leftmost Reduction Theorem: if a term beta-reduces to a normal form, the leftmost strategy reaches it too. 
The result falls out of standardization, following Copes (2018).

Builds on top of #679


lengyijun:
Can you add 2 theorems :
1.  NormalForm M <-> Normal FullBeta M
2. Decidable  Normal FullBeta

m-ow:
@lengyijun Sounds good, I'll get both into a separate PR soon :)

lengyijun:
Should we rename `Normal` to `NormalBeta` ? 
Or put it in namespace "FullBeta"


`Normal` may also refer to `Normal FullBetaEta` or `Normal FullEta`

lengyijun:
@m-ow Hi, what's your plan after this pr

m-ow:
Honestly, no fixed plan yet, but your project sounds really interesting. Feel free to point me at what's the priority on your side, happy to take a look!

lengyijun:
We will formalize "2 vars are not enough" after this pr is merged.

There will still be 1400 cases unprocessed. I haven't even studied them at all.

You can randomly pick a few and research them.

I will also pr [my fork](https://github.com/awesome-lambda-calculus/cslib/tree/sn_eta_step) step by step.
Perhaps you can help me optimize the proof.

chenson2018:
@m-ow Before I review for more detailed changes, I think I'm missing some high-level points about the defintions (some of which I maybe should have asked in earlier PRs, sorry!).

`BetaAt` seems to clearly align with the literature, but I think I'm missing where the connection to the usual definition of standard reduction is happening. I am expecting to somewhere see that if we have `BetaAt A B x` and `BetaAt B C y` with `x <= y` this is a standard reduction and that this is equivalent to `Standard`. Am I missing where this appears in a different form or otherwise misunderstanding something? (I'm following along with https://www.is.c.titech.ac.jp/~kashima/pub/C-145.pdf that's mentioned in your reference)

m-ow:
You're right, it's not there. The results here only rely on the inductive Standard relation, both sources I followed actually keep the sequence definition as a separate development. It's clearly worth having though, sorry for the gap! If you think it's best, I'll open a new PR to add the sequence definition and its equivalence with Standard.

chenson2018:
Yeah, it is pretty standard to use some equivalent inductive that's easier to work with (I assume this is why we have `Standard` not directly written in terms of `BetaAt`?), but I think we should have this proof of equivalence to make sure the definition corresponds to what we want informally.

Maybe because these are so closely related they should just go in the same module? Feel free to move things around as needed.

m-ow:
I've added the equivalence! Feel free to take another look whenever it suits you :)

lengyijun:
Is it possible to prove 
```
BetaAt M P i -> BetaAt M Q i -> P = Q
```

m-ow:
Sure, added.

m-ow:
> Is it possible to prove

Yes, I'll add it! Probably together with `BetaNormal M ↔ Normal FullBeta M​` in the next PR

Cslib/Languages/LambdaCalculus/LocallyNameless/Untyped/StandardReduction.lean
```diff
@@ -28,6 +29,44 @@ variable {Var : Type u}
 
 namespace LambdaCalculus.LocallyNameless.Untyped.Term
 
+/-- The number of β-redexes occurring in a term. -/
+def countRedexes : Term Var → Nat
+| fvar _        => 0
+| bvar _        => 0
+| abs m         => countRedexes m
+| app (abs m) n => (countRedexes (abs m) + countRedexes n) + 1
```
chenson2018: This is equivalent and I think easier to read:
```suggestion
| app (abs m) n => countRedexes m + countRedexes n + 1
```

Cslib/Languages/LambdaCalculus/LocallyNameless/Untyped/StandardReduction.lean
```diff
@@ -28,6 +29,44 @@ variable {Var : Type u}
...
+
+/-- `BetaAt M N i` reduces the redex at position `i` of `M` to obtain `N`;
+    positions are counted from left to right. -/
+inductive BetaAt : Term Var → Term Var → Nat → Prop
+/-- The outermost redex sits at position `0`. -/
+| outer : LC (abs M) → LC N → BetaAt (app (abs M) N) (M ^ N) 0
+/-- Reducing a non-abstraction operator keeps the position. -/
+| appNoAbsL : BetaAt M M' i → ¬IsAbs M → BetaAt (app M N) (app M' N) i
+/-- Reducing an abstraction operator advances the position by one. -/
+| appAbsL : BetaAt M M' i → IsAbs M → BetaAt (app M N) (app M' N) (i + 1)
```
chenson2018: What do you think about combining these like

```suggestion
| appL : BetaAt M M' i → BetaAt (app M N) (app M' N) (i + if IsAbs M then 1 else 0)
```
and similarly for the right? (Either adding a `Decidable` instance for `IsAbs` or making it boolean-valued)

Cslib/Languages/LambdaCalculus/LocallyNameless/Untyped/StandardReduction.lean
```diff
@@ -28,6 +29,44 @@ variable {Var : Type u}
...
+def countRedexes : Term Var → Nat
+| fvar _        => 0
+| bvar _        => 0
+| abs m         => countRedexes m
+| app (abs m) n => (countRedexes (abs m) + countRedexes n) + 1
+| app m n       => countRedexes m + countRedexes n
+
+/-- `IsAbs m` holds when `m` is an abstraction. -/
+@[scoped grind]
+inductive IsAbs : Term Var → Prop
```
chenson2018: Note we also have `Term.Value`, which is defined in `LcAt.lean` as locally closed abstractions. Can we consolidate these definitions? Or at minimum move this definition next to it for dependency reasons.

Cslib/Languages/LambdaCalculus/LocallyNameless/Untyped/StandardReduction.lean
```diff
@@ -28,6 +29,44 @@ variable {Var : Type u}
...
+| abs (m : Term Var) : IsAbs (abs m)
+
+/-- `BetaAt M N i` reduces the redex at position `i` of `M` to obtain `N`;
+    positions are counted from left to right. -/
+inductive BetaAt : Term Var → Term Var → Nat → Prop
```
chenson2018: For consistency with `StandardSeq` and currying reasons, I'd make this  `Nat → Term Var → Term Var → Prop`.

Cslib/Languages/LambdaCalculus/LocallyNameless/Untyped/StandardReduction.lean
```diff
@@ -28,6 +29,44 @@ variable {Var : Type u}
...
+| abs (xs : Finset Var) :
+    (∀ x ∉ xs, BetaAt (M ^ fvar x) (M' ^ fvar x) i) → BetaAt (abs M) (abs M') i
+
+/-- A standard β-reduction sequence from `M` to `N`: the contracted redex positions are
+    non-decreasing and all at least `n`. -/
+inductive StandardSeq : Nat → Term Var → Term Var → Prop
+/-- The empty sequence is standard for any lower bound. -/
+| refl : StandardSeq n M M
+/-- A first step at position `i ≥ n`, followed by a standard sequence bounded below by `i`. -/
+| head : BetaAt M P i → n ≤ i → StandardSeq i P N → StandardSeq n M N
```
chenson2018: I think it would be more intuitive for this to follow `Relation.ReflTransGen` and have the contructors be `refl` and `tail` (not just switching the naming, but the side you build the reduction on) and have `StandardSeq.head` as an additional theorem/induction principle.

Cslib/Languages/LambdaCalculus/LocallyNameless/Untyped/StandardReduction.lean
```diff
@@ -91,7 +130,179 @@ lemma Standard.cbn_trans (h1 : M ↠ₙ P) (h2 : P ⭢ₛ N) : M ⭢ₛ N := by
 lemma Standard.of_cbn (step : M ↠ₙ N) (lc_N : LC N) : M ⭢ₛ N :=
   cbn_trans step (lc_refl N lc_N)
 
-variable [DecidableEq Var] [HasFresh Var]
+/-- Opening with a free variable preserves the number of redexes. -/
+lemma countRedexes_openRec_fvar (M : Term Var) (k : Nat) (x : Var) :
```
chenson2018: You should have a theorem specializing this to `countRedexes_open_fvar` to avoid the casting for notation you later do.

Cslib/Languages/LambdaCalculus/LocallyNameless/Untyped/StandardReduction.lean
```diff
@@ -91,7 +130,179 @@ lemma Standard.cbn_trans (h1 : M ↠ₙ P) (h2 : P ⭢ₛ N) : M ⭢ₛ N := by
...
+  | app _ step_M ih => exact .appNoAbsL ih (cbn_not_isAbs step_M)
+
+/-- A standard sequence remains standard under a smaller lower bound. -/
+lemma StandardSeq.mono (h : StandardSeq n M N) (hmn : m ≤ n) : StandardSeq m M N := by
+  cases h with
+  | refl => exact .refl
+  | head step hni tail => exact .head step (le_trans hmn hni) tail
+
+/-- Standard sequences preserve being an abstraction. -/
+lemma StandardSeq.isAbs_r (h : StandardSeq n M N) (ha : IsAbs M) : IsAbs N := by
```
chenson2018: I wonder if there is/we should add a standard way to say "this relation preserves a property". (No action required, just a stray thought)

Cslib/Languages/LambdaCalculus/LocallyNameless/Untyped/LeftmostReduction.lean
```diff
@@ -0,0 +1,142 @@
...
+namespace Cslib
+
+universe u
+
+variable {Var : Type u}
+
+namespace LambdaCalculus.LocallyNameless.Untyped.Term
+
+/-- A term is in normal form when it contains no β-redexes. -/
+def BetaNormal (m : Term Var) : Prop := countRedexes m = 0
```
chenson2018, line 33: I believe you said in a comment you'll prove this really is `Relation.Normal`? Can be a follow-up PR.

chenson2018 (commented):
A few more questions/comments about the definitions added:

Cslib/Languages/LambdaCalculus/LocallyNameless/Untyped/StandardReduction.lean
m-ow: Added, thanks.

Cslib/Languages/LambdaCalculus/LocallyNameless/Untyped/LeftmostReduction.lean
m-ow: Yes, already have it, coming in the follow-up PR

Cslib/Languages/LambdaCalculus/LocallyNameless/Untyped/StandardReduction.lean
m-ow: Agreed, definitely worth doing later

Cslib/Languages/LambdaCalculus/LocallyNameless/Untyped/StandardReduction.lean
m-ow: Done

Cslib/Languages/LambdaCalculus/LocallyNameless/Untyped/StandardReduction.lean
m-ow: Moved IsAbs next to Value, kept separate since BetaAt uses IsAbs without LC.
Would you prefer Value as `IsAbs ∧ LC`?

chenson2018: Let's leave it for now to keep a cleaner diff and I'll think about what is preferred.

Cslib/Languages/LambdaCalculus/LocallyNameless/Untyped/StandardReduction.lean
m-ow: Good idea, did it. Went with Decidable IsAbs over boolean

lengyijun:
```lean
/-- Values are irreducible terms. -/
inductive Value : Term Var → Prop
| abs (e : Term Var) : e.abs.LC → e.abs.Value

/-- `IsAbs m` holds when `m` is an abstraction. -/
@[scoped grind]
inductive IsAbs : Term Var → Prop
| abs (m : Term Var) : IsAbs (abs m)

instance (m : Term Var) : Decidable (IsAbs m) := by
  cases m
  case abs => exact isTrue (.abs _)
  all_goals exact isFalse (by intro _; contradiction)
```

Why put in lcat.lean ? Unrelated, strange.

chenson2018:
> Why put in lcat.lean ? Unrelated, strange.

I very briefly explained why I asked for this above, but to be more explicit:
- having both `Value` and `IsAbs` seems redundant, or at least one should be defined in terms of the other
- I didn't want to prompt that unrelated refactor in this PR, so asked for them to be placed next to each other as a reminder of this 
- In terms of dependencies, this is the highest spot in the imports where that is possible

Edit: and lastly while you could place `IsAbs` on its own in the `Basic.lean` module, as a practical matter we almost always need to import `LC` anyway.

lengyijun:
I know, the position of Value is quite strange before this pr. 
I really like the elegance of lcat.lean.
I don't like anything strange here.
I just really surprised that lcat need re-compiled after `git pull`

chenson2018:
> I know, the position of Value is quite strange before this pr.

I don't understand the reasoning here. It makes sense for this definition to occur as soon as local closure is available. A single definition like this is not a detriment to the structure of the module.

(Let's maybe table this discussion for later, it's gotten a bit off topic for this PR)

lengyijun:
@m-ow Could you please group the theorems ?

https://github.com/awesome-lambda-calculus/cslib/commit/ab40e9d3fb2413cc638791de6fb781a80af00eff

lengyijun:
My more aggressive opinion is to separate countredex to a new file.
Because personally I don't care anything about standard reduction.

BTW StandardReduction.lean is the largest file in untyped lambda calculus now.
Worth splitting up

lengyijun:
Can you add theorm of fv

```
#[grind]
lemma step_fv (step : BetaAt i M N) : N.fv ⊆ M.fv := by
```

Cslib/Languages/LambdaCalculus/LocallyNameless/Untyped/StandardReduction.lean
m-ow: Addressed. StandardSeq now follows ReflTransGen, with refl, tail, and a derived head theorem

m-ow:
@lengyijun Addressed everything except the free-variable theorem, which I'll handle in a follow-up PR to keep this diff smaller

lengyijun:
@m-ow @chenson2018  Unable to cases  t.abs ⭢ℓ t'

```
theorem foo {t t': Term String}
(step : t.abs ⭢ℓ t') : exists a, t' = a.abs := by
  cases step with
  | appR _ => sorry  -- error here
  | abs xs _ => sorry
```

Expected: fall into the `abs` case, derive t' must be Term.abs

chenson2018:
> @m-ow @chenson2018 Unable to cases t.abs ⭢ℓ t'
> 
> ```
> theorem foo {t t': Term String}
> (step : t.abs ⭢ℓ t') : exists a, t' = a.abs := by
>   cases step with
>   | appR _ => sorry  -- error here
>   | abs xs _ => sorry
> ```
> 
> Expected: fall into the `abs` case, derive t' must be Term.abs

Yes, with the typing of `BetaAt` now, this is not surprising. We should have an induction principle that matches the original set of constructors that @m-ow wrote to avoid the dependent typing.

Cslib/Languages/LambdaCalculus/LocallyNameless/Untyped/StandardReduction.lean
chenson2018: How do you feel about the ergonomics of this change, since this now carries a lower bound as well? Do you think it is a useful addition?

Cslib/Languages/LambdaCalculus/LocallyNameless/Untyped/StandardReduction.lean
chenson2018: Marking as unresolved in light of the dependent elimination issue discussed below. I do still like it, I think we just need an additional induction principle.

lengyijun:
@m-ow Can you group the theorems ? 

https://github.com/awesome-lambda-calculus/cslib/commit/f015ac330f414c0b9dfbd5cc9ee5ca3b15fd3c8f

lengyijun:
> > @m-ow @chenson2018 Unable to cases t.abs ⭢ℓ t'
> > ```
> > theorem foo {t t': Term String}
> > (step : t.abs ⭢ℓ t') : exists a, t' = a.abs := by
> >   cases step with
> >   | appR _ => sorry  -- error here
> >   | abs xs _ => sorry
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Expected: fall into the `abs` case, derive t' must be Term.abs
> 
> Yes, with the typing of `BetaAt` now, this is not surprising. We should have an induction principle that matches the original set of constructors that @m-ow wrote to avoid the dependent typing.

Thx, I got it. I should generalize 0 to variable

lengyijun:
```lean4
Leftmost.normalization (lc : LC M) 
```

LC M is redundant

lengyijun:
Can you add 2 theorems : 
1. Relation.Normalizable FullBeta M <-> Relation.Normalizable Leftmost M 
2.  Relation.Normal FullBeta M <-> Relation.Normal Leftmost M

lengyijun:
@m-ow Lc is redundant : 

```
| outer : LC (abs M) → LC N → BetaAt 0 (app (abs M) N) (M ^ N)
| outer : BetaAt 0 (app (abs M) N) (M ^ N)
```

As far as I tested, BetaAt.lc_l is not strong enough.

I usually require M.LC with BetaAt i M M'`

lengyijun:
@m-ow @chenson2018  The induction rule for leftmost 

```lean4

theorem Leftmost.induction_rule
    {motive : ∀ {a b : Term String}, Leftmost a b → Prop}
    {a b : Term String}
    (h : Leftmost a b)
    (h_outer :
      ∀ (M N : Term String)(hm : M.abs.LC)(hn : N.LC), motive (BetaAt.outer (M:=M) (N:=N) hm hn))
      (h_appL :
  ∀ {M M' N : Term String}
    (h : Leftmost M M')
    (hi : ¬ IsAbs M),
    motive h →
    @motive (M.app N) _ (by simpa [Leftmost, hi] using (BetaAt.appL h)))
      (h_appR :
  ∀  {M M' N : Term String}
    (h : Leftmost M M')
    (hi : ¬ IsAbs N)
    (g : N.countRedexes = 0),
    motive h →
    @motive (N.app M) (N.app M') (by
      have hidx : (N.countRedexes + if N.IsAbs then 1 else 0) = 0 := by grind
      unfold Leftmost
      simpa [hidx] using (@BetaAt.appR _ _ _ _ N h)
      ))
    (h_abs :
      ∀ (M M': Term String) (xs : Finset String)
        (h :
          ∀ x ∉ xs,
            Leftmost (M ^ fvar x) (M' ^ fvar x)),
      (∀ x hx, motive (h x hx)) →
      motive (BetaAt.abs xs h))
    : motive h := by
  unfold Leftmost at h
  generalize hi : 0 = i
  rw [hi] at h
  induction h with
  | outer => grind
  | appL h ih =>
      rename_i i _ _ _ _
      have : i = 0 := by omega
      subst i
      apply h_appL
      grind
      grind
      exact h
  | appR h =>
      rename_i i _ _ _ _ _
      have : i = 0 := by omega
      subst i
      apply h_appR
      grind
      grind
      grind
      exact h
  | abs xs h ih =>
      subst_vars
      exact h_abs _ _ xs
        (by
          intro x hx
          exact h x hx)
        (by
          intro x hx
          apply ih _ hx
          omega)
```

And I have tested it with one theorem : 

```lean4
theorem leftstar_cases {M N Q : Term String} (h : M.app N ↠ℓ Q) :
 (∃ M' N', M ↠ℓ M' /\ N ↠ℓ N' /\ Q = M'.app N') \/ ∃ M', M ↠ℓ Term.abs M' :=  by
  induction h with
  | refl => grind
  | tail _ g ih => cases ih with
    | inr h => grind
    | inl h => induction g using Leftmost.induction_rule <;> grind
```

Also the stronger version : 
```lean4
theorem leftstar_cases {M N Q : Term String} (h : M.app N ↠ℓ Q) (m_lc : M.LC) :
 (∃ M' N', M ↠ℓ M' /\ N ↠ℓ N' /\ Q = M'.app N') \/
  ∃ M' : Term String, M ↠ℓ M'.abs /\ M'.abs.app N ↠ℓ Q :=  by
  generalize hp: M.app N = P
  rw [hp] at h
  induction h using Relation.ReflTransGen.head_induction_on generalizing M N with
  | refl => grind
  | head h' h ih => subst_vars
                    generalize hmn : M.app N = mn
                    rw [hmn] at h'
                    induction h' using Leftmost.induction_rule generalizing M N with cases hmn
                    | h_outer M N hm hn =>  right
                                            refine ⟨_, .refl, .trans (.single (.outer hm hn)) h⟩
                    | h_appL h hi _ =>
                      rename_i hmm' _
                      cases ih (BetaAt.lc_r hmm' m_lc) rfl with
                      | inl h =>
                        left
                        obtain ⟨M'', N', h, _, _⟩ := h
                        refine ⟨M'', N', .trans (.single ?_) h, by grind, by grind⟩
                        assumption
                      | inr h =>
                        right
                        obtain ⟨M'', h, _⟩ := h
                        refine ⟨M'', .trans (.single ?_) h, by grind⟩
                        assumption
                    | h_appR h hi g _ =>
                    cases ih m_lc rfl with
                      | inl h =>  obtain ⟨M'', N', _, hn, _⟩ := h
                                  subst Q
                                  left
                                  refine ⟨_, _, ?_, .trans (.single ?_) hn, rfl⟩
                                  grind
                                  assumption
                      | inr h =>  exfalso
                                  rename_i N' _ _ _
                                  obtain ⟨M'', h, _⟩ := h
                                  have g : BetaNormal N' := by grind [BetaNormal]
                                  rw [betanormal_iff] at g
                                  induction h using Relation.ReflTransGen.head_induction_on with
                                  | refl => grind
                                  | head h' h _ =>
                      apply g
                      apply BetaAt.to_step at h'
                      rename_i M _ _ _ _
                      specialize h' m_lc
                      grind

```

m-ow:
`StandardSeq` now follows Copes's definition: single index, `refl`/`tail`. That removes the extra lower bound.

chenson2018:
@m-ow Can you fix the merge conflict? With the changes to `StandardSeq` this seems much nicer, I think we're close here. I think after the merge conflict we can merge and then consider some of the further suggestions from @lengyijun.

lengyijun:
@m-ow Hi, just wanted to check if you’re still interested in continuing to push this PR forward.

If you’re busy or no longer planning to work on it, no problem at all — I’ll adjust my development plans accordingly.

Thanks!

chenson2018 (approved):
I've gone ahead and resolved the merge conflict. I'm going to go ahead and merge and improvements suggested in the above discussions can become follow-up PRs.

m-ow:
Thanks for resolving the conflict and getting this merged! Sorry I didn't reply sooner, and thanks everyone for the help and feedback.

lengyijun:
@m-ow Do you plan to PR more and complete the unfinished theorems?

m-ow:
Not at the moment, I'm planning to focus on other areas of CSLib, so I'll leave future lambda calculus work to you. Seems like it's in good hands! I'd still be happy to discuss the design if anything comes up.

lengyijun:
@m-ow Why does this PR mix two different code styles? Is that required by Mathlib4?

```lean4
  induction h with
  | refl =>
  | tail _ step hni ih =>
```

```
  induction h
  case refl => 
  case tail Q _ _ seq step hni ih =>
```

chenson2018:
Either style is permitted in Mathlib, and CSLib by extension. (Let's please not continue extended discussion in a GitHub PR that has been merged for some time)

lengyijun:
@m-ow Following pr: https://github.com/leanprover/cslib/pull/800

PR #701, opened 2026-07-08
feat: add AUTHORS.md

fmontesi:
Adds an AUTHORS.md file explaining where copyright information can be found.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -8,7 +8,7 @@ Each file in CSLib includes a copyright header that lists the authors who contri
 
 To see more people who have contributed to CSLib and own copyright to parts of the codebase, please refer to its Git history (on a web browser, this can be accessed at <https://github.com/leanprover/cslib/commits/main>).
 
-**Note**: Git history may not always be fully comprehensive. In cases where code is co-written, only the person who commits the changes may appear in the Git history, even if multiple people contributed to the work.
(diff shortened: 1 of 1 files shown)


AUTHORS.md
```diff
@@ -0,0 +1,15 @@
...
+
+## Copyright and Authorship
+
+**Copyright in CSLib is held by the individual authors** who contributed to the code.
+
+Each file in CSLib includes a copyright header that lists the authors who contributed significantly to that specific file (in the opener and in the optional `Authors` field, when there are multiple significant authors). Examples of significant contributions are original creation, major refactoring, or important additions.
+
+To see more people who have contributed to CSLib and own copyright to parts of the codebase, please refer to its Git history (on a web browser, this can be accessed at <https://github.com/leanprover/cslib/commits/main>).
+
+**Note**: Git history may not always be fully comprehensive. In cases where code is co-written, only the person who commits the changes may appear in the Git history, even if multiple people contributed to the work.
```
eric-wieser: I think there are two things at play here:
* If a contributor does not declare who they collaborated with, then the history will have no record of it.
* If a contributor writes a `Co-authored-by` line incorrectly, then GitHub will not link it to a GitHub account. Either way it's still in "the git history"

eric-wieser: I guess more precise would be to say "Commit messages may attribute additional authors beyond the single commit author name"

fmontesi: Good catch, this should be clearer. How does it look now?

PR #705, opened 2026-07-10
feat(LocallyNameless/Untyped): FullEta.steps_open_cong_l

lengyijun:


lengyijun:
Mirror of FullBeta.step_open_cong_l

Cslib/Languages/LambdaCalculus/LocallyNameless/Untyped/FullEta.lean
```diff
@@ -156,6 +156,13 @@ lemma steps_open_cong_r {s t t' : Term Var} (lc_s : LC s.abs) (lc_t : LC t) (ste
   case refl => rfl
   case head _ _ st _ ih => exact .trans (step_open_cong_r lc_s lc_t st) (ih (step_lc_r st))
 
+theorem steps_open_cong_l {s s' t : Term Var} (xs : Finset Var)
```
chenson2018, line 159: Similar to what I suggested in #738, I think it is useful to also have the single-step version of this even if it is technically less general.

PR #714, opened 2026-07-13
feat: add READMEs to main directories

fmontesi:
This PR adds README files to some of the main source code directories, covering explanations on principles and vision for CSLib that I've found myself repeating often in private and public conversations.

### DIFF
--- a/Cslib/Algorithms/README.md
+++ b/Cslib/Algorithms/README.md
@@ -1,0 +1,30 @@
+<pre>
+Copyright (c) 2026 Fabrizio Montesi. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Clark Barrett, Swarat Chaudhuri, Jim Grundy, Fabrizio Montesi, Leonardo de Moura, Alexandre Rademaker, Sorrachai Yingchareonthawornchai
+</pre>
+
+# Algorithms
+
+This directory hosts **algorithms and their properties**. These properties concern functional correctness, complexity, and other relevant results. The directory also includes dedicated facilities for reasoning about algorithms written in Lean.
+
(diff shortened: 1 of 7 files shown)


Cslib/Foundations/README.md
```diff
@@ -0,0 +1,32 @@
+<pre>
+Copyright (c) 2026 Fabrizio Montesi. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+</pre>
+
+# Foundations
```
ctchou, line 6: `Cslib/Foundations` is also where we put additional results about data objects defined in mathlib, such as `Nat` and `Set`.

Cslib/Algorithms/README.md
```diff
@@ -0,0 +1,30 @@
+<pre>
+Copyright (c) 2026 Fabrizio Montesi. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Clark Barrett, Swarat Chaudhuri, Jim Grundy, Fabrizio Montesi, Leonardo de Moura, Alexandre Rademaker, Sorrachai Yingchareonthawornchai
+</pre>
+
+# Algorithms
```
ctchou, line 7: Will algorithms expressed in `Boole` be placed in `Cslib/Algorithms/Boole` or in `Cslib/Languages/Boole`?

ctchou (commented):
I think the top-level `ORGANISATION.md` should reference these files.

Cslib/Foundations/README.md
fmontesi: Right. Maybe we should qualify it as 'foundational' data objects? Some extensions might live in Probability or similar. (Though these look indeed like corner cases to me.)

Cslib/Algorithms/README.md
fmontesi: The current plan is to use the former.

ctchou: Maybe it's a good idea to point that out.

fmontesi: Done

fmontesi:
All done, @ctchou, let me know how it looks.

ctchou (approved):
LGTM.

PR #718, opened 2026-07-13
feat(FLP): define pseudo-consensus and prove that it is implied by consensus

ctchou:
This PR defines the notion of a fault-tolerant "pseudo-consensus" algorithm, which is central to Völzer's proof, and proves that every f-tolerant consensus algorithm is also a f-tolerant pseudo-consensus algorithm.  This resilt is intuitively obvious and is stated without proof in Völzer's paper, but it turns out to require quite a bit of formal machinery (given in `FairSchedular.ean`) to prove.

Zulip discussion: [#CSLib > Impossibility of distributed consensus](https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/Impossibility.20of.20distributed.20consensus/with/604941788)


Cslib/Computability/Distributed/FLP/PseudoConsensus.lean
```diff
@@ -0,0 +1,122 @@
...
+public import Cslib.Computability.Distributed.FLP.CanReachVia
+public import Cslib.Computability.Distributed.FLP.FairScheduler
+
+/-! # Fault-tolerant pseudo-consensus
+
+A central idea of [Volzer2004] is the notion of pseudo-consensus, which weakens the notion of
+consensus by replacing the requirement of termination, which is stated in terms of infinite
+executions, by that of pseudo-termination, which is stated in terms of finite executions.
+This makes the notion of pseudo-consensus easier to work with than consensus.  This file
+defines pseudo-consensus and proves that it is implied by consensus.  This resilt is intuitively
```
fmontesi:
```suggestion
defines pseudo-consensus and proves that it is implied by consensus.  This result is intuitively
```

fmontesi (changes requested):
I've just found a typo, rest looks good and can be queued once fixed.

fmontesi (approved):

Cslib/Computability/Distributed/FLP/PseudoConsensus.lean
fmontesi: Huh, I could just apply it myself. It's just a typo, @ctchou, so I'm taking the liberty of applying the fix and merging directly, hope that's ok. :-)

ctchou: Thanks for spotting and fixing the typo!

PR #726, opened 2026-07-17
feat(StackTape): lemmas for mapSome

BoltonBailey:
Add lemmas describing how `StackTape.mapSome` and `cons` interact with `head`/`tail` and the empty tape, 

These are split out of #192 where they are used for reasoning about a left-to-right tape scan.

I'm not going to add any grind lemmas on these, since it seems like it's already covered by the definition.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Cslib/Foundations/Data/StackTape.lean
+++ b/Cslib/Foundations/Data/StackTape.lean
@@ -167,6 +167,10 @@ lemma cons_head?_mapSome (l : List Symbol) :
 @[scoped grind =]
 lemma mapSome_nil : mapSome ([] : List Symbol) = nil := rfl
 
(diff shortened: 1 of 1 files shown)


SamuelSchlesinger:
Are you still planning work that needs these? They look mostly reasonable, but I am not sure if they will have users.

BoltonBailey:
Yes, IIRC, these lemmas were needed in some of the complexitylib containment results, so I'd like to get them in eventually.

SamuelSchlesinger:
I think we can delay adding them when we actually get to those results, but if you really think we should add them I am open to it.

PR #731, opened 2026-07-18
feat(PFunctor): redefine `PFunctor.FreeM` in terms of `PFunctor.W`

dtumad:
Currently `PFunctor.FreeM P A` and `PFunctor.W (P.add (.const A))` are equivalent but not definitionally equal, so proofs about one can't be translated over to the other automatically, they need to factor through a `FreeM.toW` and `FreeM.ofW` definition. 

This PR redefines `PFunctor.FreeM` so that they are the same by construction. Case analysis lemmas are tagged with `@[cases_eliminator]` and `@[induction_eliminator]` so those tactics don't break (although manual pattern matching does). It also adds definitions of `PFunctor.monomial` and `PFunctor.add` that are used in the definition, and a short test file checking that `cases` and `induction` don't break.

Definitions in `Basic.lean` mostly come from downstream [PolyFun](https://github.com/Verified-zkEVM/PolyFun) library.


Cslib/Foundations/Data/PFunctor/Free.lean
```diff
@@ -69,42 +78,58 @@ universe u v uA uB
...
-This extends `WType` with an extra `pure` constructor. -/
-inductive FreeM (P : PFunctor.{uA, uB}) : Type v → Type (max uA uB v)
-  /-- A leaf node wrapping a pure value. -/
-  | protected pure {α} (a : α) : P.FreeM α
-  /-- Invoke the operation `a : P.A` with continuation `cont : P.B a → P.FreeM α`. -/
-  | liftBind {α} (a : P.A) (cont : P.B a → P.FreeM α) : P.FreeM α
-deriving Inhabited
+/-- The free monad on a polynomial functor: the W-type of the polynomial functor obtained by\
+adjoining a constant shape for each pure value to `P`. -/
+def FreeM (P : PFunctor.{uA, uB}) (α : Type v) : Type (max uA uB v) :=
```
eric-wieser, line 83: Mathlib has increasingly come to the conclusion that newer lean versions are making it a bad idea to ever define a type with a `def`; either you should embrace the instances on the underyling type and use an `abbrev`, or make a one field structure with conversion constructors and projectors (toW and ofW).

dtumad: Done, that makes sense to me especially with the module system now. I also used AI in trying to identify standard API  surface for similar definitions (injectivity, equivalence, etc.) but let me know if there's anything else that seems obvious.

I did also check that this doesn't break [PolyFun definitions](https://github.com/Verified-zkEVM/PolyFun) really, and I think we are probably the main consumer still.

Cslib/Foundations/Data/PFunctor/Free.lean
```diff
@@ -69,42 +86,112 @@ universe u v uA uB
...
+lemma ofW_injective : Function.Injective (ofW : _ → P.FreeM α) :=
+  fun _ _ h => congrArg toW h
+
+@[simp] lemma toW_inj {x y : P.FreeM α} : x.toW = y.toW ↔ x = y := toW_injective.eq_iff
+
+@[ext] lemma ext {x y : P.FreeM α} (h : x.toW = y.toW) : x = y := toW_injective h
+
+/-- The equivalence between the free monad and its underlying W-type, given by `FreeM.toW`
+and `FreeM.ofW`. Both round-trips hold definitionally. -/
+def equivW : P.FreeM α ≃ PFunctor.W (P.add (.const α)) where
```
eric-wieser, line 117: With the casting functions in place, my question now becomes: is there any advantage to actually changing the definition of FreeM vs just building this equivalence from the old definition?

Cslib/Foundations/Data/PFunctor/Basic.lean
```diff
@@ -0,0 +1,101 @@
...
+@[expose] public section
+
+universe uA uB uA₁ uA₂
+
+namespace PFunctor
+
+section monomial
+
+/-- The monomial `PFunctor` with head type `A` and constant `B` for any `a : A`. -/
+@[reducible] def monomial (A : Type uA) (B : Type uB) : PFunctor := ⟨A, fun _ => B⟩
```
eric-wieser: May as well use the shorter
```suggestion
abbrev monomial (A : Type uA) (B : Type uB) : PFunctor := ⟨A, fun _ => B⟩
```
though the difference is just `@[inline]` which is irrelevant since this has no computation content. Same throughout.

Cslib/Foundations/Data/PFunctor/Basic.lean
```diff
@@ -0,0 +1,101 @@
...
+
+end zero
+
+section one
+
+/-- The unit polynomial functor, defined as `A = PUnit` and `B _ = PEmpty`, is the identity with
+  respect to product (up to equivalence) -/
+@[reducible] protected def one : PFunctor := monomial PUnit PEmpty
+
+instance instOnePFunctor : One PFunctor where one := PFunctor.one
```
eric-wieser: I'd probably argue for inlining these, to avoid needing two spellings for the same thing:
```suggestion
/-- The unit polynomial functor is defined as `A = PUnit` and `B _ = PEmpty`, and is the identity with
  respect to product (up to equivalence) -/
instance : One PFunctor where one := monomial PUnit PEmpty
```

eric-wieser: (same below)

Cslib/Foundations/Data/PFunctor/Free.lean
dtumad: I guess I'd still think it is nice for the inverse laws of the equivalence to be definitional `rfl` rather than just propositional by induction, so that if we have say `motive : P.FreeM A -> Sort u` then we can use something of type `motive (ofW (toW x))` as a `motive x` without needing a `▸` call.

I don't think this really escalates to `Heq` annoyance more generally, but it does seem maybe nice for that narrower situation? Not sure if that's enough to justify the change now.

Cslib/Foundations/Data/PFunctor/Basic.lean
dtumad: This is usually my feeling but I see sometimes monads for example defining a manual `pure` operation seperate from the instance so not sure on the usual conventions (I guess maybe for `pure` it is because the generic `bind` can be more universe generic than `>>=`).

Cslib/Foundations/Data/PFunctor/Basic.lean
```diff
@@ -0,0 +1,97 @@
...
+section add
+
+/-- The sum of two polynomial functors `P` and `Q`, written as `P + Q`,
+defined as the sum of the head types and the sum case analysis for the child types. -/
+def add (P : PFunctor.{uA₁, uB}) (Q : PFunctor.{uA₂, uB}) :
+    PFunctor.{max uA₁ uA₂, uB} := ⟨P.A ⊕ Q.A, Sum.elim P.B Q.B⟩
+
+instance instHAddPFunctor :
+  HAdd PFunctor.{uA₁, uB} PFunctor.{uA₂, uB} PFunctor.{max uA₁ uA₂, uB} where
+  hAdd := add
```
eric-wieser, line 85: Similarly, do we need both here?

Cslib/Foundations/Data/PFunctor/Basic.lean
```diff
@@ -0,0 +1,97 @@
...
+
+@[simp] lemma add_A (P Q : PFunctor) : (add P Q).A = (P.A ⊕ Q.A) := rfl
+
+@[simp] lemma add_B_inl (P : PFunctor.{uA₁, uB}) (Q : PFunctor.{uA₂, uB}) (a : P.A) :
+    (add P Q).B (.inl a) = P.B a := rfl
```
eric-wieser, line 90: This is ill-typed, because `(add P Q).B` is not a function from `Sum _ _` but from `(add P Q).A` which isn't reducible

Cslib/Foundations/Data/PFunctor/Basic.lean
eric-wieser, line 1: Can I suggest you first make a PR that just adds this file, without the link to FreeM yet?

Cslib/Foundations/Data/PFunctor/Basic.lean
```diff
@@ -0,0 +1,97 @@
...
+
+section one
+
+/-- The unit polynomial functor, defined as `A = PUnit` and `B _ = PEmpty`, is the identity with
+  respect to product (up to equivalence) -/
+instance instOnePFunctor : One PFunctor where one := monomial PUnit PEmpty
+
+@[simp] lemma one_A : (1 : PFunctor).A = PUnit := rfl
+
+@[simp] lemma one_B (a : (1 : PFunctor).A) : (1 : PFunctor).B a = PEmpty := rfl
```
eric-wieser, line 60: Does `simps` generate these?

Cslib/Foundations/Data/PFunctor/Basic.lean
dtumad: Split this off into #803

Cslib/Foundations/Data/PFunctor/Basic.lean
dtumad: I think so, or at least it seems useful with universe level changes because `+` notation seems to unify earlier (see tests in #803).

dtumad:
THe seperated PR has the more useful first changes, and it's not clear how usefull these changes are with better modularity of definitions. Some changes alread encourage defEq abuse, should resolve the connections in a more idiomatic way

PR #738, opened 2026-07-21
refactor: remove FullBeta.step_subst_cong_r

lengyijun:
`FullBeta.steps_subst_cong_r` is the more general version.

Updated `steps_subst_cong_r` to perform induction on `LC s` (instead of on the multi-step reduction) and inlined the previous single-step logic into the `app` case.

### DIFF
--- a/Cslib/Languages/LambdaCalculus/LocallyNameless/Untyped/FullBeta.lean
+++ b/Cslib/Languages/LambdaCalculus/LocallyNameless/Untyped/FullBeta.lean
@@ -156,28 +156,17 @@ lemma steps_open_cong_l_abs
     specialize ih s
     cases step with grind [invert_steps_abs, step_open_cong_l (L := free_union Var)]
 
-/- `t ↠βᶠ t'` implies `s[x := t] ↠βᶠ s[x := t']`.
-   There is no single step lemma in this case because x
-   may be substituted for n times, so a single step t ↠βᶠ t
-   in general requires n steps in `s[x := t] ↠βᶠ (s[x := t'])` -/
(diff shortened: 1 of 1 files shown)


Cslib/Languages/LambdaCalculus/LocallyNameless/Untyped/FullBeta.lean
```diff
@@ -156,28 +156,17 @@ lemma steps_open_cong_l_abs
...
-
 /- `step_subst_cong_r` can be generalized to multiple reductions `t ↠βᶠ t'`.
    This requires s to be locally closed, locally closedness of t and t'
    can be inferred by the fact t reduces to t' -/
 lemma steps_subst_cong_r {x : Var} (s t t' : Term Var) (step : t ↠βᶠ t') (h_lc : LC s) :
     (s[x := t]) ↠βᶠ (s[x := t']) := by
-  induction step with
-  | refl => rfl
-  | tail steps step ih => grind [Relation.ReflTransGen.trans, step_subst_cong_r]
+  induction h_lc with
```
chenson2018, line 164: While it is technically true that this is more general, I think it is reasonable to want this theorem stated for both single and multiple reductions for usability? It seems like you've just inlined the proof of `step_subst_cong_r` here, which I'm not sure is an improvement.

lengyijun: One of my motivations was to clean up duplicate code and make room to write new theorems.

chenson2018: I don't understand what you mean by "make room to write new theorems". While they are often interchangeable, I think we'd still like distinct theorems for single steps. I'm going to close this for now.

chenson2018:
Closing for the reasons described above.

PR #739, opened 2026-07-21
Added IsRegular.iff_dfa' to avoid using equiv with Fin n

chiyunhsu:


ctchou:
I just want to point out that mathlib already has the infrastructure for "reindex"ing the state space of a DFA via an equivalence:
https://leanprover-community.github.io/mathlib4_docs/Mathlib/Computability/DFA.html#DFA.reindex

chiyunhsu:
Thanks for pointing it out! The PR was a mistake during our WIP progress though.

PR #741, opened 2026-07-22
feat(LambdaCalculus/Named/Untyped): Alpha equivalence equalities

chris-anto-froeschl:
Formalizes the definitions and proofs from Roy L. Crole's paper [_Alpha equivalence equalities_](https://www.sciencedirect.com/science/article/pii/S0304397512000667).

This PR currently contains the formalization of Theorem 4.1 and its prerequisites to keep the review manageable. The remaining results are intended as follow-up PRs.

**Status:** blocked-by-other-PR. This PR depends on the open [Basics of named representation](https://github.com/leanprover/cslib/pull/458) PR and currently targets an intermediate revision that will need rebasing once that PR is merged.

The overall structure closely follows the original paper to simplify cross-referencing between the formalization and the published proofs. Feedback on the organization of the files and the placement of helper definitions (e.g. `swap` and `permute`) is welcome.

### `AlphaEquivDefs.lean`

Defines the α-equivalence relations studied in the paper. The definition corresponding to Definition 3.1 reuses the α-equivalence relation introduced in [#458](https://github.com/leanprover/cslib/pull/458) (`Basic.lean`).

### `AlphaEquivEquiv.lean`

Formalizes the main equivalence results between the different α-equivalence definitions.

### `SwapProperties.lean`

Contains helper lemmas about swaps and permutations used throughout the proofs.

## Notes

Early development used a [custom implementation of `swap`](https://github.com/chris-anto-froeschl/cslib/commit/ed8d98b854d38d202933fe08eedc540346e37d5b). This has since been replaced by [`Equiv.swap`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Logic/Equiv/Basic.html#Equiv.swap) from `Mathlib.Logic.Equiv.Basic`. I think the current version should be preferred.

## AI usage

Aristotle was used during exploration of alternative definitions and auxiliary swap lemmas. The final proofs were substantially rewritten, manually verified, and checked against the original paper.


chris-anto-froeschl:
After the merge of #458, I took the liberty to refactor some operator-related properties into separate files. The current version therefore distinguishes between properties specific to swaps (`SwapProperties.lean`), renaming (`RenameProperties.lean`), and (main def) alpha equivalence (`AlphaEquivProperties.lean`).

I am open to discussion as to whether this is indeed the most elegant split. A single, larger `Properties.lean` could allow for some more elegant proofs. For example, in my local version of the follow-up PR, I prove [reflexivity of `AlphaEquiv`](https://github.com/chris-anto-froeschl/cslib/blob/alpha-equivalence-equalities-rest/Cslib/Languages/LambdaCalculus/Named/Untyped/Properties.lean#L493) using [some swap properties](https://github.com/chris-anto-froeschl/cslib/blob/alpha-equivalence-equalities-rest/Cslib/Languages/LambdaCalculus/Named/Untyped/Properties.lean#L381). This is possible because the local `Properties.lean` freely mixes swap and `AlphaEquiv` properties as needed.

In general, however, most swap properties themselves depend on some `AlphaEquiv` properties, which prevents a cycle-free import structure. A more fine-grained separation of the files might be able to resolve this dependency issue.

On the other hand, collecting everything in `Properties.lean` would result in a rather large file.

chris-anto-froeschl:
Furthermore, I'm considering cleaning up some of the proofs introduced in #458. Let me know whether you'd prefer me to leave those for a separate PR to avoid adding unnecessary noise to this one.

Cslib/Languages/LambdaCalculus/Named/Untyped/AlphaEquivEquiv.lean
```diff
@@ -0,0 +1,131 @@
...
+/-! ## Theorem 4.5 [Crole2012] -/
+theorem alphaEquiv_iff_alphaEquivRFresh (m n : Term Var) :
+    AlphaEquiv m n ↔ AlphaEquivRFresh m n := by
+  sorry
+
+/-! ## Theorem 4.6 [Crole2012] -/
+theorem alphaEquivR_iff_alphaEquivRFresh (m n : Term Var) :
+    AlphaEquivR m n ↔ AlphaEquivRFresh m n := by
+  sorry
+-/
```
yinhaoxuan: Are these theorems left as future work? I think it's better to finish them in this PR.

Cslib/Languages/LambdaCalculus/Named/Untyped/RenameProperties.lean
```diff
@@ -0,0 +1,100 @@
...
+    y ∉ m.vars ∪ {x, z} → w ∉ m.vars ∪ {x, y, z} →
+    (m.rename x y).rename (if z = x then y else z) w = (m.rename z w).rename x y := by
+  grind [rename_comm_fresh]
+
+omit [DecidableEq Var] in
+theorem induction_by_sizeOf {C : Term Var → Prop}
+    (step : ∀ m : Term Var, (∀ m1 : Term Var, sizeOf m1 < sizeOf m → C m1) → C m ) :
+    ∀ m : Term Var, C m :=
+  WellFounded.fix (r := sizeOfWFRel.rel) sizeOfWFRel.wf step
+
```
yinhaoxuan: This should be put in AlphaEquivProperties because it is not about rename

Cslib/Languages/LambdaCalculus/Named/Untyped/SwapProperties.lean
```diff
@@ -0,0 +1,511 @@
...
+  rw [← swap_rename_comm' (by grind) (by grind)] at h2
+  rw [← swap_rename_comm' (by grind) (by grind)] at h2
+  have ha : a = u ∨ a = v ∨ (a ≠ u ∧ a ≠ v) := by grind
+  have hb : b = u ∨ b = v ∨ (b ≠ u ∧ b ≠ v) := by grind
+  rcases ha with h' | h' | ⟨hau, hav⟩
+  · rcases hb with h'' | h'' | ⟨hbu, hbv⟩ <;> simp_all
+  · rcases hb with h'' | h'' | ⟨hbu, hbv⟩ <;> simp_all
+  · rcases hb with h'' | h'' | ⟨hbu, hbv⟩ <;> simp_all
+
+-- example 1: use z as witness
```
yinhaoxuan: What are the examples 1, 2, 3, 4, ...?

Cslib/Languages/LambdaCalculus/Named/Untyped/Basic.lean
```diff
@@ -63,6 +64,25 @@ def vars : Term Var → Finset Var
...
+  | app m n => app (m.permute π) (n.permute π)
+
+/-- The action of the transposition `(x y)` on a term: simultaneously swaps all occurrences
+of `x` and `y`. Corresponds to `(x y) · E` in [Crole2012] (Section 2).
+
+`swap` is is one special case of a permutation: the transposition that exchanges exactly two atoms
+a and b and fixes everything else.
+-/
+def swap (m : Term Var) (x y : Var) : Term Var := m.permute (Equiv.swap x y)
+
```
yinhaoxuan: I understand that these definitions are only used to define the alternative alpha equivalence definitions. If so, they can be put into AlphaEquivDefs.

yinhaoxuan (commented):
Thank you very much, Chris, for making the changes. The technical parts look fine to me. I have made some comments on the placement of some code.

I agree with you that currently it's not so clear what's the best way to arrange all the files. I like your idea of splitting the previously very big Properties.lean into two files, AlphaEquivProperties.lean and RenameProperties.lean. My complaint with the current arrangement is that when a newcomer arrives at this codebase, it is not clear to them what's the order of reading the files. It is not clear which files are the "mainstream", "must read" (in this case, definition of "the" alpha equivalence and its properties such as the substitution lemma), and which are "nice to have" features (in this case, alternative definition of alpha equivalences and the equivalences between them). So I suggest putting the latter into a subfolder named e.g. "AlternativeAlphaEquivalences", or renaming the files such that this division is clear at first sight.

Cslib/Languages/LambdaCalculus/Named/Untyped/RenameProperties.lean
chris-anto-froeschl: Fixed in [49c4726](https://github.com/leanprover/cslib/pull/741/commits/49c47266bca1914a9c8d58b48a2c537cffdcb152)

Cslib/Languages/LambdaCalculus/Named/Untyped/SwapProperties.lean
chris-anto-froeschl: My bad. This is just the order of the the representative examples displayed for non-trivial cases of Lemma 6.1 in the paper proof. I kept these comments to make a reader aware of the which cases falls into which implicit case of the paper. I adjusted the comments to make this more explicit in [d546f15](https://github.com/leanprover/cslib/pull/741/commits/d546f15bed098d442247974eda59c6f9565bffb0).

Cslib/Languages/LambdaCalculus/Named/Untyped/Basic.lean
chris-anto-froeschl: Fixed in [98218f4](https://github.com/leanprover/cslib/pull/741/commits/98218f4f7b297e359e272b2700c2070fafa5fc97).

I think an older version not doing the <Operation>Property.lean split had some dependency cycles here.
For the long term, I'm not even sure if AlphaEquivDefs is the most ideal content based split. I suppose time will tell as the named representation grows.

PR #743, opened 2026-07-24
fix(MultiTapeTM): clamp multi-tape input-head movement

SamuelSchlesinger:
There is a bug where the input position wraps around instead of getting properly clamped. This PR fixes it and ensures it can't happen again.

### DIFF
--- a/Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
+++ b/Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
@@ -197,12 +197,8 @@ lemma moveInputPos_neg_of_ne_left {n : ℕ} (p : Fin (n + 2)) (h : p ≠ 0) :
     moveInputPos p .neg = ⟨p.val - 1, by have := p.isLt; omega⟩ := by
   have hp : 0 < p.val := Nat.pos_of_ne_zero (fun hz => h (Fin.ext hz))
   unfold moveInputPos
-  rw [dif_pos]
-  · apply Fin.ext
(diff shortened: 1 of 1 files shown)


Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
```diff
@@ -165,12 +165,55 @@ deriving Inhabited
...
+lemma moveInputPos_neg_of_ne_left {n : ℕ} (p : Fin (n + 2)) (h : p ≠ 0) :
+    moveInputPos p .neg = ⟨p.val - 1, by have := p.isLt; omega⟩ := by
+  have hp : 0 < p.val := Nat.pos_of_ne_zero (fun hz => h (Fin.ext hz))
+  unfold moveInputPos
+  rw [dif_pos]
+  · apply Fin.ext
+    simp
+    omega
+  · simp
+    omega
```
crei:
```suggestion
  apply Fin.ext
  rw [dif_pos] <;> simp <;> omega
```

crei (commented):
Thanks for spotting! I just have a proof golfing suggestion.

SamuelSchlesinger:
Suggestion applied! @fmontesi I think we should get this in quick so folks don't build on top of this with the bug, fixing Turing machine proofs even for minor changes can be a PITA.

SamuelSchlesinger:
I have a multi-tape TM composition PR that I will make once this is in, as well.

crei (approved):

PR #745, opened 2026-07-24
fix(MultiTapeTM): Remove `output` from the configuration type of a Turing machine

crei:
The configuration should only contain data that influences the future behaviour of the Turing machine. While it is convenient to put the "output so far" in the configuration, it leads to more problems down the line, for example when the number of reachable configurations is determined or if we want to reason about loops in the configuration graph.

NB: claude was used for a review of this PR and it did some refactoring.


SamuelSchlesinger (approved):
LGTM, just remove the period from the PR title.

PR #748, opened 2026-07-25
refactor(LTS): refactor LTS/Relation properties

SamuelSchlesinger:
The current definition requires a global bound, which seems wrong. At least it is misleading as a reader of the documentation, perhaps it is intended and the documentation simply confused me.

### DIFF
--- a/Cslib/Foundations/Semantics/LTS/Basic.lean
+++ b/Cslib/Foundations/Semantics/LTS/Basic.lean
@@ -25,11 +25,12 @@ relation `Tr` between states. We follow the style and conventions in [Sangiorgi2
 - `LTS.MTr` extends the transition relation of any LTS to a multistep transition relation,
 formalising the inference system and admissible rules for such relations in [Montesi2023].
 
-- `LTS.Bounded`, `LTS.Terminating`, and `LTS.Acyclic` distinguish globally bounded execution
-length, absence of infinite executions, and absence of nonempty cycles.
+- `LTS.BoundedUpTo` records an explicit global execution-length bound. `LTS.Bounded`,
+`LTS.Terminating`, and `LTS.Acyclic` distinguish globally bounded execution length, absence of
+infinite executions, and absence of nonempty cycles.
 
 - Definitions for all the common classes of LTSs: image-finite, finitely branching, finite-state,
-finite, and deterministic.
+and deterministic.
 
 ## Main statements
 
@@ -363,9 +364,13 @@ attribute [instance] FinitelyBranching.image_finite FinitelyBranching.finite_sta
 /-- Every LTS with finite types for states and labels is also finitely branching. -/
 instance FinitelyBranching.of_finite [Finite State] [Finite Label] : lts.FinitelyBranching where
 
+/-- An LTS is bounded up to `n` if every finite execution has length strictly less than `n`. -/
+def BoundedUpTo (lts : LTS State Label) (n : ℕ) : Prop :=
+  ∀ s1 μs s2, lts.MTr s1 μs s2 → μs.length < n
+
 /-- An LTS is bounded if there is a global bound on the length of all of its finite executions. -/
 class Bounded (lts : LTS State Label) where
-  bounded : ∃ n, ∀ s1 μs s2, lts.MTr s1 μs s2 → μs.length < n
+  bounded : ∃ n, lts.BoundedUpTo n
 
 /-- An LTS is terminating if its underlying unlabelled transition relation is terminating,
 equivalently if it admits no infinite execution. -/
@@ -378,12 +383,6 @@ class Acyclic (lts : LTS State Label) where
   [acyclic : Relation.Acyclic lts.toRelation]
 
 attribute [instance] Acyclic.acyclic
-
-/-- An LTS is finite if it is finite-state and acyclic.
-
-We call this `FiniteLTS` instead of just `Finite` to avoid confusion with the standard `Finite`
-class. -/
-class FiniteLTS [Finite State] (lts : LTS State Label) extends lts.Acyclic
 
 end Classes
 
--- a/Cslib/Foundations/Semantics/LTS/Basic.lean
+++ b/Cslib/Foundations/Semantics/LTS/Basic.lean
@@ -68,7 +68,7 @@ structure LTS (State : Type u) (Label : Type v) where
 namespace LTS
 
 /-- The unlabelled transition relation underlying an LTS. -/
-def toRelation (lts : LTS State Label) : State → State → Prop :=
+def UnlabelledTr (lts : LTS State Label) : State → State → Prop :=
   fun s1 s2 => ∃ μ, lts.Tr s1 μ s2
 
 section MultiStep
@@ -375,12 +375,12 @@ class Bounded (lts : LTS State Label) where
 /-- An LTS is terminating if its underlying unlabelled transition relation is terminating,
 equivalently if it admits no infinite execution. -/
 class Terminating (lts : LTS State Label) where
-  terminating : Relation.Terminating lts.toRelation
+  terminating : Relation.Terminating lts.UnlabelledTr
 
 /-- An LTS is acyclic if its underlying unlabelled transition relation contains no nonempty
 cycle. -/
 class Acyclic (lts : LTS State Label) where
-  [acyclic : Relation.Acyclic lts.toRelation]
+  [acyclic : Relation.Acyclic lts.UnlabelledTr]
 
 attribute [instance] Acyclic.acyclic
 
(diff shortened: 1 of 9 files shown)


ctchou:
I don't know @fmontesi's original intention when he made this definition.  But, prima facie, shouldn't `Acyclic` be defined by something like this?
```
class Acyclic (lts : LTS State Label) where
  acyclic : ¬ ∃ s μs, lts.MTr s μs s ∧ 0 < μs.length
```
It seems to me that the current definition should be called "Bounded" (perhaps with `n` as an explicit parameter) and the new definition that @SamuelSchlesinger proposes should be called "WellFounded".

SamuelSchlesinger:
Yes, your proposal makes the most semantic sense to me.

SamuelSchlesinger:
I think the bounded definition also can choose to order the quantifiers differently and those are different types of `Bounded`. `GloballyBounded` vs `LocallyBounded` or something.

thomaskwaring:
i agree both that the name is confusing (it should be more like @ctchou's suggestion) and that the docs don't match the code — fwiw in Sangiorgi's book "acyclic" is only ever applied to finite-state lts's for which all of these notions agree

also given the swapping of arguments, perhaps `WellFounded` could be replaced by cslib's `Terminating`?

ctchou:
Indeed, since transition labels don't really matter in these definitions, perhaps they should be defined on `Relation` rather than `LTS`?

fmontesi:
> i agree both that the name is confusing (it should be more like @ctchou's suggestion) and that the docs don't match the code — fwiw in Sangiorgi's book "acyclic" is only ever applied to finite-state lts's for which all of these notions agree

Formalising that book is where the def came from, but indeed we should improve this.

Summarising the thoughts expressed so far and my own, I'd suggest:
- Rename the current `Acyclic` to `Bounded` (we can later separate globally from locally, but I think `Bounded` would make sense as interim since it is the obvious global one that says 'the lts has a bound to the length of all possible executions').
- Rename @SamuelSchlesinger's proposal into WellFounded.
- Make a new `Acyclic` that actually checks that there are no loops (as in graphs). We can do this in another PR, if you prefer.

SamuelSchlesinger:
I think that makes sense. I'll reform this PR into what I believe is the synthesis of this discussion, and we can do another round of discussion.

Cslib/Foundations/Semantics/LTS/Basic.lean
```diff
@@ -356,9 +363,21 @@ attribute [instance] FinitelyBranching.image_finite FinitelyBranching.finite_sta
 /-- Every LTS with finite types for states and labels is also finitely branching. -/
 instance FinitelyBranching.of_finite [Finite State] [Finite Label] : lts.FinitelyBranching where
 
-/-- An LTS is acyclic if there are no infinite multistep transitions. -/
+/-- An LTS is bounded if there is a global bound on the length of all of its finite executions. -/
+class Bounded (lts : LTS State Label) where
+  bounded : ∃ n, ∀ s1 μs s2, lts.MTr s1 μs s2 → μs.length < n
```
ctchou: It may be useful to define an `LTS.BoundedUpTo n` to make the bound `n` into an explicit parameter and then define `LTS.Bounded` in terms of it.  For example, it would allow a more refined version of the theorem `Acyclic.toBounded` below.

Cslib/Foundations/Semantics/LTS/Termination.lean
```diff
@@ -20,6 +25,83 @@ universe u v
...
+/-- A multistep transition admits a chain of its visited states. -/
+private theorem exists_state_chain (h : lts.MTr s1 μs s2) :
+    ∃ states : List State,
+      (s1 :: states).length = μs.length + 1 ∧
+        (s1 :: states).IsChain lts.toRelation := by
+  induction h with
+  | refl => exact ⟨[], by simp⟩
+  | @stepL s1 μ s2 μs s3 htr hmtr ih =>
+      obtain ⟨states, hlength, hchain⟩ := ih
+      exact ⟨s2 :: states, by simp [hlength], .cons_cons ⟨μ, htr⟩ hchain⟩
```
ctchou: I suspect you can use the notion of `LTS.Execution` and related theorems in `LTS.Execution.lean` to replace this private theorem.

Cslib/Foundations/Semantics/LTS/Termination.lean
```diff
@@ -20,6 +25,83 @@ universe u v
...
+  · rintro h ⟨s, μs, hmtr, hlength⟩
+    exact h.acyclic.irrefl s (hmtr.toTransGen lts (List.ne_nil_of_length_pos hlength))
+  · intro h
+    refine { acyclic := ⟨fun s hcycle => ?_⟩ }
+    obtain ⟨μs, hne, hmtr⟩ := (transGen_toRelation_iff lts).mp hcycle
+    exact h ⟨s, μs, hmtr, List.length_pos_of_ne_nil hne⟩
+
+/-- Finite LTSs are bounded. -/
+instance finiteLTS_bounded [Finite State] [lts.FiniteLTS] : lts.Bounded :=
+  (inferInstance : lts.Acyclic).toBounded
```
ctchou: I think we should get rid of the definition `FiniteLTS`.  The name is very misleading.  Anyone who has not read the definition would have been very puzzled by this result, which is really just `Acyclic.toBounded` above.

Cslib/Foundations/Semantics/LTS/Termination.lean
```diff
@@ -20,6 +25,83 @@ universe u v
...
+/-- An LTS is acyclic exactly when it has no nonempty multistep cycle. -/
+theorem acyclic_iff_no_nonempty_mTr :
+    lts.Acyclic ↔ ¬ ∃ s μs, lts.MTr s μs s ∧ 0 < μs.length := by
+  constructor
+  · rintro h ⟨s, μs, hmtr, hlength⟩
+    exact h.acyclic.irrefl s (hmtr.toTransGen lts (List.ne_nil_of_length_pos hlength))
+  · intro h
+    refine { acyclic := ⟨fun s hcycle => ?_⟩ }
+    obtain ⟨μs, hne, hmtr⟩ := (transGen_toRelation_iff lts).mp hcycle
+    exact h ⟨s, μs, hmtr, List.length_pos_of_ne_nil hne⟩
```
ctchou: I think this theorem belongs to `LTS/Basic.lean`.

CslibTests/LTS.lean
```diff
@@ -83,6 +85,65 @@ example : natDivLTS.Divergent n := by
     constructor
   · simp [natInfiniteExecution]
 
+-- The three execution properties are distinct on infinite-state LTSs.
```
ctchou: I think these results should be put under `Cslib`, rather than `CslibTests`, in a new file `.../LTS/ExampleXXX.lean`, though I don't have a good idea about what `XXX` should be.

ctchou (changes requested):

ctchou (changes requested):
BTW, the title of this PR should change, too.

SamuelSchlesinger:
@ctchou addressed your comments and deleted a noisy theorem with no use sites.

ctchou (approved):
LGTM.

SamuelSchlesinger:
@fmontesi can you re-review and/or merge?

Cslib/Foundations/Semantics/LTS/Basic.lean
```diff
@@ -63,6 +67,10 @@ structure LTS (State : Type u) (Label : Type v) where
 
 namespace LTS
 
+/-- The unlabelled transition relation underlying an LTS. -/
+def toRelation (lts : LTS State Label) : State → State → Prop :=
```
fmontesi: The entire PR looks good to me now but I have a minor annoyance here:

This has exactly the same name as LTS.Tr.toRelation and LTS.MTr.toRelation, which have the 'expected' meaning from the literature -- fixing the (multi)transition relation on a label. This new LTS.toRelation should probably go in Relation.lean, be namespaced under LTS.Tr, and have a different name to distinguish it from the other ones.

fmontesi (changes requested):

Cslib/Foundations/Semantics/LTS/Basic.lean
SamuelSchlesinger: Good catch. Busy today but when I get online I'll patch this.

SamuelSchlesinger: If anyone else wants this merged, feel free to push this last commit and merge it.

SamuelSchlesinger: I prefer to keep it in `LTS` for dot notation, but instead we'll rename it to `UnlabelledTr`. If we move it to `Relation.lean` it has annoying ripple effects for the rest of this PR that I think are undesirable. I'll make that commit for now and see what you think.

fmontesi: Good enough for now, I think!

fmontesi (approved):

PR #713, opened 2026-07-13
feat(LTS/Spectrum): van Glabbeek spectrum as a Galois connection

patchwright:
## Summary

Organises behavioural equivalences as the fixed points of an antitone Galois connection ("polarity") between test classes (ordered by ⊆) and equivalences (ordered by refinement). The Galois-closed equivalences — the "testable" ones — are exactly the image of `induced`. Two named behavioural equivalences are exhibited as closed elements of the SAME polarity: **trace equivalence** (kernel-trivial) and **bisimilarity** (Hennessy–Milner-mediated) — the actual "spectrum as Galois" statement.

## What's added

New module `Cslib.Foundations.Semantics.LTS.Spectrum` (4 files):

- **`Galois.lean`** — the framework (`Cslib.Init` only): `induced Ω T`, `respects Ω E`, closure operator `cl Ω`, `polarity` (the antitone Galois connection), `induced_testable`, `spectrum_eq_closed_elements`.
- **`TracePoint.lean`** — `HomTraceEq_testable`: homogeneous trace equivalence is Galois-closed, via the trace-set observer test class. (Kernel-trivial bottom point.)
- **`BisimPoint.lean`** — `HomBisimilarity_testable`: homogeneous bisimilarity is Galois-closed for image-finite LTS, via the Hennessy–Milner theorem (`theoryEq_eq_bisimilarity`). (**Non-trivial** top point — bisimilarity is not a kernel by definition; HM makes it one.)
- **`Antichain.lean`** — `exists_incomparable_closed`: the closed-element lattice contains an antichain (a lattice, not a chain).

## Why it's additive

CSLib has `Bisimilarity`, `Simulation`/`SimulationEquiv`, `TraceEq`, HML, and the Hennessy–Milner theorem (`theoryEq_eq_bisimilarity`), but no organisation of the spectrum as a Galois structure. This provides that organisation, reusing CSLib's existing equivalences + HM as named closed instances. It does not redefine any existing equivalence.

## Scope notes / honesty

- The trace point is **kernel-trivial**: the test class IS the trace-set map, so `induced = HomTraceEq` is near-definitional. It's the honest bottom point; the per-trace `Bool`-test refinement (decidability-gated) is a finer witness for the same closed element.
- The bisim point is the **non-trivial** one: it relies on HM (`theoryEq_eq_bisimilarity`) to connect bisimilarity to the HML-theory kernel. That step does real work — it's what earns the word "spectrum."
[...]

PR #752, opened 2026-07-26
feat(LTS/Spectrum): may- and must-testing equivalence as Galois-closed points

patchwright:
## Summary

A follow-up to #713 (van Glabbeek spectrum as a Galois connection). Adds two
more named points to the spectrum: **may-testing** and **must-testing**
equivalence [De Nicola & Hennessy, 1984] — the classical testing-equivalence
pair that gives the framework in #713 its name.

## What's added

Two new modules in `Cslib.Foundations.Semantics.LTS.Spectrum` (imports wired
into `Cslib.lean`, `lake exe mk_all --check` confirms no diff):

- **`MayTestingPoint.lean`** — `MayEquiv_testable`: may-testing equivalence
  (`p` may-passes the same tests as `q`) is Galois-closed. Definitional, like
  `TracePoint`: it *is* `induced` of the may-test observer class.
- **`MustTestingPoint.lean`** — `MustEquiv_testable`: must-testing equivalence
  (every maximal computation of `p ∥ T` succeeds) is Galois-closed. Also
  definitional. "Maximal computation" covers both finite-and-stuck and
  infinite (`OmegaExecution`) runs, so divergence-without-success and
  deadlock-without-success both correctly fail `must`.

Both are instantiated concretely over CCS (`Cslib.Languages.CCS`), matching
the existing point files' pattern.

## Why it's additive

`#713` established the Galois-connection framework and two points (trace,
bisimilarity). May- and must-testing are the equivalences the *testing*
literature is actually about — De Nicola–Hennessy is the paper the framework
is named after — so this is the natural next pair, not a new direction. It
does not touch `Galois.lean` or either existing point file.

## Scope notes / honesty

- Both points are **definitional**, like `TracePoint` — no theorem does the
  connective work `BisimPoint`'s Hennessy–Milner step does. That's expected:
  may/must-testing equivalences are *defined* as induced-by-test-observers in
  the source literature, unlike bisimilarity.
- `must`'s "maximal computation" clause requires both the finite-stuck case
  and the `OmegaExecution` case to succeed, matching the standard must-testing
  definition (no silent divergence escape).
- `mustPass`'s "stuck" condition reuses `Cslib.LTS.Stuck` (from
  `Foundations/Semantics/LTS/Termination.lean`), instantiated with
  `Terminated := fun _ => False`, rather than introducing a second local
  notion — testing has no independent designated-terminal-state concept, so
  a state is a testing endpoint exactly when it has no outgoing transition,
  and `¬False ∧ P ↔ P` makes the instantiation propositionally exact.

## Verification

Full `lake build`: 0 errors, 2764/2764 jobs. `lake exe mk_all --check`: no
diff. `lake exe checkInitImports`: clean. 0 `sorry` / 0 `axiom` in both new
files.

## AI use disclosure

Per CSLib's AI policy (following Mathlib): this contribution was developed
with AI assistance, same disclosure as #713.

## Open questions for maintainers

- Given #713 is still under review, happy to hold this or adjust scope if a
  `Galois.lean` API change lands there first — this PR doesn't depend on any
  pending change to #713, but a rename there would need a follow-up here.

### DIFF
--- a/Cslib.lean
+++ b/Cslib.lean
@@ -101,6 +101,8 @@ public import Cslib.Foundations.Semantics.LTS.Simulation
 public import Cslib.Foundations.Semantics.LTS.Spectrum.Antichain
 public import Cslib.Foundations.Semantics.LTS.Spectrum.BisimPoint
 public import Cslib.Foundations.Semantics.LTS.Spectrum.Galois
+public import Cslib.Foundations.Semantics.LTS.Spectrum.MayTestingPoint
+public import Cslib.Foundations.Semantics.LTS.Spectrum.MustTestingPoint
 public import Cslib.Foundations.Semantics.LTS.Spectrum.TracePoint
 public import Cslib.Foundations.Semantics.LTS.Termination
 public import Cslib.Foundations.Semantics.LTS.Total
--- a/Cslib.lean
+++ b/Cslib.lean
@@ -98,6 +98,9 @@ public import Cslib.Foundations.Semantics.LTS.Notation
 public import Cslib.Foundations.Semantics.LTS.OmegaExecution
 public import Cslib.Foundations.Semantics.LTS.Relation
 public import Cslib.Foundations.Semantics.LTS.Simulation
+public import Cslib.Foundations.Semantics.LTS.Spectrum.Antichain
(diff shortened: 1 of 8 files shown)


chenson2018:
Please see my comments on #713

PR #753, opened 2026-07-29
feat(Logic): improvements to the inference system and congruence frameworks, with associated refactorings

fmontesi:
This PR introduces several improvements to the frameworks based on `InferenceSystem` and `Congruence`.

1. `InferenceSystem.Default⇓a` is now pretty-printed to `⇓a` (which is what users write).
2. Introduces a parametric congruence notation, since terms of the same type can have multiple congruences. Consequently, the same type can now also have multiple logical equivalences (which is important, since these may be parametric as we have, e.g., in modal and propositional logics).
3. Introduces a `DefaultCongruence` class to define the 'canonical' congruence for a type.
3. Separates classes for accessing congruence notation from that to prove that a congruence is lawful (`LawfulCongruence`).
4. Simplifies instantiating `Context` by making the term type parameter implicit.
5. Introduces the first examples of rewritings based on `LogicalEquivalence` in `CslibTests`. Both a manual proof based on contexts and one based on grind are provided. In the future, we should explore ways to plug such rewritings into Lean's and Mathlib's frameworks for congruences (perhaps by autogenerating lemmas and/or checks from inductive types).

The PR also refactors CLL, HML, and Modal Logic to be in line with all these recent developments. This is a step towards the formulation of general patterns for modal logics (left to future work).

Some tests to prevent future breakage are introduced as well.

Depends on #607.


thomaskwaring (approved):
seems like a reasonable framework to me :-)

PR #754, opened 2026-07-29
feat(Languages): Stateful Processes and Mech directory

fmontesi:
This PR adds a first formalisation of Stateful Processes (for now without recursion), the textbook process calculus for reasoning about implementation of choreographies (syntactic descriptions of concurrent/distributed protocols). It includes both syntax and semantics of the calculus.

The semantics is split into symbolic (e.g., `x := e; P` evolves to `P` by performing a transition with label `x := e`) and concrete (which equips programs with memory and thus the same action would update variable `x` with the value computed from `e`).
One reason for this modularisation is future reuse: having a symbolic semantics is the basis to formulate extraction algorithms that infer the protocol executed by a network of process programs (choreography extraction in the literature). Another reason is that it is known to simplify some proofs of correctness for compilers from choreographies to processes, which is a future aim (see the README in `Mech`).

Choreographic languages will be part of future PRs.


Cslib/Languages/StatefulProcesses/Basic.lean
```diff
@@ -0,0 +1,198 @@
...
+-- The next syntax would be nice to have to avoid having trailing 0s in examples.
+-- scoped syntax:min "`(SP| " spPre ")" : term
+
+@[inherit_doc Process]
+scoped syntax "`(SP| " spProc ")" : term
+
+scoped macro_rules
+  | `(`(SP| 0)) => `(0)
+  | `(`(SP| $prf:spPre; $pr:spProc)) => `(Process.pre `(SPpre| $prf) `(SP| $pr))
+  -- | `(`(SP| $prf:spPre)) => `(Process.pre `(SPpre| $prf) 0)
```
XYUnknown: Should this line be deleted?

CslibTests/StatefulProcesses.lean
XYUnknown, line 1: Should we add an example for the symbolic lts for processes and networks, or you think one example for the transition of configurations is sufficient?

Cslib/Languages/StatefulProcesses/Network.lean
```diff
@@ -0,0 +1,165 @@
...
+/-! ## Stores, evaluation, and concrete semantics of networks -/
+
+/-- A local store represents the memory state of a process, mapping variables to values. -/
+abbrev LocalStore Var Val := (x : Var) → Val
+
+/-- Type of (potentially nondeterministic) evaluation relations for function calls. -/
+abbrev FunCallEval FunId Val := (f : FunId) → (args : List Val) → Val → Prop
+
+/-- Evaluation relation. -/
+inductive FunCallEval.EvalExpr (Eval : FunCallEval FunId Val) :
```
XYUnknown: It may be easier to read if we change `Eval` to `eval`?

XYUnknown (commented):
I think this is generally ready to be merged.

CslibTests/StatefulProcesses.lean
fmontesi: Agreed it'd be good to add -- good future work for the next PRs.

XYUnknown: ok, agree.

PR #755, opened 2026-07-29
feat(FLP): develop the theory of pseudo-consensus algorithms that can tolerate up to 1 fault

ctchou:
This PR develops a theory of pseudo-consensus algorithms that can tolerate up to 1 fault, which is the core of Völzer's argument.  The central notion is "non-uniformity": a state `s` is non-uniform iff for each boolean value `b`, there exists a process `p` such that there is an execution of the algorithm from `s` that ends up deciding on `b` without the participation of `p`.  Then any algorithm that can tolerate up to 1 fault is proved to have two properties:
(1) `nonUniform_inp`, which says that there must an initial non-uniform state.
(2) `nonUniform_step`,  which will be shown in the next PR to guarantee that any execution ending in a non-uniform state can be extended to a longer execution ending in a non-uniform state.

### DIFF
--- a/Cslib/Computability/Distributed/FLP/OnePseudoConsensus.lean
+++ b/Cslib/Computability/Distributed/FLP/OnePseudoConsensus.lean
@@ -1,0 +1,287 @@
+/-
+Copyright (c) 2026 Ching-Tsun Chou. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Ching-Tsun Chou
+-/
+
+module
+
+public import Cslib.Computability.Distributed.FLP.PseudoConsensus
+
+/-! # 1-tolerant pseudo-consensus
+
+This file develops the theory of pseudo-consensus algorithms that can tolerate up to 1 fault.
+It formalizes section 3 of [Volzer2004] except Theorem 1.
+-/
+
+@[expose] public section
+
+namespace Cslib.FLP
+
+open Function Set Multiset Fintype
+
+variable {P M S : Type*} [DecidableEq P] [DecidableEq M]
+
+/-- `a.OneSilentDecided s p b` means that the boolean value `b` is decided on in a state that
+is reachable from `s` without the participation of `p`. In the notation of [Volzer2004], this
+is equivalent to `b ∈ val(p,s)`. -/
+def Algorithm.OneSilentDecided (a : Algorithm P M S)
+    (s : State P M S) (p : P) (b : Bool) : Prop :=
+  ∃ s', a.CanReachVia {p}ᶜ s s' ∧ s'.Decided b
+
+/-- `a.Uniform s b` means that for every process `p`, `a.OneSilentDecided s p b` but
+not `a.OneSilentDecided s p !b`. -/
+def Algorithm.Uniform (a : Algorithm P M S) (s : State P M S) (b : Bool) : Prop :=
+  ∀ p, a.OneSilentDecided s p b ∧ ¬ a.OneSilentDecided s p !b
+
+/-- `a.NonUniform s` means that for each boolean value `b`, there is a process `p`
+such that `a.OneSilentDecided s p b`. -/
+def Algorithm.NonUniform (a : Algorithm P M S) (s : State P M S) : Prop :=
+  ∀ b, ∃ p, a.OneSilentDecided s p b
+
+namespace OnePseudoConsensus
+
+variable {a : Algorithm P M S} {inp : P → Bool}
+
+/-- Draw a consequence of `a.PseudoConsensus 1`. -/
+theorem pseudoTermination [Fintype P] (hpc1 : a.PseudoConsensus 1)
+    {s : State P M S} (hr : a.Reachable inp s) (p : P) :
+    ∃ s' b, a.CanReachVia {p}ᶜ s s' ∧ s'.Decided b := by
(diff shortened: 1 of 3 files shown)


Cslib/Computability/Distributed/FLP/OnePseudoConsensus.lean
```diff
@@ -0,0 +1,287 @@
...
+namespace Cslib.FLP
+
+open Function Set Multiset Fintype
+
+variable {P M S : Type*} [DecidableEq P] [DecidableEq M]
+
+/-- `a.OneSilentDecided s p b` means that the boolean value `b` is decided on in a state that
+is reachable from `s` without the participation of `p`. In the notation of [Volzer2004], this
+is equivalent to `b ∈ val(p,s)`. -/
+def Algorithm.OneSilentDecided (a : Algorithm P M S)
```
fmontesi: I'd call this `CanDecideWithout` or `CanDecideWithoutOne`.

Cslib/Computability/Distributed/FLP/OnePseudoConsensus.lean
```diff
@@ -0,0 +1,287 @@
...
+    (s : State P M S) (p : P) (b : Bool) : Prop :=
+  ∃ s', a.CanReachVia {p}ᶜ s s' ∧ s'.Decided b
+
+/-- `a.Uniform s b` means that for every process `p`, `a.OneSilentDecided s p b` but
+not `a.OneSilentDecided s p !b`. -/
+def Algorithm.Uniform (a : Algorithm P M S) (s : State P M S) (b : Bool) : Prop :=
```
fmontesi, line 34: I find the naming a bit confusing here:
- For a given boolean b, `Uniform` means that, (1) there's a state reachable without involving p where *some* process has decided b and (2) no state reachable without involving p has a process that's decided !b.
- `NonUniform` means that for any boolean b, there's a process p such that we can reach a state without involving p where some process has decided b.

Do these come from the paper?

Cslib/Computability/Distributed/FLP/OnePseudoConsensus.lean
```diff
@@ -0,0 +1,287 @@
...
+    a.Uniform (a.start (inpN pn (card P))) true := by
+  have h_i := Algorithm.reachable_start (a := a) (inp := inpN pn (card P))
+  obtain (h | h | h) := uniform_or_nonUniform hpc1 h_i
+  · grind [inpN_card_not_false, h (pn.symm ⟨0, by grind⟩)]
+  · exact h
+  · grind [inpN_card_not_false, h false]
+
+/-- Assuming `a.PseudoConsensus 1` and there are at least 2 processes, there must exist an input
+that gives rise to a non-uniform initial state. This theorem formalizes Lemma 1 of [Volzer2004]. -/
+theorem nonUniform_inp (hpc1 : a.PseudoConsensus 1) (hc : card P ≥ 2) :
```
fmontesi, line 259: Very nice result!

The name and content makes me wonder a bit if this lemma (and perhaps also some of the defs above, like `NonUniform`) should be extended in name or namespaced so that it's clear they're about 1-pseudo consensus.

fmontesi (changes requested):

Cslib/Computability/Distributed/FLP/OnePseudoConsensus.lean
ctchou: I'll change it to `CanDecideWithout`.

Cslib/Computability/Distributed/FLP/OnePseudoConsensus.lean
ctchou: Yes, this is from Völzer's paper:
image
The statement `b ∈ val(p,s)` in Völzer's notation is the statement `a.CanDecideWithout s p b` in our notation.  Note that for `val(p,s)={b}`, we need not only `b ∈ val(p,s)`, but also `!b ∉ val(p,s)`.  Now, the negation of `a.Uniform s b` is not exactly equivalent to `∃ p, a.CanDecideWithout s p !b` purely in terms of the definition of `Uniform`.  But they turn out to be equivalent in this context due to Proposition 2(a) (namely, the theorem `canDecideWithout_exists`, or `oneSilentDecided_exists before the renaming).

Personally I find Völzer's `val(p,s)` notation rather indirect and convoluted to reason about, so I had replace its uses by `CanDecideWithout`, `Uniform`, and `NonUniform`.

Cslib/Computability/Distributed/FLP/OnePseudoConsensus.lean
ctchou: All theorems in this file are already in the namespace `OnePseudoConsensus`.

Cslib/Computability/Distributed/FLP/OnePseudoConsensus.lean
fmontesi: Thanks for the screenshot!
Let's go ahead with this then.

Cslib/Computability/Distributed/FLP/OnePseudoConsensus.lean
fmontesi: Ah, I didn't notice. All good then.

fmontesi (approved):

PR #764, opened 2026-07-31
feat(CCS): Milner's Vending Machine

fmontesi:
This PR formalises Milner's classical Vending Machine example for CCS and adds general utilities to LTS, bisimulation, and CCS that have been very useful in developing the example.

One of the reasons to develop this is to test the new AI instructions in #763. Albeit all code is manually written, interacting with the AI to see where it got stuck has been very informative for finding missing utilities.

Depends on #763.


atixwasfound:
The example showing the machine can do a coin action only proves that one transition exists. It does not state that this is the only transition possible at that state, or that the whole deterministic machine is deterministic everywhere. Would it be worth adding an explicit lemma or instance saying the deterministic machine actually satisfies Deterministic (or is deterministic at every reachable state), so the claim in the file's description is backed by a named result and not just used implicitly inside the bigger proof?

fmontesi:
I have that result hidden in this proof:

https://github.com/leanprover/cslib/pull/764/changes#diff-ed8d7e4654d2484fa0d479ebbc2db48e43fe4b146df49f01b62a968623186365R79

But I agree in general, and yes, a proof that the deterministic vending machine is indeed deterministic (as in LTS.Deterministic) would be welcome. 

I'm introducing this specific example also because it's a minimal example to explore infrastructure to make reasoning about this kind of things easier. I think there's much space for improvement in reasoning about unique transitions, traces, etc.

atixwasfound:
makes sense.

Cslib/Foundations/Semantics/LTS/Bisimulation.lean
```diff
@@ -95,6 +95,38 @@ theorem IsBisimulation.follow_snd
...
+    (hb : IsBisimulation lts₁ lts₂ r)
+    (hr : r s₁ s₂)
+    (hdet : lts₁.DeterministicStateLabel s₁ μ)
+    (htr₁ : lts₁.Tr s₁ μ s₁')
+    (htr₂ : lts₂.Tr s₂ μ s₂') : r s₁' s₂' := by
+  grind [follow_snd hb hr]
+
+/-- If a state is deterministic for `μ`, then any transition made by a related state in a
+bisimulation is matched by a unique transition (left variant). -/
+theorem IsBisimulation.follow_fst_deterministic
```
thomaskwaring: this should be a lemma about `IsSimulation`, which generalises and avoids proving the same thing twice

Cslib/Foundations/Semantics/LTS/Bisimulation.lean
```diff
@@ -95,6 +95,38 @@ theorem IsBisimulation.follow_snd
...
+    (hb : IsBisimulation lts₁ lts₂ r)
+    (hr : r s₁ s₂)
+    (hdet : lts₂.DeterministicStateLabel s₂ μ)
+    (htr : lts₁.Tr s₁ μ s₁') : ∃! s₂', lts₂.Tr s₂ μ s₂' ∧ r s₁' s₂' := by
```
thomaskwaring: i don't see a specific note about it in mathlib's style guide, but adding hypotheses on separate lines seems non-standard

Cslib/Foundations/Semantics/LTS/Basic.lean
```diff
@@ -304,24 +313,29 @@ abbrev ImageFinite := ∀ s μ, Finite (lts.image s μ)
 
 /-- In a deterministic LTS, if a state has a `μ`-derivative, then it can have no other
 `μ`-derivative. -/
-@[scoped grind .]
-theorem deterministic_not_lto [h : lts.Deterministic] :
-  ∀ s μ s' s'', s' ≠ s'' → lts.Tr s μ s' → ¬lts.Tr s μ s'' := by grind
+@[scoped grind ⇒]
+theorem deterministicStateLabel_not_lto (hdet : lts.DeterministicStateLabel s μ)
```
thomaskwaring: what does `lto` abbreviate?

Cslib/Foundations/Semantics/LTS/Bisimulation.lean
```diff
@@ -95,6 +95,38 @@ theorem IsBisimulation.follow_snd
     ∃ s₁', lts₁.Tr s₁ μ s₁' ∧ r s₁' s₂' :=
   (hb hr μ).2 _ htr
 
+/-- If the unique transition of a state is matched by a related state in a bisimulation,
+then the derivatives are still in the bisimulation. -/
+theorem IsBisimulation.match_deterministic
```
thomaskwaring, line 117: this generalises to a plain simulation, i think

Cslib/Languages/CCS/Semantics.lean
```diff
@@ -55,6 +53,59 @@ inductive Terminated : Process Name Constant → Prop where
   | choice : Terminated p → Terminated q → Terminated (choice p q)
   | res : Terminated p → Terminated (res a p)
 
-end CCS
+open LTS
+
+/-- A terminated process has no outgoing transitions. -/
+@[scoped grind ⇒]
+theorem terminated_not_tr (h : Terminated p) : ¬(lts (defs := defs)).Tr p μ p' := by
```
thomaskwaring: i think the name convention is `not_tr_of_terminated`

thomaskwaring (changes requested):

Cslib/Foundations/Semantics/LTS/Basic.lean
fmontesi: Whoops, I usually write tr as \lto in latex... :-)

fmontesi:
@thomaskwaring thanks, very good comments. All done.

Cslib/Foundations/Semantics/LTS/Basic.lean
```diff
@@ -314,7 +314,7 @@ abbrev ImageFinite := ∀ s μ, Finite (lts.image s μ)
 /-- In a deterministic LTS, if a state has a `μ`-derivative, then it can have no other
 `μ`-derivative. -/
 @[scoped grind ⇒]
-theorem deterministicStateLabel_not_lto (hdet : lts.DeterministicStateLabel s μ)
+theorem deterministicStateLabel_neq_tr (hdet : lts.DeterministicStateLabel s μ)
```
thomaskwaring, line 317: perhaps better `DeterministicStateLabel.not_tr_of_ne` — at least [it should be](https://leanprover-community.github.io/contribute/naming.html#logic) `ne` not `neq`

Cslib/Foundations/Semantics/LTS/Bisimulation.lean
thomaskwaring, line 117: perhaps we should have the flipped (or un-flipped i guess) version where `lts₂` is deterministic? i'd also prefer a more informative name, but this one is fine

fmontesi: Added. I'd postpone finding a better name to a future PR for now, since this is blocking some other things and I think it'd trigger further renaming. But I think we could look into names such as 'deterministic_tr_fst' or 'deterministic_tr_left' (if the latter, we should rename the 'follow_..' theorems as well).

Cslib/Foundations/Semantics/LTS/Bisimulation.lean
```diff
@@ -83,30 +83,73 @@ def IsBisimulation (lts₁ : LTS State₁ Label) (lts₂ : LTS State₂ Label)
...
+
+/-- If the unique transition of a state is matched by a related state in the inverse of a
+bisimulation, then the derivatives are still in the bisimulation. -/
+theorem IsBisimulation.match_deterministic_flip
+    (hb : IsBisimulation lts₁ lts₂ r)
+    (hr : (flip r) s₂ s₁)
+    (hdet : lts₂.DeterministicStateLabel s₂ μ)
+    (htr₁ : lts₂.Tr s₂ μ s₂')
+    (htr₂ : lts₁.Tr s₁ μ s₁') : r s₁' s₂' := by
+  have hr' : r s₁ s₂ := by grind [flip]
```
thomaskwaring: (untested)
```suggestion
    (hr : r s₁ s₂)
    (hdet : lts₂.DeterministicStateLabel s₂ μ)
    (htr₁ : lts₁.Tr s₁ μ s₁')
    (htr₂ : lts₂.Tr s₂ μ s₂') : r s₁' s₂' := by
```
i ought to have been clearer in my other comment, by "flip" i just meant the symmetric version, no need to have the `flip` be explicit. this might need a name change

thomaskwaring (approved):
lgtm apart from clarifying my intention with the previous comments :)

PR #765, opened 2026-07-31
feat(FLP): prove that asynchronous distributed consensus tolerating any non-zero number of faults is impossible

ctchou:
This PR completes the proof that asynchronous distributed consensus tolerating any non-zero number of faults is impossible.  The main theorem `OnePseudoConsensus.not_terminating` shows that every 1-tolerant pseudo-consensus algorithm has a fair execution which doesn't contain any fault but never reaches a consensus, which then implies that there cannot be a consensus algorithm that can tolerate even a single fault.  This fair execution is constructed using the results obtained in `FLP/OnePseudoConsensus.lean` and the technical machinery developed in `FLP/FairScheduler.lean`.

### DIFF
--- a/Cslib/Computability/Distributed/FLP/Impossibility.lean
+++ b/Cslib/Computability/Distributed/FLP/Impossibility.lean
@@ -1,0 +1,174 @@
+/-
+Copyright (c) 2026 Ching-Tsun Chou. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Ching-Tsun Chou
+-/
+
+module
+
+public import Cslib.Computability.Distributed.FLP.OnePseudoConsensus
+public import Cslib.Foundations.Data.OmegaSequence.InfOcc
+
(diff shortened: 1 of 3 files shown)


Cslib/Computability/Distributed/FLP/Impossibility.lean
```diff
@@ -0,0 +1,174 @@
...
+
+namespace Cslib.FLP
+
+open Function Set Multiset Fintype Filter ωSequence
+
+variable {P M S : Type*} [DecidableEq P] [DecidableEq M]
+
+variable {a : Algorithm P M S}
+
+/-- `a.ReachableNonUniform inp s` means that `s` is a reahable and non-uniform state of
```
fmontesi:
```suggestion
/-- `a.ReachableNonUniform inp s` means that `s` is a reachable and non-uniform state of
```

fmontesi (approved):

PR #767, opened 2026-08-01
feat(MultiTapeTM): Prove time upper-bounds for space-bounded classes

crei:
Proves ` DSPACE s ⊆ ⋃ c, DTIME (fun n => n * 2 ^ (c * s n))`. From this, the inclusions `L ⊆ P` and `PSPACE ⊆ EXP` are proven.

This PR also defines several complexity classes based on the multi-tape Turing machine model.

The key lemma to get to the theorem is a bound of the number of reachable configurations in bounded space. The configurations are split into the actual "storage" part (state, work tape positions, work tape contents) and the input head position. This is not required to prove the result but it will help proving `DSPACE log log n = DSPACE 1` in a subsequent PR.

NB: AI was used to generate this content, but it was heavily edited and checked.

PR #768, opened 2026-08-01
feat(MultiTapeTM): Some lemmas about Turing machine tapes

crei:
Some useful lemmas about how the tapes are modified during the execution of Turing machines, some lemmas about halting behaviour and some about space usage (partly just moved).

This has been split out from #767

### DIFF
--- a/Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
+++ b/Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
@@ -251,17 +251,43 @@ def outputSymbol (cfg : Cfg k Symbol State input) : Option Symbol :=
 def initCfg (input : List Symbol) : Cfg k Symbol State input :=
   ⟨some tm.q₀, 1, fun _ _ => none, fun _ => 0⟩
 
+@[simp]
+lemma step_of_halt {cfg : Cfg k Symbol State input} (h : cfg.state = none) :
+    tm.step cfg = cfg := by
+  unfold step
+  rw [h]
+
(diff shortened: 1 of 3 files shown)


Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
```diff
@@ -251,17 +251,43 @@ def outputSymbol (cfg : Cfg k Symbol State input) : Option Symbol :=
...
+lemma step_of_halt {cfg : Cfg k Symbol State input} (h : cfg.state = none) :
+    tm.step cfg = cfg := by
+  unfold step
+  rw [h]
+
 /-- The sequence of configurations of the Turing machine starting from `cfg`.
 If the Turing machine halts, it will stay at the halting configuration. -/
 def configs (cfg : Cfg k Symbol State input) (t : ℕ) : Cfg k Symbol State input := tm.step^[t] cfg
 
-/-- Any number of steps run from a halting configuration results in the same configuration. -/
```
crei, line 258: Initially, I thought proofs should always first turn `configs` into `step^[n]` and then use the `Function.iterate_*` lemmas, but I think it's better to not unfold `configs` and add these lemmas here instead.

SamuelSchlesinger (approved):
This looks good to me. If you want to improve before merging, add docs to the lemmata that currently don't have doctext. Otherwise its fine to do as a followup.

SamuelSchlesinger:
@fmontesi please merge when you get the chance.

PR #770, opened 2026-08-03
feat(Algorithms): comparison sort lower bound via decision-tree induction

SamuelSchlesinger:
An attempt to prove the comparison sort lower bound of #685 taking, in my view, a more natural proof approach for this model. The original proof analyzes traces of the computation, while this approach takes direct advantage of the inductive structure of the free monad and essentially does induction on the program. At a high level, #685 formalizes the encoding argument whereas we formalize the decision tree argument. The arithmetic estimate of log_2 n! is adapted from #685.

Claude Fable 5 authored this code under my direction and review.

### DIFF
--- a/Cslib/Algorithms/LowerBounds/ComparisonSort.lean
+++ b/Cslib/Algorithms/LowerBounds/ComparisonSort.lean
@@ -7,13 +7,12 @@ Authors: Samuel Schlesinger
 module
 
 public import Cslib.Foundations.Control.Monad.Free
+public import Cslib.Algorithms.Lean.TimeM
 public import Mathlib.Data.Finset.Basic
 public import Mathlib.Data.Finset.Lattice.Fold
-public import Mathlib.Data.Fintype.BigOperators
 public import Mathlib.Data.Fintype.Perm
 public import Mathlib.Data.Nat.Factorial.Basic
 public import Mathlib.Data.Nat.Log
-public import Mathlib.Algebra.Order.Group.Nat
 
 /-!
 # The comparison sort lower bound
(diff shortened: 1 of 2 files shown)


Cslib/Algorithms/LowerBounds/ComparisonSort.lean
```diff
@@ -0,0 +1,232 @@
...
+variable {α : Type u} {β γ : Type}
+
+open Cslib.Algorithms.Lean in
+/-- Interpret the comparison query into `TimeM ℕ`: answer with `le x y`, at cost `1`. -/
+def sortHandler (le : α → α → Bool) : {ι : Type} → SortOps α ι → TimeM ℕ ι
+  | _, .cmpLE x y => ⟨le x y, 1⟩
+
+/-- The result of running a comparison program against a comparator. -/
+def run (P : FreeM (SortOps α) β) (le : α → α → Bool) : β :=
+  (P.liftM (sortHandler le)).ret
```
eric-wieser, line 72: I don't think `Cslib.Algorithms.run` should be about `SortOps`; this needs a better name.

eric-wieser: Alternatively, just drop the `run` altogether and inline `(P.liftM (sortHandler le)).ret` below (note that `ret` has notation)

Shreyas4991:
Large parts of this resemble #685 verbatim. It would be nice to have a diff. 
Cc: @eric-wieser

Cslib/Algorithms/LowerBounds/ComparisonSort.lean
```diff
@@ -0,0 +1,232 @@
...
+
+open scoped Nat
+
+/--
+A query type for comparison-based sorting, with a single query that compares two elements.
+-/
+inductive SortOps.{u} (α : Type u) : Type → Type _ where
+  /-- `cmpLE x y` is intended to return `true` if `x ≤ y` and `false` otherwise.
+  The specific order relation depends on the comparator the program is run against. -/
+  | cmpLE (x : α) (y : α) : SortOps α Bool
```
Shreyas4991, line 61: Lines 126-129 in #685 : https://github.com/leanprover/cslib/pull/685/changes#diff-6f48f10c58d3978e2af24c2a7078a55a823e61eea6d29faa6635269db594d5d7R126-R129

Cslib/Algorithms/LowerBounds/ComparisonSort.lean
```diff
@@ -0,0 +1,232 @@
...
+  simp [run, FreeM.liftM_bind]
+
+@[simp, grind =]
+lemma cost_pure (b : β) (le : α → α → Bool) :
+    cost (pure b : FreeM (SortOps α) β) le = 0 := rfl
+
+@[simp, grind =]
+lemma cost_lift (x y : α) (le : α → α → Bool) :
+    cost (FreeM.lift (SortOps.cmpLE x y)) le = 1 := rfl
+
```
Shreyas4991, line 102: Standard FreeM/Prog API also found in cslib#685

SamuelSchlesinger: Yes, this is the point of this PR: to change parts of the proof, not the statement nor all of the underlying proof machinery. Happy to make this on top of your work, which I have done here: https://github.com/Shreyas4991/cslib/pull/3.

Cslib/Algorithms/LowerBounds/ComparisonSort.lean
SamuelSchlesinger: Addressed in the PR on top of 685 in Shreyas' repo.

PR #771, opened 2026-08-03
feat(CCS): notation for CCS

fmontesi:
Adds notation for CCS processes and updates the vending machine example to use it.


Cslib/Languages/CCS/Basic.lean
```diff
@@ -44,14 +44,60 @@ deriving DecidableEq
...
+scoped syntax "ν" term ccsProc : ccsProc
+
+@[inherit_doc Process.const]
+scoped syntax "call" term : ccsProc
+
+@[inherit_doc Process]
+scoped syntax "(" ccsProc ")" : ccsProc
+
+@[inherit_doc Process]
+scoped syntax "`(CCS| " ccsProc ")" : term
```
XYUnknown, line 89: How should I interpret this syntax? Is it annotating that the process is a ccs process?

Cslib/Languages/CCS/Basic.lean
```diff
@@ -44,14 +44,60 @@ deriving DecidableEq
...
+scoped syntax ccsProc "+" ccsProc : ccsProc
+
+@[inherit_doc Process.par]
+scoped syntax ccsProc "|" ccsProc : ccsProc
+
+@[inherit_doc Process.res]
+scoped syntax "ν" term ccsProc : ccsProc
+
+@[inherit_doc Process.const]
+scoped syntax "call" term : ccsProc
```
XYUnknown: using `call` for `const` does not seem to be very intuitive....

XYUnknown (commented):
Good in general, though some choices of naming look a bit odd to me

Cslib/Languages/CCS/Basic.lean
fmontesi: Yes, it's a sort of standard naming convention in Lean for quotation. It's used to scope that we 'enter CCS-notation land'.

Cslib/Languages/CCS/Basic.lean
fmontesi: I know, but when I tried const I had lots of conflicts for some reason. :-|
Happy to revert it if I see a way to make it work.

fmontesi: I did find a way, so now it's 'const' as it should.

fmontesi:
@XYUnknown is it good now?

XYUnknown (commented):
looks good to me ;)

XYUnknown (approved):
should be able to me merged now.

PR #773, opened 2026-08-03
chore: bump toolchain to v4.33.0-rc2

Garmelon:


chenson2018:
I disabled automerge just so that I can look at the git history first, I'd like to understand where it fits in with the new nightly testing.

Garmelon:
This is not an rc1 and so it should not be affected by the new or old nightly-testing.

chenson2018:
> This is not an rc1 and so it should not be affected by the new or old nightly-testing.

I'm not understanding the relevance of this not being an rc1? Between any two toolchains nightly toolchains both core and library (Mathlib, batteries, etc.) breakage can be observed. I'm seeing inconsistent reporting of this between the CSLib nightly testing branch and the new downstream reports.

chenson2018 (approved):

PR #782, opened 2026-08-07
feat(CFG): Verified implementation of Kildall's algorithm

quartztz:
As discussed on [Zulip](https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/Dataflow.20analysis.20on.20CFGs/near/613595436), here is an implementation of Kildall's worklist algorithm for solving dataflow equations. The design features a small CFG api, as well as a characterization of correctness in terms of different flavors of fixpoints. Three main theorems show that

- The result of the algorithm is a postfixpoint (valid solution)
- If the transfer functions are monotone, the postfixpoint that is computed is the least postfixpoint
- If the transfer functions are monotone, the result is a fixpoint

Definitions of the structures are in `Analysis/Dataflow/CFG.lean`, and the algorithm and proofs are in `Analysis/Dataflow/Kildall.lean`. I didn't feel like it fit semantically in any other directory, but I'll be glad to move it if there's better.

TODO: fix all lints :(

### DIFF
--- a/Cslib/Analysis/Dataflow/CFG.lean
+++ b/Cslib/Analysis/Dataflow/CFG.lean
@@ -15,11 +15,12 @@ import Mathlib.Data.DFinsupp.WellFounded
 ## Main definitions
 
 - `CFG` is a structure representing Control Flow Graphs on which the dataflow
-  algorithm defined in Kildall.lean runs.
+  algorithm defined in `Kildall.lean` runs.
 -/
 
 variable {Node Edge : Type} [DecidableEq Node] [DecidableEq Edge]
 
+/-- Abstract structure defining the necessary operations on a CFG to define a Control Flow Graph. -/
 class CFG (Node Edge : Type) [DecidableEq Node] [DecidableEq Edge] where
   /-- All of the nodes in the CFG. -/
   nodes : List Node
--- a/Cslib/Analysis/Dataflow/CFG.lean
+++ b/Cslib/Analysis/Dataflow/CFG.lean
@@ -30,11 +30,11 @@ class CFG (Node Edge : Type) [DecidableEq Node] [DecidableEq Edge] where
   /-- A proof that the entry node is part of the graph's nodes. -/
   entry_mem : entry ∈ nodes
   /-- Extractor function for an edge's source node. -/
-  _srcOf : Edge -> Node
+  _srcOf : Edge → Node
   /-- Proof of correctness for the source extractor. -/
   srcOf_mem : ∀ e ∈ edges, _srcOf e ∈ nodes
   /-- Extractor function for an edge's destination node. -/
-  _dstOf : Edge -> Node
+  _dstOf : Edge → Node
   /-- Proof of correctness for the destination extractor. -/
   dstOf_mem : ∀ e ∈ edges, _dstOf e ∈ nodes
 
--- a/Cslib/Analysis/Dataflow/CFG.lean
+++ b/Cslib/Analysis/Dataflow/CFG.lean
@@ -30,12 +30,16 @@ public import Mathlib.Combinatorics.Quiver.Covering
 structure CFG where
   /-- All of the nodes in the CFG. -/
   Node : Type u
+  /-- A CFG contains a finite amount of nodes. -/
   [fintypeNode : Fintype Node]
+  /-- An ordering of nodes, to make the conversion to lists computable. -/
   [orderNode : LinearOrder Node]
+  /-- Decidable equality on nodes. -/
   [dEqNode : DecidableEq Node]
   /-- Quiver structure for the edges of the CFG. -/
   quiver : Quiver Node
-  [fintypeHom : ∀ a b, Fintype (@Quiver.Hom Node quiver a b)]
+  /-- A CFG contains a finite amount of edges. -/
+  [fintypeEdges : ∀ a b, Fintype (@Quiver.Hom Node quiver a b)]
   /-- Distinguished entry node in the CFG. -/
   entry : Node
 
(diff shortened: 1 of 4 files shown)


Cslib/Analysis/Dataflow/Kildall.lean
```diff
@@ -0,0 +1,333 @@
...
+- Minimality of the algorithm : The algorithm computes the least solution if the transfer functions
+  are monotone.
+- Correctness of the algorithm : The algorithm computes a fixpoint if the transfer functions are
+  monotone.
+
+## References
+
+* [G. Kildall, *A Unified Approach to Global Program Optimization*][Kildall73]
+* [F. Nielson, H.R. Nielson, C. Hankin, *Principles of Program Analysis*][Nielson99]
+* [R. LaSpina, *Formal Verification of WTO-based Dataflow Solvers*][LaSpina25]
```
ctchou, line 40: Please add these references to the file `references.bib`.

ctchou (commented):
@fmontesi  Do you have any opinions about the placement of these files?

Cslib/Analysis/Dataflow/CFG.lean
```diff
@@ -0,0 +1,72 @@
...
+  /-- A proof that the entry node is part of the graph's nodes. -/
+  entry_mem : entry ∈ nodes
+  /-- Extractor function for an edge's source node. -/
+  _srcOf : Edge → Node
+  /-- Proof of correctness for the source extractor. -/
+  srcOf_mem : ∀ e ∈ edges, _srcOf e ∈ nodes
+  /-- Extractor function for an edge's destination node. -/
+  _dstOf : Edge → Node
+  /-- Proof of correctness for the destination extractor. -/
+  dstOf_mem : ∀ e ∈ edges, _dstOf e ∈ nodes
```
ctchou, line 44: I have some questions about the design of `CFG`:
* Normally the nodes and edges of a graph are taken to be sets (which can be modeled by `Set` or `Finset`).  Why do you define them to be lists and then build types from them (`NodeOf` and `EdgeOf`)?  Scanning the code in `Kildall.lean`, I'm not sure you ever used the fact that `nodes` and `edges` are lists (rather than sets).
* For that matter, do you really need the generality that `nodes` and `edges` are subsets of `Node` and `Edge`.  Can they simply be the whole types?  Then you won't need constraints like `setOf_mem` and `dstOf_mem`.
* There is something called `Quiver` in mathlib:
https://leanprover-community.github.io/mathlib4_docs/Mathlib/Combinatorics/Quiver/Basic.html
Would it usable for your purpose?
* Why did you put `_` at the beginning of `_srcOf` and `dstOf`?

quartztz: I agree with the fact that the current presentation is very roundabout. I did not know about `Quiver`, but I feel like applying it would fix these comments: using a single type for `Node`s and a quiver to characterize the edges would both remove unneeded fields and make the projections easier.

The `_` was a crude way to mark that field as "internal", since the preferred API (that returns a `NodeOf g`) is defined in terms of it. However, this problem disappears when applying the other changes, so i'll draft a new structure to hopefully fix everything in one go.

Thank you for your comments and for the pointer!

quartztz:
Second try, integrating the previous comments regarding the type of `Node`, and using `Quiver` for edges.

While the definition of the CFG using `Finset`s is definitely simpler, I really wanted to keep the algorithm computable, and one way I found is to add a linear order on the nodes of the CFG. This "workaround" is relatively inobtrusive (any instantiation can supply a unique `id : Nat` field for `Node`s and give an "arbitrary" order based on that), as well as potentially useful for the definition of alternate scheduling strategies on the algorithm, for optimization (not as of right now, it would require some adaptation to the structure and the algorithm I haven't given enough deep thought to yet).

My question is therefore the following: what's the project's stance on computability?

I'm not a huge fan of this workaround, but I cannot see a way to make computability work without it. I having a noncomputable algorithm is a shame, but that means that either the `CFG` structure carries `List`s (and is ugly) or we do this sort of workaround (which is also ugly, on the same level as lists in my opinion).

If the library doesn't care about the algorithm being computable, then the `Finset` formulation can be used without issues, and I'll be glad (if a bit heartbroken) to rewrite everything in a noncomputable manner.

ctchou (changes requested):
Your patch produced a very large diff on `references.bib` which obscures the change you actually made.  Could you revert it and make your small change manually?  We are not against reformatting `references.bib`, but that should be done in a separate PR and not be mixed with this PR.

quartztz:
That's my bad, the autoformatter kicked in. Pushed a commit to fix! Sorry about that.

ctchou (changes requested):
Sorry for keeping iterating on this.  But we need to get the fundamental data structure right.

Cslib/Analysis/Dataflow/CFG.lean
```diff
@@ -0,0 +1,95 @@
...
+  /-- An ordering of nodes, to make the conversion to lists computable. -/
+  [orderNode : LinearOrder Node]
+  /-- Decidable equality on nodes. -/
+  [dEqNode : DecidableEq Node]
+  /-- Quiver structure for the edges of the CFG. -/
+  quiver : Quiver Node
+  /-- A CFG contains a finite amount of edges. -/
+  [fintypeEdges : ∀ a b, Fintype (@Quiver.Hom Node quiver a b)]
+  /-- Distinguished entry node in the CFG. -/
+  entry : Node
```
ctchou, line 44: First, in cslib (and mathlib), we prefer unbundled definitions.  That is, we prefer that types like `Node` be given as a parameter to `CFG`, rather than being bundled in the structure.  Similarly, assumptions like `FinType` should be made only when needed and given as parameters.  You can assume `[DecidableEq Node]` globally.

Second, do you need to allow for the possibility of having more than one edge between two nodes?  If not, then `Digraph` would be simpler and more appropriate than `Quiver`:
https://leanprover-community.github.io/mathlib4_docs/Mathlib/Combinatorics/Digraph/Basic.html#Digraph
Also, by using `Quiver`, you have a separate type of edges for each pair of nodes.  Do you have a need for a single type `Edge` of edges?  If so, you may want to consider using cslib's `LTS`:
https://api.cslib.io/docs/Cslib/Foundations/Semantics/LTS/Basic.html

Thrid, I do not understand your comment about needing a `LinearOrder` to make `CFG` computable.  Do you actually use this order in Kildall's algorithm?  If so, you may consider substituting `Fin n` for `Node` or assuming the existence of an `Equiv`:
https://leanprover-community.github.io/mathlib4_docs/Mathlib/Logic/Equiv/Defs.html#Equiv
between `Node` and some totally ordered finite type such as `Fin n`.  And you can make that assumption only when needed.

ctchou (changes requested):

PR #788, opened 2026-08-10
feat(ModalLogic): IsAxiom and axiom L (Löb's theorem for modal logic)

fmontesi:
This PR proves the validity of Axiom L (Löb's theorem for modal logic) under the expected conditions (transitivity and converse well-foundedness). I've also refactored axiom statements to use an abbreviation (IsAxiom), since I started finding the previous way of doing it with manual quantifications cumbersome and error-prone.

PR #799, opened 2026-08-14
feat(ModalLogic+Congruence): modal reasoning for Lean

fmontesi:
This PR develops a concept of modal reasoning for Lean theorems, powered by the laws of modal logic. This development led to some polishing for `Congruence` and some definitions for `Relation`.

Specifically, the PR:
- Introduces the first development where CSLib's modal logic is instantiated with Lean's `Prop` as the type of atomic propositions. We use this to construct models out of Lean predicates and containers and show some general properties in terms of logical equivalence.
- Introduces a construction that, given any scalar multiplication operator, builds an accessibility relation. If the operator is an action of some kind (as defined in mathlib), this accessibility relation has interesting properties (like transitivity or diamond) that are relevant for modal reasoning.
- Develops an example of how this framework can be used to prove mathematical theorems. In particular, we apply the new construction to obtain an accessibility relation on `Ideal`s. In this interpretation, membership of radicals of ideals is characterised by the diamond connective in modal logic. This allows for applying modal logic to proving properties about radicals.
- As examples, we show that `Ideal.radical_inf` and `Ideal.radical_idem` from mathlib can be proven by applying modal logical equivalence.

This put the ergonomics and expressive power of our infrastructure for modal logic, logical equivalence, relations, and congruence to the test. This PR also:
- Polishes a bit the ergonomics of LawfulCongruence (e.g., by making `Congruence.r` an `abbrev` so that typeclass resolution works in proofs).
- Adds a `Preserves` property for relations.
- Defines and proves soundness and completeness of modal axioms for relations with the diamond property.
- Refactors a bit LogicalEquivalence for modal logic so that it is more fundamental (based on a single model instead of a set of models, and then derives the latter on the first).

The last point not only helps with proofs, but also aligns logical equivalence with how it is handled in HML. This means that we can start working on a shared abstraction for this kind of logics, which would then derive the natural notion of logical equivalence for free.

Depends on #788


Cslib/Logics/Modal/Lean/Basic.lean
```diff
@@ -0,0 +1,123 @@
...
+theorem Proposition.equivOfPredicates_denotation_eq {r : α → α → Prop}
+    {φ₁ φ₂ : Proposition (α → Prop)} :
+    (φ₁ ≡[Equiv.OfPredicates r] φ₂) ↔
+      φ₁.denotation (ofPredicates r) = φ₂.denotation (ofPredicates r) := by
+  apply Iff.intro <;> intro h
+  case mp =>
+    apply denotation_eq_of_equiv h
+  case mpr =>
+    apply equiv_iff_denotation_eq.2
+    grind
```
thomaskwaring:
```suggestion
      φ₁.denotation (ofPredicates r) = φ₂.denotation (ofPredicates r) :=
  equiv_iff_denotation_eq  
```

Cslib/Logics/Modal/Lean/Basic.lean
```diff
@@ -0,0 +1,123 @@
...
+theorem Proposition.equivOfContainers_denotation_eq {α} [Membership α β] {r : α → α → Prop}
+    {φ₁ φ₂ : Proposition β} :
+    (φ₁ ≡[Equiv.OfContainers (β := β) r] φ₂) ↔
+      φ₁.denotation (ofContainers r) = φ₂.denotation (ofContainers r) := by
+  apply Iff.intro <;> intro h
+  case mp =>
+    apply denotation_eq_of_equiv h
+  case mpr =>
+    apply equiv_iff_denotation_eq.2
+    grind
```
thomaskwaring:
```suggestion
      φ₁.denotation (ofContainers r) = φ₂.denotation (ofContainers r) := 
  equiv_iff_denotation_eq
```

Cslib/Logics/Modal/Basic.lean
```diff
@@ -232,70 +256,117 @@ theorem Satisfies.t_box_diamond [Std.Refl m.r] : ⇓Modal[m,w ⊨ □φ → φ]
...
+    intro w'
+    refine (WellFounded.induction (C := fun w' => m.r w w' → ⇓Modal[m,w' ⊨ φ]) hwf w' ?_)
+    intro w' ih hww'
+    have hImp : ⇓Modal[m, w' ⊨ □φ → φ] := h _ hww'
+    rw [Satisfies.imp_iff_imp, Satisfies.box_iff_forall] at hImp
+    apply hImp
+    intro w'' hw'w''
+    apply ih _ hw'w''
+    exact IsTrans.trans _ _ _ hww' hw'w''
+  exact h' w' hw'
```
thomaskwaring:
```suggestion
  simp_rw [Satisfies.imp_iff_imp, Satisfies.box_iff_forall]
  intro h
  refine (hwf.induction (C := fun w' => m.r w w' → ⇓Modal[m,w' ⊨ φ]) · ?_)
  intro w' ih hww'
  have hImp : ⇓Modal[m, w' ⊨ □φ → φ] := h _ hww'
  rw [Satisfies.imp_iff_imp, Satisfies.box_iff_forall] at hImp
  apply hImp
  intro w'' hw'w''
  apply ih _ hw'w''
  exact IsTrans.trans _ _ _ hww' hw'w''
```

Cslib/Logics/Modal/Basic.lean
```diff
@@ -232,70 +256,117 @@ theorem Satisfies.t_box_diamond [Std.Refl m.r] : ⇓Modal[m,w ⊨ □φ → φ]
...
+    (φ : Proposition Atom) : IsAxiom r <| ◇□φ → □◇φ := by
+  intro v w
+  rw [Satisfies.imp_iff_imp]
+  intro h₁
+  rw [Satisfies.diamond_iff_exists] at h₁
+  rcases h₁ with ⟨_, hww₁, _⟩
+  rw [Satisfies.box_iff_forall]
+  intro _ hww₂
+  obtain ⟨w₃, hww₃⟩ := h hww₁ hww₂
+  grind
```
thomaskwaring:
```suggestion
  simp_rw [IsAxiom, Satisfies.imp_iff_imp, Satisfies.diamond_iff_exists, Satisfies.box_iff_forall]
  rintro v w ⟨_, hww₁, _⟩ _ hww₂
  obtain ⟨w₃, hww₃⟩ := h hww₁ hww₂
  grind
```

Cslib/Logics/Modal/Basic.lean
```diff
@@ -232,70 +256,117 @@ theorem Satisfies.t_box_diamond [Std.Refl m.r] : ⇓Modal[m,w ⊨ □φ → φ]
...
+
+open Relation in
+/-- Any model that admits axiom .2 has the diamond property. -/
+theorem Satisfies.pointTwo_diamond (r : World → World → Prop) [Nonempty Atom]
+    (h : ∀ φ : Proposition Atom, IsAxiom r <| ◇□φ → □◇φ) : Diamond r := by
+  intro w w₁ w₂ hww₁ hww₂
+  have a := Classical.arbitrary Atom
+  let v := fun (w' : World) (a : Atom) => r w₁ w'
+  have h' := h a v w
+  grind [Join]
```
thomaskwaring:
```suggestion
  specialize h (Classical.arbitrary Atom) (fun w' _ => r w₁ w') w
  grind [Join]
```

Cslib/Logics/Modal/Basic.lean
```diff
@@ -307,4 +378,23 @@ def Proposition.valid (S : Set (Model World Atom)) (φ : Proposition Atom) : Pro
...
+
+/-- The class of all models generated by a frame (relation). -/
+abbrev modelsOfRelation (r : World → World → Prop) : Set (Model World Atom) :=
+  {m | m.r = r}
+
+/-- A proposition is an axiom of a frame exactly when it belongs to the logic of all models over
+that frame. -/
+theorem isAxiom_iff_mem_logic_modelsOfRelation (r : World → World → Prop) (φ : Proposition Atom) :
+    IsAxiom r φ ↔ φ ∈ logic (modelsOfRelation r) := by
+  apply Iff.intro
```
thomaskwaring:
```suggestion
  constructor
```

Cslib/Logics/Modal/Denotation.lean
```diff
@@ -48,4 +49,21 @@ theorem theoryEq_denotation_eq {m : Model World Atom} {w₁ w₂ : World} :
     (∀ (φ : Proposition Atom), w₁ ∈ (φ.denotation m) ↔ w₂ ∈ (φ.denotation m)) := by
   apply Iff.intro <;> grind [_=_ satisfies_mem_denotation]
 
+/-- Logically equivalent propositions under a model have the same denotation. -/
+theorem Proposition.equiv_iff_denotation_eq :
+    (φ₁ ≡[Equiv m] φ₂) ↔ φ₁.denotation m = φ₂.denotation m := by
+  apply Iff.intro <;> intro h
```
thomaskwaring:
```suggestion
  constructor <;> intro h
```

thomaskwaring (changes requested):
preliminary review here, but overall i like this

one question (which i was thinking about as relates to the earlier axiom L PR) is whether we can state `IsAxiom` as an inference system, where the judgement is `r ⊨ φ` for "this the proposition is valid at every world for every valuation" — logically it is the same but it avoids introducing a new definition and might down-the-track mesh better with our infrastructure. what do you think?

Cslib/Logics/Modal/Lean/SMul.lean
```diff
@@ -0,0 +1,101 @@
...
+    simp
+
+/-- The relation induced by a semigroup action is transitive. -/
+instance [Semigroup M] [SemigroupAction M α] : IsTrans α (ofSMul M α) where
+  trans x y z hxy hyz := by
+    rcases hxy with ⟨m, rfl⟩
+    rcases hyz with ⟨n, rfl⟩
+    use n * m
+    rw [mul_smul]
+
```
thomaskwaring, line 53: for completeness, the following two properties (which are referenced in `Modal.Basic`) also have natural analogues in this setting.
```suggestion
/-- The relation induced by a group action is symmetric. -/
instance [Group G] [MulAction G α] : Std.Symm (ofSMul G α) where
  symm := by
    rintro x _ ⟨g, rfl⟩
    use g⁻¹
    simp

/-- The relation induced by an action of a nonempty type is serial. -/
instance [SMul M α] [Nonempty M] : Serial (ofSMul M α) where
  serial x := ⟨Classical.arbitrary M • x, Classical.arbitrary M, rfl⟩

```

Cslib/Logics/Modal/Lean/SMul.lean
```diff
@@ -0,0 +1,101 @@
...
+-/
+@[scoped grind .]
+theorem ofSMul_diamond [CommSemigroup M] [SemigroupAction M α] : Diamond (ofSMul M α) := by
+  intro x y z hxy hxz
+  rcases hxy with ⟨m, rfl⟩
+  rcases hxz with ⟨n, rfl⟩
+  refine ⟨n • (m • x), ?_, ?_⟩
+  · use n
+  · use m
+    grind only [=_ mul_smul]
```
thomaskwaring: It seems reasonable to me to use the existing typeclass for "commuting actions" — weird that the relevant instance doesn't exist in mathlib.
```suggestion
/-- TODO: upstream this generalisation of Mathlib's `smulCommClass_self`, which applies only to
monoids `M`. -/
instance {M α} [CommSemigroup M] [SemigroupAction M α] : SMulCommClass M M α where
  smul_comm m n a := by rw [←mul_smul, mul_comm, mul_smul]

/-- The relation induced by an action commuting with itself has the diamond property. -/
@[scoped grind .]
theorem ofSMul_diamond [SMul M α] [SMulCommClass M M α] : Diamond (ofSMul M α) := by
  rintro x _ _ ⟨m, rfl⟩ ⟨n, rfl⟩
  use! n • m • x, n, m, smul_comm ..
```

Cslib/Logics/Modal/Lean/SMul.lean
```diff
@@ -0,0 +1,101 @@
...
+
+/-- Preservation by an action-induced relation is exactly closure under the action. -/
+@[scoped grind =]
+theorem ofSMul_preserves_iff [SMul M α] {P : α → Prop} :
+    Preserves (ofSMul M α) P ↔ ∀ m : M, ∀ x, P x → P (m • x) := by
+  apply Iff.intro
```
thomaskwaring:
```suggestion
  constructor
```

Cslib/Logics/Modal/Lean/SMul.lean
```diff
@@ -0,0 +1,101 @@
...
+instance [Monoid M] [MulAction M α] : Std.Refl (ofSMul M α) where
+  refl x := by
+    use 1
+    simp
+
+/-- The relation induced by a semigroup action is transitive. -/
+instance [Semigroup M] [SemigroupAction M α] : IsTrans α (ofSMul M α) where
+  trans x y z hxy hyz := by
+    rcases hxy with ⟨m, rfl⟩
+    rcases hyz with ⟨n, rfl⟩
```
thomaskwaring:
```suggestion
  trans := by
    rintro x _ _ ⟨m, rfl⟩ ⟨n, rfl⟩
```

Cslib/Logics/Modal/Lean/SMul.lean
```diff
@@ -0,0 +1,101 @@
...
+def ofSMul (M α : Type*) [SMul M α] (x y : α) : Prop := ∃ m : M, m • x = y
+
+/-- The relation induced by a monoid action is reflexive. -/
+instance [Monoid M] [MulAction M α] : Std.Refl (ofSMul M α) where
+  refl x := by
+    use 1
+    simp
+
+/-- The relation induced by a semigroup action is transitive. -/
+instance [Semigroup M] [SemigroupAction M α] : IsTrans α (ofSMul M α) where
```
thomaskwaring, line 37: This doesn't use the associativity of `M`, but (strangely? maybe for good reason?) `SemigroupAction` / `MulAction` require more than `Mul M`, so best to leave it as-is I suppose.

Cslib/Logics/Modal/Lean/SMul.lean
```diff
@@ -0,0 +1,101 @@
...
+@[scoped grind =]
+theorem ofSMul_preserves_iff [SMul M α] {P : α → Prop} :
+    Preserves (ofSMul M α) P ↔ ∀ m : M, ∀ x, P x → P (m • x) := by
+  apply Iff.intro
+  case mp =>
+    intro h m x hx
+    exact h ⟨m, rfl⟩ hx
+  case mpr =>
+    intro h x y hxy hx
+    rcases hxy with ⟨m, rfl⟩
```
thomaskwaring:
```suggestion
    rintro h x y ⟨m, rfl⟩ hx
```

Cslib/Logics/Modal/Lean/Basic.lean
```diff
@@ -0,0 +1,113 @@
...
+theorem Satisfies.ofPredicates_atom_iff {P : α → Prop} (r : α → α → Prop) :
+    ⇓Modal[ofPredicates r, a ⊨ P] ↔ P a := by grind [=_ derivation_def]
+
+/-- Under `Model.ofPredicates r`, `P → □P` is an axiom iff `r` preserves `P`. -/
+@[scoped grind ⇒]
+theorem Satisfies.ofPredicates_preserves_iff {P : α → Prop} (r : α → α → Prop) :
+    (∀ a, ⇓Modal[ofPredicates r, a ⊨ P → □P]) ↔ Preserves r P := by
+  apply Iff.intro <;> intro h
+  case mp =>
+    intro a₁ a₂ hr hPa₁
```
thomaskwaring:
```suggestion
  constructor
  case mp =>
    intro h a₁ a₂ hr hPa₁
```

Cslib/Logics/Modal/Lean/Basic.lean
```diff
@@ -0,0 +1,113 @@
...
+    (fun a => P a ∧ Q a) ≡[Equiv.OfPredicates r] (P ∧ Q) := by
+  simp only [← Proposition.equiv_def]
+  grind [Proposition.Equiv]
+
+/-- Under `Equiv.OfContainers r`, if membership in the infimum of two containers is equivalent to
+membership in both containers, then atomic infimum is logically equivalent to modal conjunction. -/
+theorem Proposition.ofContainers_inf_equiv [Membership α β] [Min β] (r : α → α → Prop) (p q : β)
+    (h : ∀ x, x ∈ p ⊓ q ↔ x ∈ p ∧ x ∈ q) :
+    (↑(p ⊓ q) : Proposition β) ≡[Equiv.OfContainers r] (p ∧ q) := by
+  intro
```
thomaskwaring:
```suggestion
```

CslibTests/Modal/Ideal.lean
```diff
@@ -0,0 +1,116 @@
...
+  mul_smul m n x := by
+    dsimp [HSMul.hSMul]
+    grind [pow_mul, mul_comm]
+
+local instance [Monoid α] : MulAction ℕ+ α := posPowAction
+
+/-- Ideals are invariant under positive exponentiation. -/
+theorem Ideal.posPowSMulMemClass [Semiring R] : SMulMemClass (Ideal R) ℕ+ R where
+  smul_mem n x hx := by
+    change _ ^ (n : ℕ) ∈ _
```
thomaskwaring:
```suggestion
```

CslibTests/Modal/Ideal.lean
```diff
@@ -0,0 +1,116 @@
...
+-/
+
+/-- Positive naturals act on a monoid by exponentiation. -/
+@[instance_reducible]
+def posPowAction [Monoid α] : MulAction ℕ+ α where
+  smul n x := x ^ (n : ℕ)
+  one_smul x := pow_one x
+  mul_smul m n x := by
+    dsimp [HSMul.hSMul]
+    grind [pow_mul, mul_comm]
```
thomaskwaring:
```suggestion
    simp [HSMul.hSMul, ←pow_mul, mul_comm]
```

CslibTests/Modal/Ideal.lean
```diff
@@ -0,0 +1,116 @@
...
+    _ ≡[IdealEquiv] (◇I ∧ ◇J) := Proposition.ofSMul_diamond_and_equiv I J
+
+/-- Radicals of ideals distribute over intersection, as a consequence that `◇(I ⊓ J)` is logically
+equivalent to `◇I ∧ ◇J`. -/
+theorem Ideal.radical_inf [CommSemiring R] (I J : Ideal R) :
+    (I ⊓ J).radical = I.radical ⊓ J.radical := by
+  apply SetLike.ext'
+  rw [Submodule.coe_inf]
+  simp only [Ideal.radical_eq_modal_denotation]
+  rw [Proposition.denotation_eq_of_equiv (Ideal.inf_modelEquiv I J)]
```
thomaskwaring:
```suggestion
  simp_rw [Submodule.coe_inf, Ideal.radical_eq_modal_denotation,
    Proposition.denotation_eq_of_equiv (Ideal.inf_modelEquiv I J)]
```

Cslib/Logics/Modal/LogicalEquivalence.lean
```diff
@@ -122,22 +195,73 @@ structure Satisfies.Context (World Atom : Type*) where
...
+/-- Correspondence of equivalence and axiom validity. -/
+theorem Proposition.axiom_iff_forall_equiv (r : α → α → Prop) (φ₁ φ₂ : Proposition Atom) :
+    (Axiom r⇓(φ₁ ↔ φ₂)) ↔ ∀ v, φ₁ ≡[Equiv ⟨r, v⟩] φ₂ := by
+  apply Iff.intro <;> intro h
+  case mp =>
+    intro v w
+    exact h v w
+  case mpr =>
+    intro v w
+    exact h v w
```
thomaskwaring:
```suggestion
    (Axiom r⇓(φ₁ ↔ φ₂)) ↔ ∀ v, φ₁ ≡[Equiv ⟨r, v⟩] φ₂ := Iff.rfl
```

Cslib/Logics/Modal/LogicalEquivalence.lean
```diff
@@ -122,22 +195,73 @@ structure Satisfies.Context (World Atom : Type*) where
...
+  rw [Satisfies.iff_iff_iff]
+  apply Iff.intro
+  case mp => grind
+  case mpr =>
+    intro h
+    rw [Satisfies.and_iff_and] at h
+    rcases h with ⟨hφ₁, hφ₂⟩
+    rw [Satisfies.diamond_iff_exists] at hφ₁ hφ₂ ⊢
+    rcases hφ₁ with ⟨b, hab, hb⟩
+    rcases hφ₂ with ⟨c, hac, hc⟩
```
thomaskwaring:
```suggestion
    rw [Satisfies.and_iff_and, Satisfies.diamond_iff_exists]
    rintro ⟨⟨b, hab, hb⟩, ⟨c, hac, hc⟩⟩
```

Cslib/Logics/Modal/LogicalEquivalence.lean
```diff
@@ -122,22 +195,73 @@ structure Satisfies.Context (World Atom : Type*) where
...
+open Relation in
+/-- In a transitive diamond model, possibility distributes over conjunction for propositions
+whose satisfaction is preserved along accessibility. -/
+@[scoped grind ⇒]
+theorem Proposition.diamond_and_equiv_of_preserves {m : Model World Atom} [IsTrans World m.r]
+    {φ₁ φ₂ : Proposition Atom} (hd : Diamond m.r) (h₁ : Preserves m.r (⇓Modal[m,· ⊨ φ₁]))
+    (h₂ : Preserves m.r (⇓Modal[m,· ⊨ φ₂])) :
+    ◇(φ₁ ∧ φ₂) ≡[Equiv m] (◇φ₁ ∧ ◇φ₂) := by
+  intro a
+  rw [Satisfies.iff_iff_iff]
```
thomaskwaring: perhaps this pattern (which occurs elsewhere) could be extracted into a lemma, which would be the converse of `Proposition.equiv_iff`

Cslib/Logics/Modal/Basic.lean
```diff
@@ -307,4 +376,23 @@ def Proposition.valid (S : Set (Model World Atom)) (φ : Proposition Atom) : Pro
...
+  {m | m.r = r}
+
+/-- A proposition is an axiom of a frame exactly when it belongs to the logic of all models over
+that frame. -/
+theorem axiom_iff_mem_logic_modelsOfRelation (r : World → World → Prop) (φ : Proposition Atom) :
+    Axiom r⇓φ ↔ φ ∈ logic (modelsOfRelation r) := by
+  constructor
+  case mp =>
+    intro h m hm w
+    subst hm
```
thomaskwaring:
```suggestion
    rintro h m rfl w
```

thomaskwaring (changes requested):
i like the new `Axiom` design i think! this review is mostly some further golfing / simplifying of proofs, plus a few natural extensions to `Modal.Lean.SMul`

Cslib/Logics/Modal/LogicalEquivalence.lean
fmontesi: Right. I've generalised a theorem into equiv_iff_forall_iff.

Cslib/Logics/Modal/Lean/SMul.lean
fmontesi: Yeah I thought it weird too, but it's the smallest API I could find. :shrug:

Cslib/Logics/Modal/LogicalEquivalence.lean
```diff
@@ -89,17 +126,38 @@ lemma Proposition.Context.fill_def {c : HasContext.Context (Proposition Atom)} :
...
 
 /-- Logical equivalence is an equivalence relation. -/
-instance {World Atom} (S : Set (Model World Atom)) :
-    IsEquiv (Proposition Atom) (Proposition.Equiv S) := by
+instance (m : Model World Atom) : IsEquiv (Proposition Atom) (Proposition.Equiv m) := by
+  rw [← equivalence_iff_isEquiv]
+  constructor
+  case refl => grind [Proposition.Equiv]
+  case symm =>
+    intro φ₁ φ₂ h w
```
thomaskwaring:
```suggestion
```

Cslib/Logics/Modal/LogicalEquivalence.lean
```diff
@@ -89,17 +126,38 @@ lemma Proposition.Context.fill_def {c : HasContext.Context (Proposition Atom)} :
...
+    grind
+  case trans =>
+    intro φ₁ φ₂ φ₃ h₁ h₂ w
```
thomaskwaring:
```suggestion
```

Cslib/Logics/Modal/LogicalEquivalence.lean
```diff
@@ -89,17 +126,38 @@ lemma Proposition.Context.fill_def {c : HasContext.Context (Proposition Atom)} :
...
+  constructor
+  case refl =>
+    intro φ m hm w
+    grind [Proposition.Equiv]
+  case symm =>
+    intro φ₁ φ₂ h m hm w
+    grind [h m hm w]
+  case trans =>
+    intro φ₁ φ₂ φ₃ h₁ h₂ m hm w
+    grind [h₁ m hm w, h₂ m hm w]
```
thomaskwaring:
```suggestion
  case refl =>
    grind [Proposition.Equiv]
  case symm =>
    grind
  case trans =>
    grind
```

Cslib/Logics/Modal/LogicalEquivalence.lean
```diff
@@ -24,43 +24,80 @@ namespace Cslib.Logic.Modal
...
   grind [=_ Satisfies.iff_iff_iff]
 
+/-- A class of models, defined as a set. -/
+abbrev ModelClass World Atom := Set (Model World Atom)
+
+/-- Every class of models induces an inference system tag for reasoning within that class. -/
+inductive Within (S : ModelClass World Atom)
+
+instance (S : ModelClass World Atom) : InferenceSystem (Within S) (Judgement World Atom) where
+  derivation j := j.m ∈ S → ⇓j
```
thomaskwaring: this definition is a little strange to me — i would have expected something like
```
instance (S : ModelClass World Atom) : InferenceSystem (Within S) (Proposition Atom) where
  derivation φ := ∀ m ∈ S, ∀ w, ⇓Modal[m,w ⊨ φ]
```
which aligns better (to my eyes) with the definition of the equivalence below. can you say more about the motivation? is there any situation you expect to apply it to different worlds / models outside the class? also i would have expected it to go in `Model.Basic`, probably next to `Axiom`.

Cslib/Logics/Modal/Basic.lean
```diff
@@ -123,7 +126,7 @@ instance : HasInferenceSystem (Judgement World Atom) := ⟨Satisfies.Bundled⟩
 
 open scoped InferenceSystem Proposition
 
-@[scoped grind =_]
+@[scoped grind =]
 theorem derivation_def {m : Model World Atom} {w : World} {φ : Proposition Atom} :
   Satisfies m w φ = ⇓Modal[m,w ⊨ φ] := rfl
```
thomaskwaring, line 131: i think the following lemma is missing, and that's why the annotation on `def Satisfies ...`, and the `grind [derivation_def =_]` below are needed.
```suggestion
@[scoped grind =]
theorem Satisfies.atom_iff {a : Atom} : ⇓Modal[m,w ⊨ a] ↔ m.v w a := by rfl
```

fmontesi: We usually don't set automatic rewritings from the notation to the definition, rather the opposite (get the raw term into the notation), because theorems are typically specified on the notation.

EDIT: I misread!! See next comment.

Cslib/Logics/Modal/LogicalEquivalence.lean
fmontesi: Right, things got a bit misaligned indeed. The purpose of the `Within` inference system was to define satisfaction in a class of models, which I thought we needed, but after the last refactoring I don't think we do any longer (at least not right now). Your definition is instead (I agree) designed to support logical equivalence and be in line with `Axiom`.

I'm inclined to just remove it for now, it's not really needed anywhere and can introduce extra plumbing down the line. `EquivWithin` can still be defined without it, with no modifications -- that we could keep for now.

If we really wanna keep `Within` and define it like you say, perhaps it could be called `AxiomInClass`?

In any case, I'll remove this from this PR and we can discuss this separately, is that ok? I think we're getting a bit off topic wrt the main point of the PR.

Cslib/Logics/Modal/Basic.lean
thomaskwaring: yes, right — but this is rewriting the definition of `Satisfies` in terms of the notation, no? in the same way that `Satisfies.not_iff_not` etc do. and the `grind [derivation_def =_]` indicate that with the current setup you need to rewrite away from the notation anyway, which this avoids.

fmontesi: I'd misread it and didn't realise it's a simplification rewriting only for atoms! Looks pretty reasonable to me. Now in. Indeed it fixes lots of quirks. :)

Cslib/Logics/Modal/Basic.lean
thomaskwaring: i think we can remove this grind annotation too — the lemmas below should cover the unfolding it implies

thomaskwaring: ditto this annotation

Cslib/Logics/Modal/Basic.lean
```diff
@@ -123,10 +126,13 @@ instance : HasInferenceSystem (Judgement World Atom) := ⟨Satisfies.Bundled⟩
...
 
+@[simp, scoped grind =]
+theorem Satisfies.atom_iff {a : Atom} : ⇓Modal[m,w ⊨ a] ↔ m.v w a := by rfl
+
 /-- A world satisfies a proposition iff it does not satisfy the negation of the proposition. -/
 @[scoped grind =]
```
thomaskwaring, line 136: do these lemmas warrant a `simp` annotation as well? i don't have a strong opinion, more for consistency with `Satisfies.atom_iff`

Cslib/Logics/Modal/LogicalEquivalence.lean
```diff
@@ -24,43 +24,47 @@ namespace Cslib.Logic.Modal
...
 /-- Logical equivalence preserves validity. -/
-theorem Proposition.equiv_valid (S : Set (Model World Atom))
-    (φ₁ φ₂ : Proposition Atom) (h : φ₁ ≡[Equiv S] φ₂) :
+theorem Proposition.equivWithin_valid (S : ModelClass World Atom)
+    (φ₁ φ₂ : Proposition Atom) (h : φ₁ ≡[EquivWithin S] φ₂) :
     (φ₁.valid S ↔ φ₂.valid S) := by
-  apply Proposition.equiv_forall_der at h
-  simp only [Satisfies.iff_iff_iff] at h
-  grind
+  apply Iff.intro <;> intro h' m hm w <;> grind [h m hm w]
```
thomaskwaring:
```suggestion
  grind
```

Cslib/Logics/Modal/Lean/Basic.lean
```diff
@@ -0,0 +1,109 @@
...
+
+open Model Relation
+open scoped InferenceSystem Satisfies
+
+/-! ## Models of Lean predicates -/
+
+/-- Under `Model.ofPredicates r`, an atomic proposition `P` holds at `a` iff `P a`. -/
+@[scoped grind =]
+theorem Satisfies.ofPredicates_atom_iff {P : α → Prop} (r : α → α → Prop) :
+    ⇓Modal[ofPredicates r, a ⊨ P] ↔ P a := by grind
```
thomaskwaring:
```suggestion
    ⇓Modal[ofPredicates r, a ⊨ P] ↔ P a := Iff.rfl
```

CslibTests/Modal/Ideal.lean
```diff
@@ -0,0 +1,113 @@
...
+-/
+
+import Cslib.Logics.Modal.Denotation
+import Cslib.Logics.Modal.Lean.Basic
+import Cslib.Logics.Modal.Lean.SMul
+import Cslib.Foundations.Relation.Defs
+import Mathlib.Algebra.Group.Defs
+import Mathlib.Algebra.Ring.Defs
+import Mathlib.RingTheory.Ideal.Defs
+import Mathlib.RingTheory.Ideal.Operations
```
thomaskwaring:
```suggestion
import Cslib.Logics.Modal.Lean.SMul
import Mathlib.RingTheory.Ideal.Operations
```

CslibTests/Modal/Ideal.lean
```diff
@@ -0,0 +1,113 @@
...
+  one_smul x := pow_one x
+  mul_smul m n x := by
+    simp [HSMul.hSMul, ←pow_mul, mul_comm]
+
+local instance [Monoid α] : MulAction ℕ+ α := posPowAction
+
+/-- Ideals are invariant under positive exponentiation. -/
+theorem Ideal.posPowSMulMemClass [Semiring R] : SMulMemClass (Ideal R) ℕ+ R where
+  smul_mem n x hx := by
+    exact Ideal.pow_mem_of_mem _ hx n n.prop
```
thomaskwaring:
```suggestion
  smul_mem n _ hx := Ideal.pow_mem_of_mem _ hx n n.prop
```

thomaskwaring (approved):
thanks for looking at those grind annotations, to my understanding it's working how we'd hope now (not unfolding definitions, etc).

i'm happy with this now, i'd appreciate one more pair of eyes on it to be sure, but i'll approve in the meantime. also, am i right in assuming this supercedes #788 ?

Cslib/Logics/Modal/Basic.lean
thomaskwaring: a few more `#min_imports` to add
```suggestion
public import Cslib.Foundations.Logic.InferenceSystem
public import Cslib.Foundations.Logic.Operators
public import Cslib.Foundations.Relation.Defs
public import Mathlib.Order.BooleanAlgebra.Set
```

Cslib/Logics/Modal/Cube.lean
thomaskwaring, line 9:
```suggestion
public import Cslib.Logics.Modal.Basic
public import Cslib.Foundations.Relation.Euclidean
```

Cslib/Logics/Modal/Denotation.lean
```diff
@@ -7,6 +7,7 @@ Authors: Fabrizio Montesi
 module
 
 public import Cslib.Logics.Modal.Basic
```
thomaskwaring:
```suggestion
```

Cslib/Logics/Modal/Lean/Basic.lean
```diff
@@ -0,0 +1,109 @@
+/-
+Copyright (c) 2026 Fabrizio Montesi. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Fabrizio Montesi
+-/
+
+module
+
+public import Cslib.Logics.Modal.Basic
+public import Cslib.Logics.Modal.LogicalEquivalence
```
thomaskwaring:
```suggestion
```

Cslib/Logics/Modal/Lean/SMul.lean
```diff
@@ -0,0 +1,109 @@
...
+Authors: Fabrizio Montesi
+-/
+
+module
+
+public import Mathlib.Algebra.Group.Defs
+public import Mathlib.GroupTheory.GroupAction.SubMulAction
+public import Cslib.Foundations.Relation.Defs
+public import Cslib.Logics.Modal.LogicalEquivalence
+public import Cslib.Logics.Modal.Lean.Basic
```
thomaskwaring:
```suggestion
public import Mathlib.GroupTheory.GroupAction.SubMulAction
public import Cslib.Logics.Modal.Lean.Basic
```

Cslib/Logics/Modal/Basic.lean
fmontesi: I'm a bit unsure atm, because the ecosystem/ideas behind how to properly rewrite modal propositions is still growing in my head.

For `atom_iff` I feel confident enough instead, as there's basically no way to resolve an atom otherwise most of the time.

If we add them later I don't have strong feelings (yet?) though.

Cslib/Logics/Modal/LogicalEquivalence.lean
fmontesi: Aaaah, it *is* working much better now. :-)

chenson2018 (approved):
I reviewed very lightly. I'm suspicious of how much `abbrev` appears, but otherwise I trust Thomas's judgement.

PR #802, opened 2026-08-14
feat(MultiTapeTM): Nondeterministic multi-tape Turing machines

barni120400:
This adds nondeterministic multi-tape Turing machines next to the deterministic ones from #384. The transition function is replaced by a transition relation ([Papadimitriou94] 2.7); nothing else about the model changes, so configurations and the effect of a single step move to a shared `MultiTape/Basic.lean`. No existing statement changes meaning.

The semantics is a labelled transition system on configurations, labelled by the emitted symbol. A computation is a chain of such transitions, in the style of `SimpleGraph.Walk`, recording the trajectory that space depends on and the labels that the output depends on. `MultiTapeTM.toNTM` embeds the deterministic machine and preserves computation in bounded time and space.

### DIFF
--- a/Cslib/Computability/Machines/Turing/MultiTape/NonDeterministic.lean
+++ b/Cslib/Computability/Machines/Turing/MultiTape/NonDeterministic.lean
@@ -1,0 +1,181 @@
+/-
+Copyright (c) 2026 Aviv Bar Natan. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Aviv Bar Natan
+-/
+
+module
+
+public import Cslib.Foundations.Semantics.LTS.Execution
+public import Cslib.Computability.Machines.Turing.MultiTape.Deterministic
+
+/-!
+# Nondeterministic Multi-Tape Turing Machines
+
+Defines nondeterministic Turing machines with a read-only input tape, `k` work tapes and one
+write-only output tape, and what it means for one to compute an output within a time and space
+bound.
+
+## Design
+
+Following [Papadimitriou94], chapter 2.7, a nondeterministic machine is a `MultiTapeTM` whose
+transition function is replaced by a transition relation: `Tr q input work out` holds when `out` is
+one of the transitions permitted in that situation. Configurations, transition outputs and
+`TransitionOut.apply` are shared with the deterministic machine.
+
+The semantics is the labelled transition system `lts` on configurations, a step being labelled by
+the symbol it emits. The label is where that symbol has to live, since two transitions may lead to
+the same successor while emitting different symbols. A halted configuration steps to itself
+emitting nothing, as `MultiTapeTM.step` does.
+
+A `Computation` is a chain of such transitions, with `final`, `time`, `output` and `visited` read
+off it by recursion and `space` counting the cells `visited` touches. Space depends on the whole
+trajectory and output on the labels, so a computation records both.
+
+The transition relation may be empty at a running configuration. Every notion below asks for a
+computation whose final configuration has halted, so one that cannot continue is not a witness.
+
+## Important Declarations
+
+* `MultiTapeNTM`: the machine, an initial state and a transition relation
+* `lts`: the labelled transition system on configurations, labelled by the emitted symbol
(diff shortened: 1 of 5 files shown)


Cslib/Computability/Machines/Turing/MultiTape/NonDeterministic.lean
```diff
@@ -0,0 +1,181 @@
...
+write-only output tape, and what it means for one to compute an output within a time and space
+bound.
+
+## Design
+
+Following [Papadimitriou94], chapter 2.7, a nondeterministic machine is a `MultiTapeTM` whose
+transition function is replaced by a transition relation: `Tr q input work out` holds when `out` is
+one of the transitions permitted in that situation. Configurations, transition outputs and
+`TransitionOut.apply` are shared with the deterministic machine.
+
```
crei, line 25: Not sure if this is the right place, but I think we should clarify that in contrast to deterministic machines, it is possible that there is no successor configuration. This is not the same as halting because it does not reach a configuration with `state = none`. Maybe we should also cross-check what the textbooks say about this possibility.

Cslib/Computability/Machines/Turing/MultiTape/NonDeterministic.lean
```diff
@@ -0,0 +1,181 @@
...
+`Option Symbol` (where `none` is the blank symbol). Neither `Symbol` nor `State` is required to be
+finite.
+-/
+structure MultiTapeNTM (k : ℕ) (Symbol State : Type*) where
+  /-- initial state -/
+  q₀ : State
+  /-- transition relation: which transitions are permitted for a state, the current input symbol
+  and a tuple of work head symbols -/
+  Tr (q : State) (input : Option Symbol) (work : Fin k → Option Symbol) :
+    TransitionOut k Symbol State → Prop
```
crei, line 73:
```suggestion
  Tr (q : State) (input : Option Symbol) (work : Fin k → Option Symbol) (action : TransitionOut k Symbol State) : Prop
```

Since we are making big modifications here anyway, maybe we can find a better name for `TransitionOut` as well?

Cslib/Computability/Machines/Turing/MultiTape/NonDeterministic.lean
```diff
@@ -0,0 +1,181 @@
...
+    ntm.Step c c' ↔ c' = c := by
+  simp [Step, LTS.UnlabelledTr, lts_tr_of_halt h_halt]
+
+/-- The initial configuration corresponding to an input string. -/
+@[simp]
+def initCfg (input : List Symbol) : Cfg k Symbol State input := Cfg.init ntm.q₀ input
+
+/-- A computation of `ntm` from configuration `c`: a chain of transitions of `lts`, in the style of
+`SimpleGraph.Walk`. -/
+inductive Computation (ntm : MultiTapeNTM k Symbol State) (input : List Symbol) :
```
crei, line 109: Do you think (assuming we have https://github.com/leanprover/cslib/pull/779), this could also be defined via `List.isChain` - something like
````
structure Computation (ntm : MultiTapeNTM k Symbol State) (input : List Symbol) where
  configurations : List (Cfg k Symbol State input)
  isChain : configurations.isChain ntm.Step
```

Cslib/Computability/Machines/Turing/MultiTape/NonDeterministic.lean
```diff
@@ -0,0 +1,181 @@
...
+@[scoped grind =]
+abbrev Step (c₁ c₂ : Cfg k Symbol State input) : Prop := (ntm.lts input).UnlabelledTr c₁ c₂
+
+/-- A halted configuration steps only to itself. -/
+theorem step_of_halt {c c' : Cfg k Symbol State input} (h_halt : c.Halted) :
+    ntm.Step c c' ↔ c' = c := by
+  simp [Step, LTS.UnlabelledTr, lts_tr_of_halt h_halt]
+
+/-- The initial configuration corresponding to an input string. -/
+@[simp]
```
crei, line 104: I found this simp annotation a bit annoying for deterministic TMs, maybe we should remove it for both.

PR #803, opened 2026-08-15
feat(PFunctor): Add API of basic `PFunctor` constructions

dtumad:
PR to add various basic `PFunctor` definitions from [PolyFun](https://github.com/Verified-zkEVM/PolyFun) , split off from #731 to better isolate changes for review.

Main definitions are `PFunctor.monomial`, `PFunctor.add`, and `PFunctor.mul`, with abbreviations for common special cases of `PFunctor.monomial`. 

Discussion about the addition notation:
https://leanprover.zulipchat.com/#narrow/channel/270676-lean4/topic/Universe.20unification.20with.20.60HAdd.2EhAdd.60.20vs.2E.20.60.2B.60.20notation/with/617534952.01


Cslib/Foundations/Data/PFunctor/Basic.lean
```diff
@@ -0,0 +1,149 @@
...
+-/
+module
+
+public import Cslib.Init
+public import Mathlib.Data.PFunctor.Univariate.Basic
+
+/-!
+# Polynomial Functors
+
+This file defines additional constructions on `PFunctor` that don't belong in core mathlib.
```
eric-wieser: "Don't belong" or just "aren't provided"? I think this is fine in CSLib, but I think there is a difference between saying "mathlib is the wrong place for this" and "mathlib doesn't go this far (yet)".

Cslib/Foundations/Data/PFunctor/Basic.lean
```diff
@@ -0,0 +1,149 @@
...
+section add
+
+/-- The sum of two polynomial functors `P` and `Q`, written as `P + Q`,
+defined as the sum of the head types and the sum case analysis for the child types.
+
+This is kept as a `def` alongside the `HAdd` instance below, even though that instance is
+exactly as universe-general as `add` itself: the `binop%` elaborator behind `+` eagerly
+unifies the types of both operands with the expected type, so when the expected type carries
+universe metavariables (e.g. inside `PFunctor.W (P.add (.const α))` with `α : Type v`),
+`P + Q` can fail to elaborate where `P.add Q` succeeds; see `CslibTests/PFunctor.lean`. -/
```
eric-wieser, line 120: It would be great to raise this issue on Zulip and link the post from the PR description.

Cslib/Foundations/Data/PFunctor/Basic.lean
```diff
@@ -0,0 +1,149 @@
...
+@[simps] def add (P : PFunctor.{uA₁, uB}) (Q : PFunctor.{uA₂, uB}) :
+    PFunctor.{max uA₁ uA₂, uB} := ⟨P.A ⊕ Q.A, Sum.elim P.B Q.B⟩
+
+/-- Addition of polynomial functors, defined as the sum construction. -/
+instance instHAddPFunctor :
+    HAdd PFunctor.{uA₁, uB} PFunctor.{uA₂, uB} PFunctor.{max uA₁ uA₂, uB} where
+  hAdd := add
+
+@[simp] lemma add_def (P : PFunctor.{uA₁, uB}) (Q : PFunctor.{uA₂, uB}) :
+    P + Q = P.add Q := rfl
```
eric-wieser, line 132: I think you should explain here or above that since both spelling are allowed, we want to pick one to be simp-normal.

eric-wieser (approved):
It would be great to have another PFunctor expert review this for docstrings, but this all looks pretty reasonable to me.

Cslib/Foundations/Data/PFunctor/Basic.lean
dtumad: I guess I just remembered `PFunctor.FreeM` was decided didn't belong in Mathlib. Maybe I'm over generalizing from that. I'll change this, I don't think there's any real reason it couldn't be further upstreamed.

PR #806, opened 2026-08-16
refactor(LTS): convert LTS.Execution from a Prop to a structure

ctchou:
This PR converts `LTS.Execution` from a `Prop` with a complicated formula into a `structure` that should be easier to read.


Cslib/Foundations/Semantics/LTS/Execution.lean
```diff
@@ -19,11 +19,13 @@ namespace Cslib.LTS
...
-    (ss : List State) : Prop :=
-  ∃ _ : ss.length = μs.length + 1, ss[0] = s1 ∧ ss[ss.length - 1] = s2 ∧
-  ∀ k, {_ : k < μs.length} → lts.Tr ss[k] μs[k] ss[k + 1]
+@[scoped grind]
+structure Execution (lts : LTS State Label)
+    (s1 : State) (μs : List Label) (s2 : State) (ss : List State) where
+  length : ss.length = μs.length + 1
+  start : ss[0] = s1
+  last : ss[ss.length - 1] = s2
+  trans k (hk : k < μs.length) : lts.Tr ss[k] μs[k] ss[k + 1]
```
chenson2018: The style guide dictates this annotation be explicit:
```suggestion
  trans (k : ℕ) (hk : k < μs.length) : lts.Tr ss[k] μs[k] ss[k + 1]
```

Cslib/Foundations/Semantics/LTS/Execution.lean
```diff
@@ -42,13 +44,13 @@ theorem Execution.stepL {lts : LTS State Label} (htr : lts.Tr s1 μ s2)
...
-  · rfl
+  refine ⟨?_, ?_, ?_, ?_⟩
+  · grind
+  · grind
   · grind
-  · intro k valid
-    specialize h4 k <;> grind
+  · intro k hk
+    have := h.trans k
+    grind
```
chenson2018: You can manage one fewer calls to `grind` here:

```suggestion
  have : ss.length = μs.length + 1 := by grind
  have (k : ℕ) : (_ : k < μs.length) → lts.Tr ss[k] μs[k] ss[k + 1] := by have := h.trans k; grind
  grind
```

Cslib/Foundations/Semantics/LTS/Execution.lean
```diff
@@ -133,7 +135,13 @@ theorem Execution.split
...
   have : n + (ss.length - n - 1) = ss.length - 1 := by grind
-  simp [Execution]
-  grind
+  split_ands
+  · grind
+  · refine ⟨?_, ?_, ?_, ?_⟩
+    · grind
+    · grind
+    · simp only [List.length_drop, List.getElem_drop]
+      grind
```
chenson2018: This needs to be fixed. This works:

```suggestion
    · grind only [= List.length_drop, = List.getElem_drop]
```
but fails on removing the `only`. Can you diagnose why this is the case?

Cslib/Foundations/Semantics/LTS/Execution.lean
```diff
@@ -114,8 +116,8 @@ theorem Execution.comp
     (h1 : lts.Execution s μs1 r ss1) (h2 : lts.Execution r μs2 t ss2) :
     lts.Execution s (μs1 ++ μs2) t (ss1 ++ ss2.tail) := by
   have h0 : (ss1 ++ ss2.tail).length = (μs1 ++ μs2).length + 1 := by grind
-  use h0
-  split_ands
+  refine ⟨?_, ?_, ?_, ?_⟩
```
chenson2018: In general it is strange to write `refine ⟨?_, ?_, ?_, ?_⟩`, it seems `constructor` would be more idiomatic.

Cslib/Foundations/Semantics/LTS/Reverse.lean
```diff
@@ -105,9 +105,11 @@ theorem reverse_boundedUpTo {lts : LTS State Label} {n : ℕ} :
...
 theorem Execution.reverse {lts : LTS State Label} (h : lts.Execution s μs s' ss) :
     lts.reverse.Execution s' μs.reverse s ss.reverse := by
-  obtain ⟨_, _, _, _⟩ := h
-  use by simpa
-  grind only [reverse_tr, = List.getElem_reverse, = List.length_reverse]
+  refine ⟨?_, ?_, ?_, ?_⟩
+  · grind
+  · grind
+  · grind
+  · grind only [reverse_tr, = List.getElem_reverse, = List.length_reverse]
```
chenson2018: This is similarly a problem that is preventing us from being able to call `grind` just once. Could you diagnose what bad `grind` annotations are causing this to fail if you try to use `grind?`?

chenson2018 (changes requested):
This looks like an improvement regardless and some of this is preexisting, but could you take a look at the couple of places we are forced to use `grind only`?

Cslib/Foundations/Semantics/LTS/Execution.lean
ctchou: Fixed.

Cslib/Foundations/Semantics/LTS/Execution.lean
ctchou: Fixed.

Cslib/Foundations/Semantics/LTS/Execution.lean
ctchou: A problem with using `constructor` here is that the subgoals it generates has a different order than that in the structure definition, which is confusing.  So I would like to keep the `refine`.

chenson2018: You could also write `apply Execution.mk ..` which is perhaps more clear and retains the order.

Cslib/Foundations/Semantics/LTS/Execution.lean
ctchou: When I use `grind?` instead of `grind only`, `grind` times out.  My guess is that there are a lot of results in the `grind` database and `grind` is not smart enough to pick the right ones to apply.  `grind? only` returns basically the same command.  Personally I'm not sure why you are bothered by the `simp only; grind` sequence.  It is not uncommon that `simp` can make a goal solvable by `grind`.  I'll use the `grind only` tactic here.

chenson2018: Yes, I understand that `grind` times out. I am asking if you could examine the annotations that are applicable here to understand why.

I am bothered by `simp; grind` because it is a bit of an unknown/potentially unstable for maintenance reasons. Upstream in Mathlib it's still not settled how common this should be. My intuition is that it's often an indicator of bad `grind` annotations, especially in this case when you can move the theorem from `simp` to `grind`, but only when limiting instantiation with `only`. 

This is really out of scope for this PR, but this is the sort of concern that really bothers me about this part of the library. It's just a bit difficult to debug and I haven't had the time to figure it out.

chenson2018 (approved):
I'm going to go ahead and approve because most of my remaining comments are out of scope for this PR.

See my last comment about another alternative to refine, and feel free to merge after that.

Cslib/Foundations/Semantics/LTS/Execution.lean
ctchou: Thanks!  I didn't know that idiom.  Fixed.

chenson2018: No problem. The way that `..` works is that it expands to as many `?_` as are needed.

Cslib/Foundations/Semantics/LTS/Execution.lean
ctchou: I think I'll use this proof:
```
apply Execution.mk ..
      <;> simp only [List.length_drop, List.getElem_drop] <;> grind
```
Basically the `simp` would eliminate or simplify everything about `length` in all the subgoals and then `grind` can solve all of them.

Cslib/Foundations/Semantics/LTS/Reverse.lean
ctchou: This proof works here:
```
  apply Execution.mk .. <;>
    simp only [reverse_tr, List.getElem_reverse, List.length_reverse] <;> grind only
```

PR #809, opened 2026-08-17
feat(Crypto/Primitives/ECC): Edwards Curves

chris-anto-froeschl:
Formalizes some basic notions for (twisted) Edwards curves. As part of this, I propose introducing a new `Crypto/Primitives/ECC` subfolder.

## Contents
This PR adds two files:

- `Crypto/Primitives/ECC/TwistedEdwardsCurve.lean`: the coefficients `a`, `d` of a twisted Edwards curve, its defining equation, affine point set, bundled points, neutral point and negation, plus a validity (nonsingularity) predicate.
- `Crypto/Primitives/ECC/EdwardsCurve.lean`:  the untwisted Edwards curve as the `a = 1` specialization, with unfolding lemmas, the criterion `d ≠ 0 ∧ d ≠ 1` for a valid model, and the fact that `(0, 1)` lies on every such curve.

Everything is stated over an arbitrary commutative ring: no finite field, no cardinality assumption, and no dependence on any parameter of a particular cryptosystem.

## Context

This PR is primarily preparatory work for https://github.com/leanprover/cslib/pull/783 and subsequent PRs introducing the [Elligator cryptosystem](https://eprint.iacr.org/2013/325.pdf).

I am by no means a cryptographer by profession, so the definitions introduced here may not represent the most general or canonical formulation found in the literature. My immediate goal is simply to establish the standard notions needed for the twisted Edwards curve equation used by Elligator 1.

I would therefore welcome improvements to the terminology and generality from someone with more expertise in elliptic-curve cryptography. For now, however, this provides a small set of ECC primitives that can be built upon by the subsequent Elligator work.

## Follow-ups

Future PRs will build on these primitives with concrete curve constructions under `Crypto/Primitives/ECC/Curves`, such as `Curve1174` and `Curve25519`. The intended shape is that a curve is described purely in terms of these primitives (its base field, its Edwards coefficient, the non-squareness of that coefficient, and its base point) and that any cryptosystem-specific parameter build-up lives separately and merely proves it lands on that curve. A preview of this usage can be found [here](https://github.com/chris-anto-froeschl/elligator/blob/0bb1ed1c12a4b3d861e7adbe4f9094cb52d1be4e/Elligator/Elligator1/Curve1174.lean#L162).

I'm happy to provide further context if usage of some lemmas/defs is non-obvious by this current stage.


Cslib/Crypto/Primitives/ECC/EdwardsCurve.lean
```diff
@@ -0,0 +1,82 @@
...
+    (edwardsCurve d).IsValid ↔ d ≠ 0 ∧ d ≠ 1 := by
+  simp [edwardsCurve]
+
+/-- `edwardsCurveEquation` is the standard Edwards curve equation, with the coefficient carried as
+a subtype element recording `d ≠ 0` and `d ≠ 1`.  New generic developments should normally use
+`(edwardsCurve d).Equation x y` and carry coefficient validity separately via
+`TwistedEdwardsCurve.IsValid`; see `edwardsCurve_isValid_iff`.
+-/
+def edwardsCurveEquation (x y : R) (d : {d : R // d ≠ 0 ∧ d ≠ 1}) : Prop :=
+  (edwardsCurve d.val).Equation x y
```
SamuelSchlesinger: Given that we deprecate this in this comment and we're making this PR for the first time, lets just make these changes as we go instead and remove this definition altogether.

Cslib/Crypto/Primitives/ECC/EdwardsCurve.lean
```diff
@@ -0,0 +1,82 @@
...
+
+@[expose] public section
+
+namespace Cslib.Crypto.Primitives.ECC
+
+variable {R : Type*} [CommRing R]
+
+/-- The Edwards curve with coefficient `d`.
+This is an alias for the `a = 1` specialization of a twisted Edwards curve. -/
+def edwardsCurve (d : R) : TwistedEdwardsCurve R := TwistedEdwardsCurve.ofD d
```
SamuelSchlesinger: Lets delete this and let consumers write TwistedEdwardsCurve.ofD or keep Edwards Curve as the name there and drop ofD. Either way, this file looks more or less unnecessary.

Cslib/Crypto/Primitives/ECC/EdwardsCurve.lean
```diff
@@ -0,0 +1,82 @@
...
+/-- The set of affine points of the Edwards curve with coefficient `d`. -/
+theorem edwardsCurve_affinePoints (d : R) :
+    (edwardsCurve d).affinePoints =
+      {p : R × R | p.1 ^ 2 + p.2 ^ 2 = 1 + d * p.1 ^ 2 * p.2 ^ 2} := by
+  ext p
+  simp [TwistedEdwardsCurve.affinePoints]
+
+lemma edwardsCurveEquation_zero_one (d : {d : R // d ≠ 0 ∧ d ≠ 1}) :
+    edwardsCurveEquation (0 : R) (1 : R) d := by
+  simp
```
SamuelSchlesinger: Yeah, overall this whole file looks like a duplicate to me. If this is the interface you want, give it to yourself from the TwistedEdwardsCurve file.

Cslib/Crypto/Primitives/ECC/TwistedEdwardsCurve.lean
```diff
@@ -0,0 +1,107 @@
...
+
+/-- Negation of affine coordinates on a twisted Edwards curve. -/
+def neg (p : R × R) : R × R := (-p.1, p.2)
+
+@[simp]
+theorem neg_mem_affinePoints (E : TwistedEdwardsCurve R) (p : R × R) :
+    neg p ∈ E.affinePoints ↔ p ∈ E.affinePoints := by
+  change E.a * (-p.1) ^ 2 + p.2 ^ 2 = 1 + E.d * (-p.1) ^ 2 * p.2 ^ 2 ↔
+    E.a * p.1 ^ 2 + p.2 ^ 2 = 1 + E.d * p.1 ^ 2 * p.2 ^ 2
+  rw [neg_sq]
```
SamuelSchlesinger, line 91: This is probably the best @[simp] lemma here, but my preference is to remove the @[simp] lemmata and add them back when they have users and where it is obvious how they justify themselves.

SamuelSchlesinger (changes requested):
I like this a lot better! I think we can still remove some of it for now. Lets rename the file to EllipticCurves, the placement is fine for now.

Cslib/Crypto/Primitives/ECC/EdwardsCurve.lean
chris-anto-froeschl: Fixed in 59ce03d2d8f27b561f28759e993e5cb574695363. I decided to drop the `ofD` interface and go with the ones that are actually a bit more elegant in usage of my Elligator infrastructure.

Regarding the file rename: are you sure something like `EllipticCurves.lean` is correct? I would have rather renamed this to `EdwardsCurve.lean` (since we do not only talk about twisted ones now there) and be open for other curve families getting their own file if needed. Although this discussion is perhaps rather irrelevant if this is ultimately destined to land in mathlib anyway.

Cslib/Crypto/Primitives/ECC/EdwardsCurve.lean
chris-anto-froeschl: See 59ce03d2d8f27b561f28759e993e5cb574695363 discussion below.

Cslib/Crypto/Primitives/ECC/EdwardsCurve.lean
chris-anto-froeschl: See 59ce03d2d8f27b561f28759e993e5cb574695363 discussion below.

Cslib/Crypto/Primitives/ECC/TwistedEdwardsCurve.lean
chris-anto-froeschl: Fixed in 11c33ca169f36cd24ab0d03b9bb1f4404c9732bc.

PR #810, opened 2026-08-17
feat: graph definitions (#Attempts 3)

sorrachai:
This PR introduces basic graph definitions: `Graph`, `Digraph`, `SimpleGraph`, and `SimpleDiGraph`. 

We use the following definition of a multigraph.
A multigraph is a triple (V,E,f) where V is a vertex set, E is an edge set,
and f is a function from an edge to an (ordered/unordered) pair of vertices.
Preferably, f is a computable function. We reuse the definitions from Mathlib whenever possible.

* `MultiGraph α β`: abbrev for Mathlib `Graph α β`, adding a typeclass to keep track of endpoints explicitly. 
* `MultiDigraph α β`: a directed multigraph; the directed counterpart of Mathlib's `Graph α β` with explicit endpoints. 
* `SimpleGraph α`: a simple graph with a vertex set, extending Mathlib's `SimpleGraph α`.
* `SimpleDigraph α`: a loopless directed graph with adjacency `Adj : α → α → Prop` and a
  vertex set, extending Mathlib's `Digraph α`.

Comparison to PR #503 (https://github.com/leanprover/cslib/pull/503): This version reuses Mathlib as much as possible. All vertex sets are `Set`-valued, following the design of `Mathlib.Combinatorics.Graph`. 

Zulip discussion: Follow at 
https://leanprover.zulipchat.com/#narrow/channel/252551-graph-theory/topic/A.20Set-based.20Multigraph.20Definition/with/616948793.


Cslib/Algorithms/Lean/Graph/Basic.lean
```diff
@@ -0,0 +1,161 @@
...
+
+namespace Cslib.Algorithms.Lean
+
+
+/-- An undirected multigraph on vertex type `α` with edge labels in `β`, bundled with a
+computable map from an edge label to its ends.
+
+This is Mathlib's `Graph α β` — so parallel edges and loops are permitted, and both the
+vertex and edge sets may be infinite. -/
+abbrev Graph (α β : Type*) :=  _root_.Graph α β
```
ctchou: I would suggest replacing `α` and `β` by `V` and `E`.  When reading the code below, I have to constantly remind myself that `α` and `β` are the carrier types for vertices and edges.  Using `V` and `E` will make this transparent.

Cslib/Algorithms/Lean/Graph/Basic.lean
```diff
@@ -0,0 +1,161 @@
...
+
+/-- A computable map from an edge label of `G` to its ends. -/
+class Graph.HasEndpoints {α β : Type*} (G : Graph α β) where
+  /-- The ends of the edge labelled `e`, or `none` if `e` is not an edge of `G`. -/
+  endpoints : β → Option (Sym2 α)
```
ctchou: There is already a notion of partial functons in mathlib:
https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/PFun.html
which has a lot of API's.

Cslib/Algorithms/Lean/Graph/Basic.lean
```diff
@@ -0,0 +1,161 @@
...
+  incidence  : ∀ ⦃e x y⦄, IsArc e x y → (x ∈ vertexSet ∧ y ∈ vertexSet) := by grind
+  /-- The set of arc labels. -/
+  arcSet : Set β := { e | ∃ x y, IsArc e x y}
+  /-- A label lies in `arcSet` exactly when it is used by some arc. -/
+  arc_mem_iff_exists_isArc (e) : e ∈ arcSet ↔ ∃ x y, IsArc e x y := by exact fun _ ↦ Iff.rfl
+
+/-- A computable map from an edge label of `G` to its ends. -/
+class Digraph.HasEndpoints {α β : Type*} (G : Digraph α β) where
+  /-- The ends of the edge labelled `e`, or `none` if `e` is not an edge of `G`. -/
+  endpoints : β → Option (α × α)
```
ctchou: Same comment about partial functions as above.

Cslib/Algorithms/Lean/Graph/Basic.lean
```diff
@@ -0,0 +1,161 @@
...
+Extends Mathlib's `SimpleGraph α`, so its API is available through `toSimpleGraph`; in
+particular `edgeSet`, the unordered pairs of adjacent vertices, is inherited rather than
+redefined. Note that `Adj` is a relation on all of `α`, so `vertexSet` may be any superset
+of the vertices actually incident to an edge. -/
+structure SimpleGraph (α : Type*) extends _root_.SimpleGraph α where
+  /-- The set of vertices. -/
+  vertexSet : Set α
+  /-- The left end of every adjacent pair is a vertex. The right end then follows by
+  symmetry of `Adj`. -/
+  left_incidence : ∀ ⦃x y⦄, Adj x y → x ∈ vertexSet := by grind
```
ctchou: You don't need `:= by grind`.  There is nothing to prove here.

Cslib/Algorithms/Lean/Graph/Basic.lean
```diff
@@ -0,0 +1,161 @@
...
+`loopless` is what rules them out here. Antiparallel arcs are permitted: `Adj x y` and
+`Adj y x` may both hold. -/
+structure SimpleDigraph (α : Type*) extends _root_.Digraph α where
+  /-- The set of vertices. -/
+  vertexSet : Set α
+  /-- No vertex is adjacent to itself. -/
+  loopless : Std.Irrefl Adj
+  /-- Both ends of every adjacent pair are vertices. Unlike `SimpleGraph`, `Adj` is not
+  symmetric, so neither direction follows from the other. -/
+  incidence : ∀ ⦃x y⦄, Adj x y → x ∈ vertexSet ∧ y ∈ vertexSet := by grind
```
ctchou: You don't need `:= by grind`. There is nothing to prove here.

Cslib/Algorithms/Lean/Graph/Basic.lean
```diff
@@ -0,0 +1,161 @@
...
+theorem isLink_iff_endpoints (G : Graph α β) [G.HasEndpoints] :
+  G.IsLink e x y ↔ G.endpoints? e = some s(x, y) :=
+  Graph.HasEndpoints.endpoints_spec e x y
+
+/-- A directed multigraph on vertex type `α` with arc labels in `β`, bundled with a
+computable map from an arc label to its endpoints.
+
+The directed counterpart of Mathlib's `Graph α β`, which has no Mathlib counterpart to
+extend; the field layout mirrors it, with symmetry dropped. -/
+structure Digraph (α β : Type*) where
```
ctchou: Since you are making this definition from scratch, why not state it using `endpoints`, rather than adapting mathlib's `Graph`?  You should also consider if mathlib's `Quiver` is applicable here.

Also, since mathlib already has a `Digraph` and this definition is not an extension of it, I would suggest calling it something else.  Perhaps `DirectedMultigraph`?

Cslib/Algorithms/Lean/Graph/Basic.lean
chenson2018: I think there's some confusion here: when declaring a structure/typeclass this is how you specify a default tactic to run.

ctchou: OK.  I didn't know that.  The syntax is not helpful here.

Cslib/Algorithms/Lean/Graph/Basic.lean
chenson2018: I don't understand having this `abbrev` if they are both called `Graph`, and are defined exactly the same way.

Cslib/Algorithms/Lean/Graph/Basic.lean
sorrachai: Good point. Rename it to `MultiDigraph` and from `Graph` to `MultiGraph`.

sorrachai: (Quiver is not quite what we want; they generalize V -> V -> Prop to V -> V -> Type, which means no global edge labeling.)

Cslib/Algorithms/Lean/Graph/Basic.lean
sorrachai: (Just to add the default tactic to prove the goal)

Cslib/Algorithms/Lean/Graph/Basic.lean
sorrachai: (Just to add the default tactic to prove the goal)

Cslib/Algorithms/Lean/Graph/Basic.lean
sorrachai: I tried a bit, but I found that it allows even non-computable Prop. It would be overkill to use it, especially since we want to keep everything computable.

Cslib/Algorithms/Lean/Graph/Basic.lean
sorrachai: Just to rename things for consistency. Now, it is renamed as MultiGraph.

Cslib/Algorithms/Lean/Graph/Basic.lean
```diff
@@ -0,0 +1,137 @@
...
+  /-- The incidence predicate: `IsArc e x y` states that the arc labelled `e` runs from
+  `x` to `y`. -/
+  IsArc : E → V → V → Prop
+  /-- The ends of the edge labelled `e`, or `none` if `e` is not an edge of `G`. -/
+  endpoints : E → Option (V × V)
+  /-- `endpoints` computes `Graph.IsLink`. -/
+  endpoints_spec : ∀ e x y, IsArc e x y ↔ (x, y) ∈ endpoints e
+  /-- Both ends of every arc are vertices. `IsArc` is not symmetric, so neither direction
+  follows from the other. -/
+  incidence  : ∀ ⦃e x y⦄, IsArc e x y → (x ∈ vertexSet ∧ y ∈ vertexSet) := by grind
```
SnirBroshi:
```suggestion
  incidence : ∀ ⦃e x y⦄, IsArc e x y → (x ∈ vertexSet ∧ y ∈ vertexSet) := by grind
```

Cslib/Algorithms/Lean/Graph/Basic.lean
```diff
@@ -0,0 +1,137 @@
...
+  /-- The set of vertices. -/
+  vertexSet : Set V
+  /-- No vertex is adjacent to itself. -/
+  loopless : Std.Irrefl Adj
+  /-- Both ends of every adjacent pair are vertices. Unlike `SimpleGraph`, `Adj` is not
+  symmetric, so neither direction follows from the other. -/
+  incidence : ∀ ⦃x y⦄, Adj x y → x ∈ vertexSet ∧ y ∈ vertexSet := by grind
+
+/-- The arc set of a `SimpleDigraph`, as ordered pairs of adjacent vertices. -/
+def SimpleDigraph.arcSet (G : SimpleDigraph V) : Set (V × V) :=
```
SnirBroshi: This is diverging from Mathlib's terminology
```suggestion
/-- The edge set of a `SimpleDigraph`, as ordered pairs of adjacent vertices. -/
def SimpleDigraph.edgeSet (G : SimpleDigraph V) : Set (V × V) :=
```

Cslib/Algorithms/Lean/Graph/Basic.lean
```diff
@@ -0,0 +1,137 @@
...
+`loopless` is what rules them out here. Antiparallel arcs are permitted: `Adj x y` and
+`Adj y x` may both hold. -/
+structure SimpleDigraph (V : Type*) extends _root_.Digraph V where
+  /-- The set of vertices. -/
+  vertexSet : Set V
+  /-- No vertex is adjacent to itself. -/
+  loopless : Std.Irrefl Adj
+  /-- Both ends of every adjacent pair are vertices. Unlike `SimpleGraph`, `Adj` is not
+  symmetric, so neither direction follows from the other. -/
+  incidence : ∀ ⦃x y⦄, Adj x y → x ∈ vertexSet ∧ y ∈ vertexSet := by grind
```
SnirBroshi:
```suggestion
  adj_imp_left_mem_vertexSet : ∀ ⦃x y⦄, Adj x y → x ∈ vertexSet := by grind
  adj_imp_right_mem_vertexSet : ∀ ⦃x y⦄, Adj x y → y ∈ vertexSet := by grind
```

Cslib/Algorithms/Lean/Graph/Basic.lean
```diff
@@ -0,0 +1,137 @@
...
+an adjacent pair.
+
+Extends Mathlib's `Digraph V`, which is a bare adjacency relation and does permit loops;
+`loopless` is what rules them out here. Antiparallel arcs are permitted: `Adj x y` and
+`Adj y x` may both hold. -/
+structure SimpleDigraph (V : Type*) extends _root_.Digraph V where
+  /-- The set of vertices. -/
+  vertexSet : Set V
+  /-- No vertex is adjacent to itself. -/
+  loopless : Std.Irrefl Adj
```
SnirBroshi:
```suggestion
  irrefl_adj : Std.Irrefl Adj
```

Cslib/Algorithms/Lean/Graph/Basic.lean
```diff
@@ -0,0 +1,137 @@
...
+Extends Mathlib's `SimpleGraph V`, so its API is available through `toSimpleGraph`; in
+particular `edgeSet`, the unordered pairs of adjacent vertices, is inherited rather than
+redefined. Note that `Adj` is a relation on all of `V`, so `vertexSet` may be any superset
+of the vertices actually incident to an edge. -/
+structure SimpleGraph (V : Type*) extends _root_.SimpleGraph V where
+  /-- The set of vertices. -/
+  vertexSet : Set V
+  /-- The left end of every adjacent pair is a vertex. The right end then follows by
+  symmetry of `Adj`. -/
+  left_incidence : ∀ ⦃x y⦄, Adj x y → x ∈ vertexSet := by grind
```
SnirBroshi:
```suggestion
  adj_imp_left_mem_vertexSet : ∀ ⦃x y⦄, Adj x y → x ∈ vertexSet := by grind
```

Cslib/Algorithms/Lean/Graph/Basic.lean
```diff
@@ -0,0 +1,137 @@
...
+  IsArc : E → V → V → Prop
+  /-- The ends of the edge labelled `e`, or `none` if `e` is not an edge of `G`. -/
+  endpoints : E → Option (V × V)
+  /-- `endpoints` computes `Graph.IsLink`. -/
+  endpoints_spec : ∀ e x y, IsArc e x y ↔ (x, y) ∈ endpoints e
+  /-- Both ends of every arc are vertices. `IsArc` is not symmetric, so neither direction
+  follows from the other. -/
+  incidence  : ∀ ⦃e x y⦄, IsArc e x y → (x ∈ vertexSet ∧ y ∈ vertexSet) := by grind
+  /-- The set of arc labels. -/
+  arcSet : Set E := { e | ∃ x y, IsArc e x y}
```
SnirBroshi: This is diverging from Mathlib's terminology
```suggestion
  /-- The set of edge labels. -/
  edgeSet : Set E := { e | ∃ x y, IsArc e x y}
```

Cslib/Algorithms/Lean/Graph/Basic.lean
ctchou: That a non-computable `Prop` is allowed doesn't mean you have to use a non-computable `Prop`.  It just means that Lean will not insists that the `Prop` is computable.  Note that you can build a `Set` from a non-computable `Prop` as well.  Do you want to replace `Set` by `Finset` everywhere in your code?

The main reason for my suggesting `PFun` is that it already has a lot of API support, including ones that work via `Option`.

sorrachai: Sure, let's do it.

Cslib/Algorithms/Lean/Graph/Basic.lean
```diff
@@ -0,0 +1,141 @@
...
+
+/-- A map from an edge label of `G` to its ends. -/
+class MultiGraph.HasEndpoints {V E : Type*} (G : MultiGraph V E) where
+  /-- The ends of the edge labelled `e`, or `none` if `e` is not an edge of `G`. -/
+  endpoints : E →. (Sym2 V)
+  /-- `endpoints` computes `Graph.IsLink`. -/
+  endpoints_spec : ∀ e x y, G.IsLink e x y ↔ s(x, y) ∈ endpoints e
+
+/-- The ends of `e` in `G`; undefined when `e ∉ E(G)`. -/
+def MultiGraph.endpoints (G : MultiGraph α β) [inst : G.HasEndpoints] : β →. Sym2 α :=
```
ctchou: α, β -> V, E

Cslib/Algorithms/Lean/Graph/Basic.lean
```diff
@@ -0,0 +1,141 @@
...
+  /-- `endpoints` computes `IsLink`. -/
+  endpoints_spec : ∀ e x y, IsLink e x y ↔ (x, y) ∈ endpoints e
+  /-- Both ends of every edge are vertices. `IsLink` is not symmetric, so neither direction
+  follows from the other. -/
+  isLink_imp_left_mem_vertexSet : ∀ ⦃e x y⦄, IsLink e x y → x ∈ vertexSet := by grind
+  isLink_imp_right_mem_vertexSet : ∀ ⦃e x y⦄, IsLink e x y → y ∈ vertexSet := by grind
+  /-- The set of edge labels. -/
+  edgeSet : Set E := { e | ∃ x y, IsLink e x y}
+  /-- A label lies in `edgeSet` exactly when it is used by some edge. -/
+  edge_mem_iff_exists_IsLink (e) : e ∈ edgeSet ↔ ∃ x y, IsLink e x y := by exact fun _ ↦ Iff.rfl
```
ctchou: I don't think you need both `IsLink` and `endpoints` to be in the structure `MultiDigraph`, because `endpoints_spec` can serve as the definition of `IsLink` in terms of `endpoints`.  The contents of `isLink_imp_{left,right}_mem_vertexSet` can be expressed in terms of `endpoints` directly: namely, the `PFun.image` of `endpoints` projected to a `Set V` is a subset of `vertexSet`.  Then `isLink_imp_{left,right}_mem_vertexSet` can be derived as theorems.  Going the other way (namely, defining `endpoints` in terms of `IsLink`) is also possible, but then you'll need an analogue of mathlib's `Graph. eq_or_eq_of_isLink_of_isLink` to prevent the same edge label to be applied to multiple edges.  In either case, that analogue should probably be stated and proved/assumed.

Also, stylistically, I would prefer `edgeSet` and `edge_mem_iff_exists_IsLink` to become a separate definition and theorem outside the structure.

ctchou (changes requested):

Cslib/Algorithms/Lean/Graph/Basic.lean
sorrachai: Thanks for the comment. Indeed, these two fields are redundant. The question is which one should be in the field (and which should be derived). I would prefer endpoints because they explicitly provide a vertex pair. The reason for the IsLink relation was to [avoid quotient types (Sym2)](https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/New.20graph.20definitions/near/586415017)[#CSLib > New graph definitions @ 💬](https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/New.20graph.20definitions/near/586415017) in the case of an undirected multigraph. This is no longer the case for directed graphs.

ctchou: I think the following two `Prop`s should suffice to imply `isLink_imp_{left,right}_mem_vertexSet`:
```
left_endpoint_mem_vertexSet : endpoints.ran.MapsTo Prod.fst vertexSet
right_endpoint_mem_vertexSet : endpoints.ran.MapsTo Prod.snd vertexSet
```

sorrachai: Indeed. Added isLink_imp_left_mem_vertexSet and isLink_imp_right_mem_vertexSet.

Cslib/Algorithms/Lean/Graph/Basic.lean
```diff
@@ -0,0 +1,160 @@
...
+/-- A map from an edge label of `G` to its ends. -/
+class MultiGraph.HasEndpoints {V E : Type*} (G : MultiGraph V E) where
+  /-- The ends of the edge labelled `e`, or `none` if `e` is not an edge of `G`. -/
+  endpoints : E →. (Sym2 V)
+  /-- `endpoints` computes `Graph.IsLink`. -/
+  endpoints_spec : ∀ e x y, G.IsLink e x y ↔ s(x, y) ∈ endpoints e
+
+/-- The ends of `e` in `G`; undefined when `e ∉ E(G)`. -/
+def MultiGraph.endpoints (G : MultiGraph V E) [inst : G.HasEndpoints] : E →. Sym2 V :=
+  inst.endpoints
```
ctchou, line 75: I wonder if it is possible to define `MultiGraph.endpoints` directly.  Unfortunately, all definitions I can think of uses `Classical.choose` somewhere and hence must be marked as `noncomputable`.  But if we accept that, then we can get rid of the assumption `[G.HasEndpoints]`, which currently needs to be assumed whenever `MultiGraph.endpoints` is used.  Perhaps that can be considered an improvement and a small price to pay for using mathlib's `Graph`.

sorrachai: Alternatively, one could extend Mathlib's Graph with field endpoints and endpoints_spec, but this seems redundant, since it subsumes IsLink. I am happy with the current price to pay if we cannot define MultiGraph ourselves.

ctchou (approved):
I agree with @sorrachai's assessment.

PR #813, opened 2026-08-17
refactor(Crypto): replace bespoke constructions with Mathlib abstractions

SamuelSchlesinger:
While surveying the crypto files for things Mathlib now does better, a few cleanups fell out. Net effect is about −100 lines with no statement weakened.

- The internal one-time pad file is gone: both of its `TODO: upstream` items have since landed in Mathlib (`FinEnum (BitVec n)`, `LawfulXor`), and ciphertext uniformity is now just "XOR by the key is a permutation of the uniform distribution" via `Equiv.xor`. That theorem stays public since uniformity is stronger than perfect secrecy.
- `Internal/PerfectSecrecy.lean` is folded into `Basic.lean`, so the `Internal/` convention is retired and each theorem has exactly one public name.
- `posteriorDist` is now built on `PMF.normalize` instead of a hand-rolled `HasSum` proof, and perfect secrecy shares its "conditioning doesn't change the prior" lemma with secret sharing instead of re-deriving it.
- Smaller stuff: `coeffTranslate` was `Equiv.addRight` all along, `authorized_mono` is now stated as `Monotone`, a docstring honesty fix for `TailSampler` (translation invariance forces uniformity), namespace opens instead of fully-qualified names, and some proof golfing.

Two things reviewers might trip on: `posteriorDist_apply` is no longer `rfl` (proofs that unfolded the old definition need to go through the lemma now), and the `authorized_mono` field changed shape from `∀ {s t}, s ⊆ t → …` to `Monotone`.

Assisted by Claude Fable 5.


Cslib/Crypto/Protocols/PerfectSecrecy/OneTimePad.lean
```diff
@@ -34,18 +35,30 @@ The one-time pad (Vernam cipher) over `BitVec l`
...
 noncomputable def otp (l : ℕ) :
     EncScheme (BitVec l) (BitVec l) (BitVec l) :=
   .ofPure (PMF.uniformOfFintype _) (· ^^^ ·) (· ^^^ ·) fun k m => by
-    simp [← BitVec.xor_assoc]
+    simp [xor_cancel_left]
+
+/-- The ciphertext distribution of the OTP is uniform, regardless of the
+message: masking with a uniform key is the permutation `Equiv.xor` of the
+uniform distribution. -/
+theorem otp_ciphertextDist_eq_uniform (l : ℕ) (m : BitVec l) :
```
crei, line 50: Naming: strictly, in its syntactic form, this is not a statement about `otp`, but only about xor.

Cslib/Crypto/Protocols/SecretSharing/Shamir.lean
```diff
@@ -109,28 +111,25 @@ noncomputable def reconstruct (params : Params F Party)
 
 /-- A sampler on Shamir tail coefficients is privacy-compatible when its
 distribution is invariant under translation by any coefficient vector. This is
-the exact symmetry needed in the privacy proof. -/
+the exact symmetry needed in the privacy proof.
+
+The requirement is strong: translation invariance forces the distribution. On
```
crei: This reads a bit like an AI-generated comment in its common compressed form - is this deliberate or not (is the comment needed here)?

Cslib/Crypto/Protocols/PerfectSecrecy/Basic.lean
```diff
@@ -31,21 +35,108 @@ Characterisation theorems for perfect secrecy following
...
+    · have hne_top := PMF.apply_ne_top (scheme.marginalCiphertextDist msgDist) c
+      have := DFunLike.congr_fun (h msgDist c ((PMF.mem_support_iff _ _).mpr hc)) m
+      simp only [posteriorMsgDist_apply] at this
+      rw [← this, ENNReal.div_mul_cancel hc hne_top]
+  · ext m
+    simp only [posteriorMsgDist_apply]
+    rw [h msgDist m c, ENNReal.mul_div_cancel_right
+      ((PMF.mem_support_iff _ _).mp hc) (PMF.apply_ne_top _ c)]
+
+private theorem perfectlySecret_of_ciphertextIndist (scheme : EncScheme M K C)
```
crei: Maybe better to inline this into the proof of `perfectlySecret_iff_ciphertextIndist`?

crei (commented):
Just some minor comments.

Cslib/Crypto/Protocols/SecretSharing/Shamir.lean
SamuelSchlesinger: Good catch, no it is not needed.

Cslib/Crypto/Protocols/PerfectSecrecy/OneTimePad.lean
SamuelSchlesinger: Good point.

Cslib/Crypto/Protocols/PerfectSecrecy/Basic.lean
SamuelSchlesinger: I'll see how it looks when I do that.

crei (approved):

PR #816, opened 2026-08-18
feat(MultiTapeTM): halting-time, output-length, and visited-position lemmas for multi-tape TMs

SamuelSchlesinger:
This is in advance of a PR for Turing machine composition.

- exists_minimal_halting_config: a padded halting run has a least halting time reaching the same configuration
- outputString_length_le / output_length_le_time: the output is no longer than the elapsed time
- spaceUsedByTape_le_of_positions: space comparison between two machines via inclusion of visited head positions

### DIFF
--- a/Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
+++ b/Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
@@ -289,6 +289,26 @@ lemma configs_of_halts (cfg : Cfg k Symbol State input) (h : cfg.state = none) {
   | succ d ih =>
     rw [configs_succ_eq_step', ih, step_of_halt h]
 
+/-- Every padded halting computation has a least halting time, and the configuration there is the
+same as at the padded time. -/
+lemma exists_minimal_halting_config (tm : MultiTapeTM k Symbol State)
+    {input : List Symbol} (start : Cfg k Symbol State input) (t : ℕ)
+    (hhalt : (tm.configs start t).state = none) :
+    ∃ u ≤ t, (tm.configs start u).state = none ∧
+      (∀ m < u, (tm.configs start m).state ≠ none) ∧
+      tm.configs start u = tm.configs start t := by
+  let hex : ∃ n, (tm.configs start n).state.isNone := ⟨t, by simp [hhalt]⟩
+  let u := Nat.find hex
+  have huBool : (tm.configs start u).state.isNone := Nat.find_spec hex
+  have huHalt : (tm.configs start u).state = none := by simpa using huBool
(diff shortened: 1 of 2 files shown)


Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
```diff
@@ -289,6 +289,26 @@ lemma configs_of_halts (cfg : Cfg k Symbol State input) (h : cfg.state = none) {
   | succ d ih =>
     rw [configs_succ_eq_step', ih, step_of_halt h]
 
+/-- Every padded halting computation has a least halting time, and the configuration there is the
+same as at the padded time. -/
+lemma exists_minimal_halting_config (tm : MultiTapeTM k Symbol State)
```
crei, line 294: I think this lemma could be reworked a little. It seems to be related to `haltsAtStep` and `halting_step_unique`, so should maybe move there. Also the conjunction is rather complicated, maybe it should be split?

Maybe we could also prove `Monotone fun t => (tm.configs cfg t).state.isNone`?

Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
```diff
@@ -401,6 +421,20 @@ lemma outputString_eq_of_halt
...
+    (cfg : Cfg k Symbol State input) (t : ℕ) :
+    (tm.outputString cfg t).length ≤ t := by
+  induction t with
+  | zero => simp [outputString]
+  | succ t ih =>
+      rw [outputString_succ]
+      have hone : (tm.outputSymbol (tm.configs cfg t)).toList.length ≤ 1 := by
+        cases tm.outputSymbol (tm.configs cfg t) <;> simp
+      simp only [List.length_append]
+      omega
```
crei, line 436: There does not seem to be a `lemma Option.toList_length (x : Option α) : x.toList.length ≤ 1` (which would he hone). But maybe this one suffices (I did not test it):

```suggestion
      grind [Option.toList]
```

Cslib/Computability/Machines/Turing/MultiTape/TapeLemmas.lean
```diff
@@ -132,6 +132,28 @@ lemma spaceUsed_linear (cfg : Cfg k Symbol State input) (t : ℕ) :
     _ ≤ ∑ i, (t + 1) := Finset.sum_le_sum (fun i _ => tm.spaceUsedByTape_le cfg t i)
     _ = k * t + k := by simp [Nat.mul_succ]
 
+/--
+The space used by one tape is monotone under inclusion of its visited head positions in the
```
crei, line 136: I think this lemma deserves a bit more explanation, in particular, I don't understand the meaning of `hpos`.

PR #819, opened 2026-08-19
refactor(MultiTapeTM): Put the output tape into the configuration

barni120400:
Two independent changes to the multi-tape Turing machine, one per commit.

The first puts the output tape into the configuration: `Cfg` gains an `output` field, `step` appends the symbol the transition emits, and `initCfg` starts it empty. The output of a run is then whatever is on its final configuration's tape. That makes `outputString` unnecessary, so it and its lemmas are removed: `outputString_succ` becomes `step_output`, `outputString_eq_of_halt` is restated on configurations, and `outputString_halt` and `outputString_add_eq_append` were output-flavoured restatements of the existing `configs_of_halts` and `configs_add`. `ComputesInTimeAndSpace` now reads the output off the configuration it already mentions.

The second renames `configs` to `runFrom`. It returns the single configuration reached after `t` steps rather than a sequence, so the plural name did not match its type, and it depends on the machine only through its step function. Pure rename, no statement or proof changed.

This is step 1 of the plan discussed [on Zulip](https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/Multi-tape.20non-deterministic.20Turing.20machines/near/617383375): with the output in the configuration, a `List` of configurations determines the output string, so a computation can be such a list and reuse the machinery in `RelatesInSteps.lean`.

Note that the configuration space is now unbounded, so counting reachable configurations needs to happen on the storage configuration rather than on `Cfg`.

Supersedes #818.

NB: claude was used heavily throughout.

### DIFF
--- a/Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
+++ b/Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
@@ -83,7 +83,7 @@ We define a number of structures and concepts related to multi-tape Turing machi
 There are two ways to talk about the behaviour of a multi-tape Turing machine, and they are
 proven to be equivalent.
 
-* `MultiTapeTM.configs`: a sequence of configurations by execution step
+* `MultiTapeTM.runFrom`: the configuration reached after a given number of execution steps
 * `RelatesInSteps tm.TransitionRelation cfg cfg' t`: a proof that `tm` transforms the configuration
     `cfg` into `cfg'` in exactly `t` steps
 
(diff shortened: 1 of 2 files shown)


Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
```diff
@@ -29,7 +29,8 @@ The multi-tape Turing machine uses a read-only input tape, `k` work tapes and a
 tape.
 The input head can move freely on the input, but any move attempt beyond one cell outside the input
 results in no movement.
-The transition function can optionally output one symbol, which models the write-only output tape.
+The transition function can optionally output one symbol, which is appended to the output tape held
+in the configuration, so the output of a run can be read off its final configuration.
```
crei: I think this can stay as it is. In this high-level design section we also don't mention how the other tapes are stored exactly.

Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
crei, line 152:
```suggestion
- the contents of the write-only output tape
-/
```

crei: Just noticed this - we don't use `BiTape` any more here:
```suggestion
is the blank tape symbol). Note that it is not required that `Symbol` or `State` are finite
```

crei (approved):

crei (approved):

PR #820, opened 2026-08-19
feat(MultiTapeTM): Nondeterministic multi-tape Turing machines

barni120400:
Adds nondeterministic multi-tape Turing machines: the transition function is replaced by a transition relation, and a computation path is the list of configurations it passes through together with a proof that they form a chain of steps from the initial configuration to the one it ends at, so `List.IsChainFromTo` carries the whole validity condition. `MultiTapeTM.toNTM` embeds the deterministic machine and preserves computation in bounded time and space.

Configurations and the effect of an action move to a shared `MultiTape/Configuration.lean`. No existing statement changes meaning.

NB: claude was used heavily throughout.

### DIFF
--- a/Cslib/Computability/Machines/Turing/MultiTape/Configuration.lean
+++ b/Cslib/Computability/Machines/Turing/MultiTape/Configuration.lean
@@ -1,0 +1,187 @@
+/-
+Copyright (c) 2026 Christian Reitwiessner. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Christian Reitwiessner, Aviv Bar Natan
+-/
+
+module
+
+public import Mathlib.Algebra.Order.BigOperators.Group.Finset
+public import Mathlib.Algebra.Order.Group.Abs
+public import Mathlib.Algebra.Order.Group.Int
+public import Mathlib.Data.Finset.Dedup
+public import Mathlib.Data.Finset.Max
+public import Mathlib.Data.Int.Interval
+public import Mathlib.Basic.Sign.Defs
+public import Cslib.Init
+
+/-!
+# Configurations of Multi-Tape Turing Machines
+
+Configurations of a multi-tape Turing machine with a read-only input tape, `k` work tapes and one
+write-only output tape, together with what a single transition does to one and the space measure
+read off a list of them.
+
+## Design
+
+Nothing here mentions a machine. A step is described in two parts: an `Action`, recording
+which way the input head moves, what is written and where the work heads move, which symbol is
+emitted and which state follows; and `Action.apply`, which carries it out on a
+configuration.
+
+The output tape is part of the configuration, so the string emitted along a run can be read off
+the configuration the run ends in.
+
+## Important Declarations
+
+* `Cfg`: the configuration: the internal state, the tape contents and head positions, and the
+    output tape
+* `Action`: what a machine does in one step
+* `Action.apply`: the effect of one action on a configuration
+* `Cfg.Halted`, `Cfg.init`: halting, and the configuration a machine starts in
+* `spaceUsedOfCfgs`: work tape cells touched along a list of configurations
+-/
+
+@[expose] public section
+
+namespace Turing
+
+variable {k : ℕ} {State Symbol : Type*} {input : List Symbol}
+
+/-- What a machine does in one step. -/
+structure Action (k : ℕ) (Symbol State : Type*) where
+  /-- The movement (attempt) of the input head. -/
+  inputTape : SignType
+  /-- Actions on the work tapes: optionally a symbol to write and the head movement. -/
+  workTapes : Fin k → (Option (Option Symbol)) × SignType
+  /-- An optional symbol to output. -/
+  output : Option Symbol
+  /-- The successor state or none to halt. -/
+  state : Option State
+
+/--
+The configurations of a Turing machine is relative to the input of the machine and consist of:
+- an `Option`al state (or none for the halting state),
+- the position of the input head (shifted by one),
+- the contents of the work tape,
+- the positions of the work tape heads,
+- the contents of the write-only output tape
+-/
+@[ext]
+structure Cfg (k : ℕ) (Symbol State : Type*) (input : List Symbol) where
+  /-- the state of the TM (or none for the halting state) -/
+  state : Option State
+  /-- the position of the input head, shifted by one -/
+  inputPos : Fin (input.length + 2)
+  /-- the work tapes -/
+  workTapes : Fin k → ℤ → Option Symbol
+  /-- the positions of the heads on the work tapes -/
+  workTapePos : Fin k → ℤ
+  /-- the contents of the write-only output tape -/
+  output : List Symbol
+deriving Inhabited
+
+/-- Attempt to move the input tape head.
+The machine can only read one empty cell outside of the input,
+any attempted movement beyond that results in no movement.
+
+The addition is performed in `ℤ` before clamping. Performing it in `Fin (n + 2)` would wrap an
+outward boundary move to the opposite end of the input. -/
+@[scoped grind =]
+def moveInputPos {n : ℕ} (pos : Fin (n + 2)) (m : SignType) : Fin (n + 2) :=
+  let p := ((pos.val : ℤ) + (m.cast : ℤ)).toNat
+  if h : p < n + 2 then ⟨p, h⟩ else ⟨n + 1, by omega⟩
+
+@[simp]
+lemma moveInputPos_zero {n : ℕ} (pos : Fin (n + 2)) :
+    moveInputPos pos 0 = pos := by
+  apply Fin.ext
+  simp [moveInputPos, pos.isLt]
+
+@[simp]
+lemma moveInputPos_leftBoundary {n : ℕ} :
+    moveInputPos (0 : Fin (n + 2)) (-1) = 0 := by
+  apply Fin.ext
+  simp [moveInputPos]
+
+@[simp]
+lemma moveInputPos_rightBoundary {n : ℕ} :
+    moveInputPos (⟨n + 1, by omega⟩ : Fin (n + 2)) 1 = ⟨n + 1, by omega⟩ := by
+  unfold moveInputPos
+  rw [dite_eq_right (by simp; omega)]
+
+/-- A left move away from the left input boundary decrements the native input position. -/
+lemma moveInputPos_neg_of_ne_left {n : ℕ} (p : Fin (n + 2)) (h : p ≠ 0) :
+    moveInputPos p .neg = ⟨p.val - 1, by have := p.isLt; omega⟩ := by
+  have hp : 0 < p.val := Nat.pos_of_ne_zero (fun hz => h (Fin.ext hz))
+  unfold moveInputPos
+  apply Fin.ext
+  rw [dite_eq_left] <;> simp <;> omega
+
+/-- A right move away from the right input boundary increments the native input position. -/
+lemma moveInputPos_pos_of_ne_right {n : ℕ} (p : Fin (n + 2)) (h : p.val ≠ n + 1) :
+    moveInputPos p .pos = ⟨p.val + 1, by have := p.isLt; omega⟩ := by
+  unfold moveInputPos
+  rw [dite_eq_left]
+  · apply Fin.ext
+    simp
+  · simp
+    omega
+
+/-- The symbol currently under the input tape head. -/
+def Cfg.inputSymbol (cfg : Cfg k Symbol State input) : Option Symbol :=
+  if h₁ : cfg.inputPos = 0 then none
+  else if h₂ : cfg.inputPos = input.length + 1 then none
+  else input[cfg.inputPos.val - 1]'(by grind)
+
+@[simp]
+lemma inputSymbolInner {cfg : Cfg k Symbol State input} (p : ℕ)
+    (h₁ : cfg.inputPos.val = 1 + p)
+    (h₂ : p < input.length) :
+    cfg.inputSymbol = some input[p] := by
+  grind [Cfg.inputSymbol]
+
+/-- The symbol read by work tape `i`. -/
+def Cfg.workTapeSymbols (cfg : Cfg k Symbol State input) (i : Fin k) : Option Symbol :=
+  cfg.workTapes i (cfg.workTapePos i)
+
+/-- A configuration is halted when it has no state to continue from. -/
+abbrev Cfg.Halted (cfg : Cfg k Symbol State input) : Prop := cfg.state = none
+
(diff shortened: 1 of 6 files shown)


Cslib/Computability/Machines/Turing/MultiTape/Nondeterministic.lean
```diff
@@ -0,0 +1,143 @@
...
+lemma isChain_toList_step (p : ntm.Computation input) :
+    p.toList.IsChain (ntm.Step (input := input)) :=
+  p.toRelSeries.isChain_toList
+
+end Computation
+
+/-- `ntm` has a computation on `input` that starts at the initial configuration, halts, emits
+`output` and satisfies `P`. The notions below are its instances, so their constraints all refer to
+a single computation. -/
+def ComputesSuchThat (ntm : MultiTapeNTM k Symbol State) (input output : List Symbol)
```
crei, line 118: This kind of hides the fact that P only has to hold for some computation path. Maybe this should be a bit more visible in the name?

Cslib/Computability/Machines/Turing/MultiTape/NonDeterministic.lean
```diff
@@ -0,0 +1,143 @@
...
+lemma isChain_toList_step (p : ntm.Computation input) :
+    p.toList.IsChain (ntm.Step (input := input)) :=
+  p.toRelSeries.isChain_toList
+
+end Computation
+
+/-- `ntm` has a computation on `input` that starts at the initial configuration, halts, emits
+`output` and satisfies `P`. The notions below are its instances, so their constraints all refer to
+a single computation. -/
+def ComputesSuchThat (ntm : MultiTapeNTM k Symbol State) (input output : List Symbol)
```
avivbarnatan-air: I did it on purpose, because by definition a computation of a non-deterministic machine is existance of a computation path.
So I didn't want to carry this existence in the name outside the "ComputesSuchThat".
I think the alternative is "ExistComputationPathSuchThat" but I think "ComputesSuchThat" is better

Cslib/Computability/Machines/Turing/MultiTape/Nondeterministic.lean
barni120400: I did it on purpose, because by definition a computation of a non-deterministic machine is existance of a computation path.
So I didn't want to carry this existence in the name outside the "ComputesSuchThat".
I think the alternative is "ExistComputationPathSuchThat" but I think "ComputesSuchThat" is better

Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
```diff
@@ -137,113 +120,17 @@ section Cfg
 /-!
 ## Configurations of a Turing Machine
```
crei: This header needs to be updated.

Cslib/Computability/Machines/Turing/MultiTape/Nondeterministic.lean
```diff
@@ -0,0 +1,143 @@
...
+module
+
+public import Mathlib.Data.List.Chain
+public import Mathlib.Order.RelSeries
+public import Cslib.Computability.Machines.Turing.MultiTape.Basic
+
+/-!
+# Nondeterministic Multi-Tape Turing Machines
+
+Defines nondeterministic Turing machines with a read-only input tape, `k` work tapes and one
```
crei, line 16: Should the file also be called `Nondeterministic.lean`?

Cslib/Computability/Machines/Turing/MultiTape/NonDeterministic.lean
```diff
@@ -0,0 +1,143 @@
...
+`Option Symbol` (where `none` is the blank symbol). Neither `Symbol` nor `State` is required to be
+finite.
+-/
+structure MultiTapeNTM (k : ℕ) (Symbol State : Type*) where
+  /-- initial state -/
+  q₀ : State
+  /-- transition relation: which transitions are permitted for a state, the current input symbol
+  and a tuple of work head symbols -/
+  Tr (q : State) (input : Option Symbol) (work : Fin k → Option Symbol) :
+    TransitionOut k Symbol State → Prop
```
crei: This could be
```lean4
 Tr (q : State) (input : Option Symbol) (work : Fin k → Option Symbol)
    (action : TransitionOut k Symbol State) : Prop
```
Also, maybe we could rename `TransitionOut` to `Action`.

Cslib/Computability/Machines/Turing/MultiTape/NonDeterministic.lean
```diff
@@ -0,0 +1,143 @@
...
+
+/--
+A nondeterministic multi-tape Turing machine with `k` work tapes over the alphabet of
+`Option Symbol` (where `none` is the blank symbol). Neither `Symbol` nor `State` is required to be
+finite.
+-/
+structure MultiTapeNTM (k : ℕ) (Symbol State : Type*) where
+  /-- initial state -/
+  q₀ : State
+  /-- transition relation: which transitions are permitted for a state, the current input symbol
```
crei: Maybe "which combinations of state, current input symbol, tuple of work head symbols and resulting actions are valid transitions"

Cslib/Computability/Machines/Turing/MultiTape/NonDeterministic.lean
```diff
@@ -0,0 +1,143 @@
...
+/-- The initial configuration corresponding to an input string. -/
+@[simp]
+def initCfg (ntm : MultiTapeNTM k Symbol State) (input : List Symbol) :
+    Cfg k Symbol State input :=
+  Cfg.init ntm.q₀ input
+
+/-- A computation of `ntm` on `input`: a series of configurations starting at the initial one, in
+which each is reached from the previous by a step. `RelSeries` supplies the series itself, so
+`length` is the number of steps taken, `toList` the configurations passed through, and `last` the
+one it ends at. -/
```
crei:
```suggestion
/-- A computation path of `ntm` on `input`: a series of configurations starting at the initial one, in
which each is reached from the previous by a step. `RelSeries` supplies the series itself, so
`length` is the number of steps taken, `toList` the configurations passed through, and `last` the
one it ends at.
Note that it is not required that the machine terminates at the final configuration of the path. -/
```

Cslib/Computability/Machines/Turing/MultiTape/NonDeterministic.lean
```diff
@@ -0,0 +1,143 @@
...
+
+variable {ntm : MultiTapeNTM k Symbol State} {input : List Symbol}
+
+/-- The number of work tape cells touched. -/
+def space (p : ntm.ComputationPath input) : ℕ := spaceUsedOfCfgs p.toList
+
+/-- Consecutive configurations of a computation are joined by a step. -/
+lemma isChain_toList_step (p : ntm.ComputationPath input) :
+    p.toList.IsChain (ntm.Step (input := input)) :=
+  p.toRelSeries.isChain_toList
```
crei: This also works, and because of that, I'm not sure if we actually need this lemma.
```suggestion
  p.isChain_toList
```

Cslib/Computability/Machines/Turing/MultiTape/Configuration.lean
```diff
@@ -0,0 +1,187 @@
...
+public import Mathlib.Algebra.Order.Group.Abs
+public import Mathlib.Algebra.Order.Group.Int
+public import Mathlib.Data.Finset.Dedup
+public import Mathlib.Data.Finset.Max
+public import Mathlib.Data.Int.Interval
+public import Mathlib.Data.Sign.Defs
+public import Cslib.Init
+
+/-!
+# Configurations of Multi-Tape Turing Machines
```
crei, line 19: Then maybe this file should be called `Configuration.lean`?

Cslib/Computability/Machines/Turing/MultiTape/Nondeterministic.lean
```diff
@@ -0,0 +1,145 @@
...
+    Cfg k Symbol State input :=
+  Cfg.init ntm.q₀ input
+
+/-- A computation path of `ntm` on `input`: the configurations it passes through, forming a chain
+of steps from the initial configuration to the one it ends at. -/
+structure ComputationPath (ntm : MultiTapeNTM k Symbol State) (input : List Symbol) where
+  /-- the configurations passed through, starting with the initial one -/
+  cfgs : List (Cfg k Symbol State input)
+  /-- a path visits at least one configuration -/
+  ne_nil : cfgs ≠ []
```
crei: OK this is a bit weird. `ne_nil` is already enforced by `isChainFromTo`, but you need it for `cfg.getLast ne_nil` below. Is there a nicer way to do it? Maybe
```lean4
start : cfgs.head? = some (ntm.initCfg input)
isChain : cfgs.IsChain ntm.Step
```

Cslib/Computability/Machines/Turing/MultiTape/Nondeterministic.lean
```diff
@@ -0,0 +1,145 @@
...
+def ComputesSuchThat (ntm : MultiTapeNTM k Symbol State) (input output : List Symbol)
+    (P : ntm.ComputationPath input → Prop) : Prop :=
+  ∃ p : ntm.ComputationPath input, p.last.Halted ∧ p.last.output = output ∧ P p
+
+/-- `ntm` computes `output` from `input`, with no bound on resources. -/
+def Computes (ntm : MultiTapeNTM k Symbol State) (input output : List Symbol) : Prop :=
+  ntm.ComputesSuchThat input output fun _ => True
+
+/-- `ntm` computes `output` from `input` in exactly `t` steps. -/
+def ComputesInTime (ntm : MultiTapeNTM k Symbol State) (input output : List Symbol) (t : ℕ) :
```
crei: I would prefer to call this `ComputesInExactTime` (same for space) in this case.

Cslib/Computability/Machines/Turing/MultiTape/Nondeterministic.lean
```diff
@@ -0,0 +1,145 @@
...
+
+variable {ntm : MultiTapeNTM k Symbol State}
+
+/-- The one-step relation on configurations. A halted configuration steps to itself; a running one
+steps by any permitted transition. -/
+@[scoped grind =]
+def Step (ntm : MultiTapeNTM k Symbol State) (c₁ c₂ : Cfg k Symbol State input) : Prop :=
+  match c₁.state with
+  | none => c₂ = c₁
+  | some q => ∃ out, ntm.Tr q c₁.inputSymbol c₁.workTapeSymbols out ∧ c₂ = out.apply c₁
```
crei:
```suggestion
  | some q => ∃ action, ntm.Tr q c₁.inputSymbol c₁.workTapeSymbols out ∧ c₂ = action.apply c₁
```

Cslib/Computability/Machines/Turing/MultiTape/Nondeterministic.lean
crei: Another way would be
```lean4
last : Cfg k Symbol State input
isChainFromTo : cfgs.IsChainFromTo ntm.Step (ntm.initCfg input) last
```
then you don't need `def last` below.

barni120400: I prefer the second suggestion. It's still a bit weird that we duplicate the last element - it's written both at "last" and it's the last element in the chain.

Cslib/Computability/Machines/Turing/MultiTape/Nondeterministic.lean
barni120400: should I add a "ComputesInTime" which is computing in `<= t` time and prove that it's the same because the halting configuration can be iterated up to time `t`?

crei: I mean it sounds like a good idea but on the other hand, we already have all these definitions here that we need to unfold in proofs. Instead of `<= t`, we could already add a linear factor, but maybe we should wait with the definitions until the are used?

Cslib/Computability/Machines/Turing/MultiTape/Nondeterministic.lean
crei: Yet more options:
```lean4
structure ComputationPath (ntm : MultiTapeNTM k Symbol State) (input : List Symbol) where
  /-- the configurations passed through after the initial one -/
  tail : List (Cfg k Symbol State input)
  isChain : ((ntm.initCfg input) :: tail).IsChain ntm.Step

def ComputationPath.last (p : ntm.ComputationPath input) : Cfg k Symbol State input :=
    ((ntm.initCfg input) :: tail).getLast sorry
```
or just
```lean4
def _root_.List.IsComputationPathTo {ntm : MultiTapeNTM k Symbol State} {input : List Symbol}
    (cfgs : List (Cfg k Symbol State input)) (last : Cfg k Symbol State input) :=
  List.IsChainFromTo ntm.Step (ntm.initCfg input) last
```

crei (approved):

Cslib/Computability/Machines/Turing/MultiTape/Configuration.lean
```diff
@@ -0,0 +1,187 @@
...
+def Cfg.init (q₀ : State) (input : List Symbol) : Cfg k Symbol State input :=
+  ⟨some q₀, 1, fun _ _ => none, fun _ => 0, []⟩
+
+/--
+The effect of an action on a configuration: move the input head, write and move on the work tapes,
+append the emitted symbol to the output tape, and go to the successor state. This is the part of a
+step that does not depend on how the action was chosen.
+-/
+@[simp]
+def Action.apply (out : Action k Symbol State) (cfg : Cfg k Symbol State input) :
```
crei:
```suggestion
def Action.apply (action : Action k Symbol State) (cfg : Cfg k Symbol State input) :
```

Cslib/Computability/Machines/Turing/MultiTape/Configuration.lean
```diff
@@ -0,0 +1,187 @@
...
+  state := out.q'
+  inputPos := moveInputPos cfg.inputPos out.inputMove
+  workTapes i := match (out.workActions i).1 with
+    | none => cfg.workTapes i
+    | some s => Function.update (cfg.workTapes i) (cfg.workTapePos i) s
+  workTapePos i := cfg.workTapePos i + (out.workActions i).2
+  output := cfg.output ++ out.outS.toList
+
+/-- A work tape head moves by at most one cell when an action is applied. -/
+lemma workTapePos_apply_le (out : Action k Symbol State)
```
crei:
```suggestion
lemma workTapePos_apply_le (action : Action k Symbol State)
```

crei (approved):

PR #821, opened 2026-08-19
refactor(MultiTapeTM): derive deterministic machines from relations

barni120400:
`MultiTapeTM` is now a `MultiTapeNTM` whose transition relation permits exactly one action per state and tuple of symbols. Configurations, `Step`, paths and computation predicates are inherited directly, eliminating the deterministic-to-nondeterministic embedding and its translation proofs.

The derived `tr` and `step` functions retain the evaluation API, with graph lemmas connecting them to the shared relations. They use classical choice and remain noncomputable. `runFrom` uses Mathlib's iteration API; `run` packages its infinite execution, and `runPath` is an abbreviation for a finite prefix.

Halting, tape-space, storage and configuration-counting results are proved for nondeterministic paths or chosen runs. Deterministic corollaries and computation-predicate equivalences are removed: callers use the shared results directly or supply `runPath` as a computation witness. Whole-run bounds concern one chosen run with coherent prefixes. Tape-word helpers live in the machine-independent configuration module.

`TransformsTapes` and sequential-composition correctness now apply to nondeterministic machines. Deterministic `seq` adds only a uniqueness proof to the same composition relation; it is computable, and its correctness follows directly from the shared theorem. The almost-constant combinator and regression tests use this shared API. Single-tape machines are unchanged.

Regression examples cover relation-only deterministic construction, direct reuse of resource/counting/composition theorems, genuine branching paths, padded computations, and stuck nondeterministic machines.

Validation after rebasing onto `main` at `133d92d` (Mathlib `1cae91f`):

- `lake build --wfail --iofail`
- `lake exe mk_all --check`
- `lake test --wfail --iofail`
- `lake lint`
- `lake exe lint-style`
- `lake exe checkInitImports`
- `git diff --check`

Codex was used to investigate the design, implement the refactor, update proofs, and run validation.


Cslib/Computability/Machines/Turing/MultiTape/Configuration.lean
crei, line 82: Can you also move the references?

Cslib/Computability/Machines/Turing/MultiTape/Nondeterministic.lean
crei, line 23: I think we need to clean up the doc strings here a bit more, and maybe also the namespaces.

We should say here specifically that a deterministic TM is a special case of a nondeterministic TM. It's also a bit weird that the main design description is in `Configuration.lean` - or is it?

Now about the namespaces: I don't yet know a good solution to this, but I have a feeling that putting the definitions from `AlmostConstant.lean` in the "deterministic" namespace `MultiTapeTM` does not sound quite right. All the code here is about multi-tape Turing machines, and some of them are determinsitic, some are nondeterministic. The same about all future combinators: I guess they will almost all apply to deterministic computation models alone, but I'm not sure if they should be in the namespace of the deterministic machine. So maybe put all of that in the Turing namespace?

Cslib/Computability/Machines/Turing/MultiTape/Deterministic.lean
```diff
@@ -106,19 +71,62 @@ is the blank tape symbol). Note that it is not required that `Symbol` or `State`
...
+theorem ext {tm tm' : MultiTapeTM k Symbol State}
+    (h : tm.toMultiTapeNTM = tm'.toMultiTapeNTM) : tm = tm' := by
+  cases tm
+  cases tm'
+  cases h
+  rfl
+
+/-- The unique action permitted by the transition relation. This derived function uses classical
+choice; `tr_ofTr` recovers a supplied transition function by simplification. -/
+noncomputable def tr (tm : MultiTapeTM k Symbol State) (q : State) (input : Option Symbol)
```
crei, line 98: Could we make the name a bit more descriptive? The point here is that it is a function ad not a relation, right?

barni120400: yeah exactly. Do you think changing "tr" is a good idea? All the repo usages in "tr" will have to change. And if so, then what name? I thought about trFunction but isn't it too explicit? Another suggestion for a name is transitionFn

Cslib/Computability/Machines/Turing/MultiTape/Nondeterministic.lean
barni120400: I think that the desicions about the input, output and work tapes should be in the onfiguration file as these decisions are implemented there.

PR #822, opened 2026-08-20
feat: Terminating FullEta

lengyijun:
Adds a size measure on untyped λ-terms and uses it to prove that  η-reduction is terminating


Cslib/Languages/LambdaCalculus/LocallyNameless/Untyped/Size.lean
```diff
@@ -0,0 +1,40 @@
...
+namespace Cslib
+
+namespace LambdaCalculus.LocallyNameless.Untyped.Term
+
+universe u
+
+variable {Var : Type u}
+
+/-- Computes the size of a lambda calculus term. -/
+@[simp, grind]
```
lengyijun: not sure `grind` or `grind =` or `grind .`

PR #826, opened 2026-08-20
chore: Fix linter warnings in docstrings

crei:
Fixes linter warnings by escaping characters special for verso.


chenson2018 (approved):
Thanks!

PR #828, opened 2026-08-20
chore: Fix linter by merging with grind tactics

crei:
Merging some tactics with subsequent `grind`s as suggested by the linter.

Verified that the compilation time does not increase substantially.


chenson2018 (approved):
Thanks!

PR #839, opened 2026-08-27
feat(Language): define language homomorphism and prove that regular languages are closed through preimage under  language homomorphisms

ctchou:
This PR introduces the notion of language homomorphisms, which map finite words to finite words and preserves the empty word and concatenation.  Under a language homomorphism `f`, an LTS `lts` can be converted to another LTS `lts.mapHom f` with exact the same states but whose step on a single label `x` is the composition of the steps that `lts` takes on `f [x]`, which may be empty or consists of many labels.  This new LTS is well-defined because `f` is a language homomorphism.  The same construction is then applied to NA.FinAcc and shown that it corresponds to the preimage operation under `f` on languages.  As a consequence, regular languages are closed through preimage operations under language homomorphisms.

The image operation on languages is also defined.  Future work will show that regular languages are also closed through image operations under language homomorphisms.


fmontesi (approved):
Very nice! Thanks for keeping `mapLabel` as API compatibility.

PR #840, opened 2026-08-27
feat(ModalLogic/LambdaCalculus): modal grind set and proof of type safety for simply typed lambda calculus

fmontesi:
This PR:
- Introduces a `modal` grind set for simple modal reasoning. The set does not unfold the meaning of the modal connectives on purpose: it's supposed to be used for quick modal proof search.
- Shows how to use CSLib's modal logic to reason about the PL adage 'safety = progress + preservation'. Once progress and preservation are lifted from existing results into modal propositions, safety follows. The PR also shows how to extract the 'pure' type safety Lean statement.
- Adds a few general properties about relations that were convenient in this development.
- Removes some duplication by aligning uses of 'flip' in Relation and 'Terminating' in modal logic (instead of WellFounded on flip, now it just uses Terminating).


Cslib/Foundations/Relation/Defs.lean
```diff
@@ -90,15 +90,15 @@ abbrev Normalizing (r : α → α → Prop) : Prop :=
 
 /-- An element `x` is `SN` (for strongly-normalising) for a relation `r` if it is accessible under
 the inverse of `r`. -/
-abbrev SN (r : α → α → Prop) := Acc (fun a b => r b a)
+abbrev SN (r : α → α → Prop) := Acc (flip r)
```
chenson2018: There was pretty strong consensus [here](https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/definitions.20for.20termination.20and.20strong.20normalization/with/593370737) for the `fun a b => r b a` definition. Do you have a strong reason to change this? My reasoning is that it is very annoying for transparency reasons and because there is a split upstream between `flip` and `swap`.

Cslib/Foundations/Relation/Normal.lean
```diff
@@ -0,0 +1,27 @@
...
+public import Cslib.Foundations.Relation.Defs
+
+/-! # Relations: properties of normalisation -/
+
+@[expose] public section
+
+namespace Relation
+
+/-- Strong normalisation implies reachability of a normal form. -/
+theorem sn_reaches_normal (h : SN r a) : ∃ b, ReflTransGen r a b ∧ Normal r b := by
```
chenson2018: This theorem exists already: `SN.normalizable` in `Cslib.Foundations.Relation.Confluence`

Cslib/Logics/Modal/Attr.lean
```diff
@@ -0,0 +1,26 @@
+/-
+Copyright (c) 2026 Fabrizio Montesi. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Fabrizio Montesi
+-/
+
+module
+
+public import Cslib.Init
```
chenson2018: Can you make this follow a similar structure as https://github.com/leanprover-community/mathlib4/pull/39370/? To be clear, the parts I care about immediately are:
- Make a `Cslib.Tactic.GrindAttrs`, where all future `register_grind_attr` will go as well
- Add an `import Cslib.Tactic.GrindAttrs` to `Cslib.Init`, so that the attribute is available everywhere

(I suggest adding the `Tactic` directory now because I think we'll want this in the near future)

Cslib/Logics/Modal/Basic.lean
```diff
@@ -221,12 +222,16 @@ instance (r : World → World → Prop) : InferenceSystem (Axiom r) (Proposition
 theorem Satisfies.axiom_def (r : World → World → Prop) :
     (∀ v w, ⇓Modal[⟨r,v⟩,w ⊨ φ]) ↔ Axiom r⇓φ := by rfl
 
+@[modal .]
```
chenson2018, line 224: Most of the new attributes you're adding are exclusively tagged with the `modal` attribute and not `grind`. Does this mean you could remove `grind` attributes from some other theorems in this file?

Cslib/Logics/Modal/Basic.lean
```diff
@@ -348,6 +360,13 @@ theorem Satisfies.l (r : World → World → Prop) [IsTrans World r]
   apply ih _ hw'w''
   exact IsTrans.trans _ _ _ hww' hw'w''
 
+/-- Löb induction, via the L axiom. -/
+theorem Satisfies.l_induction (m : Model World Atom) [IsTrans World m.r]
+    (hwf : Relation.Terminating m.r) (hstep : ∀ w, ⇓Modal[m,w ⊨ □φ → φ]) (w : World) :
+    ⇓Modal[m, w ⊨ φ] := by
+  have hl := Satisfies.of_axiom m _ (Satisfies.l m.r hwf φ) w
+  grind only [modal, = box_iff_forall]
```
chenson2018, line 369: Why is this squeezed? Performance?

CslibTests/Modal/Stlc.lean
chenson2018, line 1: Why is this file a test? Because of it being a different approach to an existing theorem?

chenson2018 (changes requested):

Cslib/Foundations/Relation/Defs.lean
fmontesi: I did it for consistency: we use flip in other places. Plus, it didn't really bother me. But I don't have a very strong opinion. It was a bit confusing to me  that we're inconsistent in how we deal with flipping (e.g., we do use it in simulations and bisimulation, though I did that basically because it's in Lean core), but maybe there's something I'm not seeing?

Cslib/Foundations/Relation/Normal.lean
fmontesi: Ah, I'll remove the file then. But why is that result in a file called 'Confluence'? (This is irrelevant to this PR but I'd like to understand.)

Cslib/Logics/Modal/Attr.lean
fmontesi: Sounds reasonable. Done.

Cslib/Logics/Modal/Basic.lean
fmontesi: The opposite, I think I missed one scoped grind I should've added. :-)

But more precisely, I'm annotating some theorems only with modal as a conservative measure, because I think they risk giving grind too many options when considered together with all other theorems `scoped grind` has access to. The modal set is restricted to modal reasoning only, instead.

Cslib/Logics/Modal/Basic.lean
fmontesi: Yes, but also as a memo that `modal` should be able to prove this.

CslibTests/Modal/Stlc.lean
fmontesi: Yes, even though `type_safety` is, strictly speaking, new.

Cslib/Foundations/Relation/Defs.lean
chenson2018: I do feel strongly here for the reasons I mention above. Could we revert for now? It should be a separate PR regardless.

Cslib/Logics/Modal/Basic.lean
chenson2018: Can you leave a short comment? Someone (i.e. me!) is liable to forget and revert this otherwise.

Cslib/Foundations/Relation/Normal.lean
chenson2018: Just a historical naming quirk. I first proved some confluence results and then others added more. When I split these files I was trying to not make it too granular dependency wise.

Cslib/Foundations/Relation/Defs.lean
fmontesi: Of course!

Cslib/Logics/Modal/Basic.lean
fmontesi: Yes, good idea. Done.

chenson2018 (approved):
I cleaned up some changes that were still left over from usage of `flip`, otherwise this looks good. Thanks!

PR #841, opened 2026-08-27
feat(Computability): add generic circuit model

SamuelSchlesinger:
Adds a generic circuit model we can use for arithmetic and Boolean circuit lower/upper bounds. I have experimented with this definition for upper/lower bounds and it is quite usable, I've got some lower bounds I can PR once this is accepted.

If you want to see how it can be used, you can see: https://github.com/samuelSchlesinger/algebraic-circuits.

Used Codex with GPT 5.6 Sol to refactor this for CSLib, but the underlying definitions and many of the theorems/lemmata were originally authored by me. Some of the examples and documentation bits were added or edited by Codex as well.


Cslib/Computability/Circuits/Circuit.lean
```diff
@@ -0,0 +1,103 @@
...
+`Circuit.FanInAtMost`. Evaluation commutes with homomorphisms
+(`Circuit.map_eval`).
+-/
+
+@[expose] public section
+
+namespace Cslib.Circuits
+
+/-- A straight-line program with designated output wires. -/
+structure Circuit (σ : Signature) (n g m : Nat) where
```
BoltonBailey: Is it necessary to have the circuit be parametric over the number of internal gates? This seems a bit unintuitive to me, I usually think of types representing computations as having performance parameters like gate count as methods.

SamuelSchlesinger: It is not necessary, but it works well for the inductive definition. Basically, every gate corresponds to a line in a straightline program. If you want to have `SomeCircuit` which abstracts how many internal gates and has a function `size` you can wrap this with that.

In particular, carrying `g` around allows you to know exactly which wires you can refer to in this line.

Cslib/Computability/Circuits/Signature.lean
```diff
@@ -0,0 +1,30 @@
...
+Authors: Samuel Schlesinger
+-/
+module
+
+public import Cslib.Init
+
+/-!
+# Signatures
+
+A signature is a collection of finitary operation symbols, each with a fixed
```
crei: The finitariness (?) is not enforced by the definition, is it?

Cslib/Computability/Circuits/Signature.lean
```diff
@@ -0,0 +1,30 @@
...
+@[expose] public section
+
+namespace Cslib.Circuits
+
+/-- A collection of finitary operation symbols and their arities. -/
+structure Signature where
+  /-- The operation symbols of the signature. -/
+  Op : Type v
+  /-- The number of arguments taken by each operation symbol. -/
+  Arity : (op : Op) → Nat
```
crei: I would say either
```suggestion
  Arity (op : Op) : \N
```
or
```suggestion
  Arity : Op \to \N
```

Cslib/Computability/Circuits/Interpretation.lean
```diff
@@ -0,0 +1,28 @@
+/-
+Copyright (c) 2026 Samuel Schlesinger. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Samuel Schlesinger
+-/
+module
+
+public import Cslib.Computability.Circuits.Signature
```
crei: Maybe Signature and Interpretation should be a single file? It's quite a long import chain you have to follow from Circuit.

Cslib/Computability/Circuits/Interpretation.lean
```diff
@@ -0,0 +1,28 @@
...
+than a structure, so they can be built pointwise and specialized without any
+wrapping.
+-/
+
+@[expose] public section
+
+namespace Cslib.Circuits
+
+/-- An interpretation assigns an operation on `Carrier` to every symbol in `σ`. -/
+abbrev Interpretation (σ : Signature) Carrier :=
```
crei:
```suggestion
abbrev Interpretation (σ : Signature) (Carrier: Type*) :=
```
?

crei: What is the idea behind the term "Carrier"? Shouldn't it be "Value" or something?

Cslib/Computability/Circuits/Signature.lean
SamuelSchlesinger: Oh yeah, we don't need `op` here so I'll just do `Op \r \N`.

Cslib/Computability/Circuits/Signature.lean
SamuelSchlesinger: Nope, I don't think it should though, I'll just remove this text.

Cslib/Computability/Circuits/Program.lean
```diff
@@ -0,0 +1,553 @@
...
+    (line.mapWires wireMap).op = line.op := rfl
+
+@[simp] theorem Line.mapWires_wires
+    (line : Line σ n g)
+    (wireMap : Wire n g → Wire n' h)
+    (argument : Fin (σ.Arity line.op)) :
+    (line.mapWires wireMap).wires argument = wireMap (line.wires argument) := rfl
+
+/-- A topologically ordered straight-line program of `g` gates. -/
+inductive Program (σ : Signature) (n : Nat) : Nat → Type v where
```
crei: What about spending some more characters in general to avoid confusion?
```suggestion
inductive Program (σ : Signature) (inputCnt : Nat) : Nat → Type v where
```

crei (commented):
I haven't read everything yet. It might make sense to split up Program a bit, it is already 500 lines.

Cslib/Computability/Circuits/Circuit.lean
BoltonBailey: Right, I guess I was finding this challenging to use. To try it out I wanted to write a circuit to implement Strassen matrix multiplication, but I got stuck when I realized I didn't know the exact recursive formula for the number of gates in the circuit in advance.

Cslib/Computability/Circuits/Program.lean
SamuelSchlesinger: `n` is the blessed number for input lengths!

Cslib/Computability/Circuits/Circuit.lean
BoltonBailey: Trying this bundled model you suggested seems to work fine though, I don't see any reason to hold up the PR on this account.

SamuelSchlesinger: That is a fair point. Maybe there should be a circuit builder interface for building them.

Cslib/Computability/Circuits/Interpretation.lean
BoltonBailey: I think "carrier" is a somewhat common term for the type on which a collection of operations act ([for example](https://en.wikipedia.org/wiki/Algebraic_structure#:~:text=underlying%20set%2C-,carrier%20set,-or%20domain))

Cslib/Computability/Circuit/Program.lean
```diff
@@ -0,0 +1,553 @@
...
+  (h : Homomorphism i₁ i₂)
+  (x : Fin n → U₁) :
+  h.map ∘ p.trace i₁ x = p.trace i₂ (h.map ∘ x) := by
+  funext wire
+  refine Fin.addCases (fun input => ?_) (fun gate => ?_) wire
+  · simp [Program.trace, Function.comp_apply]
+  · simpa [Program.trace, Function.comp_apply] using congrFun (p.map_eval h x) gate
+
+/-- The scalar function computed by an internal gate. -/
+def Program.gateFunction
```
BoltonBailey, line 284: Is there a reason these can't have the `input` argument before the colon?

BoltonBailey (commented):
This is exciting, I would definitely be interested in a Circuit model like this.

Cslib/Computability/Circuit/Program.lean
SamuelSchlesinger: They certainly could!

Cslib/Computability/Circuits/Interpretation.lean
SamuelSchlesinger: That is reasonable.

BoltonBailey (approved):

crei (approved):

PR #846, opened 2026-08-29
feat(Language): prove some distributive laws for the image operation on languages

ctchou:
This PR proves the distributive laws of the image operation over most operations on languages.  Some of these laws work for any function, while others (namely, those related to concatenation) require the function to be a language homomorphism.  The distributive laws do not work for language intersection and complementation.

Once it is proved that every regular language is expressible as a regular expression, these results will enable us to show that the image of a regular language under a language homomorphism is still regular.


Cslib/Computability/Languages/LanguageHom.lean
```diff
@@ -107,18 +109,53 @@ def image (f : List α → List β) (l : Language α) : Language β :=
...
 
 @[simp, scoped grind =]
-theorem mem_preimage (l : Language β) (as : List α) :
+theorem mem_preimage (f : List α → List β) (l : Language β) (as : List α) :
     as ∈ l.preimage f ↔ f as ∈ l := by
   rfl
 
+theorem image_singleton (f : List α → List β) (as : List α) :
+    ({as} : Language α).image f = {f as} := by
+  apply Set.image_singleton
```
eric-wieser: nit:
```suggestion
    ({as} : Language α).image f = {f as} :=
  Set.image_singleton _ _
```
etc is a few characters shorter, and probably negligibly faster. (and throughout)

Cslib/Computability/Languages/LanguageHom.lean
```diff
@@ -107,18 +109,53 @@ def image (f : List α → List β) (l : Language α) : Language β :=
...
+
+theorem image_zero (f : List α → List β) : (0 : Language α).image f = 0 := by
```
eric-wieser:
```suggestion
@[simp]
theorem image_zero (f : List α → List β) : (0 : Language α).image f = 0 := by
```

Cslib/Computability/Languages/LanguageHom.lean
```diff
@@ -107,18 +109,53 @@ def image (f : List α → List β) (l : Language α) : Language β :=
...
+theorem image_mul (f : Hom α β) (l1 l2 : Language α) :
+    (l1 * l2).image f = l1.image f * l2.image f := by
+  ext bs
+  simp [mem_mul]
+
+theorem image_pow (f : Hom α β) (l : Language α) (n : ℕ) :
+    (l^n).image f = (l.image f)^n := by
+  induction n with
+  | zero => simp [image_one]
+  | succ n h_ind => simp [pow_succ, image_mul, h_ind]
```
eric-wieser, line 154: If you prove the analog of [SetSemiring.imageHom](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Set/Semiring.html#SetSemiring.imageHom) first, then `image_pow` is just `map_pow`.

eric-wieser (approved):

Cslib/Computability/Languages/LanguageHom.lean
ctchou: All fixed.

Cslib/Computability/Languages/LanguageHom.lean
ctchou: Fixed.

Cslib/Computability/Languages/LanguageHom.lean
ctchou: Very true. In fact, these theorems are basically the components of the proof of `SetSemiring.imageHom`.  But I think the proper way to proceed is to first change `Language` and `ωLanguage` to use `SetSemiring` in future PRs.  (I can volunteer to do the latter the first.). Another issue is how to express the "monoid"-ness of `List`, which is a prerequisite to state the assumption `f : List α →* List β`.  I'm working on a future PR which will touch on these issues.

PR #849, opened 2026-09-01
chore(LTS/MapHom): golf mapHom_mTr proof

jessealama:
Drop an `obtain` that the following `grind` does not need, as flagged by [the weekly `mergeWithGrind` report](https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/Weekly.20linting.20log/near/620201866).


chenson2018 (approved):
I'll assume the performance was already checked considering the other PR. Thanks!

PR #854, opened 2026-09-02
feat(MultiTapeTM): Any function constant with finitely many exceptions is computable in constant time and space

crei:
This is a starting point of a Turing machine combinator library, it adds one of the leaves:

Any function that is constant with finitely many exceptions is computable in constant time and zero space, relative to any encoding. The same holds for any function with a finite domain.

This result captures many functions we want to compose later or with a combinator library: Any function on tuples of `Bool`, for example and "equality comparison with a constant". Together with a "fold" and "composition" combinators, this already allows us to evaluate CNFs or compute `Nat.succ` on binary encoded numbers.


Cslib/Computability/Machines/Turing/MultiTape/Combinators/Finite.lean
```diff
@@ -0,0 +1,245 @@
...
+/-! ## The machine computing a function between finite types -/
+
+variable {α β : Type*} [Finite α] {encIn : α ↪ List Bool} {encOut : β ↪ List Bool} {f : α → β}
+
+/-- The prefixes of the encodings of the elements of `α`, together with the empty list.
+
+The empty list has to be added explicitly for the case where `α` is empty: `finiteFunTM` uses
+the elements of this set as its states while reading the input, so the set has to contain the
+starting state `[]` even if there is no input to read. -/
+noncomputable def encPrefixes (encIn : α ↪ List Bool) : Finset (List Bool) :=
```
eric-wieser: This one seems a bit of a shame to make noncomputable;

Cslib/Computability/Machines/Turing/MultiTape/Combinators/AlmostConstant.lean
```diff
@@ -0,0 +1,391 @@
...
+        workTapes := fun _ _ => none,
+        workTapePos := fun _ => 0,
+        output := o ++ w } := by
+  rw [runFrom_succ_eq_step', runFrom_write pos o hw le_rfl]
+  simp only [step, almostConstTM, List.drop_length, reduceIte, List.head?_nil, moveInputPos_zero,
+    Option.toList_none, List.append_nil, List.take_length]
+  exact Cfg.ext_zero_tapes rfl rfl (by simp)
+
+/-- A constant time bound for the machine `almostConstTM`. -/
+public def almostConstTime (encIn : α ↪ List Bool) (encOut : β ↪ List Bool) (f : α → β)
```
SamuelSchlesinger, line 247: Should there be an @[expose] here? Same with `finiteFunTime`.

CslibTests/Complexity/Combinators.lean
```diff
@@ -0,0 +1,47 @@
...
+open Cslib Turing MultiTapeTM
+
+/-- The Boolean `and` function is computable in constant time and zero space. -/
+example : ∀ encIn encOut, ∃ c, EncodedComputableInTimeAndSpace
+    (encIn := encIn)
+    (encOut := encOut)
+    (Function.uncurry Bool.and)
+    (fun _ => c) (fun _ => 0) := by
+  intro encIn encOut
+  apply encodedComputableInTimeAndSpace_of_finite
```
SamuelSchlesinger: Function.uncurry fullAdder leaves the output as a function of carry. Could we use fun (a, b, carry) => fullAdder a b carry so the example computes the sum/carry pair from all three input bits?

crei: Ah right, of course!

SamuelSchlesinger (approved):

Cslib/Computability/Machines/Turing/MultiTape/Configuration.lean
```diff
@@ -81,6 +81,14 @@ structure Cfg (k : ℕ) (Symbol State : Type*) (input : List Symbol) where
   output : List Symbol
 deriving Inhabited
 
+/-- Two configurations of a machine without work tapes are equal if their states, input head
```
crei, line 84: Another pull request moved this part of `Deterministic.lean` into its own file, so this lemma is also just moved.

PR #855, opened 2026-09-02
chore(Foundations/Relation/Confluence): cleanup

thomaskwaring:
Several proofs in `Foundations/Relations/Confluence` can be vastly simplified using results from Mathlib — possibly some results could be deleted entirely, though evidently the Mathlib equivalents are not particularly discoverable. I have also fixed some naming conventions, with deprecation labels for the old names.

### DIFF
--- a/Cslib/Foundations/Relation/Confluence.lean
+++ b/Cslib/Foundations/Relation/Confluence.lean
@@ -352,7 +352,7 @@ lemma Commute.join_left (c₁ : Commute r₁ r₃) (c₂ : Commute r₂ r₃) : 
 
 theorem Commute.join_confluent (c₁ : Confluent r₁) (c₂ : Confluent r₂) (comm : Commute r₁ r₂) :
     Confluent (r₁ ⊔ r₂) := by
-  rw [←Commute.toConfluent]
+  rw [← Commute.toConfluent]
   apply_rules [join_left, symm]
 
 /-- If a relation is squeezed by a relation and its multi-step closure, they are multi-step equal -/
--- a/Cslib/Foundations/Relation/Confluence.lean
+++ b/Cslib/Foundations/Relation/Confluence.lean
@@ -40,17 +40,17 @@ namespace Relation
 
 attribute [scoped grind] ReflGen TransGen ReflTransGen EqvGen
 
-theorem ReflGen.to_eqvGen (h : ReflGen r a b) : EqvGen r a b := by
-  induction h <;> grind
+theorem ReflGen.to_eqvGen (h : ReflGen r a b) : EqvGen r a b :=
+  EqvGen.reflGen_le_eqvGen r _ _ h
 
-theorem TransGen.to_eqvGen (h : TransGen r a b) : EqvGen r a b := by
-  induction h <;> grind
+theorem TransGen.to_eqvGen (h : TransGen r a b) : EqvGen r a b :=
+  EqvGen.transGen_le_eqvGen r _ _ h
 
-theorem ReflTransGen.to_eqvGen (h : ReflTransGen r a b) : EqvGen r a b := by
-  induction h <;> grind
+theorem ReflTransGen.to_eqvGen (h : ReflTransGen r a b) : EqvGen r a b :=
+  EqvGen.reflTransGen_le_eqvGen r _ _ h
 
-theorem SymmGen.to_eqvGen (h : SymmGen r a b) : EqvGen r a b := by
-  induction h <;> grind
+theorem SymmGen.to_eqvGen (h : SymmGen r a b) : EqvGen r a b :=
+  EqvGen.symmGen_le_eqvGen r _ _ h
 
 attribute [scoped grind →] ReflGen.to_eqvGen TransGen.to_eqvGen ReflTransGen.to_eqvGen
   SymmGen.to_eqvGen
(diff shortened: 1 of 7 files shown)


Cslib/Foundations/Relation/Confluence.lean
```diff
@@ -352,45 +352,20 @@ lemma Commute.join_left (c₁ : Commute r₁ r₃) (c₂ : Commute r₂ r₃) :
...
     ReflTransGen r₁ = ReflTransGen r₂ := by
   ext a b
   exact ⟨ReflTransGen.mono h₁ a b, reflTransGen_closed h₂ a b⟩
 
-lemma ReflGen.symmGen_symm : ReflGen (SymmGen r) a b → ReflGen (SymmGen r) b a
-| .refl => .refl
-| .single (.inl h) => .single (.inr h)
-| .single (.inr h) => .single (.inl h)
+lemma ReflGen.symmGen_symm : ReflGen (SymmGen r) a b → ReflGen (SymmGen r) b a :=
+  Std.Symm.symm a b
```
eric-wieser, line 422: This one seems worth deprecating

Cslib/Foundations/Relation/Confluence.lean
```diff
@@ -352,45 +352,20 @@ lemma Commute.join_left (c₁ : Commute r₁ r₃) (c₂ : Commute r₂ r₃) :
...
-      | inr h => exact ih.trans _ _ _ (.symm _ _ (.rel _ _ h))
-  · intro h
-    induction h with
-    | rel _ _ ih => exact .single (.inl ih)
-    | refl x => exact .refl
-    | symm x y eq ih =>
-      rw [symmGen_swap]
-      exact reflTransGen_swap.mp ih
-    | trans _ _ _ _ _ ih₁ ih₂ => exact ih₁.trans ih₂
+theorem reflTransGen_symmGen : ReflTransGen (SymmGen r) = EqvGen r := EqvGen.reflTransGen_symmGen r
```
eric-wieser, line 425: This one looks like it might have been named incorrectly in mathlib?

Cslib/Foundations/Relation/Confluence.lean
```diff
@@ -352,45 +352,20 @@ lemma Commute.join_left (c₁ : Commute r₁ r₃) (c₂ : Commute r₂ r₃) :
...
     Confluent (r₁ ⊔ r₂) := by
-  intro a b c ab ac
-  induction ab generalizing c with
-  | refl => exists c
-  | @tail x y ax xy ih =>
-    have h_comm : Commute (r₁ ⊔ r₂) (r₁ ⊔ r₂) := by apply_rules [join_left, symm]
-    obtain ⟨z, xz, cz⟩ := ih ac
-    obtain ⟨w, yw, zw⟩ := h_comm (.single xy) xz
-    exact ⟨w, yw, cz.trans zw⟩
+  rw [←Commute.toConfluent]
```
eric-wieser:
```suggestion
  rw [← Commute.toConfluent]
```

Cslib/Foundations/Relation/Confluence.lean
```diff
@@ -40,17 +40,17 @@ namespace Relation
...
 
-theorem ReflTransGen.to_eqvGen (h : ReflTransGen r a b) : EqvGen r a b := by
-  induction h <;> grind
+theorem ReflTransGen.to_eqvGen (h : ReflTransGen r a b) : EqvGen r a b :=
+  EqvGen.reflTransGen_le_eqvGen r _ _ h
 
-theorem SymmGen.to_eqvGen (h : SymmGen r a b) : EqvGen r a b := by
-  induction h <;> grind
+theorem SymmGen.to_eqvGen (h : SymmGen r a b) : EqvGen r a b :=
+  EqvGen.symmGen_le_eqvGen r _ _ h
```
eric-wieser, line 53: These ones look perhaps worth keeping for dot notation (and upstreaming to mathlib), but could maybe become `alias`es. I'd probably argue that the `to_` should be dropped from the name.

chenson2018: Mathlib has other pairs of theorems like `Relation.reflGen_le_reflTransGen` and `Relation.ReflGen.to_reflTransGen` that this matches, so if `to_` were to be dropped, it should be coordinated with these I guess?

(There also seems to be a bit of disagreement on if both versions are really needed for dot notation, see [this discussion](https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/Scope.20of.20issue.20.23696/near/610000105))

Cslib/Foundations/Relation/Confluence.lean
thomaskwaring: thanks! i always forget this style point

Cslib/Foundations/Relation/Confluence.lean
thomaskwaring: i've done so, though i'm not sure the line `@[deprecated Relation.ReflGen.stdSymm (since := "2026-09-03")]` is exactly correct since that instance doesn't exactly reproduce the signature of the result, is there a better way?

Cslib/Foundations/Relation/Confluence.lean
thomaskwaring: possibly, i'll leave it here and make a pr to mathlib?

Cslib/Foundations/Relation/Confluence.lean
thomaskwaring: i've left them as-is for the moment, lmk if that's appropriate. in particular i'm not sure how to `alias` them since the signature is slightly different.

Cslib/Foundations/Relation/Confluence.lean
chenson2018: If making a Mathlib PR, you might also take a look at https://github.com/leanprover-community/mathlib4/pull/40626. I didn't follow through because https://github.com/leanprover-community/mathlib4/pull/40792 superceded much of it, but looking back a couple of things got left behind. Happy for you to take that over since I'm a bit busy.

Cslib/Foundations/Relation/Confluence.lean
chenson2018: Is this an appropriate place to use the new `deprecated +typeChanged`  flag?

chenson2018 (approved):
I think the only outstanding question is about deprecation procedures, so I'll approve and feel free to merge once that is settled.

PR #856, opened 2026-09-02
feat: add a predicate for monad morphisms

eric-wieser:
We show that various list operations are preserved under monad morphisms, and that FreeM.liftM is.

Note that [PolyFun already has the _bundled_ version](https://github.com/Verified-zkEVM/PolyFun/blob/main/PolyFun/Control/Monad/Hom.lean), but having the unbundled version now does not preclude adding the bundled version later.

PR #861, opened 2026-09-03
feat: add monad-polymorphic sorts

eric-wieser:
These can be used to:
* replace the specialized timeM version (included in this PR)
* implement sorts that log as they sort
* specialize to FreeM or PFunctor.FreeM

The old copyright dates on these files are because they were written by directly copying the corresponding non-monadic code from mathlib / lean core.

As a general pattern, writing code monad generically, then specializing it, allows Lean's compiler to compile it efficiently in `Id`, but also allows it to be reasoned about for richer monads `m`. The alternative of starting with a rich monad like `FreeM` and transporting back to `Id` has two downsides:

* The algorithm itself must live inside CSLib, in order to have access to `FreeM`.
* The compiler must be able to fully optimize out `FreeM.liftM` / `TimeM.ret` to match performance characteristics.

If you have no intent of ever executing the code then these concerns don't matter all that much. For now, the main benefit to CSLib is that we can write the algorithms just once while exploring a zoo of computation models to evaluate them in.

As a bonus, this lets us reuse some proofs about `List.mergeSort`, reducing the length of some proofs.

### DIFF
--- a/Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
+++ b/Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
@@ -7,6 +7,8 @@ Authors: Sorrachai Yingchareonthawornhcai
 module
 
 public import Cslib.Algorithms.Lean.TimeM
+public import Cslib.Algorithms.Lean.Sort.Merge
+import all Cslib.Algorithms.Lean.Sort.Merge
 public import Mathlib.Data.Nat.Cast.Order.Ring
 public import Mathlib.Order.Lattice.Nat
 public import Mathlib.Data.Nat.Log
@@ -33,20 +35,10 @@ namespace Cslib.Algorithms.Lean.TimeM
 
 variable {α : Type} [LinearOrder α]
 
--- TODO: replace this with `List.mergeM`
 /-- Merges two lists into a single list, counting comparisons as time cost.
 Returns a `TimeM ℕ (List α)` where the time represents the number of comparisons performed. -/
-def merge :  List α → List α → TimeM ℕ (List α)
-  | [], ys => return ys
-  | xs, [] => return xs
-  | x::xs', y::ys' => do
-    ✓ let c := (x ≤ y : Bool)
-    if c then
-      let rest ← merge xs' (y::ys')
-      return (x :: rest)
-    else
-      let rest ← merge (x::xs') ys'
-      return (y :: rest)
+def merge (xs ys : List α) : TimeM ℕ (List α) :=
+  List.mergeM xs ys fun x y => do ✓ return x ≤ y
 
 -- TODO: replace this with `List.mergeSortM`
 /-- Sorts a list using the merge sort algorithm, counting comparisons as time cost.
@@ -68,7 +60,8 @@ open List
 /-- Our merge computes the one already in mathlib. -/
 @[simp, grind =]
 theorem ret_merge (xs ys : List α) : ⟪merge xs ys⟫ = xs.merge ys := by
-  fun_induction merge with grind [nil_merge, merge_right, cons_merge_cons]
+  unfold merge
+  fun_induction mergeM with grind [nil_merge, merge_right, cons_merge_cons]
 
 /-- A list is sorted if it satisfies the `Pairwise (· ≤ ·)` predicate. -/
 abbrev IsSorted (l : List α) : Prop := List.Pairwise (· ≤ ·) l
@@ -95,7 +88,8 @@ theorem mergeSort_sorted (xs : List α) : IsSorted ⟪mergeSort xs⟫ := by
   | case2 _ _ _ _ _ ih2 ih1 => exact sorted_merge ih2 ih1
 
 lemma merge_perm (l₁ l₂ : List α) : ⟪merge l₁ l₂⟫ ~ l₁ ++ l₂ := by
-  fun_induction merge with grind [List.merge_perm_append]
+  unfold merge
+  fun_induction mergeM with grind [List.merge_perm_append]
 
 theorem mergeSort_perm (xs : List α) : ⟪mergeSort xs⟫ ~ xs := by
   fun_induction mergeSort xs with
@@ -181,7 +175,8 @@ theorem merge_ret_length_eq_sum (xs ys : List α) :
   · grind [List.length_merge]
 
 @[simp] theorem merge_time (xs ys : List α) : (merge xs ys).time ≤ xs.length + ys.length := by
-  fun_induction merge with
+  unfold merge
+  fun_induction List.mergeM with
   | case3 =>
     grind
   | _ => simp
--- a/Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
+++ b/Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
@@ -33,6 +33,7 @@ namespace Cslib.Algorithms.Lean.TimeM
 
 variable {α : Type} [LinearOrder α]
 
+-- TODO: replace this with `List.mergeM`
 /-- Merges two lists into a single list, counting comparisons as time cost.
 Returns a `TimeM ℕ (List α)` where the time represents the number of comparisons performed. -/
 def merge :  List α → List α → TimeM ℕ (List α)
@@ -47,6 +48,7 @@ def merge :  List α → List α → TimeM ℕ (List α)
       let rest ← merge (x::xs') ys'
       return (y :: rest)
 
+-- TODO: replace this with `List.mergeSortM`
 /-- Sorts a list using the merge sort algorithm, counting comparisons as time cost.
 Returns a `TimeM ℕ (List α)` where the time represents the total number of comparisons. -/
 def mergeSort (xs : List α) : TimeM ℕ (List α) :=  do
--- a/Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
+++ b/Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
@@ -33,25 +33,37 @@ set_option autoImplicit false
 
 namespace Cslib.Algorithms.Lean.TimeM
 
-variable {α : Type} [LinearOrder α]
+variable {α : Type}
+
+open List in
+/-- `TimeM.ret` passes through `List.mergeM` into the comparator. -/
+@[simp, grind =]
+theorem ret_mergeM {T} [AddMonoid T] (xs ys : List α) (le : α → α → TimeM T Bool) :
+    ⟪List.mergeM xs ys le⟫ = List.merge xs ys (fun x y => ⟪le x y⟫) := by
+  fun_induction merge with grind [mergeM, nil_merge, merge_right, cons_merge_cons]
+
+open List in
+/-- `TimeM.ret` passes through `List.mergeSortM` into the comparator. -/
+@[simp]
+theorem ret_mergeSortM {T} [AddMonoid T] (xs : List α) (le : α → α → TimeM T Bool) :
+    ⟪List.mergeSortM xs le⟫ = List.mergeSort xs (fun x y => ⟪le x y⟫) := by
+  fun_induction List.mergeSortM with
+  | case1 | case2 => simp
+  | case3 a b xs le _ _ _ iha ihb =>
+    simp only [ret_bind, ret_mergeM, mergeSort]
+    rw [iha, ihb]
+
+variable [LinearOrder α]
 
 /-- Merges two lists into a single list, counting comparisons as time cost.
 Returns a `TimeM ℕ (List α)` where the time represents the number of comparisons performed. -/
-def merge (xs ys : List α) : TimeM ℕ (List α) :=
+abbrev merge (xs ys : List α) : TimeM ℕ (List α) :=
   List.mergeM xs ys fun x y => do ✓ return x ≤ y
 
--- TODO: replace this with `List.mergeSortM`
 /-- Sorts a list using the merge sort algorithm, counting comparisons as time cost.
 Returns a `TimeM ℕ (List α)` where the time represents the total number of comparisons. -/
-def mergeSort (xs : List α) : TimeM ℕ (List α) :=  do
-  if xs.length < 2 then return xs
-  else
-    let half  := xs.length / 2
-    let left  := xs.take half
-    let right := xs.drop half
-    let sortedLeft  ← mergeSort left
-    let sortedRight ← mergeSort right
-    merge sortedLeft sortedRight
+abbrev mergeSort (xs : List α) : TimeM ℕ (List α) :=
+  List.mergeSortM xs fun x y => do ✓ return x ≤ y
 
 section Correctness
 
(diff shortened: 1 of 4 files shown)


Cslib/Algorithms/Lean/Sort/Merge.lean
```diff
@@ -0,0 +1,80 @@
...
+    mergeM xs ys (fun x y => (pure (le x y) : m Bool)) = pure (merge xs ys le) := by
+  fun_induction mergeM with grind [merge]
+
+@[simp]
+theorem idRun_mergeM (xs ys : List α) (le : α → α → Id Bool) :
+    Id.run (mergeM xs ys le) = merge xs ys (fun x y => Id.run <| le x y) :=
+  mergeM_pure _ _ _
+
+set_option linter.unusedVariables false in
+/-- A monadic version of `List.mergeSortM` -/
```
fmontesi, line 60:
```suggestion
/-- A monadic version of `List.mergeSort` -/
```

Cslib/Algorithms/Lean/Sort/Merge.lean
```diff
@@ -0,0 +1,80 @@
...
+import Cslib.Init
+
+/-!
+# A Monadic version of the builtin `List.mergeSort`
+
+This can be instantiated with `Id` to recover the original, or with `TimeM` or `FreeM` for
+algorithmic analysis.
+-/
+
+public section
```
kim-em, line 19: I think if you put `@[expose]` here you can avoid an `import all` in the later file. Not sure it's an improvement.

Cslib/Algorithms/Lean/Sort/Merge.lean
```diff
@@ -0,0 +1,80 @@
...
+  | case3  a b xs le lr _ _ ih1 ih2 =>
+    simp only [mergeSortM]
+    rw [ih1, ih2]
+    simp
+
+@[simp]
+theorem idRun_mergeSortM (xs : List α) (le : α → α → Id Bool) :
+    Id.run (mergeSortM xs le) = mergeSort xs (fun x y => Id.run <| le x y) :=
+  mergeSortM_pure _ _
+
```
kim-em, line 90: Maybe

```
@[simp] theorem cons_mergeM_cons (x y : α) (xs ys : List α) (le : α → α → m Bool) :
    mergeM (x :: xs) (y :: ys) le = do
      if ← le x y then return x :: (← mergeM xs (y :: ys) le)
      else return y :: (← mergeM (x :: xs) ys le) := by
  simp [mergeM]

@[simp] theorem mergeSortM_nil (le : α → α → m Bool) : mergeSortM [] le = pure [] := by
  simp [mergeSortM]
@[simp] theorem mergeSortM_singleton (a : α) (le : α → α → m Bool) :
    mergeSortM [a] le = pure [a] := by
  simp [mergeSortM]
```

too?

Cslib/Algorithms/Lean/Sort/Insertion.lean
```diff
@@ -0,0 +1,57 @@
...
+@[simp]
+theorem insertionSortM_pure [LawfulMonad m] (xs : List α) (r : α → α → Bool) :
+    insertionSortM (fun x y => (pure (r x y) : m Bool)) xs = pure (insertionSort (r · ·) xs) := by
+  fun_induction insertionSortM with simp_all
+
+@[simp]
+theorem idRun_insertionSortM (xs : List α) (r : α → α → Id Bool) :
+    Id.run (insertionSortM r xs) = insertionSort (fun x y => Id.run <| r x y) xs :=
+  insertionSortM_pure _ _
+
```
kim-em, line 68: Maybe

```
@[simp] theorem orderedInsertM_nil (a : α) : orderedInsertM r a [] = pure [a] := rfl
@[simp] theorem orderedInsertM_cons (a b : α) (l : List α) :
    orderedInsertM r a (b :: l) = do
      if ← r a b then return a :: b :: l else return b :: (← orderedInsertM r a l) := rfl
@[simp] theorem insertionSortM_nil : insertionSortM r [] = pure [] := rfl
@[simp] theorem insertionSortM_cons (b : α) (l : List α) :
    insertionSortM r (b :: l) = (do orderedInsertM r b (← insertionSortM r l)) := rfl
```
too?

Cslib/Algorithms/Lean/Sort/Merge.lean
```diff
@@ -0,0 +1,80 @@
...
+      return y :: (← mergeM (x :: xs) ys le)
+
+@[simp] theorem nil_mergeM (ys : List α) (le : α → α → m Bool) : mergeM [] ys le = pure ys := by
+  simp [mergeM]
+@[simp] theorem mergeM_right (xs : List α) (le : α → α → m Bool) : mergeM xs [] le = pure xs := by
+  induction xs with
+  | nil => simp
+  | cons x xs ih => simp [mergeM]
+
+@[simp]
```
kim-em:
```suggestion
@[simp↓]
```
so this can fire in a concrete monad before the `pure` is simplified (e.g. into `some`), similarly the other `_pure` lemmas. Not sure it's the best solution, but e.g. for now 

```
example (xs ys : List α) (r : α → α → Bool) :
    List.mergeM (m := Option) xs ys (fun a b => pure (r a b)) = pure (List.merge xs ys r) := by
  simp   -- fails: goal becomes `xs.mergeM ys fun a b => some (r a b)`
```

Cslib/Algorithms/Lean/Sort/Merge.lean
```diff
@@ -0,0 +1,80 @@
...
+/-- A monadic version of `List.mergeSort` -/
+def mergeSortM : ∀ (xs : List α) (le : α → α → m Bool), m (List α)
+  | [], _ => return []
+  | [a], _ => return [a]
+  | a :: b :: xs, le => do
+    let lr := MergeSort.Internal.splitInTwo ⟨a :: b :: xs, rfl⟩
+    have := by simpa using lr.2.2
+    have := by simpa using lr.1.2
+    mergeM (← mergeSortM lr.1 le) (← mergeSortM lr.2 le) le
+termination_by xs => xs.length
```
kim-em:
```suggestion
/-- A monadic version of `List.mergeSort` -/
def mergeSortM (xs : List α) (le : α → α → m Bool) : m (List α) :=
  match xs with
  | [] => return []
  | [a] => return [a]
  | a :: b :: xs => do
    let lr := MergeSort.Internal.splitInTwo ⟨a :: b :: xs, rfl⟩
    have := lr.2.2
    have := lr.1.2
    mergeM (← mergeSortM lr.1 le) (← mergeSortM lr.2 le) le
termination_by xs.length
```

kim-em: and then later `fun_induction` works nicely without a generalize:

```
theorem mergeSort_time_le (xs : List α) :
    (mergeSort xs).time ≤ timeMergeSortRec xs.length := by
  unfold mergeSort
  fun_induction List.mergeSortM with
  | case1 | case2 => grind
  | case3 _ _ _ _ _ ih2 ih1 =>
    simp only [time_bind]
    grw [merge_time]
    simp only [mergeSort_same_length]
    unfold timeMergeSortRec
    grind
```

Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
```diff
@@ -31,42 +33,45 @@ set_option autoImplicit false
...
+open List in
+/-- `TimeM.ret` passes through `List.mergeSortM` into the comparator. -/
+@[simp]
+theorem ret_mergeSortM {T} [AddMonoid T] (xs : List α) (le : α → α → TimeM T Bool) :
+    ⟪List.mergeSortM xs le⟫ = List.mergeSort xs (fun x y => ⟪le x y⟫) := by
+  fun_induction List.mergeSortM with
+  | case1 | case2 => simp
+  | case3 a b xs le _ _ _ iha ihb =>
+    simp only [ret_bind, ret_mergeM, mergeSort]
+    rw [iha, ihb]
```
kim-em, line 54: These could be proved in terms of a 

```lean
variable {n} [Monad n]

theorem mergeM_hom (φ : {β : Type} → m β → n β)
    (hpure : ∀ {β} (b : β), φ (pure b) = pure b)
    (hbind : ∀ {β γ} (x : m β) (f : β → m γ), φ (x >>= f) = φ x >>= fun b => φ (f b))
    (xs ys : List α) (le : α → α → m Bool) :
    φ (mergeM xs ys le) = mergeM xs ys (fun x y => φ (le x y)) := by
  fun_induction mergeM xs ys le with
  | case1 | case2 => simp [hpure]
  | case3 x xs y ys ih1 ih2 =>
    simp only [mergeM, hbind]
    congr 1; funext b
    split <;> simp only [map_eq_pure_bind, hbind, hpure, ih1, ih2]

theorem mergeSortM_hom (φ : {β : Type} → m β → n β)
    (hpure : ∀ {β} (b : β), φ (pure b) = pure b)
    (hbind : ∀ {β γ} (x : m β) (f : β → m γ), φ (x >>= f) = φ x >>= fun b => φ (f b))
    (xs : List α) (le : α → α → m Bool) :
    φ (mergeSortM xs le) = mergeSortM xs (fun x y => φ (le x y)) := by
  fun_induction mergeSortM xs le with
  | case1 | case2 => simp [mergeSortM, hpure]
  | case3 a b xs le lr _ _ ih1 ih2 =>
    simp only [mergeSortM, hbind, ih1, ih2, mergeM_hom φ hpure hbind]
    rfl
```

and similarly for insertion sort. Not sure it is worth it.

Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
```diff
@@ -87,24 +92,16 @@ theorem sorted_merge {l1 l2 : List α} (hxs : IsSorted l1) (hys : IsSorted l2) :
...
-    rcases x with _ | ⟨a, _ | ⟨b, rest⟩⟩ <;> grind
-  | case2 _ _ _ _ _ ih2 ih1 => exact sorted_merge ih2 ih1
+  unfold mergeSort
+  simp only [bind_pure_comp, ret_mergeSortM, ret_map]
+  convert List.pairwise_mergeSort ?_ ?_ ?_ <;> grind
 
 lemma merge_perm (l₁ l₂ : List α) : ⟪merge l₁ l₂⟫ ~ l₁ ++ l₂ := by
-  fun_induction merge with grind [List.merge_perm_append]
+  unfold merge
+  fun_induction mergeM with grind [List.merge_perm_append]
```
kim-em:
```
theorem mergeSort_sorted (xs : List α) : IsSorted ⟪mergeSort xs⟫ := by
  simpa using List.pairwise_mergeSort' (r := (· ≤ ·)) xs

lemma merge_perm (l₁ l₂ : List α) : ⟪merge l₁ l₂⟫ ~ l₁ ++ l₂ := by
  simpa using List.merge_perm_append (xs := l₁) (ys := l₂) (fun x y : α => decide (x ≤ y))
```
perhaps?

Cslib/Algorithms/Lean/Sort/Insertion.lean
```diff
@@ -0,0 +1,57 @@
...
+Authors: Jeremy Avigad, Eric Wieser
+-/
+module
+
+public import Mathlib.Data.List.Sort
+
+import Cslib.Init
+
+/-!
+# A Monadic version of the builtin `List.insertionSort`
```
kim-em:
```suggestion
# A Monadic version of Mathlib's `List.insertionSort`
```

Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
eric-wieser: This is of course the motivation for #856. Once once lands, I'll add this result to the other.

Cslib/Algorithms/Lean/Sort/Insertion.lean
eric-wieser: I think the second ones are already covered by `simp` on `insertionSortM`? Should I put `simp` on `insertionSort` too, for the same reason?

Cslib/Algorithms/Lean/Sort/Merge.lean
eric-wieser: Yeah, I wasn't sure about this either.

Cslib/Algorithms/Lean/Sort/Merge.lean
eric-wieser: I thought the point of `@[simp] def` on pattern-matching defs was that we could skip writing these. Is that not the case?

Cslib/Algorithms/Lean/Sort/Merge.lean
eric-wieser: Done, good thought.

Cslib/Algorithms/Lean/Sort/Merge.lean
eric-wieser: Done, thanks!

Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
eric-wieser: Looks like I can drop all the arguments too

Cslib/Algorithms/Lean/Sort/Merge.lean
kim-em: Indeed, as long as it is `@[expose]`'d. I think I (and the AIs!) write out all the simp lemmas out of habit. Somehow often enough you want to tweak the lemma that it still feels worthwhile to me to have them.

In particular `mergeSortM (a :: b :: xs) le` does warrant custom lemmas.

I'm happy either way here, no further comment. :-)

Cslib/Algorithms/Lean/Sort/Insertion.lean
kim-em: As above.

kim-em (approved):
Looks good to me. Happy if this is merged, however Eric wants to handle the remaining @[simp] lemma questions.

Cslib/Algorithms/Lean/Sort/Insertion.lean
eric-wieser: Done.

Cslib/Algorithms/Lean/Sort/Merge.lean
eric-wieser: I've added your suggestions.

> In particular `mergeSortM (a :: b :: xs) le` does warrant custom lemmas.

I assume you are not asking for these? I think probably stating anything interesting requires using `wp` on the comparator, which I intend to declare out of scope for this PR.

Cslib/Algorithms/Lean/MergeSort/MergeSort.lean
eric-wieser: Let's put these in #856, which will allow me to test the grind annotations.

PR #863, opened 2026-09-04
feat(Logic): General Modal Logic

fmontesi:
This PR adds general modal logic (`Modal/Basic.lean`), formalising the approach in the reference textbook 'Modal Logic' by Blackburn et al. based on similarity types (operator signatures). The formalisation is actually a bit more general: operator signatures are formalised using polynomial functors (`PFunctor`), so the argument positions of different operators can be of different types, and there are no assumptions on finiteness of operator arities.

Basic modal logic, unary modal logics, dynamic modal logics, and HML are all recovered as special cases that impose assumptions on the operator signature. This allows for reusing proof infrastructure across all modal logics. All tests have been accordingly refactored to check that reuse works as intended and that abstractions do not leak too aggressively across specialisations of general modal logic.


Cslib/Foundations/Data/PFunctor/Basic.lean
```diff
@@ -0,0 +1,59 @@
...
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Fabrizio Montesi
+-/
+
+module
+
+public import Cslib.Init
+public import Mathlib.Data.PFunctor.Univariate.Basic
+
+/-! # Additional basic theory on polynomial functors -/
```
Shreyas4991, line 12: This PR seems a bit large. Maybe this API for PFunctors could be a separate PR. It seems fairly straightforward merge.

Shreyas4991 (commented):
It seems the unary poly functor stuff is useful even beyond modal logic and could be factored into a separate PR. This would shorten this PR.

Cslib/Logics/Modal/Basic.lean
```diff
@@ -28,391 +42,170 @@ necessity (`□φ`) and possibility (`◇φ`).
...
-  /-- World accessibility relation. -/
-  r : World → World → Prop
-  /-- Valuation of atoms at a world. -/
-  v : World → Atom → Prop
+namespace Cslib.Logic.Modal
 
-/-- Propositions. -/
-inductive Proposition (Atom : Type u) : Type u where
+/-- A modal proposition. -/
+inductive Proposition (τ : PFunctor) Atom where
```
ctchou: I think an explicit `(Atom : Type*)` would be clearer.

Cslib/Logics/Modal/Basic.lean
```diff
@@ -28,391 +42,170 @@ necessity (`□φ`) and possibility (`◇φ`).
...
-  /-- Possibility. -/
-  | diamond (φ : Proposition Atom)
+  | not (φ : Proposition τ Atom)
+  /-- Disjunction. -/
+  | or (φ₁ φ₂ : Proposition τ Atom)
+  /-- Generalised possibility, or triangle. -/
+  | triangle (op : τ.A) (φs : τ.B op → Proposition τ Atom)
+
+/-- A map of propositions for the operator `op` in the polynomial functor `τ`. -/
+abbrev PropositionMap τ op Atom := τ.B op → Proposition τ Atom
```
ctchou: Please spell out the parameter types.

Cslib/Foundations/Data/PFunctor/Basic.lean
XYUnknown, line 12: I agree that these basic additional theories for PFunctors could be a separate PR. But I am also happy to accept them here. One concern, should we consider integrating it to mathlib in long term?

Cslib/Logics/Modal/Basic.lean
XYUnknown: I find it would be clearer if the type of `Atom` is stated explicitly rather than being inferred, i.e., `inductive Proposition (τ : PFunctor) (Atom : Type u) where ...`

Cslib/Logics/Modal/Basic.lean
```diff
@@ -28,391 +42,170 @@ necessity (`□φ`) and possibility (`◇φ`).
...
+namespace Cslib.Logic.Modal
 
-/-- Propositions. -/
-inductive Proposition (Atom : Type u) : Type u where
+/-- A modal proposition. -/
+inductive Proposition (τ : PFunctor) Atom where
   /-- Atomic proposition. -/
   | atom (p : Atom)
+  /-- Falsehood. -/
+  | false
```
XYUnknown, line 54: nitpick on name: `falsum` or `bottom` as in [Blackburn2001]

Cslib/Logics/Modal/Basic.lean
```diff
@@ -28,391 +42,170 @@ necessity (`□φ`) and possibility (`◇φ`).
...
-  | not (φ : Proposition Atom)
-  /-- Conjunction. -/
-  | and (φ₁ φ₂ : Proposition Atom)
-  /-- Possibility. -/
-  | diamond (φ : Proposition Atom)
+  | not (φ : Proposition τ Atom)
+  /-- Disjunction. -/
+  | or (φ₁ φ₂ : Proposition τ Atom)
+  /-- Generalised possibility, or triangle. -/
+  | triangle (op : τ.A) (φs : τ.B op → Proposition τ Atom)
```
XYUnknown, line 60: Might just be me not being able to understand it properly: does this definition permits nullity modalities?

Cslib/Logics/Modal/Basic.lean
```diff
@@ -28,391 +42,170 @@ necessity (`□φ`) and possibility (`◇φ`).
...
-
-/-- A proposition is an axiom of a frame exactly when it belongs to the logic of all models over
-that frame. -/
-theorem axiom_iff_mem_logic_modelsOfRelation (r : World → World → Prop) (φ : Proposition Atom) :
-    Axiom r⇓φ ↔ φ ∈ logic (modelsOfRelation r) := by
-  constructor
-  case mp =>
-    rintro h m rfl w
-    exact h m.v w
-  case mpr => grind [Satisfies.axiom_def]
```
XYUnknown, line 416: How do you anticipate adapting these (not so general) old theorems? Are they re-encoded somewhere with this new language adapted from ML in [Blackburn2001]?

Cslib/Logics/Modal/Basic.lean
```diff
@@ -28,391 +42,170 @@ necessity (`□φ`) and possibility (`◇φ`).
...
-@[scoped grind =, modal =]
-theorem Satisfies.or_iff_or {m : Model World Atom} :
-    ⇓Modal[m,w ⊨ φ₁ ∨ φ₂] ↔ ⇓Modal[m,w ⊨ φ₁] ∨ ⇓Modal[m,w ⊨ φ₂] := by
-  grind [=_ Proposition.or_def, Proposition.or]
+@[scoped grind =]
+lemma Proposition.nabla_def {τ : PFunctor} (op : τ.A)
+    (φs : τ.B op → Proposition τ Atom) : Proposition.nabla op φs = (∇[op]φs) := rfl
 
-/-- Characterisation of the `→` connective.
+/-- The constant proposition map for `op`. -/
```
XYUnknown: It would be nice to explain what is the constant proposition map designed for.
It means all positions are supplied with `φ`, if my understanding is correct?

Cslib/Foundations/Semantics/Frame/Basic.lean
```diff
@@ -0,0 +1,105 @@
...
+@[expose] public section
+
+namespace Cslib
+
+/-- A frame is an indexed structure of potentially heterogeneous relations.
+
+Frames are typically used in combination with modal logics or akin concepts. This is why we use
+`op` (for operator) to range over relation indexes.
+-/
+structure Frame World (τ : PFunctor) where
```
XYUnknown, line 35: Again, I wonder if the readability will be improved if we have `(World : Type u)` explicitly here?
Also, I wonder if the concept of "non-empty set W" is modeled here or not? I think non-empty is not required here right?

Cslib/Foundations/Semantics/Frame/Basic.lean
```diff
@@ -0,0 +1,105 @@
...
+## References
+
+* [P. Blackburn, M. de Rijke, Y. Venema, *Modal Logic*][Blackburn2001]
+-/
+
+@[expose] public section
+
+namespace Cslib
+
+/-- A frame is an indexed structure of potentially heterogeneous relations.
```
XYUnknown, line 30: If I understand this definition correctly, it corresponds to Definition 1.23 in the book?
(In general I find it would be helpful to point out the concept in the literature, if it is taken from the book; or explain what gets modified, if it is a formalisation based on the literature but with necessary modifications)

Cslib/Foundations/Semantics/Frame/Basic.lean
```diff
@@ -0,0 +1,105 @@
...
+accesses the constant sequence at `w`. -/
+class DiagonalSymm (f : Frame World τ) (op : τ.A) where
+  symm w ws : f.r op w ws → ∃ i, f.r op (ws i) (fun _ => w)
+
+/-- A frame is transitive at `op` if accessibility can be composed through any accessible component:
+whenever `w` accesses `ws₁` and `ws₁ i` accesses `ws₂`, then `w` accesses `ws₂`.
+-/
+class Trans (f : Frame World τ) (op : τ.A) where
+  trans w ws₁ i ws₂ : f.r op w ws₁ → f.r op (ws₁ i) ws₂ → f.r op w ws₂
+
```
XYUnknown, line 69: Add comment for the instantiation of mathlib `IsTrans` for better documentation purposes.

Cslib/Foundations/Semantics/Frame/Basic.lean
```diff
@@ -0,0 +1,105 @@
...
+@[scoped grind =, modal =]
+lemma ofRelations_iff {τ : PFunctor} [τ.Unary] (r : τ.A → World → World → Prop) (i : τ.A)
+    (w : World) (ws : τ.B i → World) : (ofRelations r).r i w ws ↔ r i w (ws default) := by rfl
+
+@[simp, scoped grind =, modal =]
+lemma ofRelations_diagonal_iff {τ : PFunctor} [τ.Unary]
+    (r : τ.A → World → World → Prop) (i : τ.A) (w w' : World) :
+    (ofRelations r).diagonal i w w' ↔ r i w w' := by
+  rfl
+
```
XYUnknown, line 102: Do we care about having reflexivity for frames here? Or we can use the class in mathlib?

Cslib/Foundations/Semantics/Frame/LTS.lean
XYUnknown, line 1: Interesting, the lts is a unary frame.

Cslib/Logics/Modal/Semantics.lean
```diff
@@ -0,0 +1,367 @@
...
+@[scoped grind ., modal .]
+theorem Satisfies.of_axiom (m : Model World τ Atom) (φ : Proposition τ Atom) (h : Axiom m.toFrame⇓φ)
+    (w : World) : ⇓Modal[m,w ⊨ φ] := h m.v w
+
+@[scoped grind =]
+theorem Satisfies.subst_apply_iff [DecidableEq (τ.B op)] {φs : PropositionMap τ op Atom}
+    {i j : τ.B op} {φ : Proposition τ Atom} : ⇓Modal[m,w ⊨ φs[i := φ] j] ↔
+      (j = i ∧ ⇓Modal[m,w ⊨ φ]) ∨ (j ≠ i ∧ ⇓Modal[m,w ⊨ φs j]) :=
+  Function.pred_update (P := fun _ φ' => Satisfies m w φ') φs i φ j
+
```
XYUnknown, line 185: If my understanding is correct, `k` and `dual` here are generalised axiom K and Dual in definition 4.5?

Cslib/Logics/Modal/Semantics.lean
```diff
@@ -0,0 +1,367 @@
...
+    Axiom f⇓(Δ[op]φs ↔ ¬∇[op]¬φs) := by grind
+
+/-- Possibility preserves conjunction. -/
+@[scoped grind ., modal .]
+theorem Satisfies.triangle_and (f : Frame World τ) (φs₁ φs₂ : PropositionMap τ op Atom) :
+    Axiom f⇓(Δ[op](φs₁ ∧ φs₂) → (Δ[op]φs₁ ∧ Δ[op]φs₂)) := by grind
+
+/-- Possibility can be combined with necessity. -/
+@[modal .]
+theorem Satisfies.triangle_and_nabla {m : Model World τ Atom} [DecidableEq (τ.B op)]
```
XYUnknown, line 203: The unary case for this is `◇p ∧ □q → ◇(p ∧ q)` right?

Cslib/Logics/Modal/Semantics.lean
```diff
@@ -0,0 +1,367 @@
...
+    (h : ⇓Modal[m,w ⊨ Δ[op]φs₁ ∧ ∇[op]φs₂]) : ∃ i, ⇓Modal[m,w ⊨ Δ[op]φs₁[i := φs₁ i ∧ φs₂ i]] := by
+  grind
+
+/-- If `φ₁` is necessary and some successor exists, then some successor satisfies `φ₁`. -/
+@[scoped grind ., modal .]
+theorem Satisfies.triangle_of_nabla {φs₁ φs₂ : PropositionMap τ op Atom}
```
XYUnknown, line 209: And the unary case for this is `□p ∧ ◇q → ◇p` right?
Just to clarify understanding

Cslib/Logics/Modal/Semantics.lean
```diff
@@ -0,0 +1,367 @@
...
+    [DecidableEq (τ.B op)] (h : ⇓Modal[m,w ⊨ ∇[op]φs₁ ∧ Δ[op]φs₂]) :
+    ∃ i, ⇓Modal[m,w ⊨ Δ[op]φs₂[i := φs₁ i]] := by grind
+
+@[scoped grind .]
+theorem Satisfies.triangle_of_diagonal {m : Model World τ Atom} {op : τ.A} {w : World}
+    {φ : Proposition τ Atom} [instRefl : Std.Refl (m.toFrame.diagonal op)]
+    (h : ⇓Modal[m,w ⊨ φ]) : ⇓Modal[m,w ⊨ Δ[op](PropositionMap.const op φ)] :=
+  ⟨fun _ => w, instRefl.refl w, by grind⟩
+
+/-- Axiom T. -/
```
XYUnknown, line 219: valid for all reflexive frames right?

Cslib/Logics/Modal/Semantics.lean
```diff
@@ -0,0 +1,367 @@
...
+  grind
+
+/-- If `φ₁` is necessary and some successor exists, then some successor satisfies `φ₁`. -/
+@[scoped grind ., modal .]
+theorem Satisfies.triangle_of_nabla {φs₁ φs₂ : PropositionMap τ op Atom}
+    [DecidableEq (τ.B op)] (h : ⇓Modal[m,w ⊨ ∇[op]φs₁ ∧ Δ[op]φs₂]) :
+    ∃ i, ⇓Modal[m,w ⊨ Δ[op]φs₂[i := φs₁ i]] := by grind
+
+@[scoped grind .]
+theorem Satisfies.triangle_of_diagonal {m : Model World τ Atom} {op : τ.A} {w : World}
```
XYUnknown, line 214: Please add comment to explain this theorem.
I assume that it is for proving axiom t?

Cslib/Logics/Modal/Semantics.lean
```diff
@@ -0,0 +1,367 @@
...
+    {φ : Proposition τ Atom} [instRefl : Std.Refl (m.toFrame.diagonal op)]
+    (h : ⇓Modal[m,w ⊨ φ]) : ⇓Modal[m,w ⊨ Δ[op](PropositionMap.const op φ)] :=
+  ⟨fun _ => w, instRefl.refl w, by grind⟩
+
+/-- Axiom T. -/
+theorem Satisfies.t (f : Frame World τ) [instRefl : Std.Refl (f.diagonal op)]
```
XYUnknown, line 220: I wonder why the restriction of `PropositionMap.const` is necessary here?

I think we can basically do:
```
/-- If every argument holds at a world, a reflexive diagonal witnesses their possibility. -/
theorem Satisfies.triangle_of_diagonal_forall {m : Model World τ Atom} {op : τ.A} {w : World}
    {φs : PropositionMap τ op Atom} [instRefl : Std.Refl (m.toFrame.diagonal op)]
    (h : ∀ i, ⇓Modal[m,w ⊨ φs i]) : ⇓Modal[m,w ⊨ Δ[op]φs] :=
  ⟨fun _ => w, instRefl.refl w, h⟩

/-- A general version of axiom T -/
theorem Satisfies.t_general (f : Frame World τ) [Std.Refl (f.diagonal op)]
    (φ : Proposition τ Atom) (φs : PropositionMap τ op Atom)
    (h : ∀ i, Axiom f⇓(φ → φs i)) : Axiom f⇓(φ → Δ[op]φs) := by
  intro v w
  rw [Satisfies.imp_iff_imp]
  intro hφ
  apply Satisfies.triangle_of_diagonal_forall
  intro i
  exact (Satisfies.imp_iff_imp.mp (h i v w)) hφ
```

Cslib/Logics/Modal/Semantics.lean
```diff
@@ -0,0 +1,367 @@
...
+    (φ : Proposition τ Atom) : Axiom f⇓(φ → Δ[op](PropositionMap.const op φ)) := by grind
+
```
XYUnknown, line 222: similarly, these `Satisfies.t_refl` and `Satisfies.t_nabla_triangle` can also be generalised, I think

Cslib/Logics/Modal/Semantics.lean
```diff
@@ -0,0 +1,367 @@
...
+/-- In any model whose diagonal relation for `op` is reflexive, `∇[op]φ → φ` is equivalent to
+`φ → Δ[op]φ`. -/
+theorem Satisfies.t_nabla_triangle (f : Frame World τ) [Std.Refl (f.diagonal op)] :
+    Axiom f⇓((∇[op](PropositionMap.const op φ) → φ) ↔ (φ → Δ[op](PropositionMap.const op φ))) := by
+  intro _ w
+  have hr : f.r op w (fun _ => w) := by
+    simpa [Frame.diagonal] using (Std.Refl.refl (r := f.diagonal op) w)
+  grind
+
+/-- Axiom B, valid for diagonally symmetric frames. -/
```
XYUnknown, line 246: Again, I wonder if axiom B is necessary to be defined only with `PropositionMap.const`?

Cslib/Logics/Modal/Semantics.lean
```diff
@@ -0,0 +1,367 @@
...
+      simp only [φs] at hj
+      rcases hj with ⟨k, hk, hw⟩
+      have hjk : j = k := e.injective hk
+      simpa [hjk] using hw
+    grind only
+
+/-- A proposition is valid in a class of models `S` (modelled as a set) if it is satisfied under
+all models in `S` for all worlds. -/
+@[simp, scoped grind =]
+def Proposition.valid {World : Type*} {τ : PFunctor} {Atom : Type*} (S : Set (Model World τ Atom))
```
XYUnknown, line 340: Corresponding to Definition 1.28 right?
also the definition of logic below

Cslib/Logics/Modal/LogicalEquivalence.lean
```diff
@@ -116,84 +136,88 @@ instance {World Atom} (S : ModelClass World Atom) :
...
+theorem Proposition.axiom_iff_equivWithin_modelsOfFrame
+    (f : Frame World τ) (φ₁ φ₂ : Proposition τ Atom) :
+    Axiom f⇓(φ₁ ↔ φ₂) ↔ φ₁ ≡[EquivWithin (modelsOfFrame f)] φ₂ := by grind
+
+/-- Triangle and the negation of the dual nabla are universally logically equivalent. -/
+theorem Proposition.dual_equiv (φs : PropositionMap τ op Atom) :
+    Δ[op]φs ≡[UEquiv (World := World)] ¬∇[op]¬φs := by
+  -- We use `grind only` as a test that `modal` can lift axioms to logical equivalences
+  grind only [modal, Satisfies.dual]
+
```
XYUnknown, line 220: Do I understand correctly: these results are generalsing axioms to congruence over worlds?

Cslib/Logics/Modal/Unary/Basic.lean
```diff
@@ -0,0 +1,53 @@
...
+-/
+
+module
+
+public import Cslib.Foundations.Data.PFunctor.Basic
+public import Cslib.Logics.Modal.Basic
+public import Cslib.Logics.Modal.Semantics
+public import Cslib.Logics.Modal.LogicalEquivalence
+
+/-! # Unary Modal Logic -/
```
XYUnknown, line 14: Do I understand correctly, this file defines the modal operators that are intended to be used for encoding propositional dynamic logic?

Cslib/Logics/Modal/Lean/Basic.lean
XYUnknown, line 1: So, this is a generalisation from the old relation results to the new frame constructs, right?

Cslib/Logics/Modal/Unimodal/Basic.lean
```diff
@@ -0,0 +1,324 @@
...
+    let h' := h (v := v) (w := w₁) (φ := a)
+    grind
+
+/-- Axiom 5. -/
+theorem Satisfies.unimodal_five (f : Frame World τ) [Relation.RightEuclidean f.rel]
+    (φ : Proposition τ Atom) : Axiom f⇓(◇φ → □◇φ) := by
+  have := @Relation.RightEuclidean.rightEuclidean (r := f.rel)
+  grind
+
+/-- Any frame that admits 5 is Euclidean. -/
```
XYUnknown, line 182: nitpick: right euclidean

Cslib/Logics/HML/LogicalEquivalence.lean
XYUnknown, line 1: Just to clarify my understanding: this can be expressed with Cslib/Logics/Modal/LogicalEquivalence.lean, hence deleted.

CslibTests/Modal/Stlc.lean
XYUnknown, line 1: This is cool.

I wonder if this can be somehow modified into a part of the reusable library rather than living in the tests as an example application.

XYUnknown (changes requested):
I have not found anything wrong, the only technical concern is axiom T and and B in `Cslib/Logics/Modal/Semantics.lean` may be able to be generalised without the need of `PropositionMap.const` --- it is unclear to me why this restriction is needed in the first place. Though it can be a separate pr as this pr is already big.

Some concepts in the generalised frame can be explained better, sometimes it's non-trivial to figure out the exact correspondence between the Lean code and the defintion in the literature.

Cslib/Foundations/Data/PFunctor/Basic.lean
fmontesi: Agreed in general. In this case, I've left it here primarily because I wanna make sure that this API is in sync with the modal signature development.

fmontesi: I don't know. I'd be open to both upstreaming this or downstreaming PFunctor to CSLib. CS people seem very interested in this, there's the polyfun library and #803 as well. But if there are active maintainers for this in mathlib, it'd make good sense to upstream.

Cslib/Logics/Modal/Basic.lean
fmontesi: Yes, for example:

```
def myP : PFunctor where
  A := Unit
  B := fun _ => Empty

def φ : Proposition myP Bool := .triangle () (fun x => nomatch x)
```

Cslib/Logics/Modal/Basic.lean
fmontesi: Mmh, right.. using Latin is still in fashion in logic, but I kinda find `false` more in line with the Lean ecosystem.

fmontesi: Kinda also why I'm calling negation `not` instead of `neg`, then we get theorems with the expected names in the tradition of Lean and Mathlib.

Cslib/Logics/Modal/Basic.lean
fmontesi: Is the docstring clearer now?

XYUnknown: Yes :))

Cslib/Logics/Modal/Basic.lean
fmontesi: A bunch of them is in `Unimodal/Basic.lean`. Many things work out of the box and don't need restatement, like the definition of `logic`, etc.

Cslib/Foundations/Semantics/Frame/Basic.lean
fmontesi: Done for the first.
For the second: yes, it's a choice not to require that World is nonempty, since it is not necessary for most of the theory.

Cslib/Foundations/Semantics/Frame/Basic.lean
fmontesi: Yes, I've updated the docstring.

Cslib/Foundations/Semantics/Frame/Basic.lean
fmontesi: I managed to use the standard class in `Unimodal/Basic.lean`, but I'm open to evolving this later if we ever see the need ofc.

Cslib/Foundations/Semantics/Frame/LTS.lean
fmontesi: Yep! At least it's a useful way to look at it for modal reasoning.

Cslib/Logics/Modal/Semantics.lean
fmontesi: Definition 4.13. Docstring updated.

Cslib/Logics/Modal/Semantics.lean
fmontesi: It's an iff (see Satisfies.b_symm later), so at the very least we need extensional equality wrt PropositionMap.const. We can generalise it in the future (or make separate lemmas) to allow for extensional equality.

Cslib/Logics/Modal/Semantics.lean
fmontesi: Yes, docstring updated.

Cslib/Logics/Modal/Semantics.lean
fmontesi: Yes. Docstring updated.

Cslib/Logics/Modal/Semantics.lean
fmontesi: Yes again. (I'm updating the docstrings with comments on this kind of stuff, I think they might be useful to future readers as well.)

Cslib/Logics/Modal/Semantics.lean
fmontesi: Correct, docstring updated.

Cslib/Logics/Modal/Semantics.lean
fmontesi: This is equivalent to the current Axiom T, since it's proven that it's in an iff relationship with reflexivity (t_symm). However, we should probably come back to this at a later time because your theorem should be derivable in terms of this development and modus ponens.

Cslib/Logics/Modal/Semantics.lean
fmontesi: Yes

Cslib/Logics/Modal/Semantics.lean
fmontesi: Yes, there's something to think about regarding extensional equality to const or requiring just const. I'm not too sure about this yet, it probably merits some dedicated work.

Cslib/Logics/Modal/LogicalEquivalence.lean
fmontesi: They make axioms that state bi-implications available as logical equivalences. In most cases, they can be derived automatically by the axiom itself and `grind [modal]`. This is an important equivalence so I put it here as a test that this mechanism actually works.

Cslib/Logics/Modal/Unary/Basic.lean
fmontesi: Also, but not only. For example, the same works for HML and the basic temporal language (#927). It's the theory of all operator signatures where the operators take exactly one argument.

Cslib/Logics/HML/LogicalEquivalence.lean
fmontesi: Exactly, it's completely subsumed.

CslibTests/Modal/Stlc.lean
fmontesi: I agree. I stopped here mainly because the PR was already huge... :-)

Cslib/Logics/Modal/Lean/Basic.lean
fmontesi: Yep!

XYUnknown (approved):

PR #865, opened 2026-09-04
feat(Algorithms/CCS): Diffie-Hellman protocol in Stateful Processes

fmontesi:
Adds a simple model of the Diffie-Hellman protocol in Stateful Processes. The main purpose of this PR is to have a simple enough testbed to measure improvement on how we (i) integrate embedded DSLs like Stateful Processes with computation in Lean and (ii) reason about the semantics of concrete embedded programs such as these. This is explicitly stated in the docstring.


XYUnknown (approved):
All good, ready to be merged.

PR #873, opened 2026-09-05
feat(MultiTapeTM): add input and work tape rewinding

SamuelSchlesinger:
Add a shared controller for rewinding native input and contiguous work-tape contents. Use it to transform any machine into one that halts with its input head at the initial position, preserving output and work tapes.

Part 4/6 of the TM composition stack. Depends on #872. Next: #874.

Validation: strict build, import checks, full tests, and linters.

This PR was composed with Astra via Codex.

PR #874, opened 2026-09-05
feat(MultiTapeTM): compose machines using tape transformations

SamuelSchlesinger:
Construct machine composition from output redirection, work-tape rewind, and input substitution using the sequential combinator. Prove that the composite halts with the second machine's output on the first machine's result.

Part 5/6 of the TM composition stack. Depends on #873. Next: #875.

Validation: strict build, import checks, full tests, and linters.

This PR was composed with Astra via Codex.

### DIFF
--- a/Cslib/Computability/Machines/Turing/MultiTape/Plumbing/Composition.lean
+++ b/Cslib/Computability/Machines/Turing/MultiTape/Plumbing/Composition.lean
@@ -1,0 +1,52 @@
+/-
+Copyright (c) 2026 Samuel Schlesinger. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Samuel Schlesinger
+-/
+
+module
+
+public import Cslib.Computability.Machines.Turing.MultiTape.Plumbing.Composition.Rewind
+
+/-!
+# Correctness of multi-tape composition
+
+`comp_haltsWithOutput` proves that the composite machine returns the second machine's output
+on the first machine's result. The first phase takes one step per native step; the second takes
+two, following a rewind of the intermediate output and one initial classification step.
+-/
+
+@[expose] public section
+
+namespace Turing.MultiTapeTM
+
+open Composition
+
+variable {k₀ k₁ : ℕ}
+variable {Symbol State₀ State₁ : Type*}
+
+variable (tm₀ : MultiTapeTM k₀ Symbol State₀) (tm₁ : MultiTapeTM k₁ Symbol State₁)
+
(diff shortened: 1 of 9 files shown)


Cslib/Computability/Machines/Turing/MultiTape/Plumbing/README.md
```diff
@@ -11,3 +11,5 @@ These modules describe executable machines and their effects on configurations a
 - `InputFromWorkTape` simulates the native input on a work tape, preserving boundary clamping.
 - `Rewind` shares one controller between native-input and work-tape rewinding. Work-tape rewind
   starts immediately after contiguous contents and finishes at their first cell, including when empty.
+- `Composition` assembles output redirection, work-tape rewind, and input substitution with `seq`
```
crei, line 16: I think it would be better to first explain what it does and then how it does it. It reads like we already know where this is going.

Cslib/Computability/Machines/Turing/MultiTape/Plumbing/Composition.lean
```diff
@@ -0,0 +1,52 @@
...
+
+variable {k₀ k₁ : ℕ}
+variable {Symbol State₀ State₁ : Type*}
+
+variable (tm₀ : MultiTapeTM k₀ Symbol State₀) (tm₁ : MultiTapeTM k₁ Symbol State₁)
+
+/-- Composition returns the second machine's output on the first machine's result.
+The first halting time must be minimal; the second may be padded. -/
+theorem comp_haltsWithOutput
+    {input out₀ out₁ : List Symbol} {u v : ℕ}
```
crei, line 33:
```suggestion
    {input out₀ out₁ : List Symbol} {t\0 t\1 : ℕ}
```

Cslib/Computability/Machines/Turing/MultiTape/Plumbing/Composition.lean
```diff
@@ -0,0 +1,52 @@
...
+-/
+
+module
+
+public import Cslib.Computability.Machines.Turing.MultiTape.Plumbing.Composition.Rewind
+
+/-!
+# Correctness of multi-tape composition
+
+`comp_haltsWithOutput` proves that the composite machine returns the second machine's output
```
crei, line 14: This could also be a bit more stand-alone. What is "the composite machine"?

crei (changes requested):
I haven't read all of this PR yet, but I'm a bit surprised that there is still so much code to it. Maybe the plumbing is not abstract enough or does not have enough lemmas?

PR #766, opened 2026-08-01
feat(Crypto): Commitment Schemes

SamuelSchlesinger:
Defines commitment schemes and their properties. Does not include computational hiding, but builds up to a proof that no statistically hiding, perfectly binding commitment scheme can exist. This justifies the existence of the computationally binding property, which we'll hopefully define in the future.

PR #876, opened 2026-09-05
feat(Crypto): define PRGs and their security definition

SamuelSchlesinger:
Formalizes Boneh-Shoup's definitions of PRGs and proves the impossibility of making an extending PRG against an arbitrary adversary. Takes the approach of #766 by defining the construction and proving the impossibility in certain contexts, motivating the future computational security definitions.

Composed with Astra using Codex based on some code that I had written by hand.

### DIFF
--- a/Cslib/Crypto/Primitives/PRG/Asymptotic.lean
+++ b/Cslib/Crypto/Primitives/PRG/Asymptotic.lean
@@ -1,0 +1,131 @@
+/-
+Copyright (c) 2026 Samuel Schlesinger. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Samuel Schlesinger
+-/
+
+module
+
+public import Cslib.Crypto.Primitives.PRG.Basic
+public import Mathlib.Analysis.Asymptotics.SuperpolynomialDecay
+public import Mathlib.Data.FinEnum
+
+/-!
+# Asymptotic pseudorandom generator security
+
+Security for families of generators means negligible advantage for each admissible
+adversary family, following [BonehShoup2023], Definition 3.1. Negligibility uses Mathlib's
+`Asymptotics.SuperpolynomialDecay`. Admissibility is a predicate on the whole adversary
+family, so a downstream computational model can express a uniform resource restriction.
+This model has a natural-number security parameter and no sampled public system parameters.
+Efficiency of generation and sampling is not asserted by these semantic definitions.
+
+If the range test is admissible and the seed space is eventually at most half the output
+space, security is impossible: the advantage is eventually at least one half. In particular,
+no family stretching bitstrings by at least one bit is secure against arbitrary adversaries.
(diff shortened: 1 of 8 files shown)


Cslib/Crypto/Primitives/PRG/Asymptotic.lean
```diff
@@ -0,0 +1,131 @@
...
+open scoped Topology
+
+/-- A security-parameter-indexed collection of deterministic generators. -/
+abbrev Family (Seed Output : ℕ → Type*) := ∀ n, Generator (Seed n) (Output n)
+
+namespace Family
+
+variable {Seed Output : ℕ → Type*}
+variable [∀ n, Fintype (Seed n)] [∀ n, Nonempty (Seed n)]
+variable [∀ n, Fintype (Output n)] [∀ n, Nonempty (Output n)]
```
Shreyas4991, line 44: Why are Generators arbitrary types? Couldn't they just be instances of a structure. Then if the structure got a Fintype or Finite instance, there would be no need to declare [forall n, Fintype (Seed n)] etc. This is just a design comment.

Cslib/Crypto/Primitives/PRG/Basic.lean
```diff
@@ -0,0 +1,158 @@
...
+namespace Cslib.Crypto.PRG.Generator
+
+open scoped NNReal
+
+variable {Seed Output : Type*}
+
+/-- An unbounded adversary tests whether its input is in the generator's range.
+It is not assumed to be admissible for a computationally restricted class. -/
+noncomputable def rangeAdversary (G : Generator Seed Output) : Adversary Output := by
+  classical
```
Shreyas4991, line 32: Is this actually needed here? To me it sounds like 
`:= fun output => PMF.pure (decide (output ∈ Set.range G))` might work

Cslib/Crypto/Primitives/PRG/Basic.lean
```diff
@@ -0,0 +1,158 @@
...
+theorem secure_zero_of_outputDist_eq (G : Generator Seed Output)
+    (hG : G.outputDist = PMF.uniformOfFintype Output)
+    (Admissible : Adversary Output → Prop) : G.Secure Admissible 0 := by
+  intro adversary _
+  simp [advantage, realExperiment, idealExperiment, hG]
+
+/-- Zero-error security against arbitrary tests is equivalent to exactly uniform output. -/
+theorem secure_zero_iff_outputDist_eq_uniform (G : Generator Seed Output) :
+    G.Secure (fun _ => True) 0 ↔ G.outputDist = PMF.uniformOfFintype Output := by
+  classical
```
Shreyas4991, line 74: It seems better to explicitly declare what Props need to be Decidable than use `classical`.

Shreyas4991 (commented):
Initial comments. Not all are "required" of course. I would recommend taking a close took at all uses of `classical`. GPT has a tendency to play it too safe by inserting `classical` anywhere and everywhere.

PR #880, opened 2026-09-07
feat(Foundations/Relation/Confluence): generalize results from confluence to commutation

thomaskwaring:
This PR generalises many results from confluent to commuting relations, and obtains the classical case as a specialisation. We add `HJoin` and `MHJoin` — heterogenous versions of `Join` and `MJoin` — and associated API for them and related relational constructions.

NB: the theorem `confluent_equivalents` is now public, and the `TFAE` has been extended with certain other properties which generalise better to the heterogenous case.

### DIFF
--- a/Cslib/Foundations/Relation/Basic.lean
+++ b/Cslib/Foundations/Relation/Basic.lean
@@ -48,6 +48,7 @@ def transLeft (s r : α → α → Prop) [IsTrans α r] (h : s ≤ r) : Trans s 
 def transRight (s r : α → α → Prop) [IsTrans α r] (h : s ≤ r) : Trans r s r where
   trans hab hbc := _root_.trans hab (h _ _ hbc)
 
+@[grind .]
 theorem comp_le_comp {s s' r r' : α → α → Prop} (hs : s ≤ s') (hr : r ≤ r') :
     Comp s r ≤ Comp s' r' := fun a c ⟨b, hab, hbc⟩ ↦ ⟨b, hs a b hab, hr b c hbc⟩
 
@@ -59,11 +60,13 @@ theorem swap_le_iff_le_swap {r₁ r₂ : α → α → Prop} : swap r₁ ≤ r�
 
 attribute [scoped grind] ReflGen TransGen ReflTransGen EqvGen
 
+@[grind .]
 theorem ReflGen.le_reflGen : r ≤ ReflGen r := fun _ _ => ReflGen.single
 
 theorem ReflGen.to_eqvGen (h : ReflGen r a b) : EqvGen r a b :=
   EqvGen.reflGen_le_eqvGen r _ _ h
 
+@[grind .]
 theorem TransGen.le_transGen : r ≤ TransGen r := fun _ _ => TransGen.single
 
 theorem TransGen.to_eqvGen (h : TransGen r a b) : EqvGen r a b :=
@@ -72,13 +75,15 @@ theorem TransGen.to_eqvGen (h : TransGen r a b) : EqvGen r a b :=
 theorem ReflTransGen.to_eqvGen (h : ReflTransGen r a b) : EqvGen r a b :=
   EqvGen.reflTransGen_le_eqvGen r _ _ h
 
+@[grind .]
 theorem SymmGen.le_symmGen : r ≤ SymmGen r := fun _ _ => Or.inl
 
 theorem SymmGen.to_eqvGen (h : SymmGen r a b) : EqvGen r a b :=
   EqvGen.symmGen_le_eqvGen r _ _ h
 
-@[simp] theorem sup_swap_eq_symmGen : r ⊔ Function.swap r = SymmGen r := rfl
+@[simp, grind =] theorem sup_swap_eq_symmGen : r ⊔ Function.swap r = SymmGen r := rfl
 
+@[grind .]
 theorem EqvGen.le_eqvGen : r ≤ EqvGen r := EqvGen.rel
 
 theorem _root_.Equivalence.eqvGen_le (h : Equivalence r₂) (hle : r₁ ≤ r₂) : EqvGen r₁ ≤ r₂ :=
@@ -90,7 +95,7 @@ attribute [scoped grind →] ReflGen.to_eqvGen TransGen.to_eqvGen ReflTransGen.t
 
 theorem Join.single [Std.Refl r] (h : r a b) : Join r a b := ⟨b, h, refl b⟩
 
-@[simp] theorem join₂_eq_join : Join₂ r r = Join r := rfl
+@[simp, grind =] theorem join₂_eq_join : Join₂ r r = Join r := rfl
 
 @[grind =] theorem join₂_eq_comp_swap : Join₂ r₁ r₂ = Comp r₁ (swap r₂) := rfl
 
--- a/Cslib/Foundations/Relation/Basic.lean
+++ b/Cslib/Foundations/Relation/Basic.lean
@@ -71,6 +71,8 @@ theorem SymmGen.le_symmGen : r ≤ SymmGen r := fun _ _ => Or.inl
 theorem SymmGen.to_eqvGen (h : SymmGen r a b) : EqvGen r a b :=
   EqvGen.symmGen_le_eqvGen r _ _ h
 
+theorem symmGen_eq_sup_swap : SymmGen r = r ⊔ Function.swap r := rfl
+
 theorem EqvGen.le_eqvGen : r ≤ EqvGen r := EqvGen.rel
 
 theorem _root_.Equivalence.eqvGen_le (h : Equivalence r₂) (hle : r₁ ≤ r₂) : EqvGen r₁ ≤ r₂ :=
@@ -112,7 +114,7 @@ theorem _root_.Equivalence.mJoin_le (h : Equivalence r₂) (hle : r₁ ≤ r₂)
 theorem MJoin.mJoin_le_eqvGen : MJoin r ≤ EqvGen r :=
     (EqvGen.is_equivalence r).mJoin_le EqvGen.le_eqvGen
 
-theorem mHJoin_eq_mJoin : MHJoin r r = MJoin r := rfl
+@[simp] theorem mHJoin_eq_mJoin : MHJoin r r = MJoin r := rfl
 
 theorem MHJoin.mHJoin_le [Std.Refl r] [IsTrans α r] (h₁ : r₁ ≤ r) (h₂ : swap r₂ ≤ r) :
     MHJoin r₁ r₂ ≤ r := by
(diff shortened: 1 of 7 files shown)


Cslib/Foundations/Relation/Basic.lean
```diff
@@ -40,28 +42,99 @@ def transLeft (s r : α → α → Prop) [IsTrans α r] (h : s ≤ r) : Trans s
...
+  EqvGen.eqvGen_le hle
+
 attribute [scoped grind →] ReflGen.to_eqvGen TransGen.to_eqvGen ReflTransGen.to_eqvGen
   SymmGen.to_eqvGen
 
+theorem Join.single [Std.Refl r] (h : r a b) : Join r a b := ⟨b, h, refl b⟩
+
+@[simp] theorem hJoin_eq_join : HJoin r r = Join r := rfl
+
+@[grind =] theorem hJoin_eq_comp_swap : HJoin r₁ r₂ = Comp r₁ (swap r₂) := rfl
```
chenson2018: What was the thought behind the split `simp` and `grind =`? I'm always a bit suspicious when I don't see them in matching pairs.

Cslib/Foundations/Relation/Basic.lean
```diff
@@ -40,28 +42,99 @@ def transLeft (s r : α → α → Prop) [IsTrans α r] (h : s ≤ r) : Trans s
...
+
+theorem comp_self_le (r : α → α → Prop) [IsTrans α r] : Comp r r ≤ r :=
+  fun _ _ ⟨_, hab, hbc⟩ ↦ _root_.trans hab hbc
+
+theorem swap_le_iff_le_swap {r₁ r₂ : α → α → Prop} : swap r₁ ≤ r₂ ↔ r₁ ≤ swap r₂ := by
+  constructor <;> intro h a b hab <;> exact h b a hab
+
 attribute [scoped grind] ReflGen TransGen ReflTransGen EqvGen
 
+theorem ReflGen.le_reflGen : r ≤ ReflGen r := fun _ _ => ReflGen.single
```
chenson2018, line 64: Good catch noticing these. I initially found the new usage of `≤` for relations a bit strange, but I think Eric has convinced me that both of these should exist. These should find their way into Mathlib eventually, maybe leave a comment about this?

Cslib/Foundations/Relation/Basic.lean
```diff
@@ -40,28 +42,99 @@ def transLeft (s r : α → α → Prop) [IsTrans α r] (h : s ≤ r) : Trans s
...
+  (comp_le_comp h₁ h₂).trans (comp_self_le r)
+
+theorem HJoin.swap_iff {a b : α} : HJoin r₁ r₂ b a ↔ HJoin r₂ r₁ a b := by grind [HJoin]
+
 @[deprecated _root_.refl (since := "2026-09-07")]
 theorem MJoin.refl (a : α) : MJoin r a a := _root_.refl a
 
-theorem MJoin.single (h : ReflTransGen r a b) : MJoin r a b := by
-  use b
+@[deprecated Join.single (since := "2026-09-07")]
```
chenson2018, line 117: Can you check that this doesn't give a warning under the nightly toolchain? For the last PR there where some `+typeChanged` that needed to be added. (Which is fine, but I think it's nicer to handle them now)

Cslib/Foundations/Relation/Defs.lean
```diff
@@ -35,10 +35,16 @@ def dom (r : α → β → Prop) : Set α := {a | ∃ b, r a b}
 /-- Codomain of a relation, aka range. -/
 def cod (r : α → β → Prop) : Set β := {b | ∃ a, r a b}
 
+/-- Generalisation of `Join` to two relations. -/
+def HJoin (r₁ r₂ : α → α → Prop) (a b : α) : Prop := ∃ c, r₁ a c ∧ r₂ b c
```
chenson2018: Talking with Eric in person, he made a good point that this naming could be confusing in suggesting that this is about heterogeneous relations. He suggested the name `Join₂` to better differentiate. (And long-term, considered that this could replace `Join` in Mathlib)

Cslib/Foundations/Relation/Defs.lean
```diff
@@ -35,10 +35,16 @@ def dom (r : α → β → Prop) : Set α := {a | ∃ b, r a b}
...
+
 /-- The join of the reflexive transitive closure. This is not named in Mathlib, but see
   `#loogle Relation.Join (Relation.ReflTransGen ?r)` -/
 abbrev MJoin (r : α → α → Prop) := Join (ReflTransGen r)
```
chenson2018, line 43: I wonder a bit how necessary these `M` variants are. It's harmless since they are `abbrev` and it does make combinatory logic a bit nicer looking, but otherwise it could be considered redundant.

chenson2018 (changes requested):
Some minor organization comments:

Cslib/Foundations/Relation/Basic.lean
thomaskwaring: personally i quite like the LE spellings, particularly as they give slick proofs of the relations between the different `...Gen` constructions (eg `Equivalence.mJoin_le` below). will do on the upstreaming note.

Cslib/Foundations/Relation/Defs.lean
thomaskwaring: That's a good point — I was following `emptyHRelation` which is semantically entirely differet. I like `Join₂` a lot better.

Cslib/Foundations/Relation/Basic.lean
thomaskwaring: i don't think we want the latter as `simp` — in general i haven't systematically added `grind` annotations bc my proofs haven't used them (also many of the definitions are `abbrev` so they're not needed). should i at least add `grind =` to the `simp` lemmas where appropriate? i can also look for other possible `grind` lemmas.

Cslib/Foundations/Relation/Basic.lean
thomaskwaring: yes will do though you will need to tell me how (sorry!)

Cslib/Foundations/Relation/Defs.lean
thomaskwaring: i would support deprecating them — it does make some of the statements less wordy so if you think it's harmless i'm okay with keeping them.

Cslib/Foundations/Relation/Basic.lean
chenson2018: No worries, see https://github.com/leanprover/downstream-lean4/pull/54 where this was fixed

Cslib/Foundations/Relation/Defs.lean
chenson2018: If you also don't feel strongly, I think I would deprecate them just so there are fewer spellings

Cslib/Foundations/Relation/Basic.lean
chenson2018: If we don't want it as `simp`, I would probably remove `grind` as well. You don't need to review systematically, but this one stuck out.

Cslib/Foundations/Relation/Confluence.lean
```diff
@@ -29,60 +32,100 @@ variable {α : Type*} {r r₁ r₂ : α → α → Prop}
...
+@[deprecated (since := "2026-09-03")] alias DiamondCommute.toDiamond := DiamondCommute.to_diamond
+
+@[simp] theorem SemiCommute.to_semiConfluent : SemiCommute r r = SemiConfluent r := rfl
+
+@[simp] theorem LocallyCommute.to_locallyConfluent : LocallyCommute r r = LocallyConfluent r := rfl
+
+instance : Std.Symm (@Commute α) where
+  symm r₁ r₂ h x y₁ y₂ x_y₁ x_y₂ := by grind [h x_y₂ x_y₁, Join₂]
+
+lemma DiamondCommute.extend (h : DiamondCommute r₁ r₂) (h₁ : ReflTransGen r₁ a b) (h₂ : r₂ a c) :
```
lengyijun:
```lean
lemma DiamondCommute.extend (h : DiamondCommute r₁ r₂) : SemiCommute r₂ r₁ := by
  intros _ c _ h₂ h₁
  induction h₁ using ReflTransGen.head_induction_on generalizing c with
  | refl => exact Join₂.single_right (.single h₂)
  | head ha _ ih =>
    obtain ⟨d, had, hcd⟩ := h ha h₂
    obtain ⟨d', hbd', hdd'⟩ := ih had
    exact ⟨d', .head hcd hbd', hdd'⟩
```

Cslib/Foundations/Relation/Defs.lean
```diff
@@ -63,14 +69,22 @@ abbrev Commute (r₁ r₂ : α → α → Prop) := DiamondCommute (ReflTransGen
...
+/-- Generalisation of `SemiConfluent` to two relations. -/
+abbrev SemiCommute (r₁ r₂ : α → α → Prop) :=
+  ∀ {x y₁ y₂}, r₁ x y₁ → ReflTransGen r₂ x y₂ → MJoin₂ r₂ r₁ y₁ y₂
 
 /-- A relation has the Church Rosser property when equivalence implies multi-joinability. -/
-abbrev ChurchRosser (r : α → α → Prop) := ∀ {x y}, EqvGen r x y → Join (ReflTransGen r) x y
+abbrev ChurchRosser (r : α → α → Prop) := ∀ {x y}, EqvGen r x y → MJoin r x y
 
 /-- A relation is locally confluent when all reductions with a common origin are multi-joinable -/
 abbrev LocallyConfluent (r : α → α → Prop) :=
```
lengyijun: Can we group the definitions into those involving one relation and those involving two relations?

Cslib/Foundations/Relation/Basic.lean
thomaskwaring: makes sense! thanks

Cslib/Foundations/Relation/Basic.lean
thomaskwaring: okay i'm not sure how to double check on the nightly toolchain but i've added them in anyway

Cslib/Foundations/Relation/Confluence.lean
lengyijun: @thomaskwaring Can you see this comment (I'm not very familiar with GitHub's notification mechanism)

Cslib/Foundations/Relation/Defs.lean
```diff
@@ -35,10 +35,18 @@ def dom (r : α → β → Prop) : Set α := {a | ∃ b, r a b}
...
+def Join₂ (r₁ r₂ : α → α → Prop) (a b : α) : Prop := ∃ c, r₁ a c ∧ r₂ b c
+
 /-- The join of the reflexive transitive closure. This is not named in Mathlib, but see
   `#loogle Relation.Join (Relation.ReflTransGen ?r)` -/
+@[deprecated "use `Join (ReflTrasnGen ·)` instead" (since := "2026-09-12")]
 abbrev MJoin (r : α → α → Prop) := Join (ReflTransGen r)
 
+/-- Generalisation of `MJoin` to two relations. -/
+@[deprecated "use Join₂ (ReflTransGen ·) (ReflTransGen ·) instead." (since := "2026-09-12")]
+abbrev MJoin₂ (r₁ r₂ : α → α → Prop) := Join₂ (ReflTransGen r₁) (ReflTransGen r₂)
```
chenson2018: This didn't exist before this PR, so doesn't need a deprecation, right?

chenson2018 (approved):
I left one small comment but otherwise am approving and delegating to you for merging. Thanks to (and @lengyijun too) for cleaning this up!

lengyijun:
@chenson2018 updated

Cslib/Foundations/Relation/Defs.lean
thomaskwaring: ah yes you're right, thanks

PR #887, opened 2026-09-08
feat(Computability/Languages): regular languages have matching regular expressions through Kleene’s Algorithm

chiyunhsu:
We introduce `IsRegular.iff_regex`, a language is regular if and only if it matches a regular expression. It is a combination of the existing `IsRegular.regex`, a language matching a regular expression is regular and our new direction. The new direction is proven through Kleene’s algorithm.

It is the result that was anticipated in [PR#846](https://github.com/leanprover/cslib/pull/846) by @ctchou.       

The pull request introduces a new file, `KleeneAlgorithm.lean`. In it, we proved a language defined by a DFA on `Fin n` with one accepting state has a matching regular expression. The algorithm applies induction on a bound which restricts which interior states that a run may pass through.  

Implemented collaboratively by Brooke Gill and Chi-Yun Hsu


Cslib/Computability/Languages/KleeneAlgorithm.lean
```diff
@@ -0,0 +1,539 @@
...
+    {flts.mtr s xs} ∪ PathSupp flts s xs ∪ PathSupp flts (flts.mtr s xs) ys := by
+  induction xs generalizing s with
+  | nil => grind [PathSupp]
+  | cons a xs ih =>
+  rw [List.cons_append, pathSupp_head (by simp [hxs.2])]
+  by_cases hx : xs = []
+  · grind [PathSupp]
+  · grind [pathSupp_head hx]
+
+end PathSupp
```
ctchou, line 90: Instead of defining this specialized notion `PathSupp`, I think you should define the following general notion:
```
namespace Cslib.FLTS
def execution (flts : FLTS State Label) (s : State) : List Label → List State
  | [] => [s]
  | μ :: μs => s :: flts.execution (flts.tr s μ) μs
end Cslib.FLTS
```
which you can place in the file `FLTS/Basic.lean`.  So far as I can tell, you need `PathSupp` only to define `BddPath` in the following expression:
```
∀ i ∈ PathSupp p.toFLTS p.start xs, i < p.bound
```
whch you can express as:
```
let ss := p.toFLTS.execution p.start xs
∀ k, ∀ _ : 0 < k ∧ k + 1 < ss.length, ss[k] < p.bound
```
Other expressions involving membership in `PathSupp` can be similarly translated.

I make this suggestion because `execution` is a cleaner notion which will have simpler properties and can have more general applications in the future.  The above expression is also a more direct translation of the informal proof.  Lean's `grind` tactic is very good at reasoning about simple properties of natural numbers, so you shouldn't have any problem working with list indices.

Also, this whole section should be in the namespace `Cslib.FLTS`.

Cslib/Computability/Languages/KleeneAlgorithm.lean
```diff
@@ -0,0 +1,539 @@
...
+
+open Automata Acceptor
+
+variable {n : ℕ}
+
+/-- A Bounded Path (`BddPath`) has states `Fin n` and accepts strings (lists of symbols)
+starting with state `start` and ending with state `finish`
+with the interior states less than `bound`. -/
+structure BddPath (n : ℕ) (Symbol : Type*) extends FLTS (Fin n) Symbol where
```
ctchou: I think `BddPathFLTS` would be a better name.

Cslib/Computability/Languages/KleeneAlgorithm.lean
```diff
@@ -0,0 +1,539 @@
...
+
+/-- Starting at state `i`, the function `splitLast` sends a string to its longest prefix
+ending at state `k`.
+If the string ends at state `k`, then `splitLast` returns the original string.
+If the string never passes through state `k` (starting state can be `k`),
+then `splitLast` returns the empty string. -/
+def splitLast (flts : FLTS (Fin n) Symbol) (i k : Fin n) : List Symbol → List Symbol
+  | [] => []
+  | a :: x => if (splitLast flts (flts.tr i a) k x = []) ∧ flts.tr i a ≠ k then []
+  else a :: splitLast flts (flts.tr i a) k x
```
ctchou, line 140: See my comment about `splitFirst` below.

ctchou (changes requested):
This is the first batch of my comments.  I will have more in the future.

My general point is that by defining `FLTS.execution`, you can express directly the concepts in the textbook proof, rather than using bespoke recursive definitions.  This allows you to leverage the large number of results about lists in mathlib.

Cslib/Computability/Languages/KleeneAlgorithm.lean
```diff
@@ -0,0 +1,539 @@
...
+section splitFirst
+
+/-- Starting from a state `i`, the function `splitFirst` sends a string to its shortest prefix
+ending at state `k`.
+The string is empty if and only if its `splitFirst` is empty.
+If the string never passes through state `k` (starting state can be `k`),
+then `splitFirst` returns the original string. -/
+def splitFirst (flts : FLTS (Fin n) Symbol) (i k : Fin n) : List Symbol → List Symbol
+  | [] => []
+  | a :: x => if flts.tr i a = k then [a] else a :: splitFirst flts (flts.tr i a) k x
```
ctchou, line 333: Using the notion of `FLTS.execution` mentioned above, I think `splitFirst` can be defined more directly without recursion: just use `List.findIdx` or its variants to search for `k` in `FLTS.execution` and then use the returned index to extract the sub-list via `List.take`.  For `splitFirstCompl`, use `List.drop` instead of `List.take`.  Mathlib has a lot of theorems about `List.{findIdx,take,drop}` which I suspect can simplify many of your proofs or even make some of your theorems unnecessary.

For `splitLast[Compl]`, reverse `FLTS.execution` before applying `List.findIdx`.

Alternatively, `Nat.nth` and `Nat.count` let you talk about the occurrences of `k` directly:
https://leanprover-community.github.io/mathlib4_docs/Mathlib/Data/Nat/Nth.html

ctchou (changes requested):

Cslib/Computability/Languages/KleeneAlgorithm.lean
ctchou: BTW, you may want to consider doing the proof using NFA rather than DFA.  (Dexter Kozen's textbook has the proof.)  `LTS.Execution` is already defined and a number of theorems proved about it.  I suspect there may be less "noise" in that setup.

chiyunhsu: I agree with the suggestion of using `Execution` than `PathSupp`.

I am now contemplating on the reasons whether NFA approach would be more preferable than DFA approach.
More preferble:
- `LTS.Execution` is defined. Defining FLTS.Execution duplicates the code (unless there might other usages? I feel that FLTS.Execution would not be huge.)
- Explicit NFA -> Regex rather than implicit NFA -> DFA -> Regex as currently is (but is it needed?)

Less preferable:
- The current `LTS/Execution.lean` has no lemmas about bounding the states. Everything should be rewritten and strengthened, so back to the previous question, is explicit NFA -> Regex so preferable? And would this strengthened (probably longer codes) worth not defining FLTS.Execution?

I am all ears to hear what you think before we start to rewrite the codes for better applications in the future.

ctchou: It seems to me that the crux of the proof is to formalize Figure 3.3 of Hopcroft et al's textbook.  Given an execution in the sense of `LTS.Execution` and the additional assumption that no state in the execution is `> k`, you can obtain the indices of the occurrences of `k` using `Nat.nth p`, where the predicate `p` is something like:
```
fun j => ∃ _ : 0 < j ∧ j + 1 < ss.length, ss[j] = k
```
Then using those indices you can extract the relevant segments of `ss` and `xs` using `List.extract` (which degenerates to `List.take` and `List.drop` for the first and last of them).  Then you need to show that those extracted segments are `LTS.Execution` in their own right and have the necessary properties to make them members of the various `R_ij^(k)` sets.  But there are a lot of results about `Nat.nth`, `List.extract`, `LTS.Execution`, etc in mathlib and cslib that you can leverage and Lean's `grind` tactic is very good at reasoning about list indices.

Alternatively, you can continue to use concepts like `splitFirst` and `splitLast`.  Here I would suggest that instead of splitting the list directly, you use `List.findIdx` to find the index of the occurrence of `k` and then split the list using `List.take` and `List.drop` with that index.  Your theorems like `split{First,Last}_append` are then just instances of the existing theorem `List.take_append_drop`.

chiyunhsu: Thanks for making more detailed suggestions. To clarify, it sounds like there are multiple routes to change the code:
1. Define new `FLTS.Execution` to replace the role of `PathSupp`.
2. Use existing `LTS.Execution` to replace the role of `PathSupp`. This will involve `FLTS.toLTS`.
3. Upgrade the PR to prove Kleene's algorithm for NFA -> Regex. Use existing `LTS.Execution`.

Which route are you suggesting us to take?

ctchou: Number 3.  I think the determinism of FLTS is a distraction.

The hard part of the proof is to identify the occurrences of `k` in an execution and reason about them.  It doesn't really matter whether the execution is from a deterministic or nondeterministic machine.  There just happens to be more infrastructure already developed for the latter in cslib.

Cslib/Computability/Languages/KleeneAlgorithm.lean
```diff
@@ -0,0 +1,334 @@
...
+    ∃ ss : List (Fin n),
+    p.toLTS.Execution p.start xs p.last ss ∧
+    ∀ idx, ∀ _ : 0 < idx ∧ idx + 1 < ss.length, ss[idx] < p.bound
+    -- ∀ s ∈ ss.tail.dropLast, s < p.bound
+
+theorem language_bddpath_eq_nfa (lts : LTS (Fin n) Symbol) (i j : Fin n) {k : ℕ} (hk : n ≤ k) :
+    language (BddPathLTS.mk lts i j k) =
+    language (NA.FinAcc.mk {Tr := lts.Tr, start := {i}} {j}) := by
+  simp [language, Accepts]
+  grind
```
ctchou, line 81: I think this whole block of code can be replaced by:
```
namespace Cslib.LTS

variable {Symbol : Type*} {n : ℕ}

def BddExec (lts : LTS (Fin n) Symbol) (bound : ℕ)
    (start : Fin n) (xs : List Symbol) (last : Fin n) (ss : List (Fin n)) : Prop :=
  lts.Execution start xs last ss ∧ ss.AllButFirstLast (· < bound)

def bddLang (lts : LTS (Fin n) Symbol) (bound : ℕ) (start last : Fin n) : Language Symbol :=
  { xs | ∃ ss, lts.BddExec bound start xs last ss }

end Cslib.LTS
```
Note that your main proof will refer to different `start`s, `last`s, and `bound`s in the same statement, so it is better to have them as explicit parameters rather than as fields of `structure`s which you need to construct and destruct all the time.

Cslib/Computability/Languages/KleeneAlgorithm.lean
```diff
@@ -0,0 +1,334 @@
...
+  /-- The last state of the path. -/
+  last : Fin n
+  /-- The bound for interior states of the path. -/
+  bound : ℕ
+
+instance : Acceptor (BddPathLTS n Symbol) Symbol where
+  Accepts (p : BddPathLTS n Symbol) (xs : List Symbol) :=
+    ∃ ss : List (Fin n),
+    p.toLTS.Execution p.start xs p.last ss ∧
+    ∀ idx, ∀ _ : 0 < idx ∧ idx + 1 < ss.length, ss[idx] < p.bound
```
ctchou, line 74: Just a nitpick: use `i` instead of `idx`.  I understand that you want to use `i`, `j`, `k` etc to denote states from `Fin n`.  But I think it is clearer to keep denoting the latter using `s`, `t`, `r`, etc.

Cslib/Computability/Languages/KleeneAlgorithm.lean
```diff
@@ -0,0 +1,334 @@
...
+ending at state `k`. -/
+def splitFirst {lts : LTS (Fin n) Symbol} {i j : Fin n} {xs : List Symbol} {ss : List (Fin n)}
+    (_ : LTS.Execution lts i xs j ss) (k : Fin n) : List Symbol :=
+    xs.take (ss.tail.findIdx (· = k) + 1)
+
+/-- Starting at state `i`, the function `splitFirstCompl` sends a string to its longest suffix
+starting at state `k`. -/
+def splitFirstCompl {lts : LTS (Fin n) Symbol} {i j : Fin n} {xs : List Symbol} {ss : List (Fin n)}
+    (_ : LTS.Execution lts i xs j ss) (k : Fin n) : List Symbol :=
+    xs.drop (ss.tail.findIdx (· = k) + 1)
```
ctchou, line 138: It is in general a good idea to include only those parameters which are actually needed:
```
def splitFirst (xs : List Symbol) (ss : List (Fin n)) (k : Fin n) : List Symbol :=
  xs.take (ss.findIdxButFirst (· = k))

def splitFirstCompl (xs : List Symbol) (ss : List (Fin n)) (k : Fin n) : List Symbol :=
  xs.drop (ss.findIdxButFirst (· = k))
```
The same comment applies to `SplitLast[Compl]` above.

Cslib/Computability/Languages/KleeneAlgorithm.lean
```diff
@@ -0,0 +1,334 @@
...
+
+theorem language_bddpath_eq_nfa (lts : LTS (Fin n) Symbol) (i j : Fin n) {k : ℕ} (hk : n ≤ k) :
+    language (BddPathLTS.mk lts i j k) =
+    language (NA.FinAcc.mk {Tr := lts.Tr, start := {i}} {j}) := by
+  simp [language, Accepts]
+  grind
+
+open List
+
+section splitLast
```
ctchou, line 85: I think it is probably better to put the `splitLast` section after the `splitFirst` section, because the latter seems easier to reason about than the former.

ctchou (changes requested):
I think you can benefit from having the following functions about `List`:
```
namespace List

variable {α : Type*}

def AllButFirstLast (p : α → Bool) (as : List α) : Prop :=
  ∀ i, ∀ _ : 0 < i ∧ i + 1 < as.length, p as[i]

def findIdxButFirst (p : α → Bool) (as : List α) : ℕ :=
  as.tail.findIdx p + 1

def revFindIdxButLast (p : α → Bool) (as : List α) : ℕ :=
  as.length - as.reverse.findIdxButFirst p

end List
```
Then not only a lot of your definitions can be simplified (see the inline comments), but you can state and prove properties about the above `List` functions without distractions.

I have not looked into the proofs in detail.

PR #888, opened 2026-09-09
feat(Automata): Two-way automata accept exactly the regular languages

crei:
This implements Vardi's construction of a (one-way) finite automaton that accepts the complement of the language accepted by a two-way automaton. Together with closure of regular languages under complement and a simple mapping of one-way automata to two-way automata we get that two-way automata accept exactly the regular languages.

AI disclosure: Claude was used throughout bit in a tight review loop.


Cslib/Computability/Automata/TwoWayNA/Regular.lean
```diff
@@ -0,0 +1,40 @@
...
+
+namespace Cslib.Language
+
+open Automata Acceptor
+
+/-- A language is regular if and only if it is accepted by some two-way nondeterministic
+automaton with finitely many states. -/
+public theorem IsRegular.iff_twoWayNA {Symbol : Type*} {l : Language Symbol} :
+    l.IsRegular ↔ ∃ State : Type, ∃ _ : Finite State,
+      ∃ a : Automata.TwoWayNA State Symbol, language a = l := by
```
ctchou: The `Automata.` prefix is unnecessary.

Cslib/Computability/Automata/TwoWayNA/ComplToNA.lean
```diff
@@ -0,0 +1,326 @@
...
+Authors: Christian Reitwiessner
+-/
+
+module
+
+public import Cslib.Computability.Automata.NA.Basic
+public import Cslib.Computability.Automata.TwoWayNA.Basic
+public import Cslib.Foundations.Semantics.LTS.Relation
+
+/-! # A finite acceptor for the complement of the language of a two-way automaton
```
ctchou, line 14: I think this file should be renamed to `ComplToNA.lean`.

Cslib/Computability/Automata/TwoWayNA/ComplToNA.lean
```diff
@@ -0,0 +1,326 @@
...
+
+For every nondeterministic two-way automaton (`TwoWayNA`) `a`, this file constructs a
+nondeterministic finite acceptor (`NA.FinAcc`) that accepts exactly the words rejected by `a`
+(`TwoWayNA.toNAComplement`, `TwoWayNA.language_toNAComplement`). We follow Vardi's proof, which --
```
ctchou: I think `complToNA` is better than `toNAComplement`.

Cslib/Computability/Automata/TwoWayNA/Basic.lean
```diff
@@ -108,6 +109,23 @@ lemma TwoWayNA.toCfgNAFinAcc_input_eq {State Symbol : Type*} (a : TwoWayNA State
...
+theorem TwoWayNA.getElem_of_tr {a : TwoWayNA State Symbol} {input : List Symbol}
+    {c c' : TwoWayNACfg State Symbol} {x : Symbol} {m : SignType}
+    (htr : (a.toCfgNAFinAcc input).Tr c (x, m) c') (hc : c.input = input) :
+    ∃ h : (c.pos : ℕ) < input.length, input[(c.pos : ℕ)] = x := by
+  obtain ⟨-, hx, -, -⟩ := htr
+  subst hc
+  exact List.getElem?_eq_some_iff.mp hx.symm
+
 @[simp, scoped grind =]
 instance : Acceptor (TwoWayNA State Symbol) Symbol where
```
ctchou: I would move this `instance` definition to immediately after the `TwoWayNA.toCfgNAFinAcc` definition, before starting proving theorems.

Cslib/Computability/Automata/TwoWayNA/OfNA.lean
```diff
@@ -0,0 +1,105 @@
...
+    have hdrop' : input.drop (p + 1) = xs := by simp [← List.tail_drop, hdrop]
+    have hstep : ((ofNA n).toCfgNAFinAcc input).Tr
+        ⟨input, s, ⟨p, hp⟩⟩ (x, SignType.pos) ⟨input, t, ⟨p + 1, by omega⟩⟩ :=
+      ⟨rfl, by simp [← hx], ⟨rfl, htr⟩, by simp⟩
+    obtain ⟨μs, hmtr'⟩ := ih (by omega) hdrop' hmtr
+    exact ⟨(x, SignType.pos) :: μs, LTS.MTr.cons_iff.mpr ⟨_, hstep, hmtr'⟩⟩
+
+/-- A nondeterministic finite acceptor and its two-way rendering accept the same words. -/
+theorem accepts_ofNA_iff (a : NA.FinAcc State Symbol) (input : List Symbol) :
+    Acceptor.Accepts (ofNA a) input ↔ Acceptor.Accepts a input := by
```
ctchou: `open Acceptor`, so that all the `Acceptor.` prefixes become unnecessary.

Cslib/Computability/Automata/TwoWayNA/OfNA.lean
```diff
@@ -0,0 +1,105 @@
...
+
+variable {State Symbol : Type*} {input : List Symbol}
+
+namespace TwoWayNA
+
+variable {n : NA.FinAcc State Symbol}
+
+/-- The two-way automaton that performs the transitions of the nondeterministic finite acceptor
+`n`, always moving its head one symbol to the right. -/
+def ofNA (n : NA.FinAcc State Symbol) : TwoWayNA State Symbol where
```
ctchou: Instead `TwoWayNA.ofNA`, it may be better to define `NA.FinAcc.toTwoWayNA`.  Then you can use the dot-notation and write `n.toTwoWayNA`, instead of the cumbersome `(ofNA n)`.  The theorems can keep their current namespace.

Also, I find it a bit confusing to use a single letter `n` to refer to the NFA, because it is common to use `n` to refer to natural numbers.  I think `na` or even `a` would be better.

Cslib/Computability/Automata/TwoWayNA/OfNA.lean
```diff
@@ -0,0 +1,105 @@
...
+  Tr q x m q' := m = SignType.pos ∧ n.Tr q x q'
+  start := n.start
+  accept := n.accept
+
+/-- A run of `ofNA n` starting on `input` reads a multistep transition of `n` over the prefix of
+`input` scanned so far. -/
+theorem mTr_take_of_canReach {s : State} {c c' : TwoWayNACfg State Symbol}
+    (hreach : ((ofNA n).toCfgNAFinAcc input).CanReach c c') (hc : c.input = input)
```
ctchou: I'm a bit puzzled by the use of `CanReach` in this and the next theorems.  Isn't there be a direct correspondence between the prefixes of `input` and the transition label sequences of `(ofNA n).toCfgNAFinAcc input`?  Using that correspondence, you should be able to use `((ofNA n).toCfgNAFinAcc input).MTr`, which is clearer.

Cslib/Computability/Automata/TwoWayNA/ComplToNA.lean
```diff
@@ -0,0 +1,326 @@
...
+
+namespace List
+
+/-- Dropping the head of a list shifts total indexing by one. -/
+@[simp]
+private theorem getI_tail {α : Type*} [Inhabited α] (l : List α) (i : ℕ) :
+    l.tail.getI i = l.getI (i + 1) := by
+  cases l <;> simp
+
+end List
```
ctchou: Create `Foundations/Data/List/Misc.lean` and put this theorem there.

Cslib/Computability/Automata/TwoWayNA/Basic.lean
```diff
@@ -27,7 +27,8 @@ ends in an accepting state with the head just past the end of the input.
 
 * `TwoWayNA`, the automaton itself
 * `TwoWayNACfg`, a configuration of a `TwoWayNA`: Its input plus a state and the head position.
-* `TwoWayNA.Step`, The single-step relation between configurations.
+* `TwoWayNA.toCfgNAFinAcc`, the finite acceptor on configurations whose runs on a fixed input are
```
ctchou: Now that I see how it is used in practice, I think `toCfgNAFinAcc` is too long and `toCfgNA` is good enough.

Cslib/Computability/Automata/TwoWayNA/ToNA.lean
```diff
@@ -0,0 +1,326 @@
...
+
+namespace TwoWayNA
+
+/-! ## Vardi's condition of non-acceptance -/
+
+/-- Every step of `a` on `input` out of a state that `T` attaches to the head position lands in a
+state that `T` attaches to the new head position. The conjunct on the input restricts the
+invariant to the configurations that run on `input`. -/
+def IsStepClosed (a : TwoWayNA State Symbol) (input : List Symbol) (T : ℕ → Set State) : Prop :=
+  (a.toCfgNAFinAcc input).TrInv (fun c => c.input = input ∧ c.state ∈ T c.pos)
```
ctchou: Given the theorem `TwoWayNA.toCfgNAFinAcc_input_eq`, is the conjunct `c.input = input` necessary?

Cslib/Computability/Automata/TwoWayNA/Basic.lean
ctchou: Sorry, I should have noticed this earlier: the `.toLTS` is unnecessary.

Cslib/Computability/Automata/TwoWayNA/ComplToNA.lean
```diff
@@ -0,0 +1,326 @@
...
+Fix a `TwoWayNA` `a` and an input word `input` of length `n`. A *rejection certificate* is a family
+of subsets `T i ⊆ State`, one for every head position `i ∈ {0, …, n}`, subject to three
+conditions:
+
+1. `T` contains every initial state at position `0` (`IsRejectionCert.start_mem`);
+2. `T` is an invariant of the transitions of `a`: if the state `c.state` is in `T c.pos` and `a`
+   can step from the configuration `c` to the configuration `c'`, then `c'.state` is in `T c'.pos`
+   (`TwoWayNA.IsStepClosed`, `IsRejectionCert.step_closed`);
+3. no state in `T n`, i.e. at the position just past the end of the input, is accepting
+   (`IsRejectionCert.accept_notMem`).
```
ctchou, line 36: Personally I find the single-letter name `T` a bit too cryptic.  Since this is a key concept of the proof, perhaps we can have a more meaningful name?  Say `rej` or `rejCert`?  Also note that the naming convention require this name to begin with a lowercase letter.

Cslib/Computability/Automata/TwoWayNA/ComplToNA.lean
```diff
@@ -0,0 +1,326 @@
...
+  /-- The family is an invariant of the transitions of `a`. -/
+  step_closed : a.IsStepClosed input T
+  /-- No accepting state occurs past the end of the input. -/
+  accept_notMem : ∀ s ∈ T input.length, s ∉ a.accept
+
+variable {T : ℕ → Set State}
+
+/-- If a rejection certificate for `input` exists, then `a` does not accept `input`. -/
+theorem IsRejectionCert.not_accepts (hT : a.IsRejectionCert input T) :
+    ¬ Acceptor.Accepts a input := by
```
ctchou: `open Acceptor` so that you don't need the `Acceptor.` prefix.

Cslib/Computability/Automata/TwoWayNA/Regular.lean
```diff
@@ -0,0 +1,40 @@
...
+Copyright (c) 2026 Christian Reitwiessner. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Christian Reitwiessner
+-/
+
+module
+
+public import Cslib.Computability.Automata.TwoWayNA.OfNA
+public import Cslib.Computability.Automata.TwoWayNA.ToNA
+public import Cslib.Computability.Languages.RegularLanguage
```
ctchou: Since Vardi's proof doesn't use the Myhill-Nerode theorem, I think you can move the contents of this file into `Cslib.Computability.Languages.RegularLanguage` itself, rather than creating a new dependency.

Cslib/Computability/Automata/TwoWayNA/ComplToNA.lean
```diff
@@ -0,0 +1,326 @@
...
+
+/-- The subset that `T` attaches to the position to the left of `i`, and everything at position
+`0`, which has no position to its left. -/
+def prevSet (T : ℕ → Set State) : ℕ → Set State
+  | 0 => Set.univ
+  | i + 1 => T i
+
+/-- Every move of `a` out of a state in `C` while reading `x` lands in `P`, in `C` or in `N`,
+according to whether it moves the head to the left, keeps it in place, or moves it to the right. -/
+def LocalOK (a : TwoWayNA State Symbol) (x : Symbol) (P C N : Set State) : Prop :=
```
ctchou: The naming convention says names of things of type `Set State` should begin with a lowercase letter.

Also, it is a bit confusing that `.neg => P` and `.pos => N` reverse the first letters.  Perhaps instead of `P C N`, we have `ns cs ps`?

Cslib/Computability/Automata/TwoWayNA/ComplToNA.lean
```diff
@@ -0,0 +1,326 @@
...
+theorem not_accepts_iff_exists_isRejectionCert (a : TwoWayNA State Symbol)
+    (input : List Symbol) :
+    ¬ Acceptor.Accepts a input ↔ ∃ T, a.IsRejectionCert input T :=
+  ⟨fun h => ⟨_, isRejectionCert_reachable h⟩, by rintro ⟨_, hT⟩; exact hT.not_accepts⟩
+
+/-! ## Localising the closure condition -/
+
+/-- The subset that `T` attaches to the position to the left of `i`, and everything at position
+`0`, which has no position to its left. -/
+def prevSet (T : ℕ → Set State) : ℕ → Set State
```
ctchou: I think you should consider using `ωSequence (Set State)` as the type of `T`.  Then `prevSet T` can be written as `Set.univ ::ω T`, which is clearer.  See also the comment about `certToList` below.

Cslib/Computability/Automata/TwoWayNA/ComplToNA.lean
```diff
@@ -0,0 +1,326 @@
...
+
+/-! ## The finite acceptor for the complement -/
+
+/-- The nondeterministic finite acceptor that guesses a rejection certificate `T` for `a` while
+scanning the input, keeping the pair `(T (i - 1), T i)` in its state after reading `i` symbols.
+Reading the symbol at position `i` guesses `T (i + 1)` and checks local consistency at `i`. -/
+def toNAComplement (a : TwoWayNA State Symbol) : NA.FinAcc (Set State × Set State) Symbol where
+  Tr PC x PC' := PC'.1 = PC.2 ∧ a.LocalOK x PC.1 PC.2 PC'.2
+  start := {PC | PC.1 = Set.univ ∧ a.start ⊆ PC.2}
+  accept := {PC | ∀ s ∈ PC.2, s ∉ a.accept}
```
ctchou: Names of objects of type `Set State` should begin with lowercase letters.

Cslib/Computability/Automata/TwoWayNA/ComplToNA.lean
```diff
@@ -0,0 +1,326 @@
...
+
+/-- The family of subsets carried by a list, which holds the subset for the position to the left
+of `0` in front, so that position `i` is entry `i + 1` of the list. -/
+def certOfList (T : List (Set State)) (i : ℕ) : Set State :=
+  T.getI (i + 1)
+
+/-- The subsets that `T` attaches to the positions of the input head, as a list, prefixed by
+`Set.univ` for the position to the left of `0`. -/
+def certToList (input : List Symbol) (T : ℕ → Set State) : List (Set State) :=
+  Set.univ :: (List.range (input.length + 1)).map T
```
ctchou: If the type of `T` is `ωSequence (Set State)`, then we can write:
```
Set.univ :: T.take (input.length + 1)
```

ctchou (changes requested):

Cslib/Computability/Automata/TwoWayNA/ToNA.lean
crei: I think I tried and it was more complicated, but I agree, it's probably better to make the definition cleaner even if the proofs get a few lines longer.

Cslib/Computability/Automata/TwoWayNA/ComplToNA.lean
crei: The naming scheme is "Previous / Current / Next", but probably the solution is to just spend some more characters. Also they should be lowercase, right?

Cslib/Computability/Automata/TwoWayNA/OfNA.lean
crei: I changed the name, but I'm not sure about the name of the file now. `toTwoWayNA.lean` is a bit weird in a directory called 'TwoWayNA'.

Cslib/Computability/Automata/TwoWayNA/ToNA.lean
crei: Actually the conjunct is necessary because a `toCfgNAFinAcc` can have configurations/states with a different input. I think the clean fix here would be to take `input` out of the configuration and make it a parameter, so that we have a guarantee that any state of a `toCfgNAFinAcc` is on the same input.

Cslib/Computability/Automata/TwoWayNA/OfNA.lean
ctchou: I think it's fine. The full pathname of the file is `.../TwoWayNA/OfNA.lean`, which is a good description of its contents.

Cslib/Computability/Automata/TwoWayNA/ToNA.lean
ctchou: OK, I see.

ctchou (approved):
LGTM now.

ctchou:
BTW, it would be nice if @crei can also add the result that 2-way automata are essentially the same as Turing machines with read-only tapes and hence the latter also accept the regular languages.

PR #889, opened 2026-09-09
ci(nightly-testing): pass ZULIP_API_KEY through env instead of interpolating into script text

jessealama:
Move the key into the step's `env:` block and reference it as a shell variable, matching how the step already receives its other values.


chenson2018 (approved):
I think this workflow will be done away with entirely in the near future, but no harm in making it better in the meantime. Thanks!

PR #890, opened 2026-09-10
feat(Circuit): prove Lupanov upper bound

SamuelSchlesinger:
Add Boolean synthesis with sharing and prove the uniform (1 + ε) 2^n/n upper bound for De Morgan circuits.

Assisted by Codex, adapted from https://github.com/samuelSchlesinger/algebraic-circuits.


Cslib/Computability/Circuit/Boolean/Basic.lean
```diff
@@ -0,0 +1,59 @@
...
+Authors: Samuel Schlesinger
+-/
+module
+
+public import Cslib.Computability.Circuit.Basic
+
+/-!
+# Boolean circuits
+
+The De Morgan basis consists of binary AND and OR, unary NOT, and Boolean constants.
```
crei, line 13: So this is essentially boolean circuits with constant fan-in, right?

SamuelSchlesinger: That's right, fan-in 2 in particular. You can see the arities below.

Cslib/Computability/Circuit/Boolean/Synthesis.lean
```diff
@@ -0,0 +1,178 @@
...
+
+@[expose] public section
+
+namespace Cslib.Circuits.Boolean
+
+universe u
+variable {n : ℕ} {ι : Type u}
+
+/-- Functions available on the input or gate wires of a program. -/
+def available (p : Σ g, Program signature n g) : Set (BooleanFunction n) :=
```
crei: I'm a bit confused by the sum type here, also in `Synthesis`.
What about the following:
```lean4

/-- Functions available on the input or gate wires of a program. -/
def available {g : Nat} (p : Program signature n g) : Set (BooleanFunction n) :=
  {f | ∃ w, ∀ x, p.trace interpretation x w = f x}

/-- At most `cost` additional gates suffice to compute `targets` from `sources`, while
preserving all functions already available in the starting program. -/
def Synthesis (sources targets : Set (BooleanFunction n)) (cost : ℕ) : Prop :=
  ∀ g₁ (p₁ : Program _ _ g₁), sources ⊆ available p₁ →
    ∃ g₂, ∃ (p₂ : Program _ _ g₂), g₂ ≤ g₁ + cost ∧
    available p₁ ⊆ available p₂ ∧ targets ⊆ available p₂
```

Maybe `Program` could also have a function that returns its gate count (even if it is part of the type) - then we would not need g₁ and g₂?

Cslib/Computability/Circuit/Boolean/Synthesis.lean
```diff
@@ -0,0 +1,178 @@
...
+  {f | ∃ w, ∀ x, p.2.trace interpretation x w = f x}
+
+/-- At most `cost` additional gates suffice to compute `targets` from `sources`, while
```
crei: This docstring reads super compressed and the definition is rather tricky. Maybe it's better to spell out the exact definition in english terms than trying to simplify it?

Cslib/Computability/Circuit/Boolean/LupanovConstruction.lean
```diff
@@ -0,0 +1,189 @@
...
+-/
+module
+
+public import Cslib.Computability.Circuit.Boolean.Synthesis
+public import Mathlib.Data.Fintype.BigOperators
+
+/-!
+# Lupanov's block construction
+
+Split the truth table into address rows and data columns, and divide the rows into blocks.
```
crei: Could you also explain what the purpose of this file is and not just what it does?

Cslib/Computability/Circuit/Boolean/LupanovConstruction.lean
```diff
@@ -0,0 +1,189 @@
...
+
+noncomputable section
+variable {k d s : ℕ}
+
+private abbrev Assignment (k : ℕ) := Fin k → Bool
+
+private def index (k : ℕ) : Assignment k ≃ Fin (2 ^ k) :=
+  Fintype.equivOfCardEq (by simp)
+
+private def minterm : Assignment k ⊕ Assignment d → BooleanFunction (k + d)
```
crei, line 89: Can you add some documentation starting from here?

Cslib/Computability/Circuit/Boolean/Synthesis.lean
SamuelSchlesinger: Adopted your definition. The `∃ g₂` stays since the gate count is a type index, but `available` now takes a `Program` directly and is just `Set.range (p.wireFunction interpretation)`.

Cslib/Computability/Circuit/Boolean/Synthesis.lean
SamuelSchlesinger: Rewrote the docstring to spell out the quantifier structure and why the starting program is arbitrary. Module doc expanded likewise.

Cslib/Computability/Circuit/Boolean/Basic.lean
SamuelSchlesinger: Made this explicit: the module doc now states fan-in at most two, and `Program.fanInAtMost_two` / `Circuit.fanInAtMost_two` prove it against the generic `FanInAtMost` predicate.

Cslib/Computability/Circuit/Boolean/LupanovConstruction.lean
SamuelSchlesinger: Rewrote the module doc to lead with the purpose of the file, then the truth-table picture, then where the gate count comes from. `Lupanov.lean` got a matching section on the parameter choice.

Cslib/Computability/Circuit/Boolean/LupanovConstruction.lean
SamuelSchlesinger: Added docstrings from `index` on, with section headers for minterms, the block decomposition, gate counts, and correctness. Same for the lemmas in `Lupanov.lean`.

Cslib/Computability/Circuit/Boolean/LupanovConstruction.lean
```diff
@@ -0,0 +1,286 @@
...
+    {fun x => decide (∃ data ∈ support f block pattern, minterm (.inr data) x = true)} _
+  simpa using Synthesis.exists_mem (support f block pattern)
+    (fun data => minterm (.inr data)) (fun _ => 0)
+    (fun data _ => minterm_available (.inr data))
+
+/-! ### Correctness -/
+
+/-- `left` is true exactly at inputs whose address is a row of the block, at an offset where
+the pattern is `1`. -/
+private theorem left_eq_true (block : ℕ) (pattern : Assignment s) (x : Assignment (k + d)) :
```
crei: Maybe move these closer to the definition so that the meaning is clear earlier on?

Cslib/Computability/Circuit/Boolean/Lupanov.lean
```diff
@@ -0,0 +1,248 @@
...
+-/
+
+public section
+
+namespace Cslib.Circuits.Boolean.Lupanov
+
+open Filter
+
+/-- Polynomials are eventually dominated by `2 ^ n`. -/
+private theorem polynomial_le_pow (c r : ℕ) :
```
crei: If mathlib does not have this, this should go into some very fundamental foundations library in cslib and be very public :)

Cslib/Computability/Circuit/Boolean/Lupanov.lean
```diff
@@ -0,0 +1,248 @@
...
+    ∀ᶠ n : ℕ in atTop, c * n ^ r ≤ 2 ^ n := by
+  have h := Asymptotics.isLittleO_iff_nat_mul_le.mp
+    (isLittleO_pow_const_const_pow_of_one_lt (R := ℝ) r (by norm_num : (1 : ℝ) < 2)) c
+  filter_upwards [h] with n hn
+  exact_mod_cast (by simpa using hn : (c : ℝ) * n ^ r ≤ (2 : ℝ) ^ n)
+
+/-- Any constant multiple of `log₂ n` is eventually at most `n`. -/
+private theorem log_le (c : ℕ) : ∀ᶠ n : ℕ in atTop, c * Nat.log 2 n ≤ n := by
```
crei: Same with this.

Cslib/Computability/Circuit/Boolean/Lupanov.lean
```diff
@@ -0,0 +1,248 @@
...
+  have hh : c * Nat.log 2 n ≤ 2 ^ Nat.log 2 n := by
+    simpa using hN (Nat.log 2 n) (Nat.le_log_of_pow_le (by omega) hn)
+  exact hh.trans (Nat.pow_log_le_self 2 hn0)
+
+/-- With the chosen parameters, the budget is the leading term `(2 ^ k / s + 1) · 2 ^ d`
+plus error terms `3 n ^ 4`, from the address minterms, and `16 n · 2 ^ d`, from the data
+minterms, the per-pattern overhead of every block (`left` parts, constants, conjunctions and
+ORs), and the final constant. -/
+private theorem bound_le (n : ℕ) (hn : 5 * Nat.log 2 n < n) :
+    bound (3 * Nat.log 2 n) (n - 3 * Nat.log 2 n) (n - 5 * Nat.log 2 n) ≤
```
crei: Could use `Nat.log2` throughout here, it might make it a bit more readable.

Cslib/Computability/Circuit/Boolean/Lupanov.lean
```diff
@@ -0,0 +1,248 @@
...
+      (2 ^ (3 * Nat.log 2 n) / (n - 5 * Nat.log 2 n) + 1) *
+        2 ^ (n - 3 * Nat.log 2 n) + 3 * n ^ 4 +
+          16 * n * 2 ^ (n - 3 * Nat.log 2 n) := by
+  let l := Nat.log 2 n
+  let k := 3 * l
+  let d := n - 3 * l
+  let s := n - 5 * l
+  let B := 2 ^ k / s + 1
```
crei: Should be lowercase, but I assume this is the letter chose in the exposition, right?

Cslib/Computability/Circuit/Boolean/Lupanov.lean
```diff
@@ -0,0 +1,248 @@
...
+  have hn0 : 0 < n := by omega
```
crei: I wonder if some of these could be removed and replaced by `grind` at the place where they are used.

crei (changes requested):
Great contribution, I would only request some minor changes.

Cslib/Computability/Circuit/Boolean/LupanovConstruction.lean
SamuelSchlesinger: Moved `left_eq_true` and `right_eq_true` directly after their respective definitions.

Cslib/Computability/Circuit/Boolean/Lupanov.lean
SamuelSchlesinger: Added public `Nat.eventually_mul_pow_le_pow` in `Foundations/Data/Nat/Asymptotics.lean`, using Mathlib’s existing little-o result and generalizing to natural bases greater than one.

Cslib/Computability/Circuit/Boolean/Lupanov.lean
SamuelSchlesinger: Moved this alongside the polynomial bound as public `Nat.eventually_mul_log_le`, with a `Nat.eventually_mul_log2_le` corollary used here.

Cslib/Computability/Circuit/Boolean/Lupanov.lean
SamuelSchlesinger: Switched the asymptotic file to `Nat.log2` and its corresponding lemmas.

Cslib/Computability/Circuit/Boolean/Lupanov.lean
SamuelSchlesinger: Renamed it to `blockCount`.

Cslib/Computability/Circuit/Boolean/Lupanov.lean
SamuelSchlesinger: Removed the five routine opening facts and used `grind` where needed, keeping the main estimates named for readability.

Cslib/Computability/Circuit/Boolean/Synthesis.lean
barni120400, line 1: Note that almost all the content in this file is generic to circuits, and not specific to booleans.
I think that most of it should not be under the boolean directory because Synthesis and lemmas about it are for circuits in general.

SamuelSchlesinger: Yes, good point.

Cslib/Computability/Circuit/Boolean/LupanovConstruction.lean
```diff
@@ -0,0 +1,286 @@
...
+
+/-! ### Minterms -/
+
+/-- An assignment to `k` Boolean variables. Address assignments index the rows of the truth
+table and data assignments index its columns. -/
+private abbrev Assignment (k : ℕ) := Fin k → Bool
+
+/-- Number the address assignments as rows `0, …, 2 ^ k - 1`. Any bijection will do, so we
+take an arbitrary one; this is what makes the section noncomputable. -/
+private def index (k : ℕ) : Assignment k ≃ Fin (2 ^ k) :=
```
barni120400, line 84: This is the only thing forcing the `noncomputable section`, and Mathlib has a computable
bijection for it:

```lean
def index (k : ℕ) : Assignment k ≃ Fin (2 ^ k) :=
  (Equiv.arrowCongr (Equiv.refl (Fin k)) finTwoEquiv.symm).trans finFunctionFinEquiv
```

I checked and the proof can avoid noncomputable entirely

barni120400 (approved):

Cslib/Computability/Circuit/Boolean/LupanovConstruction.lean
SamuelSchlesinger: Fantastic, will make this change.

crei (approved):

SamuelSchlesinger:
paging @fmontesi

crei (approved):

PR #891, opened 2026-09-10
feat(Circuit): prove Shannon lower bound

SamuelSchlesinger:
Add semantic normalization and counting to prove the worst-case 2^n/n lower bound for De Morgan circuits.

Assisted by Codex, adapted from https://github.com/samuelSchlesinger/algebraic-circuits.


Cslib/Computability/Circuit/Boolean/Counting.lean
```diff
@@ -0,0 +1,208 @@
...
+
+public section
+
+namespace Cslib.Circuits.Boolean
+
+open scoped BigOperators
+
+variable {n g s : ℕ}
+
+/-- Boolean functions computable with at most `s` De Morgan gates. -/
```
crei:
```suggestion
/-- Boolean functions on `n` inputs computable with at most `s` De Morgan gates. -/
```

Cslib/Computability/Circuit/Boolean/Counting.lean
```diff
@@ -0,0 +1,208 @@
...
+noncomputable def computableFunctions (n s : ℕ) : Finset (BooleanFunction n) := by
```
crei: Would it be difficult to give an explicit definition (instead of using a tactic)?
I mean `mem_computableFunctions` gives a good definition so it's not too bad, but still a bit weird.

Cslib/Computability/Circuit/Boolean/Counting.lean
```diff
@@ -0,0 +1,208 @@
...
+  classical
+  exact Finset.univ.filter fun f => ∃ g ≤ s, ∃ c : Circuit signature n g 1, c.Computes f
+
+@[simp] theorem mem_computableFunctions {f : BooleanFunction n} :
+    f ∈ computableFunctions n s ↔ ∃ g ≤ s, ∃ c : Circuit signature n g 1, c.Computes f := by
+  classical
+  simp [computableFunctions]
+
+private noncomputable def irredundantFunctions (n g : ℕ) : Finset (BooleanFunction n) := by
```
crei: Can you give some intuition for this definition?

Cslib/Computability/Circuit/Boolean/Counting.lean
```diff
@@ -0,0 +1,208 @@
...
+  exact Finset.univ.filter fun f => ∃ c : Circuit signature n g 1,
+    c.Computes f ∧ Function.Injective (c.program.gateFunction interpretation)
+
+private theorem mem_irredundantFunctions {f : BooleanFunction n} :
+    f ∈ irredundantFunctions n g ↔ ∃ c : Circuit signature n g 1,
+      c.Computes f ∧ Function.Injective (c.program.gateFunction interpretation) := by
+  classical
+  simp [irredundantFunctions]
+
+private instance opFintype : Fintype Op where
```
crei: Maybe move to definition of `Op`, could be useful in general

Cslib/Computability/Circuit/Boolean/Counting.lean
```diff
@@ -0,0 +1,208 @@
...
+
+private theorem relabel_output (c : Circuit signature n g 1) (π : Equiv.Perm (Fin g))
+    (x : Fin n → Bool) :
+    Fin.addCases x (c.program.eval interpretation x ∘ π.symm) (relabel c π).2 =
+      c.eval interpretation x 0 := by
+  apply Wire.Renaming.value_apply
+  intro gate
+  simp [Wire.Renaming.ofPermutation, Function.comp_def]
+
+private noncomputable def representative (f : ↥(irredundantFunctions n g)) :
```
crei: Is it possible to do this without the `↥`?

Cslib/Computability/Circuit/Boolean/Shannon.lean
```diff
@@ -0,0 +1,135 @@
...
+      ne_of_gt (zero_lt_one.trans_le hs)] using
+      (Real.log_le_log (by positivity : 0 < (a : ℝ) * s.factorial) hcount)
+  apply (Real.log_le_iff_le_exp ha).mp
+  have hfactorial := mul_log_sub_le_log_factorial (s := s) (by omega)
+  have hlogtwo := Real.log_le_sub_one_of_pos (by norm_num : (0 : ℝ) < 2)
+  have hlogs := Real.log_le_self (by positivity : (0 : ℝ) ≤ s)
+  nlinarith
+
+private theorem eventually_inputs_le_budget :
+    ∀ᶠ n : ℕ in atTop, n + 1 ≤ 2 ^ n / n := by
```
crei: `∀ᶠ n : ℕ in atTop, n ≤ 2 ^ n / n` could also be a generic theorem

crei (changes requested):

Cslib/Computability/Circuit/Boolean/Counting.lean
SamuelSchlesinger: Applied the suggested docstring.

Cslib/Computability/Circuit/Boolean/Counting.lean
SamuelSchlesinger: Replaced both tactic blocks with direct `Finset.univ.filter` definitions using `open scoped Classical in`.

Cslib/Computability/Circuit/Boolean/Counting.lean
SamuelSchlesinger: Added a docstring explaining that the gates compute pairwise distinct functions, giving `g!` distinct relabelings for the counting bound.

Cslib/Computability/Circuit/Boolean/Counting.lean
SamuelSchlesinger: Moved the `Fintype Op` instance next to `Op` in `Boolean/Basic.lean` and made it public.

Cslib/Computability/Circuit/Boolean/Counting.lean
SamuelSchlesinger: Removed the explicit `↥` here and in the related declarations; Lean infers the coercion.

Cslib/Computability/Circuit/Boolean/Shannon.lean
SamuelSchlesinger: Extracted public `Nat.eventually_add_one_le_pow_div` into `Foundations/Data/Nat/Asymptotics.lean`, retaining the stronger `n + 1` bound and generalizing to natural bases greater than one.

Cslib/Computability/Circuit/Boolean/Basic.lean
```diff
@@ -36,6 +38,10 @@ inductive Op where
   | or
   deriving DecidableEq
 
+instance Op.instFintype : Fintype Op where
+  elems := {.const false, .const true, .not, .and, .or}
+  complete := by intro op; cases op <;> simp
```
barni120400: We can save the instance by deriving also Fintype and proving that the cardinality is 5

```suggestion
  deriving DecidableEq, Fintype

@[simp] theorem Op.card : Fintype.card Op = 5 := rfl
```

This needs the imports above swapped as well, and changing the proof in Cslib/Computability/Circuit/Boolean/Counting.lean:84 to just `simp`

SamuelSchlesinger: Fantastic! Thank you.

Cslib/Computability/Circuit/Boolean/Counting.lean
```diff
@@ -0,0 +1,207 @@
...
+  Finset.univ.filter fun f => ∃ c : Circuit signature n g 1,
+    c.Computes f ∧ Function.Injective (c.program.gateFunction interpretation)
+
+private theorem mem_irredundantFunctions {f : BooleanFunction n} :
+    f ∈ irredundantFunctions n g ↔ ∃ c : Circuit signature n g 1,
+      c.Computes f ∧ Function.Injective (c.program.gateFunction interpretation) := by
+  classical
+  simp [irredundantFunctions]
+
+private def lineEquiv (n g : ℕ) :
```
barni120400: the content of lines 64-84 is generic over `σ`. Only `Arity op ≤ 2` and `card Op = 5` are actually used.
I think that `lineEquiv` should be in Program.lean and the `Fintype` instance
plus the cardinality lemmas into a new standalone file under `Circuit/`

Cslib/Computability/Circuit/Boolean/Counting.lean
```diff
@@ -0,0 +1,207 @@
...
+    Line signature n g ≃ Σ op : Op, Fin (signature.Arity op) → Wire n g where
+  toFun l := ⟨l.op, l.wires⟩
+  invFun l := ⟨l.1, l.2⟩
+  left_inv _ := rfl
+  right_inv _ := rfl
+
+private noncomputable instance : Fintype (Line signature n g) :=
```
barni120400: The noncomputable doesn't look needed

Cslib/Computability/Circuit/Normalization.lean
barni120400, line 1: Maybe the file should define
`def Program.Irredundant (p : Program σ n g) (i : Interpretation σ U) : Prop :=
  Function.Injective (p.gateFunction i)`

and analogously for circuits?

Cslib/Computability/Circuit/Boolean/Counting.lean
crei: Did you? :)

crei (approved):

SamuelSchlesinger:
@fmontesi we can merge this one too

fmontesi (approved):

PR #892, opened 2026-09-10
ci: lint workflow files with actionlint

jessealama:
Adds a workflow that runs actionlint over the `.github/workflows` directory whenever those files change. actionlint checks expression types (used in the little GitHub Action DSL), undefined step/output references and needs/if conditions, and runs shellcheck over every `run:` shell block and [pyflakes](https://github.com/PyCQA/pyflakes) over `shell: python` chunks.

To make sure that adding this workflow does not immediately break the build, I did a manual one-off run and found one thing, which is fixed here: an `echo | sed` pipeline that is now a parameter expansion.

### DIFF
--- a/.github/workflows/actionlint.yml
+++ b/.github/workflows/actionlint.yml
@@ -1,0 +1,31 @@
+name: actionlint
+
+on:
+  push:
+    branches:
+      - main
+    paths:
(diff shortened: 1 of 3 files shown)


chenson2018 (approved):
LGTM, thanks!

PR #894, opened 2026-09-11
ci: pin zulip in a requirements file

jessealama:
Replace the bare `pip install zulip` calls with a pinned `.github/requirements.txt`. This is pretty small on its own (effectively a no-op), but pinning a version increases reproducibility and helps track issues upstream, should they arise.

This also gives Dependabot something to watch, once we've set that up.


chenson2018 (approved):
Thanks!

PR #895, opened 2026-09-11
ci: add Dependabot config for actions and pip

jessealama:
Add `.github/dependabot.yml` covering the two dependency sources CI actually has. Adding this will enable automatically created PRs if updates (checked weekly) are available, coming from two sources:

- GitHub Actions
- pip. Look only for minor and patch bumps. (Major versions are excluded, for now.)

Related to #894, which introduces the requirements file the pip entry watches. But mechanically, either one can merge first.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,25 @@
+version: 2
+updates:
+  # Keep action pins fresh. Grouped so that all bumps land in a single weekly PR,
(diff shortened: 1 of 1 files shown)


chenson2018 (approved):
Thanks!

jessealama:
Thanks! Looking forward to seeing what Dependabot says about our stuff.

PR #896, opened 2026-09-11
ci: drop manual TEST_ARGS step

jessealama:
The step where we inject `TEST_ARGS` into the `GITHUB_ENV` environment variable was a workaround for a bug in `lean-action`, in which it ignored its own `test-args` input. That fix landed in leanprover/lean-action#153 and is included in `lean-action` since v1.6.0, which is what we pick up here on the CSLib side.


chenson2018 (approved):
So many months later the bug I introduced is fixed! Thanks for remembering this.

PR #899, opened 2026-09-11
feat: Cobham Function Algebra

BoltonBailey:
This PR adds [Cobham's function algebra](https://www.cs.toronto.edu/~sacook/homepage/cobham_intrinsic.pdf).

Zulip: https://leanprover.zulipchat.com/#narrow/channel/513188-CSLib/topic/Cobham.27s.20axioms/with/622656298

Claude code was used to make this PR.

See the Design notes section of the docstring for more thoughts. I have implemented something fairly close to Cobham's paper. I would argue that we should also implement the "polynomially bounded recursion on notation" version, but perhaps that's best left for a future PR.


Cslib/Computability/FunctionAlgebras/Cobham/Defs.lean
```diff
@@ -0,0 +1,193 @@
...
+public import Mathlib.Data.Fin.Tuple.Basic
+import Cslib.Computability.Machines.Turing.SingleTape.Deterministic
+
+/-!
+# Cobham's function algebra
+
+This file defines Cobham's machine-independent characterization
+of the polynomial-time computable functions
+[Cobham, *The intrinsic computational difficulty of functions*][Cobham1965],
+as a model of computation on strings `List Symbol` over an arbitrary alphabet `Symbol`:
```
crei, line 18: Could you clarify if the original definition was on natural numbers? If yes, where is the current exposition from and maybe we want both definitions?

Cslib/Computability/FunctionAlgebras/Cobham/Defs.lean
```diff
@@ -0,0 +1,193 @@
...
+
+universe u
+
+namespace Cslib
+
+variable {Symbol : Type u}
+
+/-- **Terms of Cobham's function algebra** over the alphabet `Symbol`. A term of type
+`Cobham Symbol n` denotes an `n`-ary function on strings `List Symbol` (see
+`Cobham.eval`): the projections, the empty string, the symbol successors `x ↦ a :: x`,
```
crei, line 72: Is "symbol successor" the right term here, where is it from? Wouldn't `cons` be easier to understand?

Cslib/Computability/FunctionAlgebras/Cobham/Defs.lean
```diff
@@ -0,0 +1,193 @@
...
+          (bound.eval (Fin.cons x v)).length := Iff.rfl
+
+open Turing.SingleTapeTM in
+/-- **Cobham's characterization of polynomial time** [Cobham1965]: a binary string function
+is denoted by a limited unary term if and only if it is computable in polynomial time by a
+single-tape Turing machine. -/
+proof_wanted exists_limited_iff_polyTimeComputable (f : List Bool → List Bool) :
+    (∃ c : Cobham Bool 1, c.Limited ∧ ∀ x, c.eval (fun _ => x) = f x) ↔
+      Nonempty (PolyTimeComputable f)
+
```
crei, line 190: Would be nice to also relate that to multi-tape TMs (since we do not yet have the other bridge)

Cslib/Computability/FunctionAlgebras/Cobham/Defs.lean
BoltonBailey: Here's a [link](https://www.cs.toronto.edu/~sacook/homepage/cobham_intrinsic.pdf) from the wikipedia references section (very hard to find otherwise, so I'll mention it in the docstring too).

Yes, the original definition was on natural numbers (actually, looks like it was on decimal natural numbers rather than binary). While I have been trying to keep the implementation close to the paper, in light of the fact that he basically only treats numbers using list operations on the digits and it being easier to do recursion in a functional programming language on lists, I modified the definition to just work over lists. I was hoping people would consider this similar enough to the original definition that we wouldn't feel the need to make another version which captures the paper definition even more exactly.

Cslib/Computability/FunctionAlgebras/Cobham/Defs.lean
BoltonBailey: I guess so, strangely some of the other references refer to these operations as "successor" even when they operate on strings.

Cslib/Computability/FunctionAlgebras/Cobham/Defs.lean
BoltonBailey: Yes, I sort of wonder if there should be a directory Computability/Models to just state the models and then a separate directory about the relationships between models. Like we should maybe have

- SingleTapeTM.FP (as defined)
- MultiTapeTM.FP
- Cobham.FP
- _root_.FP as an alias to one of these
- A bunch of lemmas proving the equivalence.

There is a messy but empirically but interesting question here: Given that we want to prove SingleTapeTM.FP = MultiTapeTM.FP = Cobham.FP, what is the set of containments which imply these equalities while minimizing the total code needed to prove each of them directly.

Cslib/Computability/FunctionAlgebras/Cobham/Defs.lean
BoltonBailey: I was thinking it might be more efficient to use the definition [on page 2 here](https://mathweb.ucsd.edu/~sbuss/ResearchWeb/CRSF_paperone/paperoneRevisedAPALNov2015.pdf) which is farther from the cobham paper. I asked Claude to rewrite my complexitylib code in terms of this definition and it didn't make anything much smaller, but it feels a bit more workable as a definition to me. So if we are going to do a second version of cobham I think it would be this one.

BoltonBailey: Anyways, to answer the original question, I don't know if the idea that we can substitute strings of symbols for naturals has a specific provenance, it's just something I asked the model for at some point because I thought it would be easier, so I'll just explain the history in the module docstring.

Cslib/Computability/FunctionAlgebras/Cobham/Defs.lean
BoltonBailey: Changed

Cslib/Computability/FunctionAlgebras/Cobham/Defs.lean
```diff
@@ -0,0 +1,238 @@
...
+## Main statements
+
+- `Cslib.Cobham.exists_limited_iff_polyTimeComputable` — Cobham's characterization of
+  polynomial time: over the binary alphabet, the functions denoted by limited unary terms are
+  exactly the polynomial-time computable functions
+  (`Cslib.Turing.SingleTapeTM.PolyTimeComputable`). Proof wanted.
+- `Cslib.Cobham.exists_limited_iff_multiTapePolyTimeComputable` — the same characterization
+  against the multi-tape model
+  (`Cslib.Turing.MultiTapeTM.ComputableInTimeAndSpaceOfLength`). Proof wanted.
```
BoltonBailey:
```suggestion
  (`Turing.MultiTapeTM.ComputableInTimeAndSpaceOfLength`). Proof wanted.
```

Cslib/Computability/FunctionAlgebras/Cobham/Defs.lean
```diff
@@ -0,0 +1,238 @@
...
+### Length bounding syntax approaches
+
+There are essentially two slightly different ways of handling the part of the definition that
+ensures the recursions are bounded are bounded.
+
+In Cobham's original formulation, a function is allowed in the set of
+recursive constructions that allows us to obtain outputs whose length is a multiplication of
+lengths of input. This type of function seems to be known as a "smash function".
+Originally this smash function was `x ^ length y`, but `c^(length x * length y)` for a 1-character
+string `c` is also possible (e.g. [BBFMT16]).
```
BoltonBailey:
```suggestion
string `c` is also possible (e.g. [Beckmann2016]).
```

Cslib/Computability/FunctionAlgebras/Cobham/Defs.lean
```diff
@@ -0,0 +1,238 @@
...
+The presence of a smash function ensures we can create functions of polynomial blowup of any degree.
+The limited recursion is then length-bounded by another function of the class.
+
+The alternative is to bound the recursion not by another function of the class,
+but by a generic multivariable polynomial in the lengths of the inputs.
+This obviates the need for a smash function.
+Proof that this is equivalent to FP can be found in [Clote] Lemma 3.90, where it is called
```
BoltonBailey:
```suggestion
Proof that this is equivalent to FP can be found in [Clote1999] Lemma 3.90, where it is called
```

Cslib/Computability/FunctionAlgebras/Cobham/Defs.lean
```diff
@@ -0,0 +1,238 @@
...
+either in the syntax or in the evaluation.
+We do this to avoid mutual recursion.
+Instead, we add a predicate `Limited` which checks that the bound always holds, which we can use
+to examine the class of functions that are actually bounded.
+
+## References
+
+* [A. Cobham, *The intrinsic computational difficulty of functions*][Cobham1965]
+  [Link](https://www.cs.toronto.edu/~sacook/homepage/cobham_intrinsic.pdf)
+* [Beckmann et al., "Cobham Recursive Set Functions"][BBFMT16]
```
BoltonBailey:
```suggestion
* [Beckmann et al., "Cobham Recursive Set Functions"][Beckmann2016]
```

Cslib/Computability/FunctionAlgebras/Cobham/Defs.lean
```diff
@@ -0,0 +1,238 @@
...
+  [Link](https://mathweb.ucsd.edu/~sbuss/ResearchWeb/CRSF_paperone/paperoneRevisedAPALNov2015.pdf)
+* [Clote, *Computational Models and Function Algebras*][Clote]
```
BoltonBailey:
```suggestion
* [Clote, *Computation Models and Function Algebras*][Clote1999]
```

references.bib
```diff
@@ -406,6 +406,17 @@ @incollection{ Thomas1990
...
+  author       = {Cobham, Alan},
+  editor       = {Bar-Hillel, Yehoshua},
+  title        = {The intrinsic computational difficulty of functions},
+  booktitle    = {Logic, Methodology and Philosophy of Science: Proceedings of the 1964 International Congress},
+  pages        = {24--30},
+  publisher    = {North-Holland},
+  address      = {Amsterdam},
+  year         = {1965}
+}
+
```
BoltonBailey:
```suggestion

@article{ Beckmann2016,
  author       = {Beckmann, Arnold and Buss, Sam and Friedman, Sy-David and
                  M{\"u}ller, Moritz and Thapen, Neil},
  title        = {Cobham recursive set functions},
  journal      = {Annals of Pure and Applied Logic},
  volume       = {167},
  number       = {3},
  pages        = {335--369},
  year         = {2016},
  doi          = {10.1016/j.apal.2015.12.005}
}

@incollection{ Clote1999,
  author       = {Clote, Peter},
  editor       = {Griffor, Edward R.},
  title        = {Computation Models and Function Algebras},
  booktitle    = {Handbook of Computability Theory},
  series       = {Studies in Logic and the Foundations of Mathematics},
  volume       = {140},
  pages        = {589--681},
  publisher    = {North-Holland},
  address      = {Amsterdam},
  year         = {1999}
}

```

PR #903, opened 2026-09-14
feat(Language): characterizing regular languages using syntactic monoids

ctchou:
This PR develops yet another characterization of regular languages: a language is regular if and only if its syntactic monoid is finite.  The syntactic monoid is the quotient of the free monoid by the Myhill congruence, which is a two-sided congruence on finite words that is finer than the Nerode congruence used in the Myhill-Nerode theorem.

As part of this PR, both left and two-sided congruences on finite words are defined and all congruences on finite words are moved from the namespace `Cslib` to the more appropriate namespace `Language`.


Cslib/Computability/Languages/Congruences/Basic.lean
```diff
@@ -0,0 +1,65 @@
...
+
+namespace Language
+
+variable {α : Type*}
+
+/-- A right congruence is an equivalence relation on finite sequences (represented by lists)
+that is preserved by concatenation on the right.  The equivalence relation is represented
+by a setoid to to enable ready access to the quotient construction. -/
+class RightCongruence (α : Type*) extends eq : Setoid (List α) where
+  right_cov : CovariantClass _ _ (fun x y => y ++ x) eq
```
eric-wieser, line 29: It might be worth copying `Con` and `RingCon` here; that is, skipping `CovariantClass` and writing these more directly as `append_left` and `append_right`.

Note that those also don't override `toSetoid` to `eq`, and instead add a `FunLike` instance.

Cslib/Computability/Languages/SyntacticMonoid.lean
```diff
@@ -0,0 +1,52 @@
...
+  r := c.r
+  iseqv := c.iseqv
+  mul' {w x y z} h_wx h_yz := by
+    have h_wyxy : c.eq (w * y) (x * y) := c.right_cov.elim y h_wx
+    have h_xyxz : c.eq (x * y) (x * z) := c.left_cov.elim x h_yz
+    exact c.iseqv.trans h_wyxy h_xyxz
+
+/-- The syntactic monoid of a language `l` is the quotient of the free monoid
+by the Myhill congruence of `l`. -/
+def SyntacticMonoid (l : Language α) := l.MyhillCongruence.toCon.Quotient
```
eric-wieser: I'd suggest `abbrev` here, then you don't need the `instance` below.

ctchou: Fixed.

Cslib/Computability/Languages/Congruences/Basic.lean
ctchou: But wouldn't that require a `Mul (List α)` with `++` being the multiplication?  My impression from the following Zulip thread is that that is not considered a good idea:
[#Is there code for X? > List as a monoid](https://leanprover.zulipchat.com/#narrow/channel/217875-Is-there-code-for-X.3F/topic/List.20as.20a.20monoid/with/619146949)

ctchou:
Hi @eric-wieser  Do you have any more comments on this PR?  Thanks in advance!

fmontesi:
Gentle ping @eric-wieser for a last check, otherwise LGTM.

fmontesi (approved):

PR #904, opened 2026-09-14
feat(CCS/VendingMachine): add HML specification for the CCS vending machine

fmontesi:
This PR showcases how HML can be applied to CCS, by distinguishing the classical deterministic and nondeterministic vending machines with the formula `d[Coin](d⟨Tea⟩⊤ ∧ d⟨Coffee⟩⊤)`.

Depends on #863.


Cslib.lean
```diff
@@ -186,18 +189,21 @@ public import Cslib.Languages.Mech.LocalComputation
 public import Cslib.Languages.StatefulProcesses.Basic
```
jerryboi3338-star, line 189: 192.168.l.254

PR #908, opened 2026-09-15
ci: cancel superseded PR runs of Lean Action CI

jessealama:
Adds a [`concurrency` group](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency) to `lean_action_ci.yml` so that new pushes to a PR cancel running steps instead of doing a redundant full build behind it. This saves resources (almost certainly wasted compute and network traffic). It could even reduce the bill the `leanprover` GitHub org pays for compute. Canceling a redundant run also better reflects development reality: a canceled run shows the branch is busy.

(Mathlib does [the same thing](https://github.com/leanprover-community/mathlib4/blob/master/.github/workflows/build.yml#L16-L21). GitHub-native concurrency, replacing a third-party approach, was introduced in [this](https://github.com/leanprover-community/mathlib4/pull/20435) PR.)

Runs on `main` and `nightly-testing` are never canceled. `report_failures_nightly-testing.yml` consumes completed runs via `workflow_run`, so canceling one there would swallow a status report. This also works with the merge queue.

`docs.yml` is left alone since it only runs on pushes to `main`.


.github/workflows/lean_action_ci.yml
```diff
@@ -9,6 +9,11 @@ on:
   merge_group:
   workflow_dispatch:
 
+# Cancel superseded runs on PR branches, but never on main or nightly-testing.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' && github.ref != 'refs/heads/nightly-testing' }}
```
chenson2018: I think this may not behave as expected because of the concurrency queue, which by default allows only one job at a time. There is a setting to expand this, but the docs indicate

> Note that `queue: max` cannot be combined with `cancel-in-progress: true`, because the two options describe conflicting behaviors for handling in-progress runs.

I'm not 100% sure if that also applies to conditional `cancel-in-progress`, but it seems simplest to follow what Mathlib does: when these conditions apply the id is included in the group name, making them unique.

jessealama: Ah, good point. I thought that what I had should be safe (and maybe it is), but doing what Mathlib does is just fine.

PR #910, opened 2026-09-16
ci(weekly-lints): create [leanOptions] when the lakefile lacks it

jessealama:
If `[leanOptions]` is missing from the lakefile, `sed` append is a silent no-op. The job thus succeeds without running the linters. Here, we create the section when it's missing. A missing `[leanOptions]` is unlikely, but it's worth watching out for. (A more proper solution would be to parse the file and work with it in a more structured way. This PR keeps with the text-munging status quo.)


chenson2018:
I think this is overkill. Realistically we are never going to remove `leanOptions` completely, as this would mean we're turning off all the linting configurations.

jessealama:
Not really needed.

PR #914, opened 2026-09-16
ci(nightly-testing): guard the bump-branch steps when the nightly bump branch exists

jessealama:
The "Exit if matching branch exists" step did `exit 0`, which only ends that step. Since a zero exit is a success, the job simply moved on and the subsequent steps ran. When no `bump/v4.x.0` branch existed, the first of them threw, and the `if: failure()` Zulip warning could post a misleading message about a toolchain pattern mismatch, but the real reason was the non-existence of the proper branch.

This PR drops the no-op step and guards both fetch steps on the branch check, as other later step already are. Post a dedicated warning to Zulip when no `bump/v4.x.0` branch is found, and only post the toolchain warning when the toolchain fetch step itself failed. 

Also replace `new Exception`, which is a `ReferenceError` in JavaScript, with `new Error`. `Exception` doesn't exist in JS. So, curiously, running this code caused a _different_ kind of exception to be thrown (a `ReferenceError` for referring to something that doesn't exist), which ended up killing the step as intended, but for the wrong reason.


kim-em:
:robot: Claude says:

Narrowing `if: failure()` to `steps.bump_version.outcome == 'failure'` silences the steps that run *before* it. The old blanket warning sat just above `Setup for automatic PR creation`, so it only ever covered what preceded it: `Generate app token`, `Checkout code`, `Update the nightly-testing-YYYY-MM-DD branch`, `Install Zulip API client`, `Check last message and post if necessary`, and `check_branch` itself. Those got a misleading toolchain message before; now they get nothing. `handle_failure` doesn't pick them up either, since it keys off `workflow_run.conclusion`, which is the triggering CI run rather than this job.

Could you add a catch-all as the last step?

```yaml
    - name: Send warning message on Zulip on any other failure
      if: >-
        (failure() || cancelled())
        && steps.latest_bump_branch.outcome != 'failure'
        && steps.bump_version.outcome != 'failure'
```

with a generic "the bump job failed before reaching PR creation" body and a link to the run. Skipped steps are fine here, since a skipped step's `outcome` is `skipped` and both comparisons hold.

Two things I noticed while reading, neither blocking:

- Both warnings are less cause-specific than their wording suggests. `latest_bump_branch` can fail because `listBranches` errored rather than because no branch matched, and `bump_version` can fail fetching the branch, fetching the file, or getting empty content, all of which report "pattern doesn't match". Narrowing the `if` is already an improvement on what was there; exposing a reason from the scripts would finish the job.
- Pre-existing and separate: `handle_failure` links `${{ github.sha }}`, which on a `workflow_run` event is the default-branch commit, not the `nightly-testing` head that failed. It uses `github.event.workflow_run.id` correctly for the run link, so `github.event.workflow_run.head_sha` is presumably what was intended for the commit.

jessealama:
@kim-em I've added a catch-all as the last step, with your condition as well as a generic message and link. It'll be worth keeping our eye on the message to make sure it's useful and doesn't become recurring spam.

I also take your second point. Both scripts now record a `reason` output at each point of failure, and the two warnings report it now instead of asserting a single cause.

Agreed the `github.sha` link in `handle_failure` should be `github.event.workflow_run.head_sha`; I'll do that in a separate PR. Thanks!

chenson2018:
I'm somewhat inclined to not touch this, as it seems we could just go ahead and remove the "CSLib Status Updates" thread completely in favor of the `downstrean-lean4` reporting. Would you have any objections to this? Is there any improvements in one of your open PRs that this would end up missing?

jessealama:
I'm happy to drop this! I've also been watching the downstream-lean4 stuff and am happy to see things get consolidated.

PR #923, opened 2026-09-18
feat(Complexity): Query Complexity of Boolean Functions (aka Decision…

Lsonic233:
Defines three Boolean function complexity measures : sensitivity `s(f)`, block sensitivity `bs(f)` and certificate complexity `C(f)` with the inequality chain : `s(f) ≤ bs(f) ≤ C(f)`.

- `Defs.lean` sets up the boolean hypercube and boolean functions which is defined as a function from `{0,1}^n` to `{0,1}`. 
- `Measures.lean` defines the above three measures and proves the inequality chain. 
- Deterministic Decision Trees and `D(f)` : the decision tree complexity of boolean function `f` will follow in a second PR. 

AI usage : Used claude to help me fill in proofs in a few places and refactor my handwritten proofs.


references.bib
```diff
@@ -53,6 +53,16 @@ @article{AngluinLaird1988
   doi          = {10.1007/BF00116829}
 }
 
+@book{AroraBarak2009,
```
crei: It's already there as `AroraBarak09`, but feel free to combine them.

Cslib/Computability/QueryComplexity/Defs.lean
```diff
@@ -0,0 +1,140 @@
...
+
+namespace Cslib.QueryComplexity
+
+variable {n : ℕ}
+
+/-- Bit string of length `n`. -/
+abbrev Cube (n : ℕ) : Type := Fin n → Bool
+
+/-- `f : {0,1}ⁿ → {0,1}`. -/
+abbrev BoolFunc (n : ℕ) : Type := Cube n → Bool
```
crei, line 47: I mean it's kind of trivial but maybe you should sync some of the definitions here with https://github.com/leanprover/cslib/pull/890 ?

Cslib/Computability/QueryComplexity/Defs.lean
```diff
@@ -0,0 +1,140 @@
...
+/-!
+# Boolean functions on the hypercube
+
+A Boolean function is a map `{0,1}ⁿ → {0,1}`. This file sets up the boolean hypercube, bit and block
+flips, and partial assignments. The complexity measures are built in
+`Measures.lean` and `DecisionTree.lean`.
+
+## Main definitions
+
+- `Cube n`: bit strings of length `n`, i.e. `Fin n → Bool`.
```
crei, line 21: Do you have a reference for naming this object `Cube`? Wouldn't "Assignment" be easier to understand?

Cslib/Computability/QueryComplexity/Defs.lean
```diff
@@ -0,0 +1,140 @@
...
+- `BoolFunc n`: Boolean functions `Cube n → Bool`.
+- `Block n`: a set of coordinates.
+- `flipBit`, `flipBlock`: flipping one coordinate, or every coordinate of a block.
+- `Assignment n`: a partial assignment, fixing some coordinates and leaving others free.
```
crei, line 25: If you rename `Cube` to `Assignment`, then this could be `PartialAssignment` (and also could be in general)

Cslib/Computability/QueryComplexity/Defs.lean
```diff
@@ -0,0 +1,140 @@
...
+
+variable {n : ℕ}
+
+/-- Bit string of length `n`. -/
+abbrev Cube (n : ℕ) : Type := Fin n → Bool
+
+/-- `f : {0,1}ⁿ → {0,1}`. -/
+abbrev BoolFunc (n : ℕ) : Type := Cube n → Bool
+
+/-- A block: a set of coordinates. -/
```
crei, line 49: Sorry that I have so many remarks about terminology and please push back, but Arora-Bark talks about "bit position". I know it makes sense when talking about hypercubes and so on, but currently, "index" would be equally valid and understandable, right?

Cslib/Computability/QueryComplexity/Measures.lean
```diff
@@ -0,0 +1,312 @@
...
+def sensitiveCoords (f : BoolFunc n) (x : Cube n) : Finset (Fin n) :=
+  Finset.univ.filter (IsSensitiveCoord f x)
+
+/-- The bridge between the predicate and the `Finset` it cuts out. -/
+@[simp]
+lemma mem_sensitiveCoords {f : BoolFunc n} {x : Cube n} {i : Fin n} :
+    i ∈ sensitiveCoords f x ↔ IsSensitiveCoord f x i := by
+  simp [sensitiveCoords]
+
+/-- `s(f, x)` is the number of sensitive coordinates in input `x`. -/
```
crei: What about
```suggestion
/-- The number of sensitive coordinates in input `x`, usually denoted `s(f, x)`. -/
```
Although in Arora-Barak, this is called `sₓ(f)`, right? I don't have the other reference you cited.

Cslib/Computability/QueryComplexity/Measures.lean
```diff
@@ -0,0 +1,312 @@
...
+def pointSensitivity (f : BoolFunc n) (x : Cube n) : ℕ := (sensitiveCoords f x).card
+
+/-- `s(f)`. -/
```
crei: I think you should still explain this definition informally and only use known notation when you either introduce it in lean or when referencing textbooks.

Cslib/Computability/QueryComplexity/Defs.lean
```diff
@@ -0,0 +1,140 @@
...
+abbrev Block (n : ℕ) : Type := Finset (Fin n)
+
+/-- `x` with coordinate `i` flipped. Phrased with `Function.update` so that the
+whole `Function.update` simp set (`update_self`, `update_of_ne`, `update_idem`,
+`update_eq_self`, …) applies to it. -/
+def flipBit (x : Cube n) (i : Fin n) : Cube n := Function.update x i (!x i)
+
+/-- The flipped coordinate reads back negated. -/
+@[simp]
+lemma flipBit_self (x : Cube n) (i : Fin n) : flipBit x i i = !x i :=
```
crei: I'm wondering if we really need those lemmas if proofs using them could also be just be `simp [flipBit]`

Cslib/Computability/QueryComplexity/Defs.lean
Lsonic233: Cube is short for the boolean hypercube (or hamming cube) which views {0,1}^n as vertices of a graph such that two strings are connected by an edge if their hamming distance is 1. Analysis of Boolean Functions by O'Donnell also calls it that. 
If you prefer not to have Cube, then would something like "Input" or "BooleanInput" be better than "Assignment"?
"Assignment" is the term used for partial assignments in the survey by [BuhrmanDeWolf](https://homepages.cwi.nl/~rdewolf/publ/qc/dectree.pdf). See the section Certificate Complexity.

Cslib/Computability/QueryComplexity/Measures.lean
```diff
@@ -0,0 +1,312 @@
...
+/-- All sensitive families. -/
+def sensitiveFamilies (f : BoolFunc n) (x : Cube n) : Finset (Finset (Block n)) :=
+  Finset.univ.filter (IsSensitiveFamily f x)
+
+@[simp]
+lemma mem_sensitiveFamilies {f : BoolFunc n} {x : Cube n} {F : Finset (Block n)} :
+    F ∈ sensitiveFamilies f x ↔ IsSensitiveFamily f x F := by
+  simp [sensitiveFamilies]
+
+/-- `bs(f, x)`. -/
```
crei:
```suggestion
/-- The block sensitivity of `f` on input `x`, i.e. the maximum number `b` such that there are disjoint blocks of bit positions ... (I would just cite the definition from the text book), usually denoted `bs(f, x)`. -/
```

Cslib/Computability/QueryComplexity/Measures.lean
```diff
@@ -0,0 +1,312 @@
...
+def IsSensitiveFamily (f : BoolFunc n) (x : Cube n) (F : Finset (Block n)) : Prop :=
+  (∀ B ∈ F, IsSensitiveBlock f x B) ∧ (F : Set (Block n)).PairwiseDisjoint id
+
+instance (f : BoolFunc n) (x : Cube n) (F : Finset (Block n)) :
+    Decidable (IsSensitiveFamily f x F) :=
+  decidable_of_iff ((∀ B ∈ F, IsSensitiveBlock f x B) ∧
+    ∀ P ∈ F, ∀ Q ∈ F, P ≠ Q → Disjoint P Q) Iff.rfl
+
+/-- All sensitive families. -/
+def sensitiveFamilies (f : BoolFunc n) (x : Cube n) : Finset (Finset (Block n)) :=
```
crei, line 109: Do we really need this definition?

Cslib/Computability/QueryComplexity/Defs.lean
Lsonic233: My bad, my main reference and the place I studied this topic from is the [Buhrman and de Wolf survey](https://homepages.cwi.nl/~rdewolf/publ/qc/dectree.pdf) so most of my naming choices match that rather than Arora-Barak. I've addressed this in my previous comment.

crei:
Sorry for the many comments! In general, this is a good contribution, it just has so many definitions :)
Feel free to push back and we can also talk about in the chat or in a call.
Another final comment: Shouldn't this be "CommunicationComplexity" instead of "QueryComplexity"?

Lsonic233:
> Sorry for the many comments! In general, this is a good contribution, it just has so many definitions :) Feel free to push back and we can also talk about in the chat or in a call. Another final comment: Shouldn't this be "CommunicationComplexity" instead of "QueryComplexity"?

Sure, I'll send you a zulip dm and make a thread about this topic in a bit. 
"Communication Complexity" refers to a completely different model of computation where two parties (where the inputs are distributed between them ) try to compute a boolean function by communicating bits of info. 
"Query Complexity" or Decision Tree Complexity is the boolean decision tree model of computation (I have a second PR ready defining boolean decision trees and the decision tree complexity D(f) and proving that C(f) <= D(f)).

Cslib/Computability/QueryComplexity/Defs.lean
Lsonic233: I agree, "index" and "coordinate" are interchangeable. I'd keep "coordinate" because thats what most of the definitions about blocks use in the file.

Cslib/Computability/QueryComplexity/Defs.lean
Lsonic233: We don't - I removed them.

Cslib/Computability/QueryComplexity/Measures.lean
Lsonic233: Makes sense - I've changed the definition to look like bs_x(f) and have followed your suggestion of writing the definitions in docstrings.

Cslib/Computability/QueryComplexity/Defs.lean
crei: I don't want to argue with O'Donnell but isn't what you define rather a point in the hypercube than the hypercube itself?
I think I would prefer PartialAssignment for partial assignments. The article you link does not formally introduce the concept and only uses it twice.
I'm mostly concerned about the same definition having different names in different directories of cslib and in general creating confusion. If some terms are only used locally thisis fine but I think these definitions could be made more widely available.

Maybe @SamuelSchlesinger also has an opinion.

crei:
> Sure, I'll send you a zulip dm and make a thread about this topic in a bit. "Communication Complexity" refers to a completely different model of computation where two parties (where the inputs are distributed between them ) try to compute a boolean function by communicating bits of info. "Query Complexity" or Decision Tree Complexity is the boolean decision tree model of computation (I have a second PR ready defining boolean decision trees and the decision tree complexity D(f) and proving that C(f) <= D(f)).

Right, I confused the chapters in Arora-Barak

Cslib/Computability/QueryComplexity/Defs.lean
Lsonic233: I think the type here is the hypercube (#eval Fintype.card (Cube 3) gives 8) and `x : Cube n` is a point of `Cube n`. I should probably change the docstring above it to say something like "The boolean hypercube whose points are bit strings of length n". 
For partial assignments - I agree I'll change Assignment in my code to PartialAssignment. 
Yeah I think we should work on some infra for BooleanFunctions ({0,1}^n -> {0,1}) especially now that #890 is merged and it uses it. This would require some coordination on zulip.

Cslib/Computability/QueryComplexity/Measures.lean
Lsonic233: It's there to follow the same pattern as `sensitiveCoords` and `singletonFamily`. I could remove it and inline it - don't think it would make much of a difference.

Cslib/Computability/QueryComplexity/Defs.lean
SamuelSchlesinger: I think we should have global definitions for Boolean strings of known size somewhere. I believe we shouldn't call it `Cube` but `BitString` or `BitVec` is a reasonable name.

Cslib/Computability/QueryComplexity/Defs.lean
SamuelSchlesinger: Yeah, I'm working out the right way to do that and I'll include this refactor in a forthcoming PR.

SamuelSchlesinger:
This is great. I'm under the impression we can simplify this PR quite a bit and get something basically the right shape, but overall this is the right first direction. Thank you for the contribution! I am working on a more detailed review.

Cslib/Computability/QueryComplexity/Defs.lean
SamuelSchlesinger: I added these in #943.

Cslib/Computability/QueryComplexity/Measures.lean
```diff
@@ -0,0 +1,315 @@
...
+/-- `C` forces the value `b`: everything consistent with `C` has `f = b`. -/
+def IsCertificate (f : BoolFunc n) (C : PartialAssignment n) (b : Bool) : Prop :=
+  ∀ x, Agrees C x → f x = b
+
+instance (f : BoolFunc n) (C : PartialAssignment n) (b : Bool) : Decidable (IsCertificate f C b) :=
+  inferInstanceAs (Decidable (∀ x, Agrees C x → f x = b))
+
+/-- Assignments consistent with `x` that already force `f x`. -/
+def certificates (f : BoolFunc n) (x : Cube n) : Finset (PartialAssignment n) :=
+  Finset.univ.filter (fun C => Agrees C x ∧ IsCertificate f C (f x))
```
SamuelSchlesinger, line 234: I like keeping PartialAssignment as a general object. At a fixed input x, though, an assignment agreeing with x is determined entirely by its support. Would you consider defining pointwise certificates using a set of coordinates, with a lemma connecting that view to PartialAssignment? I think that would simplify C_x(f) and the bs_x(f) <= C_x(f) proof.

SamuelSchlesinger (changes requested):
This is great, I am very close to approval but I just want to see this one change so we can make the proofs a bit tighter. We can do the BitString and BooleanFunction reuse in a followup unless #943 lands first.

Cslib/Computability/QueryComplexity/Measures.lean
Lsonic233: I played around with your idea of defining point certificates using just coordinate sets (instead of appealing to IsCertificate) and here is the definition I am working with (`Fixes` in the code below). 

```lean4
def Fixes (f : BoolFunc n) (x : Cube n) (S : Finset (Fin n)) : Prop :=
  ∀ y, (∀ i ∈ S, y i = x i) → f y = f x

instance (f : BoolFunc n) (x : Cube n) (S : Finset (Fin n)) : Decidable (Fixes f x S) :=
  inferInstanceAs (Decidable (∀ y, (∀ i ∈ S, y i = x i) → f y = f x))

lemma fixes_univ (f : BoolFunc n) (x : Cube n) : Fixes f x Finset.univ :=
  fun _ hy => congrArg f (funext fun i => hy i (Finset.mem_univ i))

lemma fixes_nonempty (f : BoolFunc n) (x : Cube n) :
    (Finset.univ.filter (Fixes f x)).Nonempty :=
  ⟨Finset.univ, Finset.mem_filter.mpr ⟨Finset.mem_univ _, fixes_univ f x⟩⟩

def pointCertificateComplexity (f : BoolFunc n) (x : Cube n) : ℕ :=
  (Finset.univ.filter (Fixes f x)).inf' (fixes_nonempty f x) Finset.card
```

I still think we need `IsbCertificate` for theorems like [Theorem10](https://homepages.cwi.nl/~rdewolf/publ/qc/dectree.pdf) we need $C^{1}(f)$ and $C^{0}(f)$. Also I just found out that the Barak Arora textbook follows your way of defining $C_x(f)$ which is different from how the Buhrman and de Wolf defines it (this is the one I used for my implementation). 
The proofs for $bs_x(f) \leq C_x(f)$ are not significantly shorter (maybe a few lines across the main theorem and supporting lemmas). I could push a demo of these proofs if you'd like.

PR #935, opened 2026-09-22
refactor(Circuit): preserve intermediate synthesis targets

SamuelSchlesinger:
Keep both target families when composing synthesis bounds, and provide trans for retaining only the final targets. Add of_mem to reuse available functions through existing synthesis combinators, infer fold arguments from their proofs, and simplify shared-output examples and Lupanov synthesis.

This is the first in a sequence of circuit-library refactors, preparing for a simpler circuit representation and a circuit-complexity API.


barni120400 (approved):

crei (approved):
Nice improvement!

fmontesi (approved):

PR #937, opened 2026-09-22
chore: remove kim-em from the catch-all CODEOWNERS rule

kim-em:
This PR removes `@kim-em` from the catch-all `*` rule and keeps that name only on `/.github/workflows/`.

The `*` rule asks all six owners for a review on every pull request, whatever it touches. That had accumulated 47 open requests, of which the only ones genuinely needing this reviewer were two recent changes to the nightly bump pipeline. Scoping to the workflows directory keeps those and drops the rest.

The workflows rule repeats the other five owners because CODEOWNERS applies only the last matching rule, so naming one person there would have quietly removed everyone else from workflow pull requests.

Since `main` carries no code-owner approval requirement, these rules only drive review requests, so this changes who is notified and nothing about how a pull request merges.

🤖 Prepared with Claude Code


fmontesi (approved):
LGTM, but a clarification before it gets merged:

> Since main carries no code-owner approval requirement, these rules only drive review requests, so this changes who is notified and nothing about how a pull request merges.

`main` has a ruleset requiring that at least one code owner approves the pull request before it can be merged. So `CODEOWNERS` supports our policy for PR merging in `DECISION_MAKING.md`.

Removing yourself from `*` will mean that you'll be able to give the necessary maintainer approval only for PRs regarding workflows. I think this is what you want anyway, so I'm for merging this, this is FYI only. Could you confirm this, @kim-em?

I'm approving to unlock merging as soon as this is confirmed.

PR #941, opened 2026-09-23
feat(AGENTS): Lean MCP server usage guidelines

fmontesi:
Instructs agents to use the Lean MCP server during development and clarifies some command usage for building and testing.


SamuelSchlesinger:
I worry that my agents might interpret `**Use the Lean MCP server for routine Lean development whenever its tools are available.**` a bit too strongly. I am curious if you have tested with these instructions and seen an improvement in behavior?

fmontesi:
> I worry that my agents might interpret `**Use the Lean MCP server for routine Lean development whenever its tools are available.**` a bit too strongly. I am curious if you have tested with these instructions and seen an improvement in behavior?

I'm experiencing that, without a strong instruction like this, my agents default to invoking lake from the cmd line way too often and then just stick to that. With this, the behaviour improved.

But I agree in general we should be careful with this kind of instructions, so if you have ideas (now or in a later PR) I'm happy to discuss. Maybe give it a spin and see if it gives you trouble. Unfortunately this testing is really 'manual'.

SamuelSchlesinger:
Yeah, I'll give it a spin and if I see trouble I can make another PR.

SamuelSchlesinger (approved):

PR #942, opened 2026-09-23
refactor(MultiTapeTM): use Mathlib iteration lemmas

barni120400:
`runFrom` is already defined as iteration of `step`, but its API repeats Mathlib's iteration lemmas. Remove six wrappers (`runFrom_zero`, `runFrom_succ_eq_step`, `runFrom_succ_eq_step'`, `runFrom_add`, `runFrom_comm_of_step`, and `runFrom_of_halt`) and update their callers to use `Function.iterate_*` and `Function.Semiconj.iterate_right` directly.

The existing machine definitions and machine-specific halting, space, and computation statements retain their semantics. This is a standalone extraction from the larger deterministic/nondeterministic refactor.

Validation: strict library build, module import check, strict tests, environment and style linters, init-import check, and `git diff --check`.

AI assistance: OpenAI Codex was used to extract the change, update Lean proofs, and run the validation checks.


SamuelSchlesinger (approved):
Great work!

PR #943, opened 2026-09-23
refactor(Circuit): bundle gate count in circuits

SamuelSchlesinger:
Store the internal gate count as an implicit `Circuit.size` field. Enumerate programs paired with output wires for counting, while expressing membership through circuits. Add free wiring circuits and make `Circuit.Computes` describe tuples of outputs. Adapt synthesis extraction, normalization, Shannon and Lupanov statements, and their tests to the bundled representation.

Define shared `Cslib.BitString` and `Cslib.BooleanFunction` types in Foundations and import them directly into the Boolean circuit development. Code using the former `Cslib.Circuits.Boolean.BooleanFunction` name should use `Cslib.BooleanFunction`. The query-complexity development can then use the same types without a circuit dependency.


SamuelSchlesinger:
@BoltonBailey you might be interested in this, given you pointed this out on the original circuit PR.

Cslib/Computability/Circuit/Basic.lean
```diff
@@ -40,76 +38,105 @@ namespace Cslib.Circuits
...
+/-- Output `j` of a wiring circuit is input `select j`. -/
+@[simp] theorem Circuit.eval_wiring (select : Fin outputCount → Fin inputCount)
+    (interpretation : Interpretation σ U) (input : Fin inputCount → U) :
+    (Circuit.wiring σ select).eval interpretation input = input ∘ select := by
   funext output
-  exact Program.trace_input .empty interpretation input output
+  exact Program.trace_input .empty interpretation input (select output)
+
+/-- A circuit computes `F` when its outputs agree with `F` on every input. -/
+def Circuit.Computes (c : Circuit σ inputCount outputCount)
```
BoltonBailey: Maybe a nit pick, but do we actually need this at all? We could just write `c.eval interpretation = F` point free, at which point we could just drop this predicate and express the fact that a circuit computes a function as an equality.

PR #946, opened 2026-09-23
refactor(MultiTape): use RelSeries and shared space accounting

barni120400:
Replace list-based computation paths with Mathlib's `RelSeries`. Count space per tape directly on the path, so a deterministic run and its corresponding path use the same space by definition.


SamuelSchlesinger:
This looks quite nice to me. Needs a rebase.

Cslib/Computability/Machines/Turing/MultiTape/Nondeterministic.lean
```diff
@@ -90,25 +90,41 @@ def initCfg (ntm : MultiTapeNTM k Symbol State) (input : List Symbol) :
...
-  isChainFromTo : cfgs.IsChainFromTo ntm.Step (ntm.initCfg input) last
+/-- A finite nonempty list of configurations joined by steps of `ntm`. -/
+abbrev RunPath (ntm : MultiTapeNTM k Symbol State) (input : List Symbol) :=
+  RelSeries {(c, c') | ntm.Step (input := input) c c'}
 
-namespace ComputationPath
+namespace RunPath
+
+/-- The number of steps taken by a run path. -/
+def time (p : ntm.RunPath input) : ℕ := p.length
```
crei, line 100: Thus could be annoying but I’m wondering if we should not have “p.length - 1”. The time should be the number of transitions. If the RelSeries counts the number of elements then a RunPath with just one element should take zero time.

barni120400: `RelSeries.length` already counts transitions so this is ok.

barni120400:
@SamuelSchlesinger rebased

SamuelSchlesinger (approved):

PR #947, opened 2026-09-24
refactor(Boolean): share bit strings and Boolean functions

SamuelSchlesinger:
## Summary

- Define `Cslib.BitString` and `Cslib.BooleanFunction` in Foundations.
- Update Boolean circuits to import and use the shared definitions, so other Boolean complexity developments can use these types without a circuit dependency.

This is the first part of the split of #943. The stack continues in #948 and #949.

## Validation

- `lake build --wfail --iofail`
- `lake exe mk_all --check`
- `lake test`
- `lake lint`
- `lake exe lint-style`

Codex assisted with splitting #943 into three focused PRs and running these local checks.

PR #948, opened 2026-09-24
refactor(Circuit): describe output tuples with Computes

SamuelSchlesinger:
## Summary

- Generalize `Circuit.Computes` to specify all designated outputs as a function returning an output tuple.
- Add zero-gate `Circuit.wiring` for selecting, permuting, and duplicating inputs, with size, depth, evaluation, and computation lemmas.
- Rename synthesis extraction to `Synthesis.exists_circuit_outputs` and adapt scalar counting, Shannon, Lupanov, and tests to the tuple-valued API.

This is the second part of the split of #943. It builds on #947 and retains the explicit gate-count index; #949 changes that representation.

## Validation

- `lake build --wfail --iofail`
- `lake exe mk_all --check`
- `lake test`
- `lake lint`
- `lake exe lint-style`

Codex assisted with splitting #943 into three focused PRs, adapting this output change to the explicit gate-count representation, and running these local checks.


Cslib/Computability/Circuit/Basic.lean
```diff
@@ -86,24 +101,46 @@ def Circuit.outputDepths (c : Circuit σ inputCount gateCount outputCount) : Fin
...
     (x : Fin inputCount → U) : Fin outputCount → U :=
   c.program.trace i x ∘ c.outputs
 
-/-- A single-output circuit computes `f` when its output agrees with `f` on every input. -/
-def Circuit.Computes (c : Circuit σ inputCount gateCount 1)
-    (interpretation : Interpretation σ U) (f : (Fin inputCount → U) → U) : Prop :=
-  ∀ x, c.eval interpretation x 0 = f x
+/-- A circuit computes `F` when its outputs agree with `F` on every input. -/
+def Circuit.Computes (c : Circuit σ inputCount gateCount outputCount)
+    (interpretation : Interpretation σ U) (F : (Fin inputCount → U) → Fin outputCount → U) :
```
crei, line 128: Should be f instead of F

SamuelSchlesinger: I made the multi-output functions `F` on purpose but happy to shift back to `f`.

SamuelSchlesinger: Yeah, let's just use `f` everywhere it's fine.

crei (approved):

SamuelSchlesinger:
@fmontesi friendly ping to merge

